PR #2, opened 2017-09-21
fixed head printout and flight demo index

gsheni:
For the flight fix:
- added`storage_options={'anon': True}` to fix s3 issue with public files and dask
- `df = df.reset_index(drop=True)` is there because the dataframe is read in parallel with dask, and the index needs to be reset since these values are used to create the `trip_id`

For the head fix:
- there was a circular call in the `head` function of `entityset`. This is fixed by calling head on the `entity` which uses `self.df.head` to get the values. 
- `cutoff_time` for head is not implemented yet.

### DIFF
--- a/featuretools/demo/flight.py
+++ b/featuretools/demo/flight.py
@@ -5,24 +5,25 @@ import pandas as pd
 import dask.dataframe as dd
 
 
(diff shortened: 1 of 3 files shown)


kmax12:
This all looks good to me

PR #3, opened 2017-10-03
Add associative attribute to primitives

rwedge:
Some primitives, given a set of base features, will give the same feature values regardless of order. 

 `Add(feature_A, feature_B)` functions the same as `Add(feature_b, feature_A)`.  

By flagging a primitive as associative, DeepFeatureSynthesis will only create one instance of that primitive per unique set of base features.

Also subclassed the `Compare` primitive into six primitives: `Equals`, `NotEquals`, `GreaterThan`, `GreaterThanEqualTo`, `LessThan`, and `LessThanEqualTo`.


kmax12:
Looks good to me

PR #8, opened 2017-10-11
fixed hash issue with same base features

gsheni:
if there is a feature which is the square of another feature, that would not be currently possible with our current feature primitive. 
```python
Feature(es["log"]["value"]) * Feature(es["log"]["value"])
```

So this fixes this by having the `left` and `right` point to the same element in `base_features`


kmax12:
can you add a test?

kmax12:
looks good to me

PR #13, opened 2017-10-11
Custom primitive creation functions

rwedge:
Added two functions, ``make_trans_primitive`` and ``make_agg_primitive``.  Given a function to apply to input features, list of allowed input feature types, and variable type of the resulting feature, these functions will create a new primitive that can be used with DFS.


featuretools/utils/pickle_utils.py
```diff
@@ -1,4 +1,4 @@
-import pickle
+import cloudpickle
```
kmax12: do you need to add cloudpickle to requirements?

rwedge: added

PR #18, opened 2017-10-20
Select low information features

bschreck:
Removes two functions: `selection.select_high_variance_features` and `selection.select_percent_null()`, and replace them with `selection.remove_low_information_features()`.

The new function removes all features that only contain null values, or that only contain a single value.

The reason to remove these functions is that we are unsure about whether it is generally a good idea to remove these types of features. However, it is almost always a good idea to remove features that provide no information.


kmax12:
looks good to me

PR #20, opened 2017-10-23
Remove as_dir=False option from EntitySet.to_pickle()

bschreck:
`EntitySet` used to have two options for saving to disk

- `as_dir=False` was supposed to save to a single pickle file. This was broken.
- `as_dir=True` saved to a directory containing several pickle files of metadata about each `Entity`, and CSV files with the data for each `Entity`.

This PR removes the `as_dir=False` option.

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -91,8 +91,8 @@ class EntitySet(BaseEntitySet):
         """
         return [e.id for e in self.entities]
 
-    def to_pickle(self, path, as_dir=False):
-        to_pickle(self, path, as_dir=as_dir)
+    def to_pickle(self, path):
+        to_pickle(self, path)
         return self
 
     @classmethod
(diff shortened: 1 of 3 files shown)


featuretools/tests/entityset_tests/test_es.py
```diff
@@ -149,3 +151,14 @@ def test_gzip_glob_entityset(es, gzip_glob_es):
...
+
+
+def test_serialization(es):
+    dirname = os.path.dirname(integration_data.__file__)
+    path = os.path.join(dirname, 'test_entityset.p')
+    if os.path.exists(path):
+        shutil.rmtree(path)
+    es.to_pickle(path)
+    new_es = EntitySet.read_pickle(path)
+    assert es == new_es
```
kmax12: fyi, i think this check only looks to make sure the entity set ids match.

bschreck: Whoops. Changed to __eq__(deep=True)

featuretools/entityset/entityset.py
```diff
@@ -710,6 +710,8 @@ def normalize_entity(self, base_entity_id, new_entity_id, index,
 
         transfer_types = {}
         transfer_types[new_index] = type(base_entity[index])
+        for v in additional_variables + copy_variables:
```
kmax12: Is this bug fix for us currently not transferring the type of `additional_variables` and `copy_variables`?

kmax12:
Looks good to me

PR #21, opened 2017-10-23
Fixed Diff primitive for rare nan case

bschreck:
I ran into a bug with the `Diff` primitive that occurs when for a single variable and cutoff time, all values of both the variable and the feature to group by are nan. The bug occurs because Pandas drops the column names of empty dataframes, so we get a KeyError. This fix required a new variable in the mock entityset because we needed both an Id variable to group on that had a nan value and a numeric variable that had a nan.


kmax12:
Looks good to me

PR #38, opened 2017-12-07
Custom primitive kwarg fix

rwedge:
Setting custom primitive function inputs via keyword arguments no longer changes keyword arguments for all instances of that custom primitive.  Also renamed `PrimitiveBase._get_name` to `PrimitiveBase.generate_name` to avoid confusiong with `PrimitiveBase.get_name`


kmax12:
Good change. Looks good to merge

PR #40, opened 2017-12-08
Remove tests from code coverage

kmax12:
Remove tests from code coverage


kmax12:
After discussion, will keep tests under code coverage for now

PR #51, opened 2017-12-22
Add Words and Chars primitives

Seth-Rothschild:
Add two text primitives, `NumWords` and `NumCharacter` which count the number of words and the number of characters when the variable type is `Text`.

### DIFF
--- a/featuretools/primitives/transform_primitive.py
+++ b/featuretools/primitives/transform_primitive.py
@@ -20,6 +20,7 @@ from featuretools.variable_types import (
     Id,
     Numeric,
     Ordinal,
+    Text,
     Timedelta,
     Variable
 )
@@ -334,6 +335,28 @@ class Weekend(TransformPrimitive):
 class Weekday(DatetimeUnitBasePrimitive):
     """Transform Datetime feature into the boolean of Weekday"""
     name = "weekday"
+
+class NumCharacter(TransformPrimitive):
+    """
+    Return the characters in a given string.
+    """
+    name = 'characters'
+    input_types = [Text]
+    return_type = Numeric
+
+    def get_function(self):
+        return lambda array: pd.Series([len(x) for x in array])
+
+class NumWords(TransformPrimitive):
+    """
+    Returns the words in a given string by counting the spaces.
+    """
+    name = 'numwords'
+    input_types = [Text]
+    return_type = Numeric
+
+    def get_function(self):
+        return lambda array: pd.Series([len(x.split(" ")) for x in array])
 
 
 # class Like(TransformPrimitive):
--- a/featuretools/primitives/transform_primitive.py
+++ b/featuretools/primitives/transform_primitive.py
@@ -346,7 +346,7 @@ class NumCharacters(TransformPrimitive):
     return_type = Numeric
 
     def get_function(self):
-        return lambda array: pd.Series([len(x) for x in array])
+        return lambda array: pd.Series(array).str.len()
 
 
 class NumWords(TransformPrimitive):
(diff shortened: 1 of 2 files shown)


featuretools/primitives/transform_primitive.py
```diff
@@ -336,6 +337,30 @@ class Weekday(DatetimeUnitBasePrimitive):
     name = "weekday"
 
 
+class NumCharacter(TransformPrimitive):
```
kmax12: Should we name this `NumCharacters` to be consistent with `NumWords` below?

featuretools/primitives/transform_primitive.py
```diff
@@ -336,6 +337,30 @@ class Weekday(DatetimeUnitBasePrimitive):
...
+    """
+    Return the characters in a given string.
+    """
+    name = 'characters'
+    input_types = [Text]
+    return_type = Numeric
+
+    def get_function(self):
+        return lambda array: pd.Series([len(x) for x in array])
```
kmax12: I think you can use more pandas built syntax for this since the `array` variable is going to be a pandas series (please double check this though)

```
In [1]: array = pd.Series(["1","12 2","1212 3"])

In [2]: array.str.len()
Out[2]: 
0    1
1    4
2    6
dtype: int64
```

so, it would just be
```python
def get_function(self):
    return lambda array: array.str.len()
```

featuretools/primitives/transform_primitive.py
```diff
@@ -336,6 +337,30 @@ class Weekday(DatetimeUnitBasePrimitive):
...
+class NumWords(TransformPrimitive):
+    """
+    Returns the words in a given string by counting the spaces.
+    """
+    name = 'numwords'
+    input_types = [Text]
+    return_type = Numeric
+
+    def get_function(self):
+        return lambda array: pd.Series([len(x.split(" ")) for x in array])
```
kmax12: similar to above, you can do 

```python

def get_function(self):
    return lambda array: array.str.split(" ").str.len()
```

kmax12: actually, this is probably better. 

```python
def get_function(self):
    return lambda array: array.str.count(" ")  + 1
```

easier to read and might be [up to 25% faster](https://stackoverflow.com/a/37335656)

PaulHobbs: Those aren't quite equivalent - what if there's some leading or trailing whitespace, or if there's more than one space between characters?

kmax12: I think both pieces of code handle white space and more than one space the same way because they look for just " ".  To do the multiple spaces, the only way I can think of is a regex and I rather keep it simple for now. However, for the trailing or leading white space we could do

```python

def get_function(self):
    return lambda array: array.str.strip().str.count(" ") + 1

```

featuretools/primitives/transform_primitive.py
Seth-Rothschild: Changed

featuretools/primitives/transform_primitive.py
Seth-Rothschild: Array is a `numpy.ndarray`

featuretools/primitives/transform_primitive.py
Seth-Rothschild: Multiple spaces could be taken care of in the str.split(" ") case since they'll show up as empty strings. Could just remove with
```
new_list = [x for x in split_list if x != '']
```
For now, it has been changed to `str.count(" ") + 1`

featuretools/primitives/transform_primitive.py
kmax12: thanks for double checking.  I don't have strong preference, but perhaps do this to avoid the list comprehension

```python 
pd.Series(array).str.len()
```

PR #56, opened 2018-01-02
associative -> commutative

Seth-Rothschild:
Change the primitive attribute [associative](https://en.wikipedia.org/wiki/Associative_property) to the more accurate word [commutative](https://en.wikipedia.org/wiki/Commutative_property).


kmax12:
Thanks for fixing. Merging

PR #57, opened 2018-01-03
(WIP) LatLong type

Seth-Rothschild:
The issue in testing comes from `mock_ds.py` where the mock retail entityset is made with `es.entity_from_csv(entity, ` (line 292). This makes the latlong type in that entityset a string rather than a tuple. The options as I understand them are:

1. Modify `Latitude` and `Longitude` to check if the latlong is a string
2. Modify `entity_from_csv` to convert certain strings to tuples
3. Change the test to do the pandas `_from_csv`, modify the dataframe and then make `entity_from_dataframe`.
4. Leave `Latitude` and `Longitude` with no real tests for now.

My gut is to go with 3. Do you have a preference @kmax12?

### DIFF
--- a/featuretools/tests/testing_utils/mock_ds.py
+++ b/featuretools/tests/testing_utils/mock_ds.py
@@ -88,6 +88,8 @@ def make_ecommerce_files(with_integer_time_index=False, base_path=None, file_loc
                     [i * 3 for i in range(3)] +
                     [np.nan] * 2)
 
+    latlong = list([(values[i], values_2[i]) for i, _ in enumerate(values)])
+
     log_df = pd.DataFrame({
         'id': range(17),
         'session_id': [0] * 5 + [1] * 4 + [2] * 1 + [3] * 2 + [4] * 3 + [5] * 2,
@@ -97,6 +99,7 @@ def make_ecommerce_files(with_integer_time_index=False, base_path=None, file_loc
         'datetime': times,
         'value': values,
         'value_2': values_2,
+        'latlong': latlong,
         'priority_level': [0] * 2 + [1] * 5 + [0] * 6 + [2] * 2 + [1] * 2,
         'purchased': [True] * 11 + [False] * 4 + [True, False],
         'comments': [coke_zero_review()] + ['I loved it'] * 2 +
@@ -227,6 +230,7 @@ def make_variable_types(with_integer_time_index=False):
         'datetime': variable_types.Datetime,
         'value': variable_types.Numeric,
         'value_2': variable_types.Numeric,
+        'latlong': variable_types.LatLong,
         'priority_level': variable_types.Ordinal,
         'purchased': variable_types.Boolean,
         'comments': variable_types.Text
--- a/featuretools/primitives/transform_primitive.py
+++ b/featuretools/primitives/transform_primitive.py
@@ -516,7 +516,7 @@ class Latitude(TransformPrimitive):
     return_type = Numeric
 
     def get_function(self):
-        return lambda array: pd.Series(array[0])
+        return lambda array: pd.Series([x[0] for x in array] )
 
 
 class Longitude(TransformPrimitive):
@@ -529,4 +529,4 @@ class Longitude(TransformPrimitive):
     return_type = Numeric
 
     def get_function(self):
-        return lambda array: pd.Series(array[1])
+        return lambda array: pd.Series([x[1] for x in array])
--- a/featuretools/primitives/transform_primitive.py
+++ b/featuretools/primitives/transform_primitive.py
@@ -14,7 +14,6 @@ from featuretools.variable_types import (
     DatetimeTimeIndex,
     Discrete,
     Id,
-    Index,
     Numeric,
     Ordinal,
     Timedelta,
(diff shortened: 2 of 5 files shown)


kmax12:
@Seth-Rothschild I agree. Let's go with 3

Seth-Rothschild:
Changed `entity_from_csv` to `entity_from_dataframe` in `mock_ds.py` which lets us put an honest tuple into the `make_ecommerce_entityset` function. This makes the test pass for `Latitude` and `Longitude` but induces 4 new entityset type errors. Three seem to deal with compression of the entityset and the fourth is unicode. They are:
1. `test_glob_entityset` in `test_es.py`
2. `test_gzip_entityset` in `test_es.py`
3. `test_gzip_glob_entityset` in `test_es.py` and
4. `test_combine_variables` in `test_pandas_es.py`

@rwedge any idea what's causing these?

featuretools/tests/testing_utils/mock_ds.py
```diff
@@ -285,13 +295,18 @@ def make_ecommerce_entityset(with_integer_time_index=False, base_path=None, save
...
             secondary = time_index['secondary']
-        es.entity_from_csv(entity,
-                           filenames[entity],
-                           index='id',
-                           variable_types=variable_types[entity],
-                           encoding='utf-8',
-                           time_index=ti_name,
-                           secondary_time_index=secondary)
+        
+        df = pd.read_csv(filenames[entity])
```
rwedge: adding `encoding='utf-8'` to this `read_csv` call should fix problem 4.

rwedge:
Errors 1-3 stem from tests for how `entity_from_csv` handles csv paths with glob patterns, csvs compressed with gzip, and a combination of the two situations.  Changing the testing entityset to be loaded in using `pandas.read_csv` instead of  `entity_from_csv` has caused these tests to error.

Taking a step back, if we don't want to support `entity_from_csv` working with all variable types we could remove/deprecate it and have `entity_from_dataframe` be the standard way to create entities.  These test cases would no longer be necessary.

kmax12:
sounds like a good plan to me!

kmax12:
@Seth-Rothschild what do you think about adding these as default primitives in DFS? And if we do, we should probably add `NumCharacters` and `NumWords` as well

Seth-Rothschild:
@kmax12 I think `NumCharacters` and `NumWords` are probably worth adding as defaults but `Latitude`, `Longitude` and `Haversine` might not be. 

A reason to add them would be that latlongs don't show up in a feature matrix, because they're tuples. That being said, by the time a user has made a `LatLong` variable (i.e. combined the two columns and marked the variable type as `LatLong`) they're probably at a stage in processing where they'd want to actively choose the `trans_primitives` and they'd already have a choice of primitives in mind.

kmax12:
got it. what if we just add the 2 text primitives and `Haversine`? I think if someone provides to latlong's it useful to automatically return `Haversine`

Seth-Rothschild:
That sounds like a reasonable plan. Should probably implement this after merging so that I can do both at once, since `NumWords` and `NumCharacters` came after this branch.

Seth-Rothschild:
With the merges there's the new error which stems from latlong1 no longer being an array of tuples in python3. Not sure which change has this breaking, but here's the output in python3 versus python2.

python3 
```
(Pdb) latlong1
array([<map object at 0x10fcf6860>, <map object at 0x10fcf6860>,
       <map object at 0x10fcf6860>, <map object at 0x10fcf6860>,
       <map object at 0x10fcf6860>, <map object at 0x10fcf6860>,
       <map object at 0x10fcf6860>, <map object at 0x10fcf6860>,
       <map object at 0x10fcf6860>, <map object at 0x10fcf6860>,
       <map object at 0x10fcf6860>, <map object at 0x10fcf6860>,
       <map object at 0x10fcf6860>, <map object at 0x10fcf6860>,
       <map object at 0x10fcf6860>, <map object at 0x10fcf6860>,
       <map object at 0x10fcf6860>], dtype=object)
```

python2 
```
(Pdb) latlong1
array([(0.0, 0.0), (5.0, 2.0), (10.0, 4.0), (15.0, 6.0), (20.0, 8.0),
       (0.0, 0.0), (1.0, 1.0), (2.0, 2.0), (3.0, 3.0), (0.0, 0.0),
       (0.0, 0.0), (5.0, 2.0), (0.0, 0.0), (7.0, 3.0), (14.0, 6.0)], dtype=object)
```

kmax12:
in python 3 map is an iterator. look for usages of map and do `list(map(..))`.

featuretools/tests/testing_utils/mock_ds.py
```diff
@@ -301,8 +301,10 @@ def make_ecommerce_entityset(with_integer_time_index=False, base_path=None, save
 
         df = pd.read_csv(filenames[entity], encoding='utf-8')
         if entity is 'log':
-            df['latlong'] = map(lambda x: latlong_unstringify(x), df['latlong'])
-            df['latlong2'] = map(lambda x: latlong_unstringify(x), df['latlong2'])
+            df['latlong'] = list(map(
+                lambda x: latlong_unstringify(x), df['latlong']))
```
kmax12: i think this line can actually be 

`df['latlong'] = df['latlong'].apply(latlong_unstringify)`

kmax12:
can we add the new primitives to the api reference?

kmax12:
This looks good to merge (once Circle CI passes). @Seth-Rothschild can you put up a PR for removing entity_from_csv when you get a chance?

Issue #58, opened 2018-01-04
Problem with `new_entity_secondary_time_index` in dfs


Seth-Rothschild:
For the dataset I'm using, I have
```
es.normalize_entity('coordinates', 'plows', 'truck_name', 
                    make_time_index=True,
                    make_secondary_time_index={'date_fixed': []})
```
so that I can automatically calculate the first and last time that a plow is on the road. If I give the secondary time index a new name, with `new_entity_secondary_time_index`
```
es.normalize_entity('coordinates', 'plows', 'truck_name', 
                    make_time_index=True,
                    make_secondary_time_index={'date_fixed': []},
                    new_entity_secondary_time_index='last_coordinates_time')
```
then I get a `KeyError` in dfs. This seems to come from the line in `_filter_and_sort`:
```
    618             if time_last is not None and not df.empty:
--> 619                 mask = df[secondary_time_index] >= time_last
    620                 second_time_index_columns = self.secondary_time_index[secondary_time_index]
    621                 df.loc[mask, second_time_index_columns] = np.nan
```

Seth-Rothschild:
It looks like `es['plows'].secondary_time_index` stays as `{'date_fixed': 'date_fixed'}` even after `new_entity_secondary_time_index`. 

In fact, `new_entity_secondary_time_index` does
```
            if new_entity_secondary_time_index:
                secondary_df.rename(columns={secondary_time_index:new_entity_secondary_time_index},
                                    inplace=True)
```
which only changes the columns of the underlying dataframe. Setting `new_entity.secondary_time_index` explicitly would fix

Seth-Rothschild:
Closed with #59

PR #59, opened 2018-01-04
Change `normalize_entity` to update `secondary_time_index`

Seth-Rothschild:
For issue #58 the `secondary_time_index` attribute of an entityset isn't properly updated if it's set explicitly with `new_entity_secondary_tiime_index` in `normalize_entity`.  This is *a* fix but there might be a cleaner one.

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -804,8 +804,8 @@ class EntitySet(BaseEntitySet):
 
         new_entity = self.entity_stores[new_entity_id]
         if make_secondary_time_index:
-            values = make_secondary_time_index.values()[0]
-            values.remove(make_secondary_time_index.keys()[0])
+            values = list(make_secondary_time_index.values()[0])
+            values.remove(list(make_secondary_time_index.keys())[0])
             new_dict = {secondary_time_index: values}
 
             new_entity.secondary_time_index = new_dict
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -804,7 +804,7 @@ class EntitySet(BaseEntitySet):
 
         new_entity = self.entity_stores[new_entity_id]
         if make_secondary_time_index:
-            values = list(make_secondary_time_index.values()[0])
+            values = list(make_secondary_time_index.values())[0]
             values.remove(list(make_secondary_time_index.keys())[0])
             new_dict = {secondary_time_index: values}
 
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -759,6 +759,7 @@ class EntitySet(BaseEntitySet):
             if new_entity_secondary_time_index:
                 secondary_df.rename(columns={secondary_time_index: new_entity_secondary_time_index},
                                     inplace=True)
+                secondary_time_index = new_entity_secondary_time_index
             else:
                 new_entity_secondary_time_index = secondary_time_index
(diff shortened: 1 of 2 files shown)


kmax12:
@Seth-Rothschild can you write a test for this?

kmax12:
@Seth-Rothschild thanks for writing the test. this looks good to me to merge. anything concerns?

Seth-Rothschild:
The only thing I'm worried about is the standard format for the secondary time index. Right now it's set explicitly as `{'column_name': 'column_name'}` both in `normalize_entity` and the test, but since it's a dictionary I don't necessarily trust that's the only form it can take.

featuretools/tests/entityset_tests/test_pandas_es.py
```diff
@@ -716,6 +716,18 @@ def test_last_time_index(self, entityset):
         regions_lti = es["regions"].last_time_index.sort_index()
         assert (regions_lti == region_series.sort_index()).all()
 
+    def test_secondary_time_index(self, entityset):
+        es = entityset
+        es.normalize_entity('log', 'values', 'value',
+                            make_time_index=True,
+                            make_secondary_time_index={'datetime': []},
```
bschreck: This should be a dictionary mapping a date column to be used as the secondary time index column, to all the columns that became "known" or were recorded in the data at that time. To test this functionality (which I can see would not get handled properly) you should include an additional column to get added to the secondary time index of the new entity.

featuretools/entityset/entityset.py
```diff
@@ -801,6 +802,8 @@ def normalize_entity(self, base_entity_id, new_entity_id, index,
         self.delete_entity_variables(base_entity_id, additional_variables)
 
         new_entity = self.entity_stores[new_entity_id]
+        if make_secondary_time_index:
+            new_entity.secondary_time_index = {secondary_time_index: [secondary_time_index]}
```
bschreck: This should follow the logic in `BaseEntity.__init__()`. You'll want to append all the variables included in the values of the make_secondary_time_index dict here, except for the time index itself whose name changed and is already in the list. To be complete, we should probably allow for multiple `secondary_time_index`'s too, right? The dictionary format allows an arbitrary number of them. In this case the `new_entity_secondary_time_index` parameter to `normalize_entity()` would need to be a dictionary mapping the column names in the base entity to the ones you want in the new entity.

Seth-Rothschild: So, if I'm understanding correctly, these two lines should be changed to
```
new_entity.secondary_time_index = secondary_time_index or {}
for ti, cols in new_entity.secondary_time_index.items():
    if ti not in cols:
        cols.append(ti)
```
It seems like multiple secondary time index is also asserted against with 
```
            assert len(make_secondary_time_index) == 1, "Can only provide 1 secondary time index"
``` 
in `entityset.py`.

featuretools/tests/entityset_tests/test_pandas_es.py
Seth-Rothschild: Good point. I was using the `secondary_time_index` to create a "time_last" time index, but I'll pull along other columns in the test.

Seth-Rothschild:
@bschreck I think this is is probably clean enough with the latest commit. We might consider making a long term issue to handle multiple secondary time indices, but that seems out of scope for this fix.

bschreck:
Yup I agree!

kmax12:
looks good to merge to me!

PR #64, opened 2018-01-12
Remove Python 3.4 support

bschreck:
Pandas does not support Python 3.4, and it causes problems for out CI validation, so we should not support it.


requirements.txt
```diff
@@ -1,6 +1,6 @@
-numpy
+numpy>=1.14.0
 scipy>=1.0.0
-pandas==0.20.3
+pandas>=0.22.0
```
kmax12: i think we can relax this a bit. any reason not to do `pandas>=0.20.3`?

bschreck:
Assuming this makes it into the next release: D3M’s next release is using 0.22, so it would be nice to have a released version that is compatible with it

kmax12:
if we use `>=0.20.3`, we'd be fine with pandas 0.22, right?

bschreck:
hmm yeah my thought was skipping the bad 0.21 version of pandas, but I think it's fine

PR #65, opened 2018-01-12
Add a mention of default values to agg docstring

### LINKED ISSUE #55: Support default_value in make_agg_primitive
If you are using the `make_agg_primitive` to define a custom primitive, you currently cannot provide the `default_value` to use. This parameter is supported by the lower level class-based definition of primitives, so it should available through the helper method.

Seth-Rothschild:
close #55


kmax12:
thanks. merging

PR #68, opened 2018-01-15
Change : to -> in 4 places in docstrings

### LINKED ISSUE #67: autofunction documentation issues - featuretools.dfs
Hi,

I noticed that a few parameters (``entities`` and ``ignore_variables``) are missing in the documentation for [featuretools.dfs](https://docs.featuretools.com/generated/featuretools.dfs.html#featuretools.dfs). This is probably due to the input types for these parameters being specified as **dict[str: something]** in the docstring for [featuretools.synthesis.dfs](https://github.com/Featuretools/featuretools/blob/master/featuretools/synthesis/dfs.py).

I would recommend reformatting the input type in the docstring by replacing the colon with another character to fix the documentation issues.

Seth-Rothschild:
close #67 

These are very hard to catch automatically. There was the same problem in the documentation for entityset too. I would not be surprised if there are more.

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -36,9 +36,9 @@ class EntitySet(BaseEntitySet):
                 id (str) : unique identifier to associate with this instance
                 verbose (boolean)
 
-                entities (dict[str: tuple(pd.DataFrame, str, str)]): dictionary of
+                entities (dict[str -> tuple(pd.DataFrame, str, str)]): dictionary of
                     entities. Entries take the format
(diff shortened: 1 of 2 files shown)


kmax12:
Did you test rebuild the docs to verify these changes worked?

Seth-Rothschild:
Yep. Most of the docstrings already used -> for dictionary, it just looks like a couple were missed. Here's what `dfs` looks like now:
image

and here's Entityset: 
image

kmax12:
Thanks for fixing. Merging now

PR #69, opened 2018-01-16
More descriptive DFS progress bar

bschreck:
Updated DFS progress bar to display number of features actually constructed, as well features removed at the end due to restrictions on variable types and the drop_contains/drop_exact list.


featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -278,7 +284,10 @@ def filt(f):
 
         new_features.sort(key=lambda f: f.get_depth())
 
+        before_len = len(new_features)
         new_features = self._filter_features(new_features)
+        if verbose and before_len > len(new_features):
```
kmax12: what does this look like when it gets written? 

I like the main change, but not sure it's that helpful to have this print out as things are removed.

bschreck: It actually looks pretty good I think, try running it and check it out.

Basically the progress bar is like:

[45 features] 100 features/s

Eventually the feature building stops and the progress bar stops, then above the progress bar the number removed is written:

Removed 2 features with undesired variable types
Removed 2 features from drop_contains list
[45 features] 100 features/s

kmax12: can you screenshot?

Seth-Rothschild:
Here's a screenshot.
image
The top three lines are new and give some information about what DFS has done. I think that's probably a step in the right direction.

It might be worth modifying some of the words around the pbar itself through `tqdm`. The line
```
1/1 [00:03<00:00,  3.16s/cutoff time]
``` 
probably doesn't give enough information to a new user. It presupposes they know that the 1/1 is counting the cutoff times they've given, and that they don't parse `<`  to mean "elapsed time is less than remaining time". 

How about something like
```
Elapsed: 1:30 | Remaining 4:10 |+-bar-here-----| 
Calculated matrix at 3/12 cutoff times (avg. 30s/calc): 
```

bschreck:
Yup that all makes sense. I didn't test the actual feature calculation step, just the building step (because I noticed what we show for that step right now doesn't make much sense).

Seth-Rothschild:
Seems difficult to get multiple lines into the `bar_format` in tqdm. Here's a single line version
image

bschreck:
Updated to include feature exclusion line inside of pbar:
pbar

Seth-Rothschild:
I'm really enjoying using this.

A small note is that the order/phrasing of the drop contains message is a little confusing:
image
Here there are 18 (12 dropped + 6 excluded) features which aren't calculated. Is it possible to add the 12 into `features_excluded` or to move the "Removed 12 features..." line to after the calculation?

bschreck:
I didn't see that line- should be fixed now

bschreck:
Now with removed pbar from Deep Feature Synthesis.

In progress:
in_progress_pbar

Complete: 
done_pbar

kmax12:
Looks great!

PR #70, opened 2018-01-17
Last time index fixes

rwedge:
Bugfixes addressing several issues when calculating `last_time_index`:

*  Ignoring `last_time_index` instance values taken from one child entity if that parent instance was not present in another child entity  
* Not always choosing the most recent time for an instance's `last_time_index` if the entity had multiple generations of child entities


featuretools/entityset/entityset.py
```diff
@@ -1022,30 +1022,46 @@ def add_last_time_indexes(self):
 
         while len(explored) < len(self.entities):
             entity = queue.pop(0)
-            if entity.id not in children:
+
+            if entity.last_time_index is None:
                 if entity.has_time_index():
-                    entity.set_last_time_index(entity.df[entity.time_index])
-            else:
+                    lti = copy.deepcopy(entity.df[entity.time_index])
```
kmax12: any reason you can't just use `entity.df[entity.time_index].copy()`?

featuretools/entityset/entityset.py
```diff
@@ -1022,30 +1022,46 @@ def add_last_time_indexes(self):
...
+                else:
+                    lti = copy.deepcopy(entity.df[entity.index])
+                    lti[:] = None
+                lti = pd.to_datetime(lti)
```
kmax12: did we require lit to be a datetime before? why do we need to convert it here?

kmax12:
Looks good to me! BTW, great job writing tests for this.

PR #71, opened 2018-01-18
Remove `entity_from_csv`, tests, and lint

Seth-Rothschild:
Removes `entity_from_csv` as discussed in #57


featuretools/entityset/entityset.py
```diff
@@ -4,7 +4,6 @@
 from builtins import range, zip
 from collections import defaultdict
 
-import dask.dataframe as dd
```
kmax12: can we remove dask as a dependency in featuretools now?

featuretools/entityset/entityset.py
```diff
@@ -299,7 +298,8 @@ def add_column(self, entity_id, column_id, column_data, type=None):
         """
         Add variable to entity's dataframe
         """
-        self.entity_stores[entity_id].add_column(column_id, column_data, type=type)
+        self.entity_stores[entity_id].add_column(
```
kmax12: can you undo your line length changes for now? We currently allow lines to go beyond 80 characters in special cases like this one and edits below

featuretools/entityset/entityset.py
Seth-Rothschild: Looks like it. The only other use is in `load_flight` which gets used in the approximate features docs page (but with a limit on the number of rows). We could just put a hard cap on the number of rows and then not require `dask` projectwide?

kmax12:
looks good to me!

PR #77, opened 2018-01-23
Args docstring fixes

### LINKED ISSUE #75: Minor docstring inaccuracies
I came accross these two issues:
1. In [entity_from_dataframe](https://docs.featuretools.com/generated/featuretools.entityset.EntitySet.entity_from_dataframe.html) it says `secondary_time_index` is a string, when it should be a dictionary.
2. In [normalize_entity](https://docs.featuretools.com/generated/featuretools.entityset.EntitySet.normalize_entity.html) there's an undocumented kwarg `variable_types`. 
I'm thinking of using the branch to fix them to also standardize the notation in those two docstrings for lists, dicts and optional arguments to
```
list[str]
(dict[str->list[str]], optional)
```

Seth-Rothschild:
close #75 
Before:
image

After:
image

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -365,11 +365,11 @@ def approximate_features(features, cutoff_time, window, entityset,
             which entities to apply windows to, use a dictionary mapping entity
             id -> Timedelta. If None, all older data is used.
 
-        verbose (Optional(boolean)): Print progress info.
+        verbose (bool, optional): Print progress info.
 
-        profile (Optional(boolean)): Enables profiling if True
+        profile (bool, optional): Enables profiling if True
 
-        save_progress (Optional(str)): path to save intermediate computational results
+        save_progress (str, optional): path to save intermediate computational results
     '''
     if verbose:
         logger.info("Approximating features...")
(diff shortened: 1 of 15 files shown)


Seth-Rothschild:
This is probably ready for other eyes now. The changes were of 4 types:
1. `boolean -> bool`, `dict{x:y} -> dict[x -> y]`, `Optional[x] -> (x, optional)`
2. Begin argument descriptions with a capital letter, end with a period, remove shorthand in descriptions.
3. :class:\`.Variable\` -> Variable in arguments. The :class: and \` were breaking typesetting (see above)
4. Minor updates where I was sure the docstring was wrong.

There are lots of places for 4 where I was unsure, so this could definitely use a second set of eyes. In particular:
+ `variable_types` is always documented as `dict[str -> dict[str -> Variable]]`, but in `entity_from_dataframe` I know that the input is actually `dict[str -> Variable]`. Are there other places where the correct input is the latter? Which one is `normalize_entity`? (example in [Entity](https://docs.featuretools.com/generated/featuretools.Entity.html?highlight=entity))
+ Entities and relationships are almost always documented by their underlying structure rather than a class. Is that correct? (example in [dfs](https://docs.featuretools.com/generated/featuretools.dfs.html?highlight=dfs#featuretools.dfs))

kmax12:
Looks good to me! Thanks!

PR #78, opened 2018-01-24
Adding non-feature columns to calculated feature matrix

### LINKED ISSUE #74: Passing additional data through calculate feature matrix
It would be useful to be able to pass labels or other extra columns to `calculate_feature_matrix` and have them be added to the feature matrix that is returned.  Proposed implementation: include these additional columns in the `cutoff_time` argument.  The first two columns of `cutoff_time` will always be the instance ids and cutoff times.  Any extra columns will be treated as additional data columns and included in the final feature matrix.

rwedge:
Closes #74 

Extra columns in the  `cutoff_time` argument to `calculate_feature_matrix`  are included as additional columns in the resulting feature matrix.  When taking advantage of this feature the order of columns is important: the first two columns of `cutoff_time` must be the instance ids and and time columns (the order of these first two columns doesn't matter).

This PR also fixes a bug that would occur when using the `approximate` option for `calculate_feature_matrix` and every aggregation feature could be approximated. The time index values for the calculated feature matrix (see the `cutoff_time_in_index` option) were the times the aggregation features were approximated at, not the times specified in `cutoff_time`.


docs/source/automated_feature_engineering/handling_time.rst
```diff
@@ -71,8 +71,18 @@ There is one row in the feature matrix corresponding to a row in ``cutoff_times`
                                       cutoff_time_in_index=True)
     feature_matrix
 
+It is often the case that we want our labels in our calculated feature matrix so that the ordering is consistent between the labels and the rows of the feature matrix. However, adding labels to the initial dataframe means that you would have to explicitly prohibit ``dfs`` from building features with that column. To bypass this, we can provide additional columns to cutoff times which will be added directly the feature matrix. While the two columns will be used as an index and cutoff time, any additional columns will appear as features in the resulting feature matrix. 
 
+.. ipython:: python
+
+    cutoff_times['labels'] = pd.Series([0, 0, 1, 0, 1])
```
kmax12: it's more common to see this called just `label` with no s

docs/source/automated_feature_engineering/handling_time.rst
```diff
@@ -71,8 +71,18 @@ There is one row in the feature matrix corresponding to a row in ``cutoff_times`
                                       cutoff_time_in_index=True)
     feature_matrix
 
+It is often the case that we want our labels in our calculated feature matrix so that the ordering is consistent between the labels and the rows of the feature matrix. However, adding labels to the initial dataframe means that you would have to explicitly prohibit ``dfs`` from building features with that column. To bypass this, we can provide additional columns to cutoff times which will be added directly the feature matrix. While the two columns will be used as an index and cutoff time, any additional columns will appear as features in the resulting feature matrix.
```
kmax12: do you think it is explicit enough that it is the _first_ two columns that get used and that order matters

Seth-Rothschild: My understanding is that the first two columns get used but the order of those columns doesn't matter. How about the phrasing:
"While the first two columns will be used as index and cutoff time regardless of their order in the dataframe, any additional columns will appear as features in the resulting matrix."

docs/source/automated_feature_engineering/handling_time.rst
```diff
@@ -71,18 +71,18 @@ There is one row in the feature matrix corresponding to a row in ``cutoff_times`
...
 
-It is often the case that we want our labels in our calculated feature matrix so that the ordering is consistent between the labels and the rows of the feature matrix. However, adding labels to the initial dataframe means that you would have to explicitly prohibit ``dfs`` from building features with that column. To bypass this, we can provide additional columns to cutoff times which will be added directly the feature matrix. While the two columns will be used as an index and cutoff time, any additional columns will appear as features in the resulting feature matrix. 
+It is often the case that we want our labels in our calculated feature matrix so that the ordering is consistent between the labels and the rows of the feature matrix. However, adding labels to the initial dataframe means that you would have to explicitly prohibit ``dfs`` from building features with that column. To bypass this, we can provide additional columns to cutoff times which will be added directly the feature matrix. While the first two columns will be used as an index and cutoff time regardless of their order in the dataframe, any additional columns will appear as features in the resulting feature matrix.
```
kmax12: we may support mixing the order up, but let’s document instance id first, followed by cutoff time.

kmax12:
my only thought is that we should have a way to provide labels in the case where you using a single cutoff time and the `instance_ids` parameter. right now, the work around for that is to simply create a dataframe cutoff times where you repeat the cutoff time. 

because there is a work around for that situation, this is good to merge.

PR #84, opened 2018-02-05
Use multiple cores to calculate feature matrix [WIP]

rwedge:
This PR adds two more optional parameters to `calculate_feature_matrix`, `njobs` and `chunk_size`.  The `chunk_size` parameter sets the limit for how many rows of the feature matrix can be calculated at once, either as a percentage of the entire matrix or a finite number of rows.  Setting this parameter to `None` will remove this limit.  The `njobs` parameter sets the number of processes spawned to calculate the feature matrix.

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -186,9 +186,9 @@ def calculate_feature_matrix(features, cutoff_time=None, instance_ids=None,
     for _, group in grouped:
         if group.shape[0] > num_per_chunk:
             for i in range(0, group.shape[0], num_per_chunk):
-                chunks.append((_, group.iloc[i:i + num_per_chunk]))
+                chunks.append(group.iloc[i:i + num_per_chunk])
         else:
-            chunks.append((_, group))
+            chunks.append(group)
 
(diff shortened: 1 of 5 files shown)


featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -172,30 +176,74 @@ def calculate_feature_matrix(features, cutoff_time=None, instance_ids=None,
...
+    num_per_chunk = calc_num_per_chunk(chunk_size, cutoff_time.shape)
+
+    # initialize chunks
+    chunks = []
+    for _, group in grouped:
+        if group.shape[0] > num_per_chunk:
+            for i in range(0, group.shape[0], num_per_chunk):
+                chunks.append((_, group.iloc[i:i + num_per_chunk]))
+        else:
+            chunks.append((_, group))
```
bschreck: Can't you just make chunks a list of [group] instead of [(_, group)], and then just parallelize calculate_batch directly instead of having the calculate_chunk wrapper function? It seems like calculate_chunk doesn't do anything

kmax12:
working on this in a different branch. @rwedge will put up a new PR when ready

PR #86, opened 2018-02-07
Forbid stacking of expanding features in DFS

### LINKED ISSUE #85: DFS errors by stacking on expanding features
Supplying the NMostCommon and NUnique features to DFS errors, as it stacks NUnique on NMostCommon, which is an expanding feature that returns lists as output. The resulting values supplied to NUnique look like:
```
pd.Series([['v1', 'v2', v3'], ['v2, 'v3', 'v1']])
```
We don't have a good way of handling the stacking of expanding features in pandas_backend, so instead we should just not allow them in DFS.

bschreck:
This fixes #85 

Adds 2 lines to DFS to prevent stacking on top of expanding features, and provides a test to make sure this does not happen. See #85 for reasoning.

### DIFF
--- a/featuretools/synthesis/deep_feature_synthesis.py
+++ b/featuretools/synthesis/deep_feature_synthesis.py
@@ -736,6 +736,10 @@ def check_stacking(primitive, input_types):
         else:
             continue
         return False
+    # Don't stack on expanding features for now.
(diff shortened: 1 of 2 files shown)


bschreck:
using #94 to resolve this instead.

PR #91, opened 2018-02-13
Handle Transform features that need access to all values of entity

bschreck:
Adds logic to pass in all values of an entity before cutoff to features that need them.

The problem:
Features like Percentile and CumMean/CumMax/etc. need access to the all the values the entity before the cutoff time, rather than only specified instance ids. Normal Transform features are one-to-one mappings, where the output for specific instance only depends on a single input value for that instance. On the contrary, features like Percentile use all the available data to determine the rank of a particular value. Currently, they are only provided with a subset of the data- the subset that is linked to the set of specified instance ids on target entity.

This PR creates an additional `entity_frames` object inside of `pandas_backend.py`, which contains all the data before the cutoff time. It adds logic to determine whether to pass in this object of the normal `entity_frames` object to each feature calculation function. It also decides where to place the output of these functions.

Lastly, it makes sure to only extract the relevant columns from the entityset using `EntitySet.get_pandas_data_slice`. Previously, we simply used all the columns (which is a waste of memory).

### DIFF
--- a/featuretools/computational_backends/feature_tree.py
+++ b/featuretools/computational_backends/feature_tree.py
@@ -58,9 +58,8 @@ class FeatureTree(object):
         self._order_feature_groups()
 
     def _find_necessary_columns(self):
-        # TODO: If only_if_needs_all_values is False,
-        # can try to remove columns that are only used in the
-        # intermediate large_entity_frames
+        # TODO: Can try to remove columns that are only used in the
+        # intermediate large_entity_frames from self.necessary_columns
         # TODO: Can try to only keep Id/Index/DatetimeTimeIndex if actually
         # used for features
         self.necessary_columns = defaultdict(set)
@@ -72,7 +71,7 @@ class FeatureTree(object):
                                             variable_types.Id,
                                             variable_types.TimeIndex))]
             self.necessary_columns[eid] |= set(index_cols)
-        self.necessary_columns_for_all_values = copy.copy(self.necessary_columns)
+        self.necessary_columns_for_all_values_features = copy.copy(self.necessary_columns)
 
         identity_features = [f for f in self.all_features
                              if isinstance(f, IdentityFeature)]
@@ -80,9 +79,9 @@ class FeatureTree(object):
         for f in identity_features:
             self.necessary_columns[f.entity.id].add(f.variable.id)
             if self._needs_all_values(f):
-                self.necessary_columns_for_all_values[f.entity.id].add(f.variable.id)
+                self.necessary_columns_for_all_values_features[f.entity.id].add(f.variable.id)
         self.necessary_columns = {eid: list(cols) for eid, cols in self.necessary_columns.items()}
-        self.necessary_columns_for_all_values = {eid: list(cols) for eid, cols in self.necessary_columns_for_all_values.items()}
+        self.necessary_columns_for_all_values_features = {eid: list(cols) for eid, cols in self.necessary_columns_for_all_values_features.items()}
 
     def get_all_features(self):
         all_features = []
--- a/featuretools/computational_backends/feature_tree.py
+++ b/featuretools/computational_backends/feature_tree.py
@@ -232,4 +232,3 @@ def _get_ftype_string(f):
         return "identity"
     else:
         raise UnknownFeature("{} feature unknown".format(f.__class__))
-
--- a/featuretools/computational_backends/feature_tree.py
+++ b/featuretools/computational_backends/feature_tree.py
@@ -1,10 +1,10 @@
+import copy
 import itertools
 import logging
 from builtins import object
 from collections import defaultdict
 
-from ..utils import gen_utils as utils
-
+from featuretools import variable_types
 from featuretools.exceptions import UnknownFeature
 from featuretools.primitives import (
     AggregationPrimitive,
@@ -12,6 +12,8 @@ from featuretools.primitives import (
     IdentityFeature,
     TransformPrimitive
 )
+
+from ..utils import gen_utils as utils
 
 logger = logging.getLogger('featuretools.computational_backend')
 
@@ -29,21 +31,58 @@ class FeatureTree(object):
         self.feature_hashes = set([f.hash() for f in features])
 
         all_features = {f.hash(): f for f in features}
-        feature_deps = {}
+        feature_dependencies = {}
+        feature_dependents = defaultdict(set)
         for f in features:
             deps = f.get_deep_dependencies(ignored=ignored)
-            feature_deps[f.hash()] = deps
+            feature_dependencies[f.hash()] = deps
             for dep in deps:
+                feature_dependents[dep.hash()].add(f.hash())
                 all_features[dep.hash()] = dep
-                feature_deps[dep.hash()] = dep.get_deep_dependencies(
+                subdeps = dep.get_deep_dependencies(
                     ignored=ignored)
+                feature_dependencies[dep.hash()] = subdeps
+                for sd in subdeps:
+                    feature_dependents[sd.hash()].add(dep.hash())
+        # turn values which were hashes of features into the features themselves
+        # (note that they were hashes to prevent checking equality of feature objects,
+        #  which is not currently an allowed operation)
+        self.feature_dependents = {fhash: [all_features[dhash] for dhash in feature_dependents[fhash]]
+                                   for fhash, f in all_features.items()}
+        self.feature_dependencies = feature_dependencies
         self.all_features = list(all_features.values())
-        self.feature_deps = feature_deps
-        self.feature_dependents = self._build_dependents()
+        self._find_necessary_columns()
 
         self._generate_feature_tree(features)
         self._order_entities()
         self._order_feature_groups()
+
+    def _find_necessary_columns(self):
+        # TODO: If only_if_needs_all_values is False,
+        # can try to remove columns that are only used in the
+        # intermediate large_entity_frames
+        # TODO: Can try to only keep Id/Index/DatetimeTimeIndex if actually
+        # used for features
+        self.necessary_columns = defaultdict(set)
+        entities = set([f.entity.id for f in self.all_features])
+        for eid in entities:
+            entity = self.entityset[eid]
+            index_cols = [v.id for v in entity.variables
+                          if isinstance(v, (variable_types.Index,
+                                            variable_types.Id,
+                                            variable_types.TimeIndex))]
+            self.necessary_columns[eid] |= set(index_cols)
+        self.necessary_columns_for_all_values = copy.copy(self.necessary_columns)
+
+        identity_features = [f for f in self.all_features
+                             if isinstance(f, IdentityFeature)]
+
+        for f in identity_features:
+            self.necessary_columns[f.entity.id].add(f.variable.id)
+            if self._needs_all_values(f):
+                self.necessary_columns_for_all_values[f.entity.id].add(f.variable.id)
+        self.necessary_columns = {eid: list(cols) for eid, cols in self.necessary_columns.items()}
+        self.necessary_columns_for_all_values = {eid: list(cols) for eid, cols in self.necessary_columns_for_all_values.items()}
 
     def get_all_features(self):
         all_features = []
@@ -82,7 +121,7 @@ class FeatureTree(object):
             # this entity. If any of these are themselves top-level features, add
             # their entities as dependencies of the current entity.
             deps = {g.hash(): g for f in features
-                    for g in self.feature_deps[f.hash()]}
+                    for g in self.feature_dependencies[f.hash()]}
             for d in deps.values():
                 _, num_forward = self.entityset.find_path(e, d.entity.id,
                                                           include_num_forward=True)
@@ -162,7 +201,7 @@ class FeatureTree(object):
     def _build_dependents(self):
         feature_dependents = defaultdict(set)
         for f in self.all_features:
-            dependencies = self.feature_deps[f.hash()]
+            dependencies = self.feature_dependencies[f.hash()]
             for d in dependencies:
                 feature_dependents[d.hash()].add(f)
         return feature_dependents
--- a/featuretools/computational_backends/feature_tree.py
+++ b/featuretools/computational_backends/feature_tree.py
@@ -198,14 +198,6 @@ class FeatureTree(object):
 
         return list(features.values()), out
 
-    def _build_dependents(self):
-        feature_dependents = defaultdict(set)
-        for f in self.all_features:
-            dependencies = self.feature_dependencies[f.hash()]
-            for d in dependencies:
-                feature_dependents[d.hash()].add(f)
-        return feature_dependents
-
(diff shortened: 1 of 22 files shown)


Seth-Rothschild:
Trying to use this with custom primitives now: what's the expected workflow? I thought that you might be able to set `use_previous` by doing
```
def primitive1(value):
    numeric = pd.Series(value)
    return (numeric - numeric.shift(1))

Prim1 = make_trans_primitive(primitive1,
                            input_types=[vtypes.Numeric],
                            cls_attributes={'use_previous': True},
                            return_type=vtypes.Numeric)
```
but that doesn't seem to be the way to access calculating using additional data.

bschreck:
@Seth-Rothschild the way to do it would be `cls_attributes={'needs_all_values': True}`. You just have the wrong variable name.

featuretools/computational_backends/feature_tree.py
```diff
@@ -25,16 +25,21 @@ def __init__(self, entityset, features, ignored=None):
...
             deps = f.get_deep_dependencies(ignored=ignored)
             feature_deps[f.hash()] = deps
             for dep in deps:
                 all_features[dep.hash()] = dep
-                feature_deps[dep.hash()] = dep.get_deep_dependencies(ignored=ignored)
+                feature_deps[dep.hash()] = dep.get_deep_dependencies(
+                    ignored=ignored)
         self.all_features = list(all_features.values())
         self.feature_deps = feature_deps
+        self.feature_dependents = self._build_dependents()
```
kmax12: i think this is the inverse of `self.feature_deps`, but the variable name doesn't make it very clear since it is the same

kmax12: why break this out into a different function? can it be combined into the logic above that calculates `self.feature_deps`?

featuretools/computational_backends/pandas_backend.py
```diff
@@ -151,15 +169,30 @@ def calculate_all_features(self, instance_ids, time_last,
                 if not self.entityset.find_backward_path(start_entity_id=filter_eid,
                                                          goal_entity_id=eid):
                     entity_frames[eid] = eframes_by_filter[eid][eid]
+                    if (large_eframes_by_filter is not None and
+                            eid in large_eframes_by_filter and eid in large_eframes_by_filter[eid]):
+                        large_entity_frames[eid] = large_eframes_by_filter[eid][eid]
```
kmax12: what is this line about? seems a bit weird

featuretools/computational_backends/pandas_backend.py
```diff
@@ -151,15 +169,30 @@ def calculate_all_features(self, instance_ids, time_last,
...
                     if verbose:
 
                         pbar.set_postfix({'running': 0})
 
-                    handler = self._feature_type_handler(group[0])
-                    handler(group, entity_frames)
+                    test_feature = group[0]
+
+                    input_frames, output_frames = \
+                        self._determine_input_output_frames(
```
kmax12: seems like this logic can be done in a simpler way to avoid implementing a function that only gets called once. 

```
input_frames = large_entity_frames
if self.feature_tree.needs_all_values_differentiator(test_feature) == 'normal_no_dependent':
    input_frames = entity_frames
```

then you can put the output frame logic in `existing _place_in_output_frames` function you've defined.

featuretools/computational_backends/pandas_backend.py
```diff
@@ -210,6 +243,72 @@ def generate_default_df(self, instance_ids, extra_columns=None):
                     default_df[c] = [np.nan] * len(instance_ids)
         return default_df
 
+    def _find_necessary_columns(self, only_if_needs_all_values=False):
```
kmax12: is `only_if_needs_all_values` an optimization or is it required for this to work? a bit hard to follow its role. in general, I find it confusing when a function has a single boolean parameter that controls behavior like this unless it is very obvious what it does. as far as I can tell, it there is only one place in the code where we set it to true. if that is the case, I'd prefer if the optimization happens more explicitly outside the function. makes the function itself more readable and maintainable.

featuretools/computational_backends/feature_tree.py
bschreck: we would have to change the above logic to accommodate it, since we rely on get_deep_dependencies() but we don't have an equivalent get_deep_dependents() method. It would just have to be a bit more complicated as a queue/search/while loop instead of just the two for loops. But yeah might make it look more uniform

bschreck: i didn't want to change feature_deps when I was first doing it for fear it was used elsewhere, but now that everythign works it would be better to name it feature_dependencies

featuretools/entityset/entityset.py
```diff
@@ -256,6 +272,7 @@ def query_entity_by_values(self, entity_id, instance_vals, variable_id=None,
                                       variable_id=variable_id,
                                       columns=columns,
                                       time_last=time_last,
+                                      training_window=training_window,
```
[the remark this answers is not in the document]
bschreck: I think they weren't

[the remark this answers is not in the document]
bschreck: should write a test to confirm

featuretools/computational_backends/feature_tree.py
bschreck: Nevermind i think i can make it work

featuretools/entityset/entityset.py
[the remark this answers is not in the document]
bschreck: We actually never use this function except in tests... We always call the one on the Entity itself (`Entity.query_by_values`). I think we should just delete this function and change those tests to the one we actually use in EntitySet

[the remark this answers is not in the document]
bschreck: Changed in latest commit. Let me know what you think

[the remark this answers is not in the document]
kmax12: yep, let's remove if it's unused

rwedge:
Codecov has `feature_tree.get_all_features` not being hit by the test cases and it doesn't look like it's called by other parts of featuretools.  Is this a method we want to keep?

featuretools/computational_backends/pandas_backend.py
```diff
@@ -151,15 +171,48 @@ def calculate_all_features(self, instance_ids, time_last,
...
+
+                    input_frames = large_entity_frames
+                    if needs_all_values_type == "normal_no_dependent":
+                        input_frames = entity_frames
+
+                    handler = self._feature_type_handler(test_feature)
+                    result_frame = handler(group, input_frames)
+
+                    output_frames = []
+                    if needs_all_values_type.startswith('dependent'):
```
kmax12: I'd make this more explicit like the other conditionals around it

```
if needs_all_values_type in ["dependent", "dependent_and_output"]:
    pass
```

featuretools/computational_backends/feature_tree.py
```diff
@@ -25,21 +27,62 @@ def __init__(self, entityset, features, ignored=None):
...
-                feature_deps[dep.hash()] = dep.get_deep_dependencies(ignored=ignored)
+                subdeps = dep.get_deep_dependencies(
+                    ignored=ignored)
+                feature_dependencies[dep.hash()] = subdeps
+                for sd in subdeps:
+                    feature_dependents[sd.hash()].add(dep.hash())
+        # turn values which were hashes of features into the features themselves
+        # (note that they were hashes to prevent checking equality of feature objects,
+        #  which is not currently an allowed operation)
+        self.feature_dependents = {fhash: [all_features[dhash] for dhash in feature_dependents[fhash]]
```
kmax12: why do you iterate over the keys in `all_features` rather than `feature_dependents`?

bschreck: If a feature has no dependencies, it won't be put in feature_dependents by the above code. But, looking at it now, I can just explicitly add those in the for loop.

bschreck: hmm actually that doesn't make much sense because it's a defaultdict, so a feature with no deps will get treated as an empty value in the defaultdict anyway.

featuretools/computational_backends/pandas_backend.py
```diff
@@ -95,13 +97,28 @@ def calculate_all_features(self, instance_ids, time_last,
...
                                                  index_eid=self.target_eid,
                                                  instances=instance_ids,
+                                                 entity_columns=necessary_columns,
                                                  time_last=time_last,
                                                  training_window=training_window,
                                                  verbose=verbose)
+        large_eframes_by_filter = None
+        if any([f.needs_all_values for f in self.feature_tree.all_features]):
+            large_necessary_columns = self.feature_tree.necessary_columns_for_all_values_features
+            all_instances = self.entityset[self.target_eid].get_all_instances()
```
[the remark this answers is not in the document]
bschreck: workaround for issue #93. can merge with master and remove

featuretools/computational_backends/feature_tree.py
```diff
@@ -153,9 +197,44 @@ def _get_feature_depths(self, entity_id):
...
+                return True
+        return False
+
+    def _is_output_feature(self, f):
+        return f.hash() in self.feature_hashes
 
 # These functions are used for sorting and grouping features
 
+    def needs_all_values_differentiator(self, f):
+        if self._dependent_needs_all_values(f) and self._is_output_feature(f):
```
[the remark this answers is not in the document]
bschreck: yeah i agree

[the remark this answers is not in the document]
bschreck: i think they're more to help me think while I go about solving the problem, and makes sense to review them at final solution

featuretools/primitives/cum_transform_feature.py
```diff
@@ -19,6 +19,7 @@ class CumFeature(TransformPrimitive):
     allow_where = True
     agg_feature = None
     rolling_function = True
+    needs_all_values = True
```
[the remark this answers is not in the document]
bschreck: What about something like "cross_entity", or "spans_entity"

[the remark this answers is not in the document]
bschreck: "full_entity_transform"

featuretools/computational_backends/feature_tree.py
```diff
@@ -153,9 +198,53 @@ def _get_feature_depths(self, entity_id):
...
+        is_output = f.hash() in self.feature_hashes
+        # If a dependent feature requires all the instance values
+        # of the associated entity, then we need to calculate this
+        # feature on all values
+        # If the feature is one in which the user requested as
+        # an output (meaning it's part of the input feature list
+        # to calculate_feature_matrix), then we also need
+        # to subselect the output based on the desired instance ids
+        # and place in the return data frame.
+        if self._dependent_needs_all_values(f) and is_output:
```
kmax12: i think you can refactor this to get rid of extra functions

```
dependent_needs_all_values = False
for d in self.feature_dependents[f.hash()]:
        if d.needs_all_values:
            dependent_needs_all_values = True
needs_all_values = f.needs_all_values or dependent_needs_all_values
```
then use variables below

featuretools/computational_backends/pandas_backend.py
```diff
@@ -10,9 +10,9 @@
 
 import numpy as np
 import pandas as pd
+# progress bar
```
kmax12: delete?

featuretools/computational_backends/pandas_backend.py
```diff
@@ -25,9 +25,11 @@
     IdentityFeature,
     TransformPrimitive
 )
-# progress bar
 from featuretools.utils.gen_utils import make_tqdm_iterator
 
+# featuretools
```
kmax12: delete?

featuretools/computational_backends/pandas_backend.py
```diff
@@ -151,15 +170,48 @@ def calculate_all_features(self, instance_ids, time_last,
                 if not self.entityset.find_backward_path(start_entity_id=filter_eid,
                                                          goal_entity_id=eid):
                     entity_frames[eid] = eframes_by_filter[eid][eid]
+                    # precalculated features will only be placed in entity_frames,
```
kmax12: is it possible to have to have precalculated a `needs_all_values` feature? do we need to add code to handle that case?

featuretools/computational_backends/pandas_backend.py
```diff
@@ -151,15 +170,48 @@ def calculate_all_features(self, instance_ids, time_last,
...
+                    if needs_all_values_type != "dependent":
+                        # input was selected (small) set of instances, or
+                        # feature itself is an output,
+                        # so need to place result
+                        # in selected (small) output frame
+                        output_frames.append(entity_frames)
+
+                    for frames in output_frames:
+                        index = frames[entity_id].index
+                        frames[entity_id] = result_frame.loc[index]
```
kmax12: you might want to use `result_frame.reindex()` here. you want to if some of the index values are going to be in result_frame.

featuretools/primitives/cum_transform_feature.py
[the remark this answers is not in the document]
kmax12: `historical_transform`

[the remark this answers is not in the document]
kmax12: I like the idea of `full_entity_transform`. any thoughts on other versions that don't repeat transform?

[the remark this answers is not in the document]
bschreck:
```
applied_to_full_entity
applied_to_full_data
```?

File: featuretools/computational_backends/feature_tree.py [hunk 16]
```diff
@@ -25,28 +27,70 @@ def __init__(self, entityset, features, ignored=None):
...
+        feature_dependents = defaultdict(set)
         for f in features:
             deps = f.get_deep_dependencies(ignored=ignored)
-            feature_deps[f.hash()] = deps
+            feature_dependencies[f.hash()] = deps
             for dep in deps:
+                feature_dependents[dep.hash()].add(f.hash())
                 all_features[dep.hash()] = dep
-                feature_deps[dep.hash()] = dep.get_deep_dependencies(ignored=ignored)
+                subdeps = dep.get_deep_dependencies(
```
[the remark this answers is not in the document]
bschreck: get_deep_dependencies() gets all the dependencies recursively, so features + each one's deep dependencies defines every feature. To build the dependents dict, for each one of those, we need all of its dependents, not dependencies. deps is every dependency of the top level features, and subdeps is every dependency of all the other ones. So deps also contains all the dependents of all the subdeps, and top level features are all the other dependents for all the rest. kind of tricky to wrap your head around

featuretools/computational_backends/pandas_backend.py
bschreck: i thought that I considered all cases before but I did miss the case when you have a `needs_all_values` feature that depends on an approximated/precalculated feature. This will need to be reworked.

kmax12:
Looks great. Thanks for the work on this!

PR #94, opened 2018-02-17
Infer expanding features using ListVariable type

### LINKED ISSUE #85: DFS errors by stacking on expanding features

bschreck:
Resolves #85 by introducing a new ListVariable type (as well as ListDiscrete and ListNumeric). Expanding features now must state these types as their `return_type`, and `PrimitiveBase` can now infer if a feature is expanding by checking the computed `variable_type` property.

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -170,7 +170,7 @@ Location Transform Primitives
 *****************************
 .. autosummary::
    :toctree generated/
-   
+
    Latitude
    Longitude
    Haversine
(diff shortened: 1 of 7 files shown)


kmax12:
we're handing this in an upcoming PR. closing this one for now

PR #95, opened 2018-02-17
Better boolean type conversion

### LINKED ISSUE #79: Boolean type conversion does not get called
If a user supplies specific `variable_types` for a Boolean variable that has an underlying "object" dtype, and that include type information as kwargs, the associated logic in `Entity.entityset_convert_variable_type` does not get called. This is because Booleans  are a subclass of Discrete, and the "object" dtype is in "_categorical_types".

Relevant code is in `Entity.convert_variable_types()`.

What this means is that if I have a dataframe with a string column that I call a Boolean:

```
df = pd.DataFrame({'boolean': ['y', 'n', 'y'], 'id': [0,1,2]})
variable_types .= {'boolean': (ft.variable_types.Boolean, {'true_val': 'y', 'false_val': 'n'})}
es.entity_from_dataframe('test', df, index='id', variable_types=variable_types)
```

The column remains a string, does not get converted to a Boolean, and so features like `PercentTrue` fail because they expect boolean inputs.

bschreck:
Resolves #79 and does better inferring of Boolean types by checking if nunique() of the input column is 2.

### DIFF
--- a/featuretools/entityset/entity.py
+++ b/featuretools/entityset/entity.py
@@ -18,6 +18,7 @@ from featuretools.utils.wrangle import _check_timedelta
 logger = logging.getLogger('featuretools.entityset')
 
 _numeric_types = vtypes.PandasTypes._pandas_numerics
+_boolean_types = vtypes.PandasTypes._pandas_booleans
 _categorical_types = [vtypes.PandasTypes._categorical]
 _datetime_types = vtypes.PandasTypes._pandas_datetimes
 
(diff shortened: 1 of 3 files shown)


featuretools/entityset/entity.py
```diff
@@ -365,12 +370,16 @@ def infer_variable_types(self, ignore=None, link_vars=None):
...
-                        sample = df[variable].sample(min(10000, df[variable].nunique()))
-                        avg_length = sample.str.len().mean()
-                        if avg_length > 50:
-                            inferred_type = vtypes.Text
+                        nunique = df[variable].dropna().nunique()
+                        if nunique == 2:
+                            inferred_type = vtypes.Boolean
                         else:
-                            inferred_type = vtypes.Categorical
+                            sample = df[variable].sample(min(10000, nunique))
```
kmax12: can you explain why you do `min(10000, nunique)`? Say, `nunique == 5`, wouldn't that be too small of a sample?

bschreck: no because that would be all of the values... so not a sample. However, if nunique is that small then we should just default to Categorical over Text. So I think there should be another conditional for nunique < some heuristic

rwedge:
Closing this PR for now, I think the degree to which we support inferring boolean types could use discussion before implementing

PR #96, opened 2018-02-17
EntitySet.concat() reindexes relationships

### LINKED ISSUE #83: EntitySet Concat() does not reindex data
EntitySets store an internal dictionary mapping the indexes that connect entities across relationships. This speeds up computation time by enabling quicker joins, much like what a database  does. To create these indexes, inside of `EntitySet.add_relationship`, we internally call:

```
self.index_data(relationship)
```

However, when we concat two EntitySets like this:
```
es_updated = es_old.concat(es_new)
```

We update the internal dataframes of `es_old` but do not update the indexes. This affects `calculate_feature_matrix` down the road not just as a performance hit, but also potentially in the actual feature values.

An easy fix is to add the following to `EntitySet.concat`:
```
for r in es.relationships:
    es.index_data(r)
```

bschreck:
Resolves #83

### DIFF
--- a/featuretools/entityset/base_entity.py
+++ b/featuretools/entityset/base_entity.py
@@ -3,7 +3,6 @@ from __future__ import print_function
 import logging
 from builtins import map
 
-import numpy as np
 import pandas as pd
 from past.builtins import basestring
 
--- a/featuretools/entityset/base_entity.py
+++ b/featuretools/entityset/base_entity.py
@@ -138,7 +138,7 @@ class BaseEntity(FTBase):
                         # We want to check that all the elements of the lists of instances
                         # for each relationship are the same in both entities being
                         # checked for equality, but don't care about the order.
-                        if not (np.sort(related) == np.sort(other.indexed_by[v][i])).all():
+                        if not set(related) == set(other.indexed_by[v][i]):
                             return False
             return True
 
--- a/featuretools/entityset/base_entity.py
+++ b/featuretools/entityset/base_entity.py
@@ -3,10 +3,9 @@ from __future__ import print_function
 import logging
 from builtins import map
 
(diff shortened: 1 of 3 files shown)


featuretools/entityset/base_entity.py
```diff
@@ -124,6 +125,17 @@ def __eq__(self, other, deep=False):
...
+            if self.indexed_by is None and other.indexed_by is not None:
+                return False
+            else:
+                for v, index_map in self.indexed_by.items():
+                    if v not in other.indexed_by:
+                        return False
+                    for i, related in index_map.items():
+                        if i not in other.indexed_by[v]:
+                            return False
+                        if not (np.sort(related) == np.sort(other.indexed_by[v][i])).all():
```
kmax12: can you leave a comment in the code what this line is checking?

bschreck: done

featuretools/entityset/base_entity.py
```diff
@@ -134,6 +134,10 @@ def __eq__(self, other, deep=False):
                     for i, related in index_map.items():
                         if i not in other.indexed_by[v]:
                             return False
+                        # indexed_by maps instances of two entities together by lists
+                        # We want to check that all the elements of the lists of instances
+                        # for each relationship are the same in both entities being
+                        # checked for equality, but don't care about the order.
                         if not (np.sort(related) == np.sort(other.indexed_by[v][i])).all():
```
kmax12: thanks for the comment. can this line just be `set(related) != set(other.indexed_by[v][i])`?

bschreck: yeah that should work too

kmax12:
Looks good. merging

PR #97, opened 2018-02-17
Fix related_instances() to respect time_last when instance_ids not provided

### LINKED ISSUE #93: EntitySet._related_instances doesn't respect time_last if no instances provided
Calling `_related_instances` with no instance ids but with a cutoff time, with start and final as the same entity id:
```
EntitySet._related_instances(entity_id, entity_id,
                           instance_ids=None, time_last=some_actual_time)
```
does not respect the cutoff time, and instead just returns all instances of the entity
I'm not exactly sure if we intended this, but I doubt it.

It's not currently a problem because we only call `EntitySet._related_instances()` from `EntitySet.get_pandas_data_slice()`, which is only called in 2 places, both times with instance_ids provided. However I noticed it in my PR #91 when called `get_pandas_data_slice` a second time inside of `PandasBackend` with no instances. I patched it up by specified all the instances of the relevant entity, but we should fix this so it doesn't happen in the future.

Thanks to @Seth-Rothschild for finding this problem.

bschreck:
Resolves #93

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -983,18 +983,15 @@ class EntitySet(BaseEntitySet):
         training_window_is_dict = isinstance(training_window, dict)
         window = training_window
         start_estore = self.entity_stores[start_entity_id]
-        if instance_ids is None:
-            df = start_estore.df
-        else:   # instance_ids was passed in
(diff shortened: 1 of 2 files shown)


kmax12:
good catch. merging

PR #110, opened 2018-03-13
Uses full entity update for dependencies of uses_full_entity features

bschreck:
This PR fixes two issues related to dependencies of uses_full_entity features.

Issue 1:

If:

a feature X does not have the uses_full_entity attribute
a dependent feature Y does have the uses_full_entity attribute
another dependent feature Z does not have the uses_full_entity attribute, and is one of the requested output features
Then the output frame as a result of computing this feature X should be placed in both entity_frames, and large_entity_frames in pandas_backend.py. However, we previously only check if the feature X itself is a requested output feature, not if it has a dependent that is an output feature (and not a uses_full_entity feature). This means the output of X was only placed in large_entity_frames. The computation of feature Z then failed because it depended on X being in entity_frames.

Issue 2:
To make sure we don't have overlapping columns when we concatenate the output frame as a result of computing a feature with the existing entity_frames, we drop duplicated columns in feature computation output frame (called result_frame in the code). However, we removed these columns in-place, such that if we have to do 2 concats (placing the output frame in both entity_frames and large_entity_frames, it's possible that we remove some computed features that wouldn't get placed in the second output frame.

It is much cleaner to explicitly label each feature with the input frame is should be given, and which output frames its result should be placed in. This fixes issue 2 nicely

I wrote a test case that checks for both of these conditions.

### DIFF
--- a/featuretools/computational_backends/feature_tree.py
+++ b/featuretools/computational_backends/feature_tree.py
@@ -213,15 +213,20 @@ class FeatureTree(object):
 
     def uses_full_entity_differentiator(self, f):
         is_output = f.hash() in self.feature_hashes
+        normal_dependent_is_output = any([(not d.uses_full_entity and
+                                           d.hash() in self.feature_hashes)
+                                          for d in self.feature_dependents[f.hash()]])
+        need_selected_output = is_output or normal_dependent_is_output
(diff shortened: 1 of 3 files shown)


kmax12:
Looks great! Merging

PR #111, opened 2018-03-14
Keep non-feature columns when encoding feature matrix

rwedge:
#78 added the option of including non-feature columns with the calculated feature matrix, but `encode_features`  was not including them when creating an encoded feature matrix.   This PR fixes `encode_features` so it includes non-feature columns in the encoded feature matrix.

### DIFF
--- a/featuretools/synthesis/encode_features.py
+++ b/featuretools/synthesis/encode_features.py
@@ -118,7 +118,7 @@ def encode_features(feature_matrix, features, top_n=10, include_unknown=True,
                                       unit="feature")
     for c in iterator:
         if c in extra_columns:
-            pass
+            continue
         try:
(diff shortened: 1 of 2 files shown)


featuretools/synthesis/encode_features.py
```diff
@@ -65,6 +65,8 @@ def encode_features(feature_matrix, features, top_n=10, include_unknown=True,
         X = feature_matrix.copy()
 
     encoded = []
+    feature_names = [f.get_name() for f in features]
+    extra_columns = [col for col in X.columns if col not in feature_names]
```
kmax12: to make this safer, can we add in a check that all the features you pass in are columns in the feature matrix? with this change, we might silence issues where the feature matrix and features don't match

kmax12:
looks good to me. merging in.

PR #118, opened 2018-03-24
NumCharacters primitive handles nan 

gsheni:
Resolves an issue where `NumCharacters` would error if the string was a `NaN` (or could interpreted as a `Nan`) and would be unable to return the string length for the primitive.


kmax12:
Looks good. Merging

PR #120, opened 2018-03-26
Update input types for binary features

gsheni:
- swapped the order of the list of input types for `Compare`


kmax12:
Closing. Will be fixed in #142

PR #121, opened 2018-03-26
Chunking

rwedge:
Instead of calculating all rows of a feature matrix that share the same cutoff time together,  Featuretools breaks the feature matrix rows into chunks to calculate separately, prioritizing grouping rows with the same cutoff time in the same chunk.  `ft.dfs` and `ft.calculate_feature_matrix` now have a `chunk_size` parameter to allow for custom chunk sizes.  `chunk_size` accepts positive integers for explicit chunk sizes or floats between 0 and 1 for percentage-based chunk sizes.  The old 'group all rows that share a cutoff time' method can be used by setting the string "cutoff time" as the chunk size.


kmax12:
Good work! Merging in

PR #123, opened 2018-03-28
move transform to last step of dfs without creating duplicate features

### LINKED ISSUE #119: Why not build Entity features after Direct features？
I find that direct features are now built at last step in the realization of DFS algorithm, which causes the problem in #81 . 
However, I've also read the paper ‘Deep Feature Synthesis: Towards Automating DataScience Endeavors’. In this paper, you said in Section Ⅱ.C  that "We must first synthesize rfeat and dfeat features so we can apply efeat feature to the results." Actually I think the idea in your paper is much more reasonable.  So why don't you use this idea? Is there any inexplicit trouble?

zhxt95:
A try to resolve #119 
Swap the order of building transform and direct features and add a filter to avoid buiding duiplicate transform features with dfeat.

### DIFF
--- a/featuretools/synthesis/deep_feature_synthesis.py
+++ b/featuretools/synthesis/deep_feature_synthesis.py
@@ -368,12 +368,7 @@ class DeepFeatureSynthesis(object):
                                      max_depth=max_depth)
 
         """
-        Step 3 - Create Transform features
-        """
-        self._build_transform_features(all_features, entity, max_depth=max_depth)
-
-        """
-        Step 4 - Recursively build features for each entity in a forward relationship
+        Step 3 - Recursively build features for each entity in a forward relationship
         """
         forward_entities = self.es.get_forward_entities(entity.id)
         # filter entities in path and using traversal filters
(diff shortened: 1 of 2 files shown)


Seth-Rothschild:
Looks like there are a couple of formatting errors in the code (you can click the Details button on circleci or run `flake8` locally to see where they are).

zhxt95:
The number of features changed mainly because I build DirectFeature before transform step, thus there will be more features( identity features + backward features + forward features) than before( identity features + backward features ) that can be used in transform step, e.g.`<Feature: value + sessions.customers.age>`
And since the forward entities also get more features in their tansform steps, so we will also get more  dfeats which are made by adding, e.g.`<Feature: sessions.SUM(log.value_many_nans) + customers.age>`
Eventually, we'll find that both num_add_feats and num_add_as_base_feat increase.

kmax12:
@zhxt95 just wanted to give you an update that we are still looking at this and considering it for inclusion in a future release. basically, we want to be careful to make this change since it could potentially change the features created by a `ft.dfs()` call after a user upgrades. We will keep you posted as we make progress testing it in different situations. Thanks for working on this!

zhxt95:
Good to know that. I'm willing to offer help whenever you need:)

CJStadler:
Resolved by #623. Thanks for your work on this @zhxt95!

PR #124, opened 2018-03-30
Update repr to show all entities and display to row/col counts

Seth-Rothschild:
Previously the output from an `EntitySet` was the first five entities. Now it shows all entities, all relationships and all columns in an entity. I've also changed the shape output to say `(Rows: {}, Columns: {})`.

Example images below.
image

image


kmax12:
Looks good, merging.

PR #126, opened 2018-03-30
Add make_temporal_cutoffs utility function

bschreck:
Adds 2 new utility functions:

`make_temporal_cutoffs()` creates multiple evenly spaced cutoff times for each instance prior to the specified input cutoff times
`tdfs()` is a wrapper around `dfs()` that calls `make_temporal_cutoffs` and passes the resulting multiple time windows into `dfs()`. The result is a flattened _feature tensor_, with multiple values per instance. This feature tensor can be used for temporal machine learning algorithms, such as Recurrent Neural Networks.


featuretools/utils/time_utils.py
```diff
@@ -0,0 +1,23 @@
+import pandas as pd
+
+
+def make_temporal_cutoffs(instance_ids,
```
kmax12: can you add a docstring to this function?

docs/source/automated_feature_engineering/handling_time.rst
```diff
@@ -118,3 +118,48 @@ We can see that that the counts for the same feature are lower when we shorten t
...
+
+.. ipython:: python
+
+    temporal_cutoffs = ft.make_temporal_cutoffs(cutoffs['customer_id'],
+                                                cutoffs['cutoff_time'],
+                                                window_size='3d',
+                                                num_windows=2)
+    temporal_cutoffs
+
+    import featuretools as ft
```
kmax12: i don't think you need this import

Issue #125, opened 2018-03-30
Integer time index in entity turned into timestamp in feature matrix index


rwedge:
When calculating features where the target entity uses an integer-based time index, the time index for the resulting feature matrix is in datetime format.  As a side effect, additional columns in the `cutoff_time` dataframe are not passed correctly.

kmax12:
fixed by #128

Issue #99, opened 2018-02-20
Silent failure in 'cutoff_time'


Seth-Rothschild:
It seems that it's possible to pass in a string rather than a time to 'cutoff_time' which blocks `dfs` from using the cutoff time at all without warning. This can show up in the #91 branch in the form of using **all** values rather than valid ones for the cutoff time. 

Thanks to @bschreck for finding this and the error in my code that led to it.

Seth-Rothschild:
Closed with #128

PR #128, opened 2018-04-02
Improved handling of integer time index

rwedge:
Addresses #99 and #125.

* Time values in `cutoff_time` that are not numeric or datetime will raise an error
* If target entity has an integer time index, feature matrix will also have integer time index
* Extra columns in `cutoff_time` will now pass through correctly when using integer time index
* Incompatible time index and cutoff time types will raise an error

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -26,6 +26,8 @@ from featuretools.utils.wrangle import _check_timedelta
 
 logger = logging.getLogger('featuretools.computational_backend')
 
+_numeric_types = (int, np.int32, np.int64)
+
 
 def calculate_feature_matrix(features, cutoff_time=None, instance_ids=None,
                              entities=None, relationships=None, entityset=None,
@@ -110,6 +112,11 @@ def calculate_feature_matrix(features, cutoff_time=None, instance_ids=None,
     if not isinstance(cutoff_time, pd.DataFrame):
         if cutoff_time is None:
             cutoff_time = datetime.now()
+            # if integer time index, use max value as cutoff time instead
+            if target_entity.time_index:
+                target_time_index = target_entity.df[target_entity.time_index]
+                if isinstance(target_time_index.iloc[0], _numeric_types):
+                    cutoff_time = target_time_index.max()
 
         if instance_ids is None:
             index_var = target_entity.index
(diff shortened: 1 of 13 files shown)


featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -266,7 +277,10 @@ def calc_results(time_last, ids, precalculated_features=None, training_window=No
 
         # if all aggregations have been approximated, can calculate all together
         if no_unapproximated_aggs and approximate is not None:
-            grouped = [[datetime.now(), group]]
+            if isinstance(group[target_time].iloc[0], _numeric_types):
+                grouped = [[np.inf, group]]
+            else:
+                grouped = [[datetime.now(), group]]
```
kmax12: should we move this `datetime.now()` outside of `calc_results`. basically, we want every chunk that's part of a single cfm call to use the same time.

featuretools/entityset/entity.py
```diff
@@ -591,25 +591,19 @@ def _filter_and_sort(self, df, time_last=None,
         """
         if self.time_index:
             if time_last is not None and not df.empty:
```
kmax12: is it possible for time last to be none by this point in the code? If not, let's remove

rwedge: `Entity.get_all_instances` and a few others call this without specifying a time last

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -110,6 +113,11 @@ def calculate_feature_matrix(features, cutoff_time=None, instance_ids=None,
     if not isinstance(cutoff_time, pd.DataFrame):
         if cutoff_time is None:
             cutoff_time = datetime.now()
+            # if integer time index, use max value as cutoff time instead
+            if target_entity.time_index:
```
kmax12: the target entity might not have a time index, but some other entity in the entityset might. perhaps it makes sense to add a method or attribute to an entityset that returns if it is a datetime or numeric time index. 

we should assume all entities in an entityset are using the same units for the time index, so this method could do the check when being initialized or when `set_time_index` is called.

featuretools/variable_types/variable.py
```diff
@@ -291,6 +291,11 @@ class TimeIndex(Variable):
     _dtype_repr = "time_index"
 
 
+class NumericTimeIndex(TimeIndex, Numeric):
```
kmax12: can you add to API reference in documentation?

kmax12:
Looks great. Merging

PR #129, opened 2018-04-04
Primitives as strings in DFS parameters

bschreck:
dfs() and DeepFeatureSynthesis() can now accept strings (instead of Python classes) for all primitives built into Featuretools.

API looks like this:

    ft.dfs(entityset=es, target_entity='test',
              agg_primitives=['sum'],
              trans_primitives=['percentile'],
              where_primitives=['sum', 'percentile'])


featuretools/primitives/utils.py
```diff
@@ -59,7 +59,20 @@ def get_transform_primitives():
...
                     transform_primitives.add(attribute)
-    return list(transform_primitives)
+    return {prim.name: prim for prim in transform_primitives}
+
+
+def list_primitives():
+    transform_primitives = get_transform_primitives()
+    agg_primitives = get_aggregation_primitives()
+    transform_df = pd.DataFrame({'name': list(transform_primitives.keys()),
+                                 'description': [prim.__doc__.split("\n")[0] for prim in transform_primitives.values()]})
```
Seth-Rothschild: This gives different results between docstrings that are
```
"""
Words go here
"""
```
``` 
"""Words go here
"""
```
and 
```
""" Words go here
"""
```
We have all three but should probably settle on one.

kmax12: we use the Google style docstrings for all public functions. for some private functions, we may be inconsistent and should fix those as we see it. 

here's the Google style docstrings guide: http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

Seth-Rothschild: I've pushed a version which has nonempty first line for all of the primitives. It looks like there's some extra space being added by something in the description (as shown in this image)

image

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -36,38 +38,37 @@ class DeepFeatureSynthesis(object):
...
-                             :class:`Month <.primitives.Month>`, \
-                            :class:`Weekday <.primitives.Weekday>`, \
-                             :class:`Haversine <.primitives.Haversine>`, \
-                             :class:`NumWords <.primitives.NumWords>`, \
-                             :class:`NumCharacters <.primitives.NumCharacters>`]
-
-            where_primitives (list[:class:`.primitives.AggregationPrimitive`], optional):
-                only add where clauses to these types of Aggregation
-                Features.
+                Default: ["sum", \
```
kmax12: Nitpick - I dont think these all need their own line. can we reform to be 80 chars?

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -179,13 +176,50 @@ def __init__(self,
...
+                self.trans_primitives.append(t)
+
+        if where_primitives is None:
+            where_primitives = [ftypes.Count]
+        self.where_primitives = []
+        for p in where_primitives:
+            if isinstance(p, basestring):
+                prim_obj = agg_prim_dict.get(p.lower(), None)
+                if prim_obj is None:
+                    prim_obj = trans_prim_dict.get(p.lower(), None)
```
kmax12: where's only apply to aggregations, no need to check transform too

featuretools/synthesis/dfs.py
```diff
@@ -60,30 +60,30 @@ def dfs(entities=None,
...
-                         :class:`Skew <.primitives.Skew>`, \
-                         :class:`Min <.primitives.Min>`, \
-                         :class:`Mean <.primitives.Mean>`, \
-                         :class:`Count <.primitives.Count>`, \
-                         :class:`PercentTrue <.primitives.PercentTrue>`, \
-                         :class:`NUniqe <.primitives.NUnique>`, \
-                         :class:`Mode <.primitives.Mode>`]
-
-        trans_primitives (list[TransformPrimitive], optional):
+                Default: ["sum", \
```
kmax12: same comment as above about single lines

featuretools/synthesis/deep_feature_synthesis.py
Seth-Rothschild: It's 114 chars (agg defaults) and 104 chars (trans defaults). Keep it on one line anyways? Even removing "    Default:" isn't enough to get it under 80.

kmax12:
Great work. Merging in

kmax12:
Great work. Merging in

PR #135, opened 2018-04-30
Ensure config dir is writable

bschreck:
Issue: on some environments (namely, some automatic CI environments) the home directory is not writable. Featuretools thus fails to install because we attempt to create a ~/.featuretools folder.

This PR resolves this problem by checking if the config folder is writable, and if not, it creates a temporary folder instead. It also allows a user to specify which folder to use for config via an environment variable ($FEATURETOOLS_DIR).

Much of this code was modified from the IPython source:

https://github.com/ipython/ipython/blob/master/IPython/utils/path.py (get_ipython_dir())

and

https://github.com/ipython/ipython/blob/master/IPython/paths.py (get_home_dir())

Lastly, it changes config_yaml.txt to a proper yaml file: config.yaml


.circleci/config.yml
```diff
@@ -8,4 +8,4 @@ jobs:
       - checkout
       - run: pyenv local 2.7.13 3.5.2 3.6.0
       - run: make installdeps
-      - run: make lint && tox && codecov
+      - run: tox && make lint && codecov
```
kmax12: any reason for shifting order? I'd think we want things to fail faster

bschreck: My thought was that you'd want to know if your code passed the tests regardless of whether it passed linting. That is a more important indicator of how your branch is doing.

bschreck: In this case in particular, the lint error was masking the actual testing error that I cared about, and was not able to reproduce locally

kmax12: gotcha. this change makes sense to me

kmax12:
Looks good. Merging

PR #137, opened 2018-05-02
move return in test_arith and test_override

Seth-Rothschild:
There were 4 tests found that were not being run. Three had to do with symbol overrides and one was testing arithmetic on ordinals. Added functionality for adding and subtracting ordinals.


kmax12:
Great! Merging

PR #139, opened 2018-05-07
Use s3 demo csv with better column names

Seth-Rothschild:
Change the location from which we download the demo retail data. With that, we can remove the line which replaced the old column names.


kmax12:
Looks good merging

PR #146, opened 2018-05-10
Remove unused utility functions

bschreck:
Removes:
 * utils.wrangle.flatten_2d
* utils.wrangle._check_time_against_time

These functions were previously used, but are no longer needed


kmax12:
Looks good, merging!

PR #153, opened 2018-05-22
Support Pandas 0.23.0

bschreck:
Various changes to support Pandas 0.23. Issues related to the new ability to merge on a combination of index and column names, the need to supply a "sort" argument to `DataFrame.append`, as well as some issues related to MultiIndex sorting.


featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -136,10 +136,25 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
...
+            else:
+                try:
+                    cutoff_time['time'] = pd.to_datetime(cutoff_time['time'])
+                except (ValueError, TypeError):
+                    raise TypeError("cutoff_time times must be datetime type")
         pass_columns = [column_name for column_name in cutoff_time.columns[2:]]
 
     if _check_time_type(cutoff_time['time'].iloc[0]) is None:
         raise ValueError("cutoff_time time values must be datetime or numeric")
+    if cutoff_time['instance_id'].dtype == np.dtype('O') and target_entity.df[target_entity.index].dtype.name.find('int') > -1:
```
kmax12: Not sure this check is needed.

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -136,10 +136,25 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
             # take the first column that isn't instance_id and assume it is time
             not_instance_id = [c for c in cutoff_time.columns if c != "instance_id"]
             cutoff_time.rename(columns={not_instance_id[0]: "time"}, inplace=True)
+        if cutoff_time['time'].dtype == object:
+            if entityset.time_type == NumericTimeIndex:
+                try:
+                    cutoff_time['time'] = pd.to_numeric(cutoff_time['time'])
+                except (ValueError, TypeError):
+                    raise TypeError("cutoff_time times must be a numeric")
+            else:
```
kmax12: can we do an explicit check for DateTimeIndex here?

featuretools/computational_backends/calculate_feature_matrix.py
kmax12: I think we can do without these checks and conversations for now. let's assume the user has this fixed up before calling calculate feature matrix.

bschreck: These are to suppress warnings from the tests. Should we change the tests then? Or add in a filter to ignore pandas warnings when running the tests?

kmax12:
Looks good to me. Merge it!

PR #154, opened 2018-05-22
Better Pandas Backwards Compatibility

kmax12:
Pandas 0.23.0 added a `sort` parameter to `DataFrame.append`. This PR handles the case where an old version of pandas is being used.


bschreck:
Looks good!

PR #155, opened 2018-05-24
Ensure relationship variables are the same underlying dtype

bschreck:
If a parent entity has an index with integer data, and the linking variable in the child has object/string data, this will convert the child variable to integers. This is necessary for the new version of pandas (0.23), because merges on mixed dtypes are not allowed.

### DIFF
--- a/featuretools/entityset/base_entityset.py
+++ b/featuretools/entityset/base_entityset.py
@@ -201,7 +201,7 @@ class BaseEntitySet(object):
         if ((is_object_dtype(parent_e.df[parent_v]) or
                 is_string_dtype(parent_e.df[parent_v])) and
                 is_numeric_dtype(child_e.df[child_v])):
-            parent_e.df[parent_v] = pd.to_numeric(self[parent_e.id].df[parent_v])
+            parent_e.df[parent_v] = pd.to_numeric(parent_e.df[parent_v])
         if ((is_object_dtype(child_e.df[child_v]) or
                 is_string_dtype(child_e.df[child_v])) and
                 is_numeric_dtype(parent_e.df[parent_v])):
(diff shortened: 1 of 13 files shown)


featuretools/entityset/base_entityset.py
```diff
@@ -186,15 +189,23 @@ def add_relationship(self, relationship):
...
+
+        if not isinstance(parent_e[parent_v], vtypes.Discrete):
             parent_e.convert_variable_type(variable_id=parent_v,
                                            new_type=vtypes.Index,
                                            convert_data=False)
-
+        if ((is_object_dtype(parent_e.df[parent_v]) or
+                is_string_dtype(parent_e.df[parent_v])) and
+                is_numeric_dtype(child_e.df[child_v])):
+            parent_e.df[parent_v] = pd.to_numeric(self[parent_e.id].df[parent_v])
```
rwedge: The `self[parent_e.id]` here can be changed to `parent_e.df`

featuretools/tests/demo_tests/test_demo_data.py
```diff
@@ -24,3 +25,12 @@ def test_load_retail_diff():
...
     os.remove(make_retail_pathname(nrows_second))
+
+
+def test_load_mock_customer_dfs():
+    es = load_mock_customer(return_entityset=True)
+    dfs(entityset=es,
+        target_entity="sessions",
+        agg_primitives=["mean", "sum", "mode"],
+        trans_primitives=["month", "hour"],
+        max_depth=2)
```
rwedge: A little weird that this test has no assertions. Maybe assert that the normalize successfully converted both parent and child values to matching types?  Or maybe the test should go in test_es:test_add_relationship_converts_types

PR #157, opened 2018-05-25
Update documentation after recent changes / removals

rwedge:
Some examples in the docs still depended on `Entity.head` and `Primitive.head` and needed to be updated.  Likewise, the change requiring either `entityset` or `entities` and `relationships` to be passed to calculate_feature_matrix meant some documentation required updating.


kmax12:
Looks good!

PR #159, opened 2018-05-29
Check dtypes match when adding relationship

rwedge:
Also try to convert categorical dtype to int dtype before converting it to object dtype


kmax12:
Good to merge once tests pass

PR #162, opened 2018-06-04
Config does error checking to see if it can write to disk

bschreck:
`config.py` now does error checking to see if it can write to disk. Returns a fallback config dictionary with barebones entries (just logging) if reading config file from disk fails.


kmax12:
Looks good to me!

PR #165, opened 2018-06-07
Remove list of times as on option for `cutoff_time` in `calculate_feature_matrix`

rwedge:
Using a list of times for `cutoff_time` was creating some unexpected behavior if there wasn't an `instance_ids` list of the same length. Some examples:
```python
ft.calculate_feature_matrix(features=features,
                            entityset=es,
                            instance_ids=[0, 1, 2],
                            cutoff_time=[0, 1])
```
This calculates features for instance 0 at time 0, instance 1 at time 1, and doesn't calculate features for instance 2.
```python
ft.calculate_feature_matrix(features=features,
                            entityset=es,
                            instance_ids=[0, 1],
                            cutoff_time=[0, 1, 2])
```
This calculates features for instance 0 at time 0, instance 1 at time 1, and not use time 2.
```python
ft.calculate_feature_matrix(features=features,
                            entityset=es,
                            cutoff_time=[0, 1])
```
This would calculate features for the first instance in the entity's dataframe at time 0 and calculate features for the second instance in the entity's dataframe at time 1.

While requiring a list of instance_ids with the same length as the list of cutoff times is one way to resolve this issue, equal length lists of instances and times can easily be represented in DataFrame format and would leave one less input case to handle.

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -46,7 +46,7 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
         cutoff_time (pd.DataFrame or Datetime): Specifies at which time to calculate
             the features for each instance.  Can either be a DataFrame with
             'instance_id' and 'time' columns, DataFrame with the name of the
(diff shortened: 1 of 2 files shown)


kmax12:
Looks good to me!

PR #166, opened 2018-06-08
Fix docstring in training window

Seth-Rothschild:
Removes reference to no longer supported dictionary input to training window.


featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -65,11 +65,9 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
             where the second index is the cutoff time (first is instance id).
             DataFrame will be sorted by (time, instance_id).
 
-        training_window (dict[str -> Timedelta] or Timedelta, optional):
+        training_window (Timedelta, optional):
             Window or windows defining how much older than the cutoff time data
```
kmax12: I think the "or windows" should be removed too.

kmax12:
Looks good. Merging

PR #167, opened 2018-06-11
Add support for building and uploading conda package

bschreck:
This PR adds several conda config files needed to build and upload Featuretools to anaconda.org. It includes a script that launches an Ubuntu docker container, and then builds and uploads the package for 3 different Python versions.

Several dependencies needed to have their versions bumped because anaconda only lists the newer ones. Removed a duplicate pytest-runner dependency from dev-requirements.txt (it's already in test-requirements.txt).


kmax12:
Looks good, merging

don-lab-dc:
@kmax12, would be great to have something in the docs about using conda to install.  The standard method (conda install featuretools, after adding forge channel) didn't work for me. At present, docs just cover pip.

PR #168, opened 2018-06-13
Make sure index variable comes first

bschreck:
Reorders variable list in `Entity.__init__()` so that the index variable is at the front of the list.

### DIFF
--- a/featuretools/entityset/entity.py
+++ b/featuretools/entityset/entity.py
@@ -120,6 +120,11 @@ class Entity(object):
         if self.index is not None and self.index not in inferred_variable_types:
             self.add_variable(self.index, vtypes.Index)
 
+        # make sure index is at the beginning
+        index_variable = [v for v in self.variables
+                          if v.id == self.index][0]
+        self.variables = [index_variable] + [v for v in self.variables
+                                             if v.id != self.index]
         self.update_data(df=self.df,
                          already_sorted=already_sorted,
                          recalculate_last_time_indexes=False,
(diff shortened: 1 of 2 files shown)


featuretools/tests/entityset_tests/test_entity.py
```diff
@@ -22,6 +22,11 @@ def test_is_index_column(es):
     assert es['cohorts'].index == 'cohort'
 
 
+def test_index_at_beginning(es):
```
kmax12: can we add another test? 

I'm thinking create an entity from a DF where the index isn't the first column and make sure that it gets moved to be be the first in the variables list.

featuretools/entityset/entity.py
```diff
@@ -120,6 +120,11 @@ def __init__(self, id, df, entityset, variable_types=None, name=None,
         if self.index is not None and self.index not in inferred_variable_types:
             self.add_variable(self.index, vtypes.Index)
 
+        # make sure index is at the beginning
+        index_variable = [v for v in self.variables
+                          if v.id == self.index][0]
+        self.variables = [index_variable] + [v for v in self.variables
```
kmax12: do we not care if the order of variable's doesn't match the underlying dataframe? I don't believe we assume that anywhere. can you think of any places?

bschreck: Why don't we reorder the columns in the dataframe as well?

featuretools/tests/entityset_tests/test_entity.py
bschreck: Yeah makes sense, I'll add that too

featuretools/entityset/entity.py
kmax12: ya, let's do that

kmax12:
Looks good, merging

PR #169, opened 2018-06-18
No last time index update on normalize

rwedge:
This PR makes some changes to when last time indexes are recalculated.

* `recalculate_last_time_indexes` will only recalculate if the entity had a last time index
* Not recalculating last time index values after normalizing an entity.  This can really slow down EntiySet creation.  I think using `EntitySet.add_last_time_indexes` after getting the final EntitySet structure is simple enough.
* During `EntitySet.concat`, last time indexes are updated in one go instead of when updating each entity's data.


kmax12:
Looks good, merging

PR #171, opened 2018-06-19
Remove variable stats

kmax12:
Currently, entity sets calculate and store statistics about the variables in them. This was previously used internally by DFS, but is no longer needed. To clean up the code, all references to calculating variable statistics inside of Featuretools have been removed.


featuretools/tests/entityset_tests/test_es.py
```diff
@@ -312,55 +312,6 @@ def test_handles_datetime_mismatch():
...
-
-    # TODO: the below fails, but shouldn't
-    # boolean with nan have count and number of true/false labels defined
-    # assert e['boolean_with_nan'].count == 2
-
-    # assert e['boolean_with_nan'].num_true == 1
-    # assert e['boolean_with_nan'].num_false == 1
-
-
 def test_column_funcs(entityset):
```
bschreck: I think this test can be deleted now.

kmax12: I don't see that we test the stuff in this function anywhere else, so I'll keep it to be safe. I did rename it though since it no longer tests the column functions

PR #175, opened 2018-06-25
Update flight

Seth-Rothschild:
Rework the `load_flight` data to use newly uploaded csvs from 2017. New functionality:
1. Can take in a `month_filter` to restrict which months are used (e.g. [3, 5] to use March and May)
2. Can take in a `categorical_filter` to restrict which categorical values are loaded. As an example  `{'dest_city': ['Boston, MA'], 'origin_city': ['Washington, DC']}` will take all flights into Boston and all flights out of Washington DC). 
3. An optional `verbose` argument which gives a progress bar while loading the data.

### DIFF
--- a/featuretools/demo/flight.py
+++ b/featuretools/demo/flight.py
@@ -1,7 +1,6 @@
 import os
 import re
 import boto3
-import botocore
 from botocore.handlers import disable_signing
 from builtins import str
 
@@ -16,8 +15,8 @@ def load_flight(use_cache=True,
                 demo=True,
                 only_return_data=False,
                 nrows=None,
-                month_filter=[1, 2],
-                categorical_filter={'dest_city': ['Boston, MA'], 'origin_city': ['Boston, MA']}):
+                month_filter=None,
+                categorical_filter=None):
     """ Download, clean, and filter flight data from 2017.
         Input:
             month_filter (list[int]): Only use data from these months. Default is [1, 2].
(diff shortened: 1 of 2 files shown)


featuretools/demo/flight.py
```diff
@@ -1,68 +1,219 @@
...
+    # Code via SO https://stackoverflow.com/questions/1175208/elegant-python-function-to-convert-camelcase-to-snake-case
+    s1 = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', name)
+    return re.sub('([a-z0-9])([A-Z])', r'\1_\2', s1).lower()
+
+
+def make_flight_pathname(demo=True):
+    if demo:
+        filename = 'flight_dataset_sample.csv.zip'
+        key = 'bots_flight_data_2017/data_2017_jan_feb.csv.zip'
+        rows = 860457.0
```
kmax12: does this have to be a float?

featuretools/demo/flight.py
```diff
@@ -1,68 +1,219 @@
...
+    return clean_data
+
+
+def filter_data(clean_data,
+                month_filter=None,
+                categorical_filter=None):
+    if month_filter is not None:
+        tmp = False
+        for month in month_filter:
+            tmp = tmp | (clean_data['scheduled_dep_time'].apply(lambda x: x.month) == month)
```
kmax12: can this be `clean_data['scheduled_dep_time'].dt.month == month`? if that works, it's probably much faster than apply

kmax12: also, you should try to use an isin here rather than iterate over each month

featuretools/demo/flight.py
Seth-Rothschild: We do `rows/ 99` earlier. Not sure if the cleanest solution is to do this, `99.` or `float(rows)` when we divide.

kmax12: not sure it matters because you take the `ceil` of it, but move any casting as close as possible to where it is required

kmax12:
Looks good to me

PR #181, opened 2018-07-03
Add last time index paragraph to docs

Seth-Rothschild:
Add a paragraph describing last time index and how it is used in the documentation.


kmax12:
Looks good!

PR #187, opened 2018-07-06
Update load flight 2

Seth-Rothschild:
Docstring from `load_flight` needed an update to be formatted correctly, so I took the opportunity to add an example and fix a minor secondary time index problem. Here's a screenshot of how the docs page looks after the change:
image


kmax12:
looks good, merging

Issue #190, opened 2018-07-13
Passing in a string to `dfs` instead of a list silently ignores argument


WillKoehrsen:
When running `dfs`, I passed in a  single entity to ignore as a string:

```
features = ft.dfs(entityset = es, target = 'app_train', 
                            ignore_entities = 'app_test', max_depth = 2)
```

This results in features being generated from `app_test` because `ignore_entities` is a string. The documentation does specify that `ignore_entities` must be a list (even when using only one entity), but there is no warning raised when the argument is not a list. Passing in a single entity as a list works:

```
features = ft.dfs(entityset = es, target = 'app_train', 
                            ignore_entities = ['app_test'], max_depth = 2)
```
A warning would be helpful or behavior to handle a single string as in Pandas `drop` function that works with a list of column names or a single string. 

This same problem might apply to other arguments in `dfs`, but I only tested it with `ignore_entities`.

PR #192, opened 2018-07-16
Check type of `ignore_entities` in dfs

WillKoehrsen:
Addresses #190 by checking the type of `ignore_entities` (using `isinstance`) when running `dfs`. Passing in any type other than a list will result in a `TypeError`.

### DIFF
--- a/featuretools/synthesis/deep_feature_synthesis.py
+++ b/featuretools/synthesis/deep_feature_synthesis.py
@@ -120,6 +120,10 @@ class DeepFeatureSynthesis(object):
         if ignore_entities is None:
             self.ignore_entities = set()
         else:
+
+            if not isinstance(ignore_entities, list):
+                raise TypeError('ignore_entities must be a list')
+
             assert target_entity_id not in ignore_entities,\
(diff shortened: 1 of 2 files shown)


featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -120,6 +120,10 @@ def __init__(self,
         if ignore_entities is None:
             self.ignore_entities = set()
         else:
+
```
rwedge: Unnecessary whitespace

featuretools/tests/dfs_tests/test_deep_feature_synthesis.py
```diff
@@ -114,6 +114,13 @@ def find_other_agg_features(features):
 
 
 def test_ignores_entities(es):
+    with pytest.raises(TypeError):
+        dfs_obj = DeepFeatureSynthesis(target_entity_id='sessions',
```
rwedge: Since the call to DeepFeatureSynthesis will error it's not necessary to assign a variable to the result

PR #193, opened 2018-07-20
Type checking ignore entities. 

WillKoehrsen:
Addresses #192  by checking the type of ignore_entities (using `isinstance`) when running `dfs`. Passing in any type other than a list will result in a `TypeError`.

This is an update because the authorship on two of the commits for the [previous pull request](https://github.com/Featuretools/featuretools/pull/192) were incorrect.


kmax12:
Looks good, merging.

PR #198, opened 2018-08-01
Reset index of cutoff times in calculate feature matrix

WillKoehrsen:
Addresses issue where duplicated indices in `cutoff_time` dataframe results in duplicated rows in the feature matrix with `calculate_feature_matrix`. Solution is to reset the index in `cutoff_time` to remove the duplicates.


kmax12:
Looks good. Merging!

PR #200, opened 2018-08-02
Clarify `transactions_df` in "Representing Data" documentation page

Seth-Rothschild:
Following the Stack Overflow question [here](https://stackoverflow.com/questions/51581812/es-normalize-entity-error-variable-not-found-in-entity), I’ve updated the documentation so that you can see which dataframe we’re using in the example. This avoids the potential confusion from accidentally using `load_mock_customer()['transaction']` and trying to normalize a `sessions` entity from it.

This PR is the low-impact version of easing this confusion. We just unhide the hidden lines and don't remove the times. It would also be possible to modify the section more heavily to write it in terms of the `use_single_table` parameter, but those changes would necessarily change more of the associated text and flow.


kmax12:
Looks good merging

PR #201, opened 2018-08-02
Modify retail demo (v2)

Seth-Rothschild:
We change the underlying CSV of the retail demo to include the following:
1. Drop duplicate and null rows
2. Update `price` to `unit_price` and add a `total` column
3. Mark cancelled orders with a boolean column
4. Change the `unit_price` and `total` values to be in dollars

In addition to those changes to the CSV, there were some changes to the `demo.load_retail` function:
1. Add a `return_single_table` optional argument to match the other demos
2. Add a dataset source to the docstring as well as a changelog
3. Move the time index to `order_products` (which is what it represented) and make new time indices for all of the entities.
4. Remove `country` from the customers entity and move it to the `orders` entity.

### DIFF
--- a/featuretools/demo/retail.py
+++ b/featuretools/demo/retail.py
@@ -8,14 +8,23 @@ import featuretools.variable_types as vtypes
 from featuretools.config import config as ft_config
 
 
-def load_retail(id='demo_retail_data', nrows=None, use_cache=True):
+def load_retail(id='demo_retail_data', nrows=None, return_single_table=False, use_cache=True):
     '''
-    Returns the retail entityset example.
+    Returns the retail entityset example which is a modified version of
+    the data found `here <https://archive.ics.uci.edu/ml/datasets/online+retail>`_.
+    In version 2 of our CSV, we have
+
+    1. updated the column names,
+    2. dropped null and duplicate rows and,
+    3. added descriptive columns like a boolean "cancelled" and a numeric "total" which is ``quantity`` times ``unit_price``.
 
     Args:
         id (str):  Id to assign to EntitySet.
-        nrows (int):  Number of rows to load of item_purchases
-            entity. If None, load all.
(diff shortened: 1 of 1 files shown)


Seth-Rothschild:
Changing this seems to have unexpected consequences on tests. Should not be merged until resolved. Investigating today.

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -176,6 +176,7 @@ def test_saveprogress(entityset):
     cutoff_time = pd.DataFrame({'time': times, 'instance_id': range(17)})
     property_feature = IdentityFeature(entityset['log']['value']) > 10
     save_progress = tempfile.mkdtemp()
+    property_feature.default_value
```
Seth-Rothschild: The default value of `property_feature` seems to no longer be tested, in spite of no associated changes to that file. I'm testing it directly here, but we should probably find out the root cause. (tagging @rwedge)

Seth-Rothschild:
The [binary transform coverage loss](https://codecov.io/gh/Featuretools/featuretools/compare/0e098d712e4c299aef67ab22d6891c484343279c...6d75725cdec954bb7b3e1caa21905aceb94317da/src/featuretools/primitives/binary_transform.py) is still completely mysterious, but aside from that the changes are likely ready for inclusion. 

The [third test](https://github.com/Featuretools/featuretools/blob/master/featuretools/tests/computational_backend/test_calculate_feature_matrix.py#L169) in `test_calculate_feature_matrix` creates a feature matrix with `property_feature`. In `master` this triggers [generate_default_df](https://github.com/Featuretools/featuretools/blob/master/featuretools/computational_backends/pandas_backend.py#L265) while in this branch, with only `demo_retail` changes, it seems to not.

kmax12:
Looks good. Merging

PR #206, opened 2018-08-09
Set class name for Min

gsheni:
- Changed primitive `Min` so the class name is set after creation.

### DIFF
--- a/featuretools/primitives/aggregation_primitives.py
+++ b/featuretools/primitives/aggregation_primitives.py
@@ -7,9 +7,7 @@ import numpy as np
 import pandas as pd
 from scipy.stats import skew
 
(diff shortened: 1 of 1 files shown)


featuretools/primitives/aggregation_primitives.py
```diff
@@ -95,25 +92,25 @@ def pd_mode(x):
...
 
 
-Min = make_agg_primitive(
-    np.min,
-    [Numeric],
-    None,
-    name="min",
-    stack_on_self=False,
-    description="Finds the minimum non-null value of a numeric feature.")
+# Min = make_agg_primitive(
```
kmax12: might as well delete this code

kmax12:
Looks good. Merging

PR #207, opened 2018-08-10
Use filenames from S3 for Demos

Seth-Rothschild:
This PR would update our CSV naming convention to use the name of the CSV on S3. I've marked it as WIP because we should consider adding a step to explicitly hunt down old file names from their `csv_save_location` and delete old CSVs. The current branch only downloads a new file with the new name.

Deleting old files might be a reasonable thing to do given the size of the full flight dataset.


rwedge:
Looks good to me

PR #208, opened 2018-08-10
Parallel memory warnings

rwedge:
There are two main parts to this pull request
* **Change to how Featuretools sets worker memory limit by default:**  previously, reducing the number of distributed workers would not increase the maximum memory allotted to each worker.  With this change, Featuretools will set memory limits by splitting the total memory on the system equally between workers instead of equally between total cores on the system.
* **Raise errors or warnings about worker memory limits:** Featuretools will now raise an error if it is about to transmit the EntitySet to a worker and realizes the worker's memory limit is smaller than the size of the EntitySet.  Featuretools will also raise a warning if the worker's memory limit is less than twice the size of the EntitySet.


kmax12:
Looks good, but let's add a test case for a user supplied memory limit

PR #209, opened 2018-08-14
Improve performance of aggregation feature calculation

kmax12:
We had logic that checked for the case where the dataframe we were aggregating on was empty or had no related instances. This logic involved a slow python loop. Due to how we fill in missing values later in the function, this check was unnecessary and could be removed.

In some benchmarks, I saw 25% speeds ups for calculations. 

Also, added test for the empty child dataframe case, which was previously untested.


rwedge:
Looks good!

PR #210, opened 2018-08-14
Agg rename refactor

kmax12:
We were creating an unnecessary wrapping when computing aggregation features. Removing that now


kmax12:
closing because this will be addressed in #215

PR #212, opened 2018-08-15
Removing slow indexing 

kmax12:
Based on benchmarking with larger real world datasets, we've observed that calls to `Entity.query_by_value` are frequently a bottleneck in feature calculation. This is due to a python loop that we used to look up values in an index. 

Rather than optimize the python loop, I am simply removing the pre-indexing of data and using a cython optimized pandas merge within `Entity.query_by_value`.


kmax12:
This was rolled into #215. Closing

PR #213, opened 2018-08-15
Make primitive lookup case insensitive 

kmax12:
With this PR, we now handle the case where the name of the primitive isn't already all lower case


rwedge:
Looks good!

PR #217, opened 2018-08-17
(WIP) Support Serializing to CSV

kmax12:
We have parquet and pickle, but the ability to serialize to CSV would be useful in some cases because of easy human readability

Todo

- [ ] tests
- [ ] potentially refactor serialization for better code reuse


kmax12:
this was implemented in #361

PR #219, opened 2018-08-17
remove convert_links_to_integers

Seth-Rothschild:
Removes the keyword argument `convert_links_to_integers` from normalize entity.

Some of the `last_time_index` tests were using a normalization with `convert_links_to_integers` and sorting values, leading to a resort of the "true" values in the test.

Original ordering was 
```
0.0, 5.0, 10.0, 15.0, 20.0, 1.0, 2.0, 3.0, 7.0, 14.0, NaN
```
if we convert these to range(11) and sort them according to the value in the list above, the new ordering of that range would become
```
0, 5, 6, 7, 1, 8, 2, 9, 3, 4, 10
```
which aligns exactly with the changes we had to make in `true_values_lti`. 

The other change, changing the `value` in `test_parent_some_missing`, is really a convenience measure to ensure that it ends up second to last on the sorted list. If we put it earlier, we'll have to move **all** the rows instead of just one.

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -741,7 +741,6 @@ class EntitySet(object):
 
     def normalize_entity(self, base_entity_id, new_entity_id, index,
                          additional_variables=None, copy_variables=None,
-                         convert_links_to_integers=False,
                          make_time_index=None,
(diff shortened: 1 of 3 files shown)


kmax12:
This looks good to me!

PR #228, opened 2018-08-23
Switch flight CSV variables

Seth-Rothschild:
Previous PR used the larger CSV when the `demo=True` option was selected in `load_flight`. This fixes that.


kmax12:
Looks good. Merging

PR #229, opened 2018-08-23
Optimize metadata calculation

kmax12:
Currently, we recalculate the metadata for an entityset every time the property is accessed. This slows done many operations in Featuretools.

This pull request refactors things so we only calculate the metadata if it hasn't been created yet or if an operation invalidated it. 

Locally this change makes the tests run in 134 seconds vs 513 without this change.


rwedge:
Wouldn't entity level operations like `Entity.add_variable` and `Entity.update_data` also invalidate the EntitySet metadata?

PR #233, opened 2018-08-27
Use randomly generated names in retail csv

Seth-Rothschild:
In this PR we replace the `customer_id` floats with randomly generated names in the retail demo dataset.


Seth-Rothschild:
Changing `customer_id` to `customer_name` in csv and modifying the associated step in normalize entity.

kmax12:
Looks good. Merging

PR #234, opened 2018-08-27
More robust handling when no data at a cutoff time

kmax12:
This PR updates how we merge dataframes internally during feature computations. 

Also added a new test case for the case this merge code now handles.


rwedge:
Looks good!

PR #236, opened 2018-08-29
Improve Consistency of Transform Primitives

WillKoehrsen:
Due to the use of a `frozenset` which is an [unordered Python object](https://docs.python.org/3/library/stdtypes.html?highlight=frozenset#set-types-set-frozenset), the order of commutative transform primitives that used multiple columns, such as `AND`, could change when running Deep Feature Synthesis in different Python sessions. This behavior did not affect the feature values because the primitives are commutative. It did change the feature definitions which made comparing different sets of feature definitions a challenge.

The behavior is addressed by sorting the features used in commutative transform primitives alphabetically by name. For example, if the transform primitive `AND` is used with the features (columns) `FLAG_A` and `FLAG_B` the resulting feature will be `AND(FLAG_A, FLAG_B)`.

### DIFF
--- a/featuretools/synthesis/deep_feature_synthesis.py
+++ b/featuretools/synthesis/deep_feature_synthesis.py
@@ -749,4 +749,7 @@ def match(input_types, features, replace=False, commutative=False):
                 new_match = tuple(new_match)
             matching_inputs.add(new_match)
 
+    if commutative:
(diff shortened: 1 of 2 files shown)


kmax12:
Looks good. Merging

PR #238, opened 2018-09-01
Avoid stacking "expanding" primitives

Seth-Rothschild:
This PR prevents DFS from trying to stack on primitives which output "expanding" features. It is resolves the error from #85.

### DIFF
--- a/featuretools/synthesis/deep_feature_synthesis.py
+++ b/featuretools/synthesis/deep_feature_synthesis.py
@@ -690,6 +690,9 @@ def check_stacking(primitive, input_types):
             if type(f) in primitive.stack_on_exclude:
                 return False
 
+    if primitive.expanding:
+        return False
+
     for f in input_types:
         if f.base_of_exclude is not None:
             if primitive in f.base_of_exclude:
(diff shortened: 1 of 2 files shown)


featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -690,8 +690,9 @@ def check_stacking(primitive, input_types):
             if type(f) in primitive.stack_on_exclude:
                 return False
 
-    if primitive.expanding:
-        return False
+    for f in input_types:
+        if f.expanding:
```
kmax12: we can move this check inside of the loop a couple lines down where we also loop over each input.

also, let's add a test case for this.

Seth-Rothschild: Working on the test case.

For moving this inside the `for` loop, want me to also move the check for [`base_of_exclude`](https://github.com/Featuretools/featuretools/pull/238/files#diff-e5a7e95a13580e8dd6309952a9b9c5e0R698) which can also be placed inside the larger loop?

kmax12:
Looks good. Merging

PR #249, opened 2018-09-10
Update deep_feature_synthesis.py

gsheni:
- change docstring which shows wrong string it should be `num_unique`, not `n_unique`, see [here](https://github.com/Featuretools/featuretools/blob/master/featuretools/primitives/aggregation_primitives.py#L106)


kmax12:
looks good merging

PR #250, opened 2018-09-12
Clean up merging of aggregation primitives

### LINKED ISSUE #252: “IndexError: Too many levels” when running Featuretools dfs after upgrade
Featuretools' `dfs()` method fails to run on my entity set after upgrading from v0.1.21 to v0.2.x and v0.3.0.

The error is raised when the Pandas backend tries to calculate the aggregate features `_calculate_agg_features()`. In particular:

`--> 442 to_merge.reset_index(1, drop=True, inplace=True)`
`...`
`IndexError: Too many levels: Index has only 1 level, not 2`

This is working fine in v0.1.x and the entity set hasn't changed after the upgrade. The entity set is composed of 7 entities and 6 relationships. Each entity (dataframe) is added via `entity_from_dataframe`.

kmax12:
This pull request makes a minor refactor to how we calculate and then merge some aggregation primitives. By returning a series instead of a dataframe, so we can avoid a reset index. 

I also noticed we don't test the trend primitive, which uses this code path, so I added another test for good measure. 

fixes #252 

Overall this PR simplifies code without changing any functionality.

### DIFF
--- a/featuretools/computational_backends/pandas_backend.py
+++ b/featuretools/computational_backends/pandas_backend.py
@@ -438,8 +438,6 @@ class PandasBackend(ComputationalBackend):
             # the column (in actuality grouping by both index and group would
             # work)
             to_merge = base_frame.groupby(base_frame[groupby_var], observed=True, sort=False).apply(wrap)
-
-            to_merge.reset_index(1, drop=True, inplace=True)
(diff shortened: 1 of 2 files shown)


rwedge:
Looks good to me!

PR #255, opened 2018-09-13
Handle non-string column names

### LINKED ISSUE #254: Integer column names in dataframe
### Bug Description
Currently, if your dataframe has an integer column name, running `es.entity_from_dataframe` will throw an error. There are three ways we might consider dealing with this:
1. We convert non-strings to strings automatically 
2. We detect the existence of non-strings and raise an error with a message for how to fix
3. We natively support integers. This is challenging, because we'll have to support integers when generating feature names in primitive stacking as well.

Of these, there are arguments for both 1 and 2. I'd lean towards 2 because it encourages the better result: if you have named columns the features generated will be easier to parse. 

Thanks to @WacoHolve_gitlab on gitter for finding this issue and providing a MWE there.

-----

Seth-Rothschild:
close #254 

This is an implementation of option 2 from #254 where we raise an error whenever we see a non `str` column name. There is also unit test with an integer column name to see that the error is raised.

Since this is one of a few possible solutions, it's marked as WIP until we determine which path we want to follow.

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -14,7 +14,7 @@ from .relationship import Relationship
 from .serialization import load_entity_data, write_entityset
 
 import featuretools.variable_types.variable as vtypes
-from featuretools.utils.gen_utils import make_tqdm_iterator
+from featuretools.utils.gen_utils import make_tqdm_iterator, is_string
 
 pd.options.mode.chained_assignment = None  # default='warn'
 logger = logging.getLogger('featuretools.entityset')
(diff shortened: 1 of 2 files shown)


Seth-Rothschild:
Removing WIP. Note that the main check

``if not isinstance(c, (''.__class__, u''.__class__)):``

is one of [many ways](https://stackoverflow.com/a/33699705/9458191) we might want to check for strings in both Python 2 and 3. @rwedge has suggested that we have used
```
if type(ret) != str:
    ret = ret.encode("utf-8")
```
elsewhere in Featuretools to ensure that we're consistently encoding things. This solution used fewer lines of code than a try/except block, but I don't have a strong preference for the way we should be doing this.

featuretools/entityset/entityset.py
```diff
@@ -1211,6 +1211,9 @@ def _import_from_dataframe(self,
                                  r.child_entity.id == entity_id]
 
         for c in dataframe.columns:
+            if not is_string(c):
+                raise ValueError("All column names must be strings. (Column has name {})".format(c))
```
kmax12: Change language to "All column names must be strings (Column {} is not a string)"

kmax12:
Looks good. Merging

PR #258, opened 2018-09-15
Handle where clause when no child data

### LINKED ISSUE #256: when no data at a cutoff time "where feature" will cause keyerror
in pandas_backend.py

line 382: base_frame = base_frame[base_frame[where.get_name()]]

line 454:to_merge = base_frame.groupby(base_frame[groupby_var],
                                          observed=True, sort=False).agg(to_agg)

when base_frame is a empty dataframe, after line 382 base_frame become no columns dataframe,
so line 454 raise keyerror

kmax12:
Handles case where there is no child data for a aggregation feature with a where clause. They issue is that getting zero rows from the empty dataframe returned a new dataframe without column names resulting in an error later in the calculation.

Fixes #256

### DIFF
--- a/featuretools/computational_backends/pandas_backend.py
+++ b/featuretools/computational_backends/pandas_backend.py
@@ -378,8 +378,8 @@ class PandasBackend(ComputationalBackend):
 
         # handle where clause for all functions below
         where = test_feature.where
-        if where is not None:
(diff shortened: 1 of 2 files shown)


WillKoehrsen:
Looks good to me!

PR #259, opened 2018-09-17
Update retail.py

gsheni:
This fixes the following error
```python
featuretools.entityset - WARNING    index order_product_id not found in dataframe, creating new integer column
```


kmax12:
Looks good. merging

PR #260, opened 2018-09-18
Add `is_string` utils function

Seth-Rothschild:
This PR replaces the python2 style `isinstance(value, basestring)` with a utils function `is_string(value)`. It replaces the import `from past.builtins import basestring` with our own function `from featuretools.utils import is_string`. The only other use of `past.builtins` is in `serialization.py` where we import `unicode`.

This is related to #255 and the associated issue #254.


kmax12:
Looks good. merging

PR #261, opened 2018-09-18
Update python versions to match docker container

Seth-Rothschild:
Due to [upstream changes](https://github.com/themattrix/docker-pyenv), our circle ci config is out of date. We now test

- Python 2.7.15 (up from 2.7.13)
- Python 3.5.6 (up from 3.5.2)
- Python 3.6.6 (up from 3.6.0) and
- ~~Python 3.7.0 (previously untested)~~ [pyarrow does not support 3.7](https://issues.apache.org/jira/browse/ARROW-1661)

We can also remove the line `apt update && apt install git -y` which installs git.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,6 @@ jobs:
         - image: themattrix/tox
     steps:
       - checkout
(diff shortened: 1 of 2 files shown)


kmax12:
Looks good merging

PR #262, opened 2018-09-19
Add tests for Entity methods

Seth-Rothschild:
This PR adds three tests and fixes the `add_variable` method for Entity.

+ Fixes a faulty assertion and expected column types in `entity.add_variable`
+ Test `entity.add_variable`
+ Adds tests for two value errors in `entity.update_data`
+ Test the previously untested `elif unit == 'Y'` block in `add_td`

### DIFF
--- a/featuretools/tests/entityset_tests/test_entity.py
+++ b/featuretools/tests/entityset_tests/test_entity.py
@@ -82,16 +82,3 @@ def test_update_data(es):
     with pytest.raises(ValueError) as excinfo:
         es['customers'].update_data(df)
     assert 'Updated dataframe contains 13 columns, expecting 12' in str(excinfo)
-
-
-def test_add_variable(es):
-
-    es['customers'].add_variable('age')
-    assert es['customers'].shape[1] == 12
-
-    with pytest.raises(AssertionError) as excinfo:
-        es['customers'].add_variable('new')
-    assert 'Must provide data to infer type' in str(excinfo)
-
-    es['customers'].add_variable('new', data=[1, 2, 3])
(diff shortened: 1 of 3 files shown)


featuretools/tests/entityset_tests/test_entity.py
```diff
@@ -72,5 +72,26 @@ def test_parents(es):
...
+    with pytest.raises(ValueError) as excinfo:
+        es['customers'].update_data(df.drop(columns=['cohort']))
+    assert 'Updated dataframe is missing new cohort column' in str(excinfo)
+
+    with pytest.raises(ValueError) as excinfo:
+        es['customers'].update_data(df)
+    assert 'Updated dataframe contains 13 columns, expecting 12' in str(excinfo)
+
+
+def test_add_variable(es):
```
kmax12: rather than add a test for add_variable, let's just remove the function. it's only used in one place internally and we can probably avoid needing it at all

Seth-Rothschild:
It seems it's not possible to satisfy the conditional
```
if self.index is not None and self.index not in inferred_variable_types
```
which used `add_variables`. Either you set the index as the index of the entity and `self.infer_variable_types` will catch it or you set the variable type as `variable_types.Index`, in which case it will be automatically included in the `inferred_variable_types` dictionary.

If there is a situation which can happen, it would be easy to add the relevant lines from `add_variable` under this conditional.

kmax12:
@Seth-Rothschild ya, that makes sense to me. let's remove that code block and the add_variables function

kmax12:
Looks good. Merging

PR #264, opened 2018-09-19
Handle no child data when calculating aggregation features with multiple arguments

kmax12:
This pull requests updates how we handle calculating aggregation features when there is no child data. We fixed this for aggregations with where features in #258, but this fixes it for when there are multiple arguments to the agg feature.

This PR fixes the second issue reported in #252

### DIFF
--- a/featuretools/tests/computational_backend/test_pandas_backend.py
+++ b/featuretools/tests/computational_backend/test_pandas_backend.py
@@ -13,7 +13,6 @@ from featuretools import Timedelta
 from featuretools.computational_backends.pandas_backend import PandasBackend
 from featuretools.primitives import (
     And,
-    Trend,
     Count,
     DirectFeature,
     Equals,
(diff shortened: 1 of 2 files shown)


featuretools/tests/computational_backend/test_pandas_backend.py
```diff
@@ -515,11 +515,18 @@ def test_where_clause_empty_dataframe():
...
+    # create agg feature that requires multiple arguments
+    trend = Trend([es["child"]['value'], es["child"]['time_index']], es["parent"])
 
     # cutoff time before all rows
-    ft.calculate_feature_matrix(entityset=es, features=[count], cutoff_time=pd.Timestamp("12/31/2017"))
+    ft.calculate_feature_matrix(entityset=es, features=[count, count_where, trend], cutoff_time=pd.Timestamp("12/31/2017"))
 
     # cutoff time after all rows, but where clause filters all rows
-    ft.calculate_feature_matrix(entityset=es, features=[count], cutoff_time=pd.Timestamp("1/4/2018"))
+    ft.calculate_feature_matrix(entityset=es, features=[count, count_where, trend], cutoff_time=pd.Timestamp("1/4/2018"))
```
rwedge: Could we add assertions here to confirm we get expected feature values from these matrices?

kmax12: done

Issue #265, opened 2018-09-20
AttributeError: 'float' object has no attribute 'total_seconds'


alexelgier:
I've been getting an error while running calculate_feature_matrix, I'm using a branch that was suggested to me in (https://github.com/Featuretools/featuretools/issues/252), related to a different problem.

It was suggested that perhaps one of my entities timeindex was not datetype, or that there might be missing values in that column, but I've checked my entityset and that's not the case. Sadly I can't share data because of legal reasons, although at the moment I'm trying to reproduce with an entityset filled with random data.

Any help debugging or finding the root of this problem would be appreciated.

Regards,
Alex

Stacktrace: 
```
File "/home/mlgroup/NRM/venv/src/featuretools/featuretools/computational_backends/calculate_feature_matrix.py", line 258, in calculate_feature_matrix
    pass_columns=pass_columns)
  File "/home/mlgroup/NRM/venv/src/featuretools/featuretools/computational_backends/calculate_feature_matrix.py", line 520, in linear_calculate_chunks
    backend=backend)
  File "/home/mlgroup/NRM/venv/src/featuretools/featuretools/computational_backends/calculate_feature_matrix.py", line 342, in calculate_chunk
    training_window=window)
  File "/home/mlgroup/NRM/venv/src/featuretools/featuretools/computational_backends/utils.py", line 34, in wrapped
    r = method(*args, **kwargs)
  File "/home/mlgroup/NRM/venv/src/featuretools/featuretools/computational_backends/calculate_feature_matrix.py", line 316, in calc_results
    profile=profile)
  File "/home/mlgroup/NRM/venv/src/featuretools/featuretools/computational_backends/pandas_backend.py", line 196, in calculate_all_features
    result_frame = handler(group, input_frames)
  File "/home/mlgroup/NRM/venv/src/featuretools/featuretools/computational_backends/pandas_backend.py", line 313, in _calculate_transform_features
    values = feature_func(*variable_data)
  File "/home/mlgroup/NRM/venv/src/featuretools/featuretools/primitives/transform_primitive.py", line 207, in pd_diff
    return grouped_df[bf_name].apply(lambda x: x.total_seconds())
  File "/home/mlgroup/NRM/venv/lib/python3.6/site-packages/pandas/core/series.py", line 3194, in apply
    mapped = lib.map_infer(values, f, convert=convert_dtype)
  File "pandas/_libs/src/inference.pyx", line 1472, in pandas._libs.lib.map_infer
[...]

PR #266, opened 2018-09-24
(WIP) Handling return type when creating features from DatetimeTimeIndex

kmax12:
This PR fixes the issue reported in #265. 

That issue surfaced as a calculation error, but the the underlying problem was incorrectly stacking features that shouldn’t exist. This PR updates the return type so the invalid feature doesn’t get created.

### DIFF
--- a/featuretools/primitives/primitive_base.py
+++ b/featuretools/primitives/primitive_base.py
@@ -11,7 +11,7 @@ from featuretools.utils.wrangle import (
     _check_time_against_column,
     _check_timedelta
 )
-from featuretools.variable_types import Variable
+from featuretools.variable_types import Variable, Datetime, DatetimeTimeIndex
 
 logger = logging.getLogger('featuretools')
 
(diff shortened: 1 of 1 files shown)


featuretools/primitives/primitive_base.py
```diff
@@ -101,6 +101,11 @@ def variable_type(self):
             feature = feature.base_features[0]
             return_type = feature.return_type
 
+            # only the original time index should exist
+            # so make this feature's return type just a Datetime
+            if return_type == DatetimeTimeIndex:
```
rwedge: We'd encounter a similar issue with NumericTimeIndex, I think

kmax12: you're right. added

featuretools/tests/feature_function_tests/test_primitive_base.py
```diff
@@ -96,3 +96,7 @@ def test_return_type_inference_direct_feature(es):
 def test_return_type_inference_time_index(es):
     last = Last(es["log"]["datetime"], es["customers"])
     assert last.variable_type == Datetime
+
+    nti = NumericTimeIndex("datetime", es["log"])  # create a numeric time index variable
```
rwedge: maybe a little cleaner to make a second test using the _with_integer_time_index_ option for `make_ecommerce_entityset` instead

kmax12: done

rwedge:
Looks good!

PR #271, opened 2018-10-01
Load retail demo data using compressed file

WillKoehrsen:
Changed `load_retail` to load a `gz` compressed version of the retail detail from S3. This reduces the size of the downloaded file from 43 MB to 7.3 MB.


WillKoehrsen:
Added a fallback to uncompressed version. This might be necessary if `pandas.read_csv()` cannot read the `gz` compressed version.

kmax12:
Looks good. Merging

Issue #275, opened 2018-10-02
Provide more helpful error message when cutoff time rows are duplicated


WillKoehrsen:
Currently, the error produced when a `cutoff_time` dataframe has duplicate rows and is used in `ft.dfs` is 

`Exception: cannot handle a non-unique multi-index!`

This is not very helpful and it can be frustrating trying to figure out the real issue. 

A better error message would highlight the problem:

`Exception: cutoff time dataframe has duplicate rows.`

WillKoehrsen:
I think providing a helpful error message is better than internally dropping duplicates in `cutoff_time` because the user should probably investigate why the rows are duplicated.

kmax12:
Fixed in #276  closing

PR #276, opened 2018-10-03
Check duplicate rows cutoff times

WillKoehrsen:
Addresses #275 by raising assertion error if there are duplicated rows in `cutoff_time` dataframe.


featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -141,7 +141,8 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
         cutoff_time = pd.DataFrame(map_args, columns=['instance_id', 'time'])
     else:
         cutoff_time = cutoff_time.reset_index(drop=True)
-
+        assert (cutoff_time.duplicated().sum() == 0), \
```
kmax12: Let's make this check more specific. it should check if the instance id and time are duplicated regardless of the pass through columns being different.

WillKoehrsen:
Updated check to make specific for `instance_id` and `time`.

kmax12:
Looks good. Merging

PR #277, opened 2018-10-03
Update tests for python 3.7

gsheni:
- CircleCi will now run tests in python 3.7
- use `fastparquet` instead of `pyarrow` for tests because `pyarrow` does not currently support python 3.7
- fix warning for `pd.concat`
- updated test_requirements.txt package versions to most recent. 
 - ignore `DeprecationWarning` and `PendingDeprecationWarning` in pytest

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -3,4 +3,4 @@ mock==2.0.0
 pytest-xdist==1.23.2
 pytest-cov==2.6.0
 pyarrow>=0.10.0; python_version != '3.7'
-fastparquet>=0.10.0; python_version == '3.7'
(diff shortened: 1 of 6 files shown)


test-requirements.txt
```diff
@@ -2,4 +2,4 @@ pytest==3.5.1
 mock==2.0.0
 pytest-xdist==1.20.1
 pytest-cov==2.5.1
```
rwedge: Can we try updating these packages to their most recent version?

gsheni: Fixed.

rwedge:
Looks good

PR #279, opened 2018-10-07
Allow where clauses on direct features in Deep Feature Synthesis

kmax12:
DFS will automatically add where clauses to aggregation features based on the values in the `interesting_values`property of another variable within that entity. 

This PR allows DFS to add where clauses using the interesting values of a direct feature. To accomplish this I added a variable property to direct features that used to only be defined for identity features.

First reported by @favstats on stackoverflow: https://stackoverflow.com/questions/52673694/specifying-interesting-variables-with-featuretools-does-not-work


favstats:
This fixed my issue! Thank you so much for the very quick help, this is really amazing!

featuretools/tests/testing_utils/mock_ds.py
```diff
@@ -34,6 +34,8 @@ def make_ecommerce_files(with_integer_time_index=False, base_path=None, file_loc
     product_df = pd.DataFrame({'id': ['Haribo sugar-free gummy bears', 'car',
                                       'toothpaste', 'brown bag', 'coke zero',
                                       'taco clock'],
+                               'department': ["food", "electronics", "food",
```
rwedge: Maybe "health" for the toothpaste's department type?

featuretools/tests/dfs_tests/test_deep_feature_synthesis.py
```diff
@@ -544,6 +549,7 @@ def test_where_different_base_feats(es):
         assert hashed not in where_feats
 
 
+# TODO: not clear what this tests
```
rwedge: Should we add this as a backlog issue?

kmax12:
@rwedge addressed your comments. does this look good to merge?

PR #282, opened 2018-10-10
Parallel verbose fixes

rwedge:
This PR adds checks to see if `verbose` is `True` before printing messages about scattering EntitySets to dask workers or persisting them on the scheduler.  It also upgrades the dask and distributed package versions.


kmax12:
Looks good to me

PR #288, opened 2018-10-15
change to fullargsspec

gsheni:
- This fixes the following warning in python 3.7
> primitives/utils.py:100 
DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
```

### DIFF
--- a/featuretools/primitives/utils.py
+++ b/featuretools/primitives/utils.py
@@ -7,10 +7,11 @@ from .primitive_base import PrimitiveBase
 import featuretools.primitives
 from featuretools.utils import is_string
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: featuretools/primitives/utils.py [hunk 1]
```diff
@@ -7,6 +7,11 @@
 import featuretools.primitives
 from featuretools.utils import is_string
 
+try:
```
kmax12: can you add a comment in the code to this block about why were doing this?

gsheni: Done.

kmax12:
looks good. merging

PR #292, opened 2018-10-24
Update circleci tests to check for warnings when building docs

rwedge:
This PR adds a CI test to check the documentation builds without warnings to catch when changes would introduce new errors in the documentation.   It also moves the linting check higher in the order of commands so linting errors will be caught quickly.


rwedge:
This PR depends on the next release of IPython so it won't be ready to merge until then.

kmax12:
Looks good to me. Issue with checks, but the tests are passing if I go directly to circle

PR #293, opened 2018-10-25
Clean up unused methods

kmax12:
Made minor changes that clean up the code. No api changes.

Changes

* removed `Relationship.get_other_variable` because the method wasn't used
* removed `Relationship.get_other_entity` because it was only used one. Added logic to `Entity.related_instances` instead
* removed `Relationship.get_entity_variable` because it was only used one. Added logic to `Entity.related_instances` instead
* removed `Relationship._get_link_variable_name` and `EntitySet. gen_relationship_var` because they did the same thing and it isn't really a method of either of those classes. Created method `get_relationship_variable_id` in our utils file instead. 
* removed unused methods `_check_variable_list`,  `_check_variable` , and `_check_entity` from wrangle.py
* remove training_window_is_dict handling in related instances that is not possible
* removed unused methods `Entity.is_child_of` and `Entity.is_parent_of` 
* remove unused method `EntitySet.get_relationships`
* remove unused method `make_index_variable_name`
* remove unused class `RedirectStdStreams`
* remove invalid property `Discrete.percent_unique`. this property access other properties that are no longer defined on `Variable`

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -25,7 +25,10 @@ from .utils import (
 )
 
 from featuretools.primitives import AggregationPrimitive, PrimitiveBase
-from featuretools.utils.gen_utils import make_tqdm_iterator
+from featuretools.utils.gen_utils import (
+    get_relationship_variable_id,
+    make_tqdm_iterator
+)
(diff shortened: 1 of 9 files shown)


rwedge:
Did the `training_window_is_dict` change make it into the PR?

kmax12:
you're right, it didn't. just fixed

PR #299, opened 2018-10-30
Move pympler to test requirements

rwedge:
* Remove pympler import in `gen_utils` file and import it directly in testing files that use it
* Move pympler requirement from general requirements to testing requirements.


kmax12:
Looks good to me. Go ahead and merge

PR #303, opened 2018-10-31
Fixed #297 update tests to check error strings

### LINKED ISSUE #297: Update tests to check error strings
To test if Featuretools errors correctly, we have many tests like this
```
def test_raise_key_error_missing_entity(es):
    with pytest.raises(KeyError):
        es["this entity doesn't exist"]
```

The test checks to make sure the right error type is thrown but now the message itself. We can test for the message itself like this

```
def test_find_path_no_path_found(es):
    es.relationships = []
    error_text = "No path from products to customers. Check that all entities are connected by relationships"
    with pytest.raises(ValueError, match=error_text):
        es.find_path('products', 'customers')
```

We should update all instances of ` pytest.raises(` to include the match parameter.

jeff-hernandez:
- On windows platform, there is an [open issue](https://github.com/pandas-dev/pandas/issues/15086) currently in pandas where it raises an error when reading a file with accents in the file path (i.e. `régions.csv`). So, I resolved it with the following:

```python
# featuretools\tests\testing_utils\mock_ds.py:334
df = pd.read_csv(open(filenames[entity], 'r', encoding='utf8'), encoding='utf-8')
```

- This snippet `np.dtype((np.integer, np.floating)).type` was causing this [issue](https://github.com/Featuretools/featuretools/issues/232). So, I resolved it by changing it to the following:
```python
np.issubdtype(time, np.integer) or np.issubdtype(time, np.floating)
```
- Not sure how to get the error text for `test_not_enough_memory`


featuretools/tests/entityset_tests/test_timedelta.py
```diff
@@ -61,9 +62,9 @@ def test_delta_with_observations(es):
     large_delta = Timedelta(99999, 'observations', 'log')('customers',
                                                           instance_id=0,
                                                           entityset=es)
-    with pytest.raises(NotEnoughData):
+    with pytest.raises(NotEnoughData, match=''):
```
kmax12: no need to have match text here. In this case `NotEnoughData` is already specific enough

jeff-hernandez:
Thanks @kmax12. I updated the commit. Let me know if anything else needs to be changed.

kmax12:
@jeff-hernandez looks like python 2.7 tests are failing because `encoding` isn't a valid parameter to `open` in python 2.7.

kmax12:
does the following work for `test_add_relationship_errors_on_dtype_mismatch` to handle unicode?

```python
error_text = u'Unable to add relationship because id in régions is Pandas dtype object and session_id in log2 is Pandas dtype int64.'
with pytest.raises(ValueError, match=error_text):
    mismatch = Relationship(entityset[u'régions']['id'], entityset['log2']['session_id'])
    entityset.add_relationship(mismatch)
```
other than that, it's good to merge?

jeff-hernandez:
@kmax12 I tried the code snippet. This is the error I get for python 2.7:

```
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe9' in position 42: ordinal not in range(128)
```

kmax12:
@jeff-hernandez im looking into this now. weirdly it's only a problem in python 2.7.

kmax12:
actually, here's an easier fix. let's just change the entity for this test case. I think that this solution is fine since we aren't trying to test unicode stuff with this test case. 

So, it can just be 
```
error_text = u'Unable to add relationship because id in customers is Pandas dtype category and session_id in log2 is Pandas dtype int64.'
    with pytest.raises(ValueError, match=error_text):
        mismatch = Relationship(entityset[u'customers']['id'], entityset['log2']['session_id'])
        entityset.add_relationship(mismatch)
```

jeff-hernandez:
Okay sounds good. I made the adjustment.

jeff-hernandez:
Okay sounds good. I made the adjustment.

kmax12:
this is look good. actually one more stylistic change. In many places, you removed a blank line before the error test. can you add those back in? 

for example

screen shot 2018-11-02 at 1 00 14 pm

after that, we're good to merge

jeff-hernandez:
No problem. I updated the adjustment. Let me know if anything else needs modification. Thanks.

PR #305, opened 2018-10-31
Importing from SQL#248

bentona:
Addresses https://github.com/Featuretools/featuretools/issues/248

Adds `featuretools.io.from_sql()`

 * This function takes an `id` and `connection_string` or an SQLAlchemy connection object `connection`.
 * Ingests each table (all tables by default, or a subset specified by `tables` parameter) as an entity, using database fk constraints to infer and generate relationships.
 * Uses the primary key of the table as the `id` for the Entity, or generate one if not available.

Other parameters for `entity_from_dataframe()` can also be passed using a table-keyed dictionary. Ex: 
```python
> from_sql(id='my_es', connection=my_connection, variable_types={"widgets": {'name': ft.variable_types.Categorical}}
```

There are a few requests for feedback in-line, but I'm especially eager for thoughts about:
 * Handling generating an index for Entities without corresponding primary keys (right now I'm just creating a 'featuretools_sql_import_id' field)
 * Opinions on using an instantiated helper class (current) vs. one function vs. many functions


bentona:
> rename it more simply to just `io` (input/output). [...] We can also move (serialization) to the `io` submodule in the future.

Makes sense, I like that plan.

> it's not immediately clear to me the advantage of the class here. why not just put all that implementation inside of the `entity_set_from_sql` function?

I'm not too picky about the specific abstraction, I just reached for a class since there is a good amount of state manipulation. It's a ton of functionality to fit in one function, but I think other functions would work to break it down as well as a class, there would just be some state passing around. I'm just hesitant to leave it all in a `do_it_all_function` since those are typically rough to maintain.

> you can add the fixture to a new file in`tests/testing_utils/`, if you'd like.

👍 

> 1. We should raise an error if a table's primary key is actually the combination of two columns.
> 2. perhaps the function should just be called `featuretools.io.from_sql`. I think the entity set part is probably implied. we can discuss this later
> 3. We should add a verbose parameter to the method so that it prints out it's progress of loading tables. we can do this easily by wrapping the looping over tables with TQDM (https://github.com/tqdm/tqdm)

Sounds good - I'll add these pieces towards the end, probably.

featuretools/io/sql.py
```diff
@@ -0,0 +1,79 @@
...
+    loader.load_all_tables()
+    loader.load_relationships()
+    return loader.es
+
+
+def _find_pk_name(table):
+    column_names = table.columns.keys()
+    pks = [name for name in column_names if table.columns[name].primary_key]
+    if len(pks) > 1:
+        raise RuntimeError('Composite primary key detected')
```
bentona: a) is there a better idiom than this `if`/`else`?
b) is there a more specific type of error we can raise?

featuretools/io/sql.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def _get_table_names(self):
+        return list(self.metadata.tables.keys())
+
+    def _load_table(self, table_name):
+        table = sa.Table(table_name, self.metadata, autoload=True)
+        index_name = _find_pk_name(table)
+        make_index = index_name is None
+        if make_index:
+            index_name = 'featuretools_sql_import_id'
```
bentona: I'm supposing it is best to generate an index in the case where we don't detect a pk - thoughts about this or the naming convention?

bentona:
Pushed more updates, I believe we're getting close now. A few questions for you in-line - other outstanding TODOs, let me know if I missed anything.

1. Add TQDM 
2. Test against larger datasets
3. Determine if we should
  a) Leave the class/object helper
  b) Move functionality into a collection of pure-ish functions
  c) Move all functionality into `from_sql`
4. Update docs

re: 3, I'm partial to (a) or (b), but can move towards (c) if you feel strongly.

featuretools/tests/io_tests/test_from_sql.py
```diff
@@ -0,0 +1,39 @@
+# flake8: noqa: F811
```
bentona: It looks like using fixtures in this way causes a linting error, and I see some other people on SO with the same issue :/

featuretools/tests/io_tests/test_from_sql.py
```diff
@@ -0,0 +1,39 @@
...
+import pytest
+
+import featuretools as ft
+
+from ..testing_utils import sqlite, sqlite_composite_pk  # noqa: F401; pylint: disable=unused-variable
```
bentona: see above

featuretools/io/sql.py
```diff
@@ -4,9 +4,15 @@
 import featuretools as ft
 
 
-def from_sql(id, connection):
-    loader = EntitySetLoaderFromSQL(connection=connection, id=id)
-    loader.load_all_tables()
+def from_sql(id, connection=None, connection_string=None, tables=None, **kwargs):
```
bentona: Is using `**kwargs` here the appropriate approach for passthrough?

featuretools/tests/io_tests/test_from_sql.py
rwedge: Instead of defining the fixtures in `testing_utils` and importing them we could create a `featuretools/tests/io_tests/conftest.py` file and define the fixtures there. From the looks of [this documentation](https://docs.pytest.org/en/latest/example/simple.html#package-directory-level-fixtures-setups) we could then use the fixtures in this file without needing to import them.

bentona: Awesome - that worked, updated.

bentona:
@kmax12 I've been busy and this fell off my radar for a bit. I think I've address the remaining outstanding issues before merge. We now explicitly list the pass-through parameters at the first layer, but retain `kwargs` internally. I figure it's best to prevent as much duplication as possibly while ensuring maximum usability. 

This branch has gotten a bit out of hand, I'm not used to working off an upstream repo, I'd like to pull these changes in to a fresh branch before merging, but I wanted to leave all the comments/context until I get the OK.

Let me know how everything looks!

featuretools/io/sql.py
```diff
@@ -0,0 +1,100 @@
+import pandas as pd
+import sqlalchemy as sa
+
+import featuretools as ft
+
+
+def from_sql(id, connection=None, connection_string=None, tables=None, index={},
```
kmax12: for the api, let's combine `connection` and `connection_string` into one argument. Then we can infer if it is a connection object or string. For example, see how pandas does it here: https://pandas.pydata.org/pandas-docs/stable/generated/pandas.read_sql_table.html#pandas.read_sql_table.

kmax12: need to add a docstring for this method. Look at any some of our public methods for the formatting (`encode_features` is a good one to look at). Would be nice to include an example section based on one of the test cases, as well. 

Also, add this method to `api_reference.rst` in a new section title "Input/Output". By including it the api reference, a page will be automatically for it when build the docs, so you can see how it looks. Can you also somehow indicate that this method is "Beta" when you add?

featuretools/io/sql.py
```diff
@@ -0,0 +1,100 @@
...
+    pks = [name for name in column_names if table.columns[name].primary_key]
+    if len(pks) > 1:
+        raise RuntimeError('Composite primary key detected')
+    elif len(pks) == 1:
+        return pks[0]
+    else:
+        return None
+
+
+class EntitySetLoaderFromSQL:
```
kmax12: this is an internal class, but brief documentation of each method would be helpful for maintainablility

featuretools/io/sql.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def _get_args_for_table(self, table_name):
+        return {k: v.get(table_name) for k, v in self.passthrough_args.items()}
+
+    def _load_table(self, table_name):
+        table = sa.Table(table_name, self.metadata, autoload=True)
+        df = pd.read_sql(table.select(), table.metadata.bind)
+
+        table_args = self._get_args_for_table(table_name)
+        print("table_args:", table_args)
```
kmax12: either remove this print statement or wrap it in `if verbose:` like we do in other places

featuretools/io/sql.py
```diff
@@ -0,0 +1,100 @@
...
+
+        return self.es.entity_from_dataframe(
+            entity_id=table_name,
+            dataframe=df,
+            **table_args
+        )
+
+    def _column_fk_list(self, table, column_name):
+        fks = table.columns[column_name].foreign_keys
+        return list(filter(lambda fk: fk, fks))
```
kmax12: not sure i understand this line.

i think something might be more pythonic / easier to read `[fk for fk in fks if fk is not None]`

featuretools/io/sql.py
```diff
@@ -0,0 +1,100 @@
...
+        fks = {column_name: self._column_fk_list(table, column_name) for column_name in column_names}
+        return fks
+
+    def load_relationships(self):
+        for table_name in self.table_names:
+            fk_dict = self._load_table_fks(table_name)
+            child_table = table_name
+            for column_name, fks in fk_dict.items():
+                child_column = column_name
+                child_table = table_name
```
kmax12: this looks like the same line as two above. not sure it does anything since table name doesn't change inside this lop

featuretools/io/sql.py
```diff
@@ -0,0 +1,100 @@
...
+        fks = table.columns[column_name].foreign_keys
+        return list(filter(lambda fk: fk, fks))
+
+    def _load_table_fks(self, table_name):
+        table = sa.Table(table_name, self.metadata, autoload=True)
+        column_names = table.columns.keys()
+        fks = {column_name: self._column_fk_list(table, column_name) for column_name in column_names}
+        return fks
+
+    def load_relationships(self):
```
kmax12: do we handle the case where you only include one of the tables that's part of a relationship? 

say we have table A which is a child of table B. but a user only want to load in table A. do we handle the situation where we try to add the relationship for A to B, but B doesn't exist?

featuretools/io/sql.py
```diff
@@ -0,0 +1,100 @@
...
+                child_column = column_name
+                child_table = table_name
+                for fk in fks:
+                    parent_column = fk.column.name
+                    parent_table = fk.column.table.name
+                    self._add_relationship(parent_table, parent_column, child_table, child_column)
+
+    def load_tables(self, included_tables):
+        if included_tables is not None:
+            tables_to_load = set(self.table_names) & set(included_tables)
```
kmax12: rather than take the &, let's just check to make sure everything in `included_tables` is in `self.table_names`. if it's not raise a helpful error pointing that out.

featuretools/io/sql.py
bentona: good catch

featuretools/io/sql.py
bentona: I believe this testcase handles what you're thinking about:
https://github.com/Featuretools/featuretools/pull/305/files#diff-9d3216b92b716cc50a8429e1d3350b4bR44

specifically, since we load widgets & factories, but don't load customers, only the widgets <> factories relationship is loaded. 

Let me know if there's another case we should test.

kmax12: here is the case

```
def test_load_only_child_in_relationship(sqlite):
    es = ft.io.from_sql(id="widget_limited_tables", connection=sqlite, tables=["customers_widgets"])
    assert(len(es.relationships) == 0)
    assert(len(es.entities) == 1)
```

when I run this I get the following error 
```
*** KeyError: 'Entity customers does not exist in widget_limited_tables'
```

featuretools/io/sql.py
```diff
@@ -0,0 +1,107 @@
+import pandas as pd
+import sqlalchemy as sa
+
+import featuretools as ft
+from featuretools.utils.gen_utils import is_string
+
+
+def from_sql(id, connection=None, tables=None, index={},
```
kmax12: shouldn't use empty dictionaries as arguments to python functions

more info why here: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

rwedge:
Closing this PR for now due to inactivity

Issue #248, opened 2018-09-10
Integration with SQL?


kmatusz:
### Feature Request Description
The concept of EntitySet is similar to SQL database, with relations as foreign keys and entities as tables. I think a good feature would be to connect featuretools directly to an SQL database, or at least provide conversion of  a database to EntitySet automatically.

kmax12:
@kmatusz this makes a lot of sense and is something on our list to add to Featuretools. One approach could be to use SQLAlchemy to implement this. 

If you (or anyone else reading this) is interested in contributing this functionality to Featuretools, please let us know so we can discuss further.

jkgenser:
@kmax12 I would be very interested help with contributing this functionality to featuretools.

gsheni:
@kmatusz Please checkout our add-on library which allows you to create EntitySets from a database:
- https://github.com/alteryx/featuretools-sql

PR #308, opened 2018-11-07
Resolve bug preventing using first column as index by default

bentona:
While working on #248, I hit a snag on handling index creation, and I noticed that this `elif` clause seems to never get hit.

It looks like `None` is not `in` every list, so when `index` is `None`, `index not in dataframe.columns` is always false unless `None` is explicitly in the list. 

I took the opportunity to flatten out the logic a little bit, it seems safe to pull the index setting into the previous check for a `None` index.


kmax12:
good find, thanks! 

before merging, we should add a test case for inferring the first column is the index. This should be defined in `test_es.py` near the other tests about the index

bentona:
@kmax12 Hey, just added a (failing) test. The actual error is a bit incomprehensible, so given our earlier discussion about confusing error messages, along with the fact it wasn't getting used enough to trigger any other bug reports, I thought you might just want to yank the feature. Either way I didn't want to spin my wheels too much on this before wrapping up the rest of the SQL feature - let me know how to proceed.

kmax12:
it looks like it's only failing test because of linting

just three issues related to whitespace

```
featuretools/tests/entityset_tests/test_es.py:214:53: E231 missing whitespace after ','
featuretools/tests/entityset_tests/test_es.py:221:62: E231 missing whitespace after ','
featuretools/tests/entityset_tests/test_es.py:221:66: E231 missing whitespace after ','
```

Issue #315, opened 2018-11-12
Add details about serialization features to the documentation


kmax12:
Features saved in one version of Featuretools may not be able to be loaded in another version of Featuretools. 

We should add this to the documentation for `ft.save_features` and `ft.load_features` as well as the [deployment](https://docs.featuretools.com/guides/deployment.html) page.

WillKoehrsen:
I opened a pull request  #316 to address this. It updates the `save_features` and `load_features` functions docstrings as well as the deployment page. Is there anywhere else we should alert users to this behavior?

WillKoehrsen:
Closed by #316. This provides documentation alerting the user to behavior but does not address the stability issue.

Issue #311, opened 2018-11-10
Feature definitions not compatible between v0.3.0 and newer versions.


WillKoehrsen:
This might be a known issue already, but I have a set of features created in Featuretools v0.3.0 that do not work when I use `ft.load_features` in v0.3.1. I have uploaded the set of feature definitions. 

The following code works fine in v0.3.0 but not in v0.3.1:

`f = ft.load_features('features-2')`

The error is : `TypeError: 'str' object does not support item assignment`. I can post a full stack trace if necessary. 

[features-defs.zip](https://github.com/Featuretools/featuretools/files/2568401/features-defs.zip)

WillKoehrsen:
The features also do not load in v0.4.0.

kmax12:
Features saved in one version of Featuretools may not be able to be loaded in another version of Featuretools due to our usage pickle.

We should add this to the documentation for `ft.save_features` and `ft.load_features` as well as the deployment page.

Open issue #315 to fix this.

WillKoehrsen:
#316 helps on this issue by alerting users to incompatibility, however, do we want to try and make features compatible between versions? At the very least, the warning when users try to load in features from a different version should mention the incompatibility. Currently, the error is not very useful.

rwedge:
I think the serialization changes and use of schema versioning added since the issue was opened have addressed the problem.

PR #316, opened 2018-11-14
Feature stability between versions documentation

WillKoehrsen:
Temporarily addresses #315 and #311 by alerting users that features created in one version of Featuretools are not necessarily compatible with other versions. A longer-term fix would be advised.


kmax12:
Looks good. Merging

PR #318, opened 2018-11-14
Handle return type when creating features from Id variables

kmax12:
When we create a feature from an Id variable, its return type should get converted to a Categorical because it is no longer an Id variable unless it is a direct feature. 

This will remove features like `CUM_SUM(SUM(order_products.total) by MODE(order_products.product_id))`

Same as #266, but for Id variable types

### DIFF
--- a/featuretools/primitives/primitive_base.py
+++ b/featuretools/primitives/primitive_base.py
@@ -12,8 +12,10 @@ from featuretools.utils.wrangle import (
     _check_timedelta
 )
 from featuretools.variable_types import (
+    Categorical,
     Datetime,
     DatetimeTimeIndex,
+    Id,
     Numeric,
     NumericTimeIndex,
     Variable
@@ -113,6 +115,8 @@ class PrimitiveBase(object):
                 return_type = Datetime
             elif return_type == NumericTimeIndex:
                 return_type = Numeric
+            elif return_type == Id:
+                return_type = Categorical
 
         return return_type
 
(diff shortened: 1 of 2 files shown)


featuretools/tests/feature_function_tests/test_primitive_base.py
```diff
@@ -106,3 +106,8 @@ def test_return_type_inference_datetime_time_index(es):
 def test_return_type_inference_numeric_time_index(es_numeric):
     last = Last(es_numeric["log"]["datetime"], es_numeric["customers"])
     assert last.variable_type == Numeric
+
+
+def test_return_type_inference_id(es_numeric):
+    mode = Mode(es_numeric["log"]["session_id"], es_numeric["customers"])
+    assert mode.variable_type == Categorical
```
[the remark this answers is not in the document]
kmax12: sure. added

featuretools/primitives/primitive_base.py
```diff
@@ -100,12 +102,14 @@ def entity(self):
     # P TODO: this should get refactored to return_type
     @property
     def variable_type(self):
+        from . import direct_feature
```
rwedge: Should we move this import up to the top?

kmax12: direct_feature imports this file, so we have to include this here to avoid the circular reference. 

in #326, the directory structure changes things around and actually moves `DirectFeature` into this file, so we'll likely be able to update this when that PR gets merged in.

rwedge: makes sense

PR #321, opened 2018-11-16
Handle primitives with same function being applied to same column

kmax12:
Pandas relies on the function name when calculating aggregations. This means if a two primitives with same function name are applied to the same column, pandas can't differentiate them. We have a work around to this based on the name property that we test here.


WillKoehrsen:
Looks good. Merging.

PR #324, opened 2018-11-20
fixed #294 - optional id parameter of entityset

### LINKED ISSUE #294: Make id an optional parameter to EntitySet constructor
Currently, users must provide an `id` to initialize an `EntitySet`.

```
import featuretools as ft
ft.EntitySet(id="my_entity_set")
```

This is only used by the `repr` of the EntitySet, so we shouldn't require users to provide it. For most users, this is the first line of Feautretools code they write, so we shouldn't have arguments that aren't necessary.  

To fix this issue we should 

1. make `id` an optional argument to the `EntitySet` constructor 
2. allow methods of entityset that use the id to function without a provided id
3. Edit examples in the Index and "Representing Data with EntitySets" of the documentation to not require using an `id`. 
4. make sure we have a test case when `id` isn't provided

jeff-hernandez:
I am not sure if I missed any examples in the documentation. Let me know if anything needs modification. Thanks.

### DIFF
--- a/featuretools/tests/entityset_tests/test_es.py
+++ b/featuretools/tests/entityset_tests/test_es.py
@@ -934,3 +934,35 @@ def test_sizeof(entityset):
         total_size += entity.last_time_index.__sizeof__()
 
     assert entityset.__sizeof__() == total_size
+
+
+def test_construct_without_id():
+    assert ft.EntitySet().id is None
+
+
+def test_repr_without_id():
+    match = 'Entityset: None\n  Entities:\n  Relationships:\n    No relationships'
+    assert repr(ft.EntitySet()) == match
+
+
+def test_getitem_without_id():
+    error_text = 'Entity test does not exist'
+    with pytest.raises(KeyError, match=error_text):
+        ft.EntitySet()['test']
+
+
+def test_metadata_without_id():
+    es = ft.EntitySet()
+    assert es.create_metadata_dict().get('id') is None
+
+
+def test_to_pickle_id_none():
+    entityset = ft.EntitySet()
+    dirname = os.path.dirname(integration_data.__file__)
+    path = os.path.join(dirname, 'test_entityset.p')
+    if os.path.exists(path):
+        shutil.rmtree(path)
+    entityset.to_pickle(path)
+    new_es = ft.read_pickle(path)
+    assert entityset.__eq__(new_es, deep=True)
+    shutil.rmtree(path)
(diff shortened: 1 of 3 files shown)


kmax12:
The only place in the code base I see an entityset's id used is in serialization/deserialization and in the repr methods. Can you also add tests to ensure those work?  

also, looks like linting is failing. run `make lint` locally to see error

jeff-hernandez:
A list of methods of the class `EntitySet` I found that contain the `id` property (e.g. `self.id`):
- `__init__`
- `__repr__`
- `__getitem__`
- `create_metadata_dict`

I wrote tests for the methods listed above. I hope that I got all the methods. Let me know if I missed anything. Thanks.

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -934,3 +934,23 @@ def test_sizeof(entityset):
...
+    assert ft.EntitySet().id is None
+
+
+def test_repr_without_id():
+    match = 'Entityset: None\n  Entities:\n  Relationships:\n    No relationships'
+    assert repr(ft.EntitySet()) == match
+
+
+def test_getitem_without_id():
+    error_text = 'Entity test does not exist in None'
```
kmax12: Let change this error text to not even reference `self.id`. So it can just say `'Entity test does not exist'`

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -934,3 +934,23 @@ def test_sizeof(entityset):
...
+    with pytest.raises(KeyError, match=error_text):
+        ft.EntitySet()['test']
+
+
+def test_metadata_without_id():
+    es = ft.EntitySet()
+    assert es.create_metadata_dict().get('id') is None
```
kmax12: the metadata dict is also used when serializing and loading an entityset. can we add one test that serializes and loads an entityset without an id? 

something like

```python
def test_to_pickle_id_none(entityset):
    entityset = ft.EntitySet()
    dirname = os.path.dirname(integration_data.__file__)
    path = os.path.join(dirname, 'test_entityset.p')
    if os.path.exists(path):
        shutil.rmtree(path)
    entityset.to_pickle(path)
    new_es = ft.read_pickle(path)
    assert entityset.__eq__(new_es, deep=True)
    shutil.rmtree(path)
```

jeff-hernandez: Yes not a problem. I made the adjustments described above. Let me know if anything else needs modification. Thanks.

kmax12:
Looks good. Merging

Issue #313, opened 2018-11-11
Import error


Fannasankh:
module 'tornado.platform.asyncio' has no attribute 'AnyThreadEventLoopPolicy'

-----
*Issues created here on Github are for bugs or feature requests. For usage questions and questions about errors, please ask on Stack Overflow with the [featuretools](https://stackoverflow.com/questions/tagged/featuretools) tag. Check the [documentation](https://docs.featuretools.com/#get-help) for further guidance on where to ask your question.*

kmax12:
@Fannasankh can you share a reproducible example and details on your environment?

Fannasankh:
I use Google Colab
example https://colab.research.google.com/drive/1pmXx4pEn4yw_pI5SU_wd20vPlVQl4Dvx

Fannasankh:
> @Fannasankh can you share a reproducible example and details on your environment?
I use Google Colab
example https://colab.research.google.com/drive/1pmXx4pEn4yw_pI5SU_wd20vPlVQl4Dvx

Fannasankh:
> @Fannasankh can you share a reproducible example and details on your environment?

I use Google Colab
example https://colab.research.google.com/drive/1pmXx4pEn4yw_pI5SU_wd20vPlVQl4Dvx

stobasa:
Am having the same error. I use Anaconda environment on Ubuntu.

rwedge:
Someone posted about this on stack overflow: https://stackoverflow.com/questions/53237312/import-featuretools-and-attributeerror-module-tornado-platform-asyncio-has-no

If I install a version of tornado lower than 5 I get this error when I try to import featuretools

Fannasankh:
I tried version 5.1, but it did not help on Google Colab

kmax12:
@Fannasankh this looks like an issue with the Google Colab environment supporting tornado 5.1. Looks like after you upgrade there is problem simply importing Tornado. 

@rwedge put up a PR to distributed (https://github.com/dask/distributed/issues/2351), which is the library we depend on, to fix the source of this issue. 

Is it an option to run the code locally or in a different environment for the time being?

Fannasankh:
It works on version 4.5 and 5.1 locally, but updating a version of a tornado on google colab does not help

Fannasankh:
And when i updating tornado on Google Colab to version 5.0/5.1 i have error "ImportError: cannot import name 'future_set_exc_info'"

Fannasankh:
[...]

PR #328, opened 2018-11-28
Update requirements

rwedge:
This updates the required versions of two dependencies, distributed and pympler.  This PR downgrades Pympler from 0.6 to 0.5 because 0.6 is not available on conda.  This PR upgrades the distributed requirement to 1.24.2, which includes the bugfix that addresses #313


kmax12:
Looks good to me

PR #320, opened 2018-11-14
Update properties of cumulative transform primitives

kmax12:
Cumulative transform primitives had an incorrectly define property. It was named `needs_all_values` when it should have been `uses_full_entity`. 

This PR fixes that and adds a test.

fixes #317  reported by @MarekHauzr

PR #329, opened 2018-11-29
v0.4.1

### LINKED ISSUE #297: Update tests to check error strings

rwedge:
**v0.4.1** Nov 29, 2018
    * Resolve bug preventing using first column as index by default (#308)
    * Handle return type when creating features from Id variables (#318)
    * Make id an optional parameter of EntitySet constructor (#324)
    * Handle primitives with same function being applied to same column (#321)
    * Update requirements (#328)
    * Clean up DFS arguments (#319)
    * Clean up Pandas Backend (#302)
    * Update properties of cumulative transform primitives (#320)
    * Feature stability between versions documentation (#316)
    * Add download count to GitHub readme (#310)
    * Fixed #297 update tests to check error strings (#303)
    * Remove usage of fixtures in agg primitive tests (#325)

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -2,6 +2,20 @@
 
 Changelog
 ---------
+**v0.4.1** Nov 29, 2018
+    * Resolve bug preventing using first column as index by default (:pr:`308`)
+    * Handle return type when creating features from Id variables (:pr:`318`)
+    * Make id an optional parameter of EntitySet constructor (:pr:`324`)
+    * Handle primitives with same function being applied to same column (:pr:`321`)
+    * Update requirements (:pr:`328`)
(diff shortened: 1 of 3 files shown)


kmax12:
Looks good to release!

PR #332, opened 2018-11-30
Add readme to pypi page

### LINKED ISSUE #331: Fix Featuretools PyPI Page
The [current PyPI page](https://packaging.python.org/guides/making-a-pypi-friendly-readme/) for Featuretools doesn't how our readme.

There are some instructions to fix that here: https://packaging.python.org/guides/making-a-pypi-friendly-readme/

jeff-hernandez:
I updated `setup.py` to include `README.md`. Let me know if anything needs modification. Thanks.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,5 @@
 include *.txt
 include LICENSE
+include README.md
 recursive-exclude * __pycache__
 recursive-exclude * *.py[co]
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,10 @@
+from os import path
 from setuptools import find_packages, setup
 from setuptools.command.build_ext import build_ext as _build_ext
+
+dirname = path.abspath(path.dirname(__file__))
+with open(path.join(dirname, 'README.md')) as f:
+    long_description = f.read()
 
 
 # Bootstrap numpy install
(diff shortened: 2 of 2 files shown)


jeff-hernandez:
During continuous integration with Python 3.5, the shell doesn't seem to find the correct directory for the `README.md`. I used the suggested method below:
```
this_directory = path.abspath(path.dirname(__file__))
```
I get the following error:

```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pip-req-build-omei4hwd/README.md'
```

I will experiment with different methods to input the `README.md`

kmax12:
I would expect `__file__` to point to `__init__.py`, which is one level lower then the readme.

jeff-hernandez:
On other python versions, there isn't an error with finding `open('README.md')`, but there is a problem in python version 3.5. The error occurs during the command `tox -e py35`.

kmax12:
does it occur if you `pip install` in python 3.5 or is it specific to tox?

jeff-hernandez:
It might be specific to tox. The occurs during the `tox` command in the continuous integration process on CircleCI.

kmax12:
can you replicate locally with or without tox?

jeff-hernandez:
I could try to replicate the same error locally with and without tox.

jeff-hernandez:
Before replicating the error, I found a solution on [stackoverflow](https://stackoverflow.com/a/47941626/5386567). Essentially, we need to include the `README.md` inside the  `MANIFEST.in`. Let me know if anything needs modification. Thanks.

kmax12:
good find. I just tested uploading it to the test pypi server: https://test.pypi.org/project/featuretools/0.4.2.dev0/. I can confirm it works. 

Merging now. Will go out in next release!

PR #333, opened 2018-11-30
Remove parse_date_cols parameter from EntitySet._import_from_dataframe

rwedge:
This keyword argument is in a private method but is never used by other Featuretools methods.  All it does is add the columns to `variable_types` as Datetime types, but `variable_types` is already a parameter so they could just be included in `variable_types` instead.


kmax12:
Looks good to merge

PR #336, opened 2018-12-03
Clean up entity creation logic

rwedge:
The idea behind this PR is to make it easier to find where in the code we handle data type and variable type conversion when creating a new entity.  Currently it's handled in two places: `EntitySet._import_from_dataframe` and `Entity.__init__`.  

This PR moves the relevant logic in `EntitySet._import_from_dataframe` into `Entity.__init__` and removes a couple statements that try to infer variable types that can be handled by `Entity.infer_variable_types` instead.

### DIFF
--- a/featuretools/entityset/entity.py
+++ b/featuretools/entityset/entity.py
@@ -69,25 +69,21 @@ class Entity(object):
         _validate_entity_params(id, df, time_index)
         created_index, index, df = _create_index(index, make_index, df)
 
-        self.data = {"df": df,
-                     "last_time_index": last_time_index,
-                     }
         self.encoding = encoding
         self._verbose = verbose
         self.created_index = created_index
         self.id = id
         self.entityset = entityset
+        self.data = {'df': df, 'last_time_index': last_time_index}
 
         variable_types = variable_types or {}
         secondary_time_index = secondary_time_index or {}
         self._create_variables(variable_types, index, time_index, secondary_time_index)
 
+        self.df = df[[v.id for v in self.variables]]
         self.set_index(index)
         self.set_time_index(time_index, already_sorted=already_sorted)
         self.set_secondary_time_index(secondary_time_index)
-        self.update_data(df=self.df,
-                         already_sorted=already_sorted,
-                         recalculate_last_time_indexes=False)
 
     def __repr__(self):
         repr_out = u"Entity: {}\n".format(self.id)
--- a/featuretools/entityset/entity.py
+++ b/featuretools/entityset/entity.py
@@ -67,7 +67,7 @@ class Entity(object):
                 Otherwise, assume index exists in dataframe.
         """
         _validate_entity_params(id, df, time_index)
-        created_index, index, df = create_index(index, make_index, df)
+        created_index, index, df = _create_index(index, make_index, df)
         if index not in variable_types:
             variable_types[index] = vtypes.Index
 
@@ -675,7 +675,8 @@ def col_is_datetime(col):
     return False
 
 
-def create_index(index, make_index, df):
+def _create_index(index, make_index, df):
+    '''Handles index creation logic'''
     created_index = None
     if index is None:
         assert not make_index, "Must specify an index name if make_index is True"
--- a/featuretools/entityset/entity.py
+++ b/featuretools/entityset/entity.py
@@ -66,15 +66,7 @@ class Entity(object):
                 dataframe, and create a new column of that name using integers the (0, len(dataframe)).
                 Otherwise, assume index exists in dataframe.
         """
-        assert is_string(id), "Entity id must be a string"
-        assert len(df.columns) == len(set(df.columns)), "Duplicate column names"
-        for c in df.columns:
-            if not is_string(c):
-                raise ValueError("All column names must be strings (Column {} "
-                                 "is not a string)".format(c))
-        if time_index is not None and time_index not in df.columns:
-            raise LookupError('Time index not found in dataframe')
-
+        _validate_entity_params(id, df, time_index)
         created_index, index, df = create_index(index, make_index, df)
         if index not in variable_types:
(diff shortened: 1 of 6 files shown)


featuretools/entityset/entityset.py
```diff
@@ -1142,40 +1141,6 @@ def _import_from_dataframe(self,
...
-            created_index = index
-
-        if time_index is not None and time_index not in dataframe.columns:
-            raise LookupError('Time index not found in dataframe')
-
-        for c in dataframe.columns:
-            if not is_string(c):
-                raise ValueError("All column names must be strings (Column {} is not a string)".format(c))
-
-            if dataframe[c].dtype.name.find('category') > -1:
```
kmax12: in `infer_variable_types` this logic is just 

```
elif df[variable].dtype.name == "category":
      inferred_type = vtypes.Categorical
```

did this handle a case that the other didn't or were we just being extra cautious to just look for "category" anywhere in the name?

featuretools/entityset/entity.py
```diff
@@ -62,61 +62,39 @@ def __init__(self, id, df, entityset, variable_types=None,
...
+                raise ValueError("All column names must be strings (Column {} "
+                                 "is not a string)".format(c))
+        if time_index is not None and time_index not in df.columns:
+            raise LookupError('Time index not found in dataframe')
+
+        created_index, index, df = create_index(index, make_index, df)
+        if index not in variable_types:
+            variable_types[index] = vtypes.Index
+
         self.data = {"df": df,
```
kmax12: can we avoid setting the data like this? perhaps by just making df an argument to some of the other functions? 

it feels wrong to set the data here and then call `self.update_data(...)` a second time at the end

featuretools/entityset/entity.py
```diff
@@ -62,61 +62,39 @@ def __init__(self, id, df, entityset, variable_types=None,
...
-                _v = vtype[0](v, self, **vtype[1])
-            else:
-                _v = inferred_variable_types[v](v, self)
-            self.variables += [_v]
 
-        # do one last conversion of data once we've inferred
-        self.convert_all_variable_data(inferred_variable_types)
+        variable_types = variable_types or {}
+        secondary_time_index = secondary_time_index or {}
+        self.create_variables(variable_types, index, time_index, secondary_time_index)
```
kmax12: i think this should be `self._create_variables` since a user should never need call it, right?

kmax12: i think `secondary_time_index` is getting modified by reference within this function (actually when it gets passed to `infer_variable_types`). if that's true, i'd say let's avoid that or at least leave a comment.

featuretools/entityset/entityset.py
rwedge: I'm not sure.  It looks like from [this documentation](https://pandas.pydata.org/pandas-docs/stable/categorical.html#equality-semantics) that if we check for `df[variable].dtype == 'category'` instead we'll catch all CategoricalDtype.

kmax12: cool. let's refactor to that

featuretools/entityset/entity.py
rwedge: should we just call the relevant parts of `self.update_data` separately and skip calling `update_data`?.  It already expects the indexes to be set and it's a little strange to be "updating" the dataframe on initialization

featuretools/entityset/entity.py
```diff
@@ -669,3 +681,21 @@ def col_is_datetime(col):
         else:
             return True
     return False
+
+
+def create_index(index, make_index, df):
```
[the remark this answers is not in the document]
rwedge: Maybe this should also be private

featuretools/entityset/entity.py
kmax12: ya, i agree with that approach

featuretools/entityset/entity.py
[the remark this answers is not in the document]
kmax12: yep

featuretools/entityset/entity.py
rwedge: This code here, right?
```python
secondary_time_index = secondary_time_index or {}
for ti, cols in secondary_time_index.items():
    if ti not in cols:
        cols.append(ti)
```

The first line should be thrown out and mabye the rest should go in `_validate_entity_params` with either a warning or an error if the secondary time index column isn't included.

featuretools/entityset/entity.py
```diff
@@ -359,7 +323,34 @@ def query_by_values(self, instance_vals, variable_id=None, columns=None,
...
+        # convert data once we've inferred
+        self.convert_all_variable_data(inferred_variable_types)
+        # make sure index is at the beginning
+        index_variable = [v for v in variables
+                          if v.id == index][0]
+        self.variables = [index_variable] + [v for v in variables
+                                             if v.id != index]
+
+    def infer_variable_types(self, variable_types, time_index, secondary_time_index):
         """Extracts the variables from a dataframe
```
kmax12: update doc string

featuretools/entityset/entity.py
```diff
@@ -60,29 +55,32 @@ def __init__(self, id, df, entityset, variable_types=None,
...
 
         variable_types = variable_types or {}
         secondary_time_index = secondary_time_index or {}
         self._create_variables(variable_types, index, time_index, secondary_time_index)
 
         self.df = df[[v.id for v in self.variables]]
         self.set_index(index)
-        self.set_time_index(time_index, already_sorted=already_sorted)
+
+        if time_index:
```
kmax12: more compact
```
self.time_index = None
if time_index:
    self.set_time_index(time_index, already_sorted=already_sorted)
elif not already_sorted:
    self.df.sort_index(kind="mergesort", inplace=True)
```

kmax12:
Looks good to merge

PR #344, opened 2018-12-12
Removed time_index_reduce parameter

### LINKED ISSUE #301: Remove time_index_reduce parameter from `EntitySet.normalize_entity`
`time_index_reduce` can be used to determine the time_index of the new entity. According to the docs,

> If making a time_index, choose either the 'first' time or the 'last' time from the associated children instances. If creating a secondary time index, then the primary time index always reduces using 'first', and secondary using 'last'.

Using the `last` option could cause label leakage, so it should be removed. It's also untested in the code base and rarely, if ever used by end users. 

Thank you to @dah33 for pointing this out

jeff-hernandez:
I finished the adjustment. Let me know if anything needs modification. Thanks.

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -702,8 +702,7 @@ class EntitySet(object):
                          make_time_index=None,
                          make_secondary_time_index=None,
                          new_entity_time_index=None,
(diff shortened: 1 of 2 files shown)


kmax12:
Looks good. Merging

PR #348, opened 2018-12-13
Add tests for duplicate additional_variables/copy_variables in normal…

georgewambold:
…ize_entity

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -753,9 +753,15 @@ class EntitySet(object):
             raise TypeError("'additional_variables' must be a list, but received type {}"
                             .format(type(additional_variables)))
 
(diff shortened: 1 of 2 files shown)


kmax12:
even though this is caught now, it is sort of just an fortunate "accident" that we now check for duplicate columns later in the process of loading an entity. 

I'd say it is still worth it to add a check in `normalize_entity()`, so that we can give a more specific error of "Duplicate columns in copy_variables" or "Duplicate columns in additional_variables" when the user does this. 

Want to add that? the test cases look good though

georgewambold:
Ah, ok-- Yeah I'm down to add the more specific errors.

kmax12:
Looks good, merging. Thanks for the contribution, we appreciate it!

PR #350, opened 2018-12-17
[WIP] Primitives with multiple columns

rwedge:
This pull request updates the API for handling features that output multiple columns of values.  In the past these were referred to in the code as "expanding" features.  A brief rundown of the changes:

* **Multiple columns in the feature matrix can now correspond to the same feature.**  Where before a feature with multiple outputs would have a list of values per instance, these features now get one column per value.

* **Column names do not always match a feature name.**  With features having multiple columns it's necessary to modify the column names to distinguish between the various output columns of a feature.  

* **feature.expanding (bool) replaced with feature.number_output_features (int).** Instead of a feature either being expanding or not, it has an integer corresponding to the number of output features (columns) it has.

* **feature.get_expanded_names() replaced with feature.get_feature_names().**  Feature.get_feature_names is now the method to call when generating a feature matrix's column names.

* **feature.default_value is a singular value that becomes the default for all output features.**  Instead of a list of default values, one for each output feature, each output features shares the same default value.
* **two new keyword arguments for make_primitive functions**
    * number_output_features (int): sets the number of output features for a custom primitive
    * number_output_features_keyword (str): if the number of output features a custom primitive has is set by a keyword argument when the feature is initialized, providing the name of that keyword argument here will allow featuretools to set the correct number of output features.

* **features with multiple outputs can now be used as direct features**

### DIFF
--- a/featuretools/computational_backends/pandas_backend.py
+++ b/featuretools/computational_backends/pandas_backend.py
@@ -537,10 +537,9 @@ def agg_wrapper(feats, time_last):
 
             if f.expanding:
                 names = f.get_expanded_names()
-                for i, name in enumerate(names):
-                    d[name] = None
-                    if i < len(values):
-                        d[name] = values[i]
+                d.update({name: None for name in names})
+                assert len(values) <= len(names)
+                d.update({name: value for name, value in zip(names, values)})
             else:
                 d[f.get_name()] = values
         return pd.Series(d)
--- a/featuretools/computational_backends/pandas_backend.py
+++ b/featuretools/computational_backends/pandas_backend.py
@@ -327,10 +327,9 @@ class PandasBackend(ComputationalBackend):
             if f.number_output_features > 1:
                 names = f.get_feature_names()
                 assert len(names) == len(values)
+                values = [strip_values_if_series(value) for value in values]
                 for name, value in zip(names, values):
-                    value = strip_values_if_series(value)
                     frame[name] = value
-
             else:
                 values = strip_values_if_series(values)
                 frame[f.get_name()] = values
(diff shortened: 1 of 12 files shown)


featuretools/computational_backends/pandas_backend.py
```diff
@@ -518,17 +521,21 @@ def wrap(df):
...
             if f.uses_calc_time:
-                d[f.get_name()] = func(*args, time=time_last)
+                values = func(*args, time=time_last)
             else:
-                d[f.get_name()] = func(*args)
+                values = func(*args)
 
+            if f.number_output_features > 1:
+                names = f.output_feature_names()
+                d.update({name: None for name in names})
```
kmax12: why is this needed? seems like every entry in names will get add to dict 2 lines below

featuretools/computational_backends/pandas_backend.py
```diff
@@ -518,17 +521,21 @@ def wrap(df):
...
+                assert len(values) <= len(names)
+                d.update({name: value for name, value in zip(names, values)})
```
kmax12: i think you can just do

`d.update(dict(zip(names, values)))`

kmax12:
let's also support returning multiple features for transform primitives

featuretools/primitives/base/aggregation_primitive_base.py
```diff
@@ -80,7 +80,7 @@ def make_agg_primitive(function, input_types, return_type, name=None,
                        stack_on_exclude=None, base_of=None,
                        base_of_exclude=None, description='A custom primitive',
                        cls_attributes=None, uses_calc_time=False,
-                       commutative=False):
+                       commutative=False, number_output_features_keyword=None):
```
kmax12: what if the number of output features isn't dynamic. can the user simply say "there are 4 output features for this primitive"? maybe there needs to be a `number_output_features` parameter too that gets checked before `number_output_features_keyword`

rwedge:
Moved to #376

PR #360, opened 2018-12-20
use circleci docker images on circleci

rwedge:
Since we no longer use tox for testing this PR switches to using a different docker image for each set of tests to match the version of python being tested.


kmax12:
This looks good me

PR #364, opened 2019-01-07
Primitive refactor

kmax12:
This PR is separates the concept of a Primitive from a Feature. The internals of Featuretools change, but it is has minimal impact on the external API. 

Compared to before, a primitive is now only aware of the data it take in. A feature is then defined by input variables and/or features, as well as the primitive that will be applied. Put another way, a feature takes the specific entities and variables of an entityset and the primitive to be applied as its input so the primitive doesn't have to work about it. 

This has several advantages

1. It is easier to unit test primitives. There is no need to have an entity set to test a primitive
2. Primitive definitions become more reusable since they are not tied to the concept of entity set.
3. Conceptually the user defining a primitive has to only think about input and output data which is just numpy arrays.

To give an example, here is how a feature is currently defined using the count primitive

```
from featuretools.primitive import Count
f = Count(es["logs"]["value"], parent_entity=es["customers"])
```

Now, you define the inputs to the feature and provide the primitive as an input.

```
import featuretools as ft
from featuretools.primitive import Count
f = ft.Feature(es["logs"]["value"], parent_entity=es["customers"], primitive=Count)
```

if a primitive has parameters it can be used like this

```
f = ft.Feature(es["logs"]["comment"], parent_entity=es["customers"], primitive=CountString(string="coffee"))
```

the API for calling DFS doesn't change with the exception of being able to provide primitive with arguments

```
ft.dfs(target_entity="customers",
       entityset=es,
       agg_primtives=["count", Sum],
       trans_primtiives=CountString(string="coffee"))
```


featuretools/feature_base/feature_base.py
```diff
@@ -0,0 +1,495 @@
...
+    _name = None
+    expanding = False
+
+    def __init__(self, entity, base_features, primitive):
+        """Base class for all features
+
+        Args:
+            entity (Entity): entity this feature is being calculated for
+            base_featres (list[FeatureBase]): list of base features for primitive
+            primitive (): primitive to calculate. if not initilized when passed, gets initialized with no arguments
```
rwedge: missing the type information for primitive

featuretools/tests/dfs_tests/test_deep_feature_synthesis.py
```diff
@@ -360,13 +351,13 @@ def test_seed_features_added_with_identity_features(es):
 
 
 def test_dfs_builds_on_seed_features_more_than_max_depth(es):
-    seed_feature_sessions = Count(es['log']["id"], es['sessions']) > 2
-    seed_feature_log = Hour(es['log']['datetime'])
-    session_agg = Last(seed_feature_log, es['sessions'])
+    seed_feature_sessions = ft.Feature(es['log']["id"], parent_entity=es['sessions'], primitive=Count)
```
rwedge: I don't think it changes the test but `seed_feature_sessions` is missing the `> 2`

featuretools/feature_base/feature_base.py
```diff
@@ -0,0 +1,495 @@
...
+            assert len(set([bf.entity for bf in base_features])) == 1, msg
+        else:
+            base_features = [_check_feature(base_features)]
+
+        if any(bf.expanding for bf in base_features):
+            self.expanding = True
+
+        base_entity = set([f.entity for f in base_features])
+        assert len(base_entity) == 1, \
+            "More than one entity for base features"
```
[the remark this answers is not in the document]
kmax12: good catch. fixing

featuretools/tests/dfs_tests/test_deep_feature_synthesis.py
kmax12: ya, i noticed that. i dont think adding the `>2` improved the test so i simplified it

featuretools/feature_base/feature_base.py
```diff
@@ -0,0 +1,495 @@
...
+        for d in self.base_features[:]:
+            deps += [d]
+
+        if hasattr(self, "where") and self.where:
+            deps += [self.where]
+
+        # if self.use_previous and self.use_previous.is_absolute():
+            # entity = self.entity
+            # time_var = IdentityFeature(entity[entity.time_index])
+            # deps += [time_var]
```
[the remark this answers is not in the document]
kmax12: the time index automatically gets added, so we don't need to put the time index feature as a dependent. removed

PR #367, opened 2019-01-09
Add ZipCode Variable Type 

gsheni:
- This adds a zipcode variable type
- Fixed some import statement order and spacing


kmax12:
@gsheni can you add this new variable type to the api reference? after that, good to merge

gsheni:
@kmax12 Added to the api_reference.rst
Also, 
- I changed it to a subtype of Text because zipcodes might begin with 0, ex 02116.
- Added zipcode column to mock_ds

kmax12:
Looks good to merge

PR #368, opened 2019-01-10
Update docs to entityset for wordcount primitive example

RogerTangos:
- corrects an error in the WordCount example in the docs
- WordCount example now gives the expected result


kmax12:
Looks good merging

PR #372, opened 2019-01-15
Allow user to pass in desired feature return types

RogerTangos:
- `dfs()` can be passed a list of `allowed_variable_types`, `None` or `'all'`. This filters returned features based on their output types.
- `None` defaults to `[Numeric, Discrete, Boolean]`
- Docstrings updated accordingly
- Completes work begun (but unexposed in the api) in https://github.com/Featuretools/featuretools/commit/ff70cbe9a9de7d69e37244ff3b23fc6b21596c86, but
- Tests changes


RogerTangos:
There were some initial strange linting errors here, which are now fixed. (strange, because the lines in question were from the master branch.)

Now this seems to be failing on pandas 0.24.0. Once #383 fixes that, I'll merge master back in.

RogerTangos:
I fixed a bunch of indentation errors that were being picked up by the linter update, and now codecov/patch is knocking the coverage score for it. Many of the fixes it's penalizing were on commented lines.

Would you like me to fix these, or can we go ahead and merge? On the original PR, the coverage was unchanged.

kmax12:
@RogerTangos thanks for investigating the linting issues and resolving. we still need to review this before it's ready to merge. we're currently preparing the next release of featuretools (likely to go out this week), but after that we will return to review this.

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -175,14 +175,14 @@ def __init__(self,
         self.drop_contains = drop_contains or []
         self.where_stacking_limit = where_stacking_limit
 
-    def build_features(self, variable_types=None, verbose=False):
+    def build_features(self, allowed_variable_types=None, verbose=False):
```
kmax12: would a name like `return_variable_types` be more logical?

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -491,16 +498,15 @@ def _build_transform_features(self, all_features, entity, max_depth=0):
...
 
         if max_depth is not None and max_depth < 0:
             return
 
         features = self._features_by_type(all_features=all_features,
                                           entity=parent_entity,
-                                          variable_type=[Numeric,
-                                                         Categorical,
-                                                         Ordinal],
+                                          variable_type=allowed_variable_types,
```
kmax12: I think we actually shouldn't be filtering any features based on type here since happens on line 216.

That means the correct change here would be to actually to update  `_features_by_type` to not filter based on variable_type if `variable_type` == None.

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -266,7 +268,8 @@ def _filter_features(self, features):
 
         return f_keep
 
-    def _run_dfs(self, entity, entity_path, all_features, max_depth):
+    def _run_dfs(self, entity, entity_path, all_features, max_depth,
+                 allowed_variable_types=None):
```
kmax12: once we remove `allowed_variable_types` from `_build_forward_features`, we no longer need this parameter to `_run_dfs`

featuretools/synthesis/dfs.py
```diff
@@ -31,7 +31,8 @@ def dfs(entities=None,
         chunk_size=None,
         n_jobs=1,
         dask_kwargs=None,
-        verbose=False):
+        verbose=False,
+        allowed_variable_types=None):
```
kmax12: rename to `return_variable_types` per comment in `deep_feature_synthesis.py`

kmax12: please add test cases for this change before we can merge

featuretools/synthesis/deep_feature_synthesis.py
RogerTangos: Good point. I'll update this.

RogerTangos:
Thanks @kmax12. It's the end of the day here, so I'll get to these updates and the tests tomorrow. Thank you for the review.

RogerTangos:
@kmax12 - requested changes made.

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -208,25 +208,20 @@ def build_features(self, variable_types=None, verbose=False):
                 all_features[e.id] = {}
 
         self.where_clauses = defaultdict(set)
+
+        if return_variable_types is None:
```
kmax12: let's move this logic inside of if block on line 235. that way it is closer to where it's used and we don't have the double check for `return_variable_types == 'all'` case

featuretools/tests/dfs_tests/test_deep_feature_synthesis.py
```diff
@@ -694,6 +694,33 @@ def test_initialized_agg_prim(es):
...
+def test_return_variable_types(es):
+    dfs_obj = DeepFeatureSynthesis(target_entity_id="sessions",
+                                   entityset=es,
+                                   agg_primitives=[Count, NMostCommon],
+                                   trans_primitives=[Absolute, Hour, IsIn])
+
+    discrete = ft.variable_types.Discrete
+    numeric = ft.variable_types.Numeric
+
+    f1 = dfs_obj.build_features(return_variable_types=None)
```
kmax12: we should add a test where you go outside of the default type. for example, if you pass in `Datetime` or `DatetimeTimeIndex` it should return some additional features since sessions has datetime columns.

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -606,7 +608,9 @@ def _build_agg_features(self, all_features,
 
                     self._handle_new_feature(new_f, all_features)
 
-    def _features_by_type(self, all_features, entity, variable_type, max_depth):
+    def _features_by_type(
```
kmax12: to keep the style of our other code - let's keep this as one line or break up the arguments over two lines

featuretools/synthesis/deep_feature_synthesis.py
RogerTangos: Isn't the assertion meant to provide a check for back input and give feedback? To do that, it needs to go before `self._run_dfs`.

RogerTangos:
Updates to everything but line 212. I did actually write some code for this, but IMO, it made the file harder to understand, not easier. If you have an opinion on how that should be done, feel free to copy past and I'll add it.

featuretools/synthesis/deep_feature_synthesis.py
kmax12: ahh, i didn't see the assert. okay. let's leave as is

kmax12:
Looks good. Merging!

Issue #371, opened 2019-01-15
test_install_primitives failing on master branch


RogerTangos:
### Bug/Feature Request Description
On the first run of `python test featuretools/tests/`, [test_install_primitives](https://github.com/Featuretools/featuretools/blob/master/featuretools/tests/primitive_tests/test_install_primitives.py) fails when `install_path` is `"INSTALL_VIA_CLI"` or `"INSTALL_VIA_MODULE"`.  

featuretools 0.5.1
python 3.6.6
OSX 10.13.6

Does this have to do with testing locally instead of through CircleCI?

kmax12:
@RogerTangos can you share the stacktrace?

RogerTangos:
```

==> arcarter . 5 . featuretools: pytest featuretools/tests
======================================== test session starts ========================================
platform darwin -- Python 3.6.6, pytest-3.8.2, py-1.7.0, pluggy-0.8.0
rootdir: /Users/arcarter/code/featuretools, inifile: setup.cfg
plugins: xdist-1.23.2, remotedata-0.3.1, openfiles-0.3.0, forked-0.2, doctestplus-0.1.3, cov-2.6.0, arraydiff-0.2
collected 325 items

featuretools/tests/computational_backend/test_calculate_feature_matrix.py ................... [  5%]
....................                                                                          [ 12%]
featuretools/tests/computational_backend/test_dask_features.py .                              [ 12%]
featuretools/tests/computational_backend/test_encode_features.py ......                       [ 14%]
featuretools/tests/computational_backend/test_pandas_backend.py ........................      [ 21%]
featuretools/tests/demo_tests/test_demo_data.py ...                                           [ 22%]
featuretools/tests/dfs_tests/test_deep_feature_synthesis.py ................................. [ 32%]
.                                                                                             [ 32%]
featuretools/tests/dfs_tests/test_dfs_method.py ........                                      [ 35%]
featuretools/tests/entityset_tests/test_entity.py .......                                     [ 37%]
featuretools/tests/entityset_tests/test_es.py ............................................... [ 52%]
..............                                                                                [ 56%]
featuretools/tests/entityset_tests/test_es_metadata.py ...............                        [ 60%]
featuretools/tests/entityset_tests/test_last_time_index.py ............                       [ 64%]
featuretools/tests/entityset_tests/test_timedelta.py ..........                               [ 67%]
featuretools/tests/primitive_tests/test_agg_feats.py ................                         [ 72%]
[...]

PR #373, opened 2019-01-15
add testing heading to readme

RogerTangos:
- add testing instructions to readme
- apparently... also remove a blank space at the end of a paragraph. (Sorry about that. It's a linter thing.)

Related to #371


README.md
```diff
@@ -78,6 +78,13 @@ In this demonstration, we use a multi-table dataset of 3 million online grocery
 
 For more examples of how to use Featuretools, check out our [demos](https://www.featuretools.com/demos) page.
 
+## Testing
```
kmax12: These instructions should get added to the development section in the install page (https://docs.featuretools.com/getting_started/install.html#development).

You can create a development section in the readme and like to that url too.

RogerTangos: Ah. I hadn't looked in the docs, and had checked the readme instead. To avoid duplication, would it make sense just to link from the readme to the docs url?

kmax12: get_data_path

RogerTangos:
Pushed. Python tests will fail likely fail because of the pandas 0.24.0 and linter update, but no code has changed. Should be good to merge.

kmax12:
Made one change to the capitalization of Featurefools. Otherwise, looks good. Thanks!

PR #374, opened 2019-01-15
Pin Numpy Requirement

kmax12:
The 1.16.0 release of numpy is breaking the dask import in python 2.7. This has been fixed in dask, but hasn't been released yet. Here is the PR: https://github.com/dask/dask/pull/4346. 

To workaround this for the time being, this PR pins Featuretools to the previous version of Numpy.

Once next release of dask goes out with the fix we can unpin this.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 boto3>=1.9.51
 botocore>=1.12.51
-numpy>=1.13.3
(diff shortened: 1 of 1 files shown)


kmax12:
Looks like dask is doing a release on friday, so we hold off with this fix

https://github.com/dask/dask/issues/4378

PR #376, opened 2019-01-16
Multiple output column features with primitive refactor

rwedge:


featuretools/feature_base/feature_base.py
```diff
@@ -327,9 +334,8 @@ class DirectFeature(FeatureBase):
 
     def __init__(self, base_feature, child_entity):
         base_feature = _check_feature(base_feature)
-        if base_feature.expanding:
-            self.expanding = True
         self.base_feature = base_feature
+        self.primitive = base_feature.primitive
```
kmax12: hmm, not sure about this. i would this our code shouldn't ever try access the primitive property of a direct feature

featuretools/primitives/standard/aggregation_primitives.py
```diff
@@ -132,6 +132,31 @@ def percent_true(s):
...
+
+    def __init__(self, n=3):
+        self.number_output_features = n
+
+    @property
+    def default_value(self):
+        return np.nan
+
+    def get_function(self):
+        def pd_topn(x, n=self.number_output_features):
```
kmax12: i think we can rename `pd_topn` to just `n_most_common`

featuretools/primitives/standard/aggregation_primitives.py
```diff
@@ -132,6 +132,31 @@ def percent_true(s):
...
+    input_types = [Discrete]
+    return_type = Discrete
+    stack_on = []
+    stack_on_exclude = []
+
+    def __init__(self, n=3):
+        self.number_output_features = n
+
+    @property
+    def default_value(self):
```
kmax12: do we need this? i think primitive base already has default value as np.nan

featuretools/primitives/standard/aggregation_primitives.py
```diff
@@ -132,6 +132,31 @@ def percent_true(s):
         return percent_true
 
 
+class NMostCommon(AggregationPrimitive):
+    """Finds the N most common elements in a categorical feature."""
+    name = "n_most_common"
+    input_types = [Discrete]
+    return_type = Discrete
+    stack_on = []
```
kmax12: i think we want to let NMostCommon stack

featuretools/primitives/standard/aggregation_primitives.py
```diff
@@ -132,6 +132,31 @@ def percent_true(s):
...
+
+    @property
+    def default_value(self):
+        return np.nan
+
+    def get_function(self):
+        def pd_topn(x, n=self.number_output_features):
+            array = np.array(x.value_counts()[:n].index)
+            if len(array) < n:
+                filler = np.full(n - len(array), self.default_value)
```
kmax12: let's not use self.default_value here

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -736,3 +734,9 @@ def match(input_types, features, replace=False, commutative=False):
         return set([tuple(sorted(s, key=lambda x: x.get_name().lower())) for s in matching_inputs])
 
     return set([tuple(s) for s in matching_inputs])
+
+
+def ready_primitive(primitive):
+    if type(primitive) == type:
```
kmax12: this is how we do this check in FeatureBase

```
# initialize if not already initialized
if not isinstance(primitive, PrimitiveBase):
    primitive = primitive()
```

not sure which i like more. do you have thoughts on which way is better?

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -658,19 +653,22 @@ def _max_hlevel(self, f):
 def check_stacking(primitive, input_types):
```
kmax12: i think we should rename the the `input_types` variable to just `inputs` since it is a list of feature base objects not variable types

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -658,19 +653,22 @@ def _max_hlevel(self, f):
...
     """checks if features in input_types can be used with supplied primitive
        using the stacking rules"""
+    assert primitive.__class__ != str.__class__
```
kmax12: do we need this line?

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -658,19 +653,22 @@ def _max_hlevel(self, f):
...
     if primitive.stack_on_exclude is not None:
         for f in input_types:
             if isinstance(f.primitive, tuple(primitive.stack_on_exclude)):
                 return False
 
+    # R TODO: handle this
     for f in input_types:
-        if f.primitive.expanding:
+        feats = f.get_dependencies(deep=True) + [f]
+        if any([feat.primitive.number_output_features > 1 for feat in feats]):
```
kmax12: i don't think you need to check the deep dependencies. if the input feature doesn't have `number_output_features > 1` it shouldn't matter what it's dependents have.

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -736,3 +734,9 @@ def match(input_types, features, replace=False, commutative=False):
         return set([tuple(sorted(s, key=lambda x: x.get_name().lower())) for s in matching_inputs])
 
     return set([tuple(s) for s in matching_inputs])
+
+
+def ready_primitive(primitive):
```
kmax12: let's name this method better. maybe just `handle_primitive`?

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -582,7 +576,8 @@ def _build_agg_features(self, all_features,
                     continue
 
                 # limits the aggregation feature by the given allowed feature types.
-                if not any([issubclass(agg_prim, feature_type) for feature_type in self.where_primitives]):
+                if not any([issubclass(type(agg_prim), type(feature_type))
```
kmax12: `feature_type` here should be renamed to `primitive` or something like that

featuretools/synthesis/encode_features.py
```diff
@@ -83,7 +83,10 @@ def encode_features(feature_matrix, features, top_n=10, include_unknown=True,
         iterator = features
 
     for f in iterator:
-        if (f.expanding or (not issubclass(f.variable_type, Discrete))):
+        # TODO: features with multiple columns are not encoded by this method,
```
kmax12: can we add a warning when this is the case with `number_output_features > 1`

featuretools/computational_backends/pandas_backend.py
```diff
@@ -341,9 +358,11 @@ def _calculate_direct_features(self, features, entity_frames):
             # Sometimes entityset._add_multigenerational_links adds link variables
             # that would ordinarily get calculated as direct features,
             # so we make sure not to attempt to calculate again
-            if f.get_name() in child_df.columns:
-                continue
-            col_map[f.base_features[0].get_name()] = f.get_name()
+            base_names = f.base_features[0].get_feature_names()
```
kmax12: you can do just `f.base_feature.get_feature_names()`

featuretools/computational_backends/pandas_backend.py
```diff
@@ -518,17 +532,20 @@ def wrap(df):
...
-                d[f.get_name()] = func(*args, time=time_last)
+                values = func(*args, time=time_last)
             else:
-                d[f.get_name()] = func(*args)
+                values = func(*args)
 
+            if f.primitive.number_output_features > 1:
+                names = f.get_feature_names()
+                assert len(names) == len(values)
+                d.update(dict(zip(names, values)))
```
kmax12: any reason use use DataFrame.update here instead of the loop approach from `_calculate_transform_features`? 

```
for name, value in zip(names, values):
    frame[name] = strip_values_if_series(value)
```

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -518,11 +515,8 @@ def _build_forward_features(self, all_features, parent_entity,
 
             new_f = DirectFeature(f, child_entity)
 
-            if f.expanding:
```
kmax12: is this removed because we now support direct features with multiple outputs?

featuretools/tests/computational_backend/test_pandas_backend.py
```diff
@@ -499,3 +524,8 @@ def test_empty_child_dataframe():
     fm2 = ft.calculate_feature_matrix(entityset=es, features=[count_where, trend_where], cutoff_time=pd.Timestamp("1/4/2018"))
     names = [count_where.get_name(), trend_where.get_name()]
     assert_array_equal(fm2[names], [[0, np.nan]])
+
+
+def test_computes_direct_features_of_multi_ouput_features():
```
kmax12: let's create an issue for this

featuretools/tests/primitive_tests/test_direct_features.py
```diff
@@ -44,3 +58,66 @@ def test_direct_rename(es):
     assert feat.get_name() != copy_feat.get_name()
     assert feat.base_features[0].generate_name() == copy_feat.base_features[0].generate_name()
     assert feat.entity == copy_feat.entity
+
+
+def test_direct_of_multi_output_transform_feat(es):
```
kmax12: maybe add that this is just testing naming

featuretools/tests/primitive_tests/test_direct_features.py
```diff
@@ -44,3 +58,66 @@ def test_direct_rename(es):
...
+    alt_features = [Feature(es["customers"]["signup_date"], primitive=Year),
+                    Feature(es["customers"]["signup_date"], primitive=Month),
+                    Feature(es["customers"]["signup_date"], primitive=Day),
+                    Feature(es["customers"]["signup_date"], primitive=Hour),
+                    Feature(es["customers"]["signup_date"], primitive=Minute),
+                    Feature(es["customers"]["signup_date"], primitive=Second)]
+    fm, fl = dfs(
+        entityset=es,
+        target_entity="sessions",
+        trans_primitives=[TestTime, Year, Month, Day, Hour, Minute, Second])
```
kmax12: i think you can do `features_only=True` here. same for test case below

featuretools/synthesis/deep_feature_synthesis.py
rwedge: yes

featuretools/tests/computational_backend/test_pandas_backend.py
rwedge: this got implemented in test_directo_features.py so I'll remove it here

featuretools/computational_backends/pandas_backend.py
rwedge: Changing this causes an error related to this property direct features have that `DirectFeature.base_feature` points to the desired feature while `DirectFeatures.base_features[0]` points to the next feature on the path between the base feature and the child.  So `sessions.customers.loves_ice_cream` has `<Feature: loves_ice_cream>` as the `base_feature` but `customers.loves_ice_cream` for `base_features[0]`.

I saw a TODO above the logic for how `DirectFeature` does this, maybe I should add an explanation?

PR #379, opened 2019-01-22
Mean ignore NaNs

gsheni:
- Modified the Mean primitive so it can be user controlled to ignore NaNs or not. 
- Wrote simple test case to test the function.

### DIFF
--- a/featuretools/tests/primitive_tests/test_agg_feats.py
+++ b/featuretools/tests/primitive_tests/test_agg_feats.py
@@ -126,18 +126,21 @@ def test_check_input_types(es):
 
 
 def test_mean_nan():
-    mean_func = Mean().get_function()
     array = np.array([5, 5, 5, 5, 5])
-    assert mean_func(array) == 5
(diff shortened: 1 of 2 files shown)


featuretools/tests/primitive_tests/test_agg_feats.py
```diff
@@ -124,6 +125,16 @@ def test_check_input_types(es):
     assert mean._check_input_types()
 
 
+def test_mean_nan():
+    mean_func = Mean().get_function()
+    array = np.array([5, 5, 5, 5, 5])
+    assert mean_func(array) == 5
+    array = np.array([5, np.nan, np.nan, np.nan, np.nan, 10])
```
kmax12: you should add a test case where there are nans in the list, but `ignore_nans=False`

gsheni: Fixed.

kmax12:
Looks good to me!

PR #380, opened 2019-01-24
Add primitive.get_data_path and example of primitive loading data from external files

rwedge:
This PR adds the method `PrimitiveBase.get_data_path`.  This helper function creates an absolute file path by joining the path to `featuretools/primitives/data` with the filename supplied by the user.


featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -1055,3 +1055,41 @@ def gen_feat_names(self):
 def test_feature_names_inherit_from_make_trans_primitive():
     # R TODO
     pass
+
+
+def test_get_data_path(es):
+    class Mod4(TransformPrimitive):
+        '''Replace each integer with a word from a shakespearean sonnet'''
```
kmax12: is this doc string accurate?

Issue #381, opened 2019-01-25
Plotting EntitySets


floscha:
Hey everyone,

I thought it would be great if it was possible to plot a visual representation of an *EntitySet*, like one can for example [plot the structure of a neural network in Keras](https://www.tensorflow.org/api_docs/python/tf/keras/utils/plot_model):

```python
es = ft.demo.load_mock_customer(return_entityset=True)
es.plot()
```

So I build some quick prototype which looks like the following:

![screeshot](https://user-images.githubusercontent.com/33159547/51755945-16cf3a00-20c0-11e9-8ce5-104a4e7431d6.png)

If you think that such a method could add value to the **featuretools** library, then I'll happily create a proper pull request for this.

kmax12:
That is a great idea and something that has been requested by users. A PR would be great.

bukosabino:
I have some problems with graphviz when I use es.plot().

At first, I get this error:

`ImportError: Please install graphviz to plot entity sets. (See https://pypi.org/project/graphviz/ for details)`

It is resolved with `pip install graphviz`. So, we need to add this package to requirements.txt. But, then I get:

`ExecutableNotFound: failed to execute ['dot', '-Tsvg'], make sure the Graphviz executables are on your systems' PATH`

I have tried using conda:

`conda install python-graphviz`

But, it's not fixed on my Mac. Let me know if you have any hint.

bukosabino:
Ok. So, to use `es.plot()` people need to install graphviz as backend:

Mac OS:
`brew install graphviz`

Linux (Ubuntu example):
`sudo apt-get install graphviz`

Windows:
`conda install python-graphviz`

Now, we need to add the graphviz package on requirements.txt too.

kmax12:
@bukosabino because graphviz can be a bit challenging to install on different platforms, we made it an optional requirement for people who want to use it. your comment on this issue is good documentation for it and we can add more to the docs after we get some feedback on people's experience installing it.

kmax12:
closing since #382 has been merged in and released in featuretools v0.6.0

rwedge:
The plotting tests require graphviz, so we should probably add graphviz to the test requirements or flag them as optional in some way.

kmax12:
ah, yes let's do that. can you open a PR?

PR #382, opened 2019-01-25
Plotting entitysets

floscha:
This PR adds a `plot()` method to the `EntitySet` class as described in #381, as well as a suite of relevant unit tests.


kmax12:
thank you for the contribution - let us know when it's ready for review. 

note: don't worry about the failing codecov checks. those seem to be errors. if you update from master, it may fix

floscha:
Thanks for the hint. I was already having a hard time figuring out how my changes could've increased the number of misses... Other than that, I'd say the PR is ready for review now

kmax12:
great. someone on our side will review soon.

requirements.txt
```diff
@@ -1,5 +1,6 @@
 boto3>=1.9.51
 botocore>=1.12.51
+graphviz>=0.10.1
```
kmax12: can we make this an optional requirement? 

my recommended approach would be to import inside of `plot` some like this 

```
try:
    import graphviz
except ImportError:
    raise ImportError('Please install graphviz to install use plotting functionality')
```

featuretools/entityset/entityset.py
```diff
@@ -1094,6 +1095,59 @@ def related_instances(self, start_entity_id, final_entity_id,
...
+        for entity in self.entities:
+            variables_string = '\l'.join([var.id + ' : ' + var.dtype  # noqa: W605
+                                          for var in entity.variables])
+            label = '{%s|%s\l}' % (entity.id, variables_string)  # noqa: W605
+            graph.node(entity.id, shape='record', label=label)
+
+        # Draw relationships
+        for rel in self.relationships:
+            label = '%s -> %s' % (rel._child_variable_id, rel._parent_variable_id)
+            graph.edge(rel._parent_entity_id, rel._child_entity_id, label=label)
```
kmax12: i believe this edge is going in wrong direction. the arrow should point from the child to the parent. 

for example in the attached image the orders entity should points to the customers, rather than the other way around

test

requirements.txt
floscha: Fair enough. It's now solved with a779771

featuretools/entityset/entityset.py
floscha: You're right regarding the direction. Also, [this page](https://developer.intacct.com/entity-relationship-diagrams/ap-payment/) suggests that there's a dedicated "one-to-many"-relationship arrowhead, which I also added with 1216fb4 and made all lines orthogonal. Furthermore, if the key is the same for both related entities, the key is now only displayed once.

The result now looks like this:
download

I'm not 100% satisfied with the aesthetics but I guess that'll have to do for now.

floscha:
Thanks for your feedback @kmax12. I've extended the docs with 11ce114 (I hope I've done that part right) and incorporated your comments (see above).

kmax12:
@floscha looks like there are a few linting errors being caught in CI. they are 

```
featuretools/entityset/entityset.py:8:1: F401 'graphviz' imported but unused
featuretools/entityset/entityset.py:1111:13: F811 redefinition of unused 'graphviz' from line 8
featuretools/entityset/entityset.py:1136:55: E231 missing whitespace after ':'
```

you should be able to reproduce locally by running `make lint`.

floscha:
My bad. They're fixed now

kmax12:
This looks great. I just made a some changes to the docs and reverted the arrow back to what you had before. 

Merging now.

PR #386, opened 2019-01-30
Update test and dev requirements

gsheni:
- Upgraded all pypi packages in dev, test, and setup requirements. 
- `flake8`, and `pympler` do not have their latest version on conda, so only upgraded to the latest on conda. 
- lint fixes 
- pytest threw an error that fixtures cannot be called directly, so this was fixed.

### DIFF
--- a/featuretools/tests/primitive_tests/test_install_primitives.py
+++ b/featuretools/tests/primitive_tests/test_install_primitives.py
@@ -33,15 +33,39 @@ def bad_primitives_files_dir(this_dir):
     return os.path.join(this_dir, "bad_primitive_files")
 
 
+@pytest.fixture
+def amazon_path_s3():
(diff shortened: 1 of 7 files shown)


featuretools/tests/primitive_tests/test_install_primitives.py
```diff
@@ -33,15 +33,39 @@ def bad_primitives_files_dir(this_dir):
...
+    return "s3://featuretools-static/primitives_to_install.tar.gz"
+
+
+@pytest.fixture
+def amazon_path_http():
+    return "https://s3.amazonaws.com/featuretools-static/primitives_to_install.tar.gz"
+
+
+@pytest.fixture
+def install_via_cl():
```
rwedge: `install_via_cli`

Same thing inside the parameterize list

PR #394, opened 2019-02-01
Fix transform stacking on multi-output aggregation

gsheni:
- Fixes an assertion issue
- test case to check this.

### DIFF
--- a/featuretools/tests/primitive_tests/test_transform_features.py
+++ b/featuretools/tests/primitive_tests/test_transform_features.py
@@ -43,6 +43,7 @@ from featuretools.primitives import (  # CumCount,; CumMax,; CumMean,; CumMin,; 
     NotEqualScalar,
     NumCharacters,
     NumWords,
+    NMostCommon,
     Percentile,
     ScalarSubtractNumericFeature,
     Second,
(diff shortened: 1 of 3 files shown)


gsheni:
@rwedge fixed with dfs test that uses a mutli output agg and transform

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -1052,6 +1053,18 @@ def gen_feat_names(self):
...
+def test_tranform_stack_agg(es):
+    topn = ft.Feature(es['log']['product_id'],
+                      parent_entity=es['customers'],
+                      primitive=NMostCommon(n=3))
+    with pytest.raises(AssertionError):
+        ft.Feature(topn, primitive=Percentile)
+    ft.dfs(entityset=es,
+           target_entity="log",
+           agg_primitives=[NMostCommon],
+           trans_primitives=[Percentile])
```
rwedge: can you move this dfs call to a separate test case in `test_deep_feature_synthesis` and add an assert statement to confirm no Percentile(NMostCommon()) features we created?

PR #397, opened 2019-02-01
check that agg_primitives and trans_primitives are right primitive type

rwedge:
This PR adds a check in DFS to `agg_primitives` and `trans_primitives` to give a clearer error when primitives are included in the wrong list (putting `Hour` in `agg_primitives`, for exmaple)

### DIFF
--- a/featuretools/tests/dfs_tests/test_deep_feature_synthesis.py
+++ b/featuretools/tests/dfs_tests/test_deep_feature_synthesis.py
@@ -683,7 +683,7 @@ def test_initialized_agg_prim(es):
     assert (feature_with_name(features, "N_MOST_COMMON(log.product_id)"))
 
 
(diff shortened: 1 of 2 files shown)


featuretools/tests/dfs_tests/test_deep_feature_synthesis.py
```diff
@@ -681,3 +681,19 @@ def test_initialized_agg_prim(es):
                                    trans_primitives=[])
     features = dfs_obj.build_features()
     assert (feature_with_name(features, "N_MOST_COMMON(log.product_id)"))
+
+
+def test_checks_agg_primitives_are_aggs(es):
```
kmax12: the method name should be something like `test_check_primitives_correct_type`

PR #400, opened 2019-02-04
add comments to feature_tree

RogerTangos:
- describe what some of the instance variables are
- Add one line break in code, bringing max line size to 100
- rename `all_features` in `__init__`, since it's confusing to have `all_features` be different from `self.all_features`


kmax12:
Looks good! I suspect more work can be done here, but need to start somewhere. Thank you for the contribution. Merging now

PR #401, opened 2019-02-04
Add new configuration object

CharlesBradshaw:
Adds `data_path` as a static variable and a class method `set_data_path` to update `data_path`


featuretools/tests/primitive_tests/test_feature_base.py
```diff
@@ -109,3 +112,33 @@ def test_return_type_inference_id(es):
     # also test direct feature of aggregation
     mode_direct = ft.Feature(mode, es["sessions"])
     assert mode_direct.variable_type == Categorical
+
+
+def test_set_data_path(es):
```
kmax12: this test should just init primitives, no need to create features

PR #403, opened 2019-02-05
Install data and dependencies

rwedge:
In addition to primitives files, `featuretools install` will now look for a folder named "data" and copy it to `featuretools/primitives/data/`.  It will also look for "requirements.txt" and run `pip install -r requirements.txt` if it finds it.


featuretools/primitives/install.py
```diff
@@ -31,45 +33,73 @@ def install_primitives(directory_or_archive, prompt=True):
...
+        # Iterate over all the files and determine the primitives to install
+        with open(os.path.join(directory, "info.json"), 'r') as f:
+            info = json.load(f)
+
+        # before installing, confirm with user
+        primitives_list = ", ".join(info['primitives'])
+        if prompt:
+            while True:
+                resp = input("Install primitives: %s? (Y/n) " % primitives_list)
+                if resp == "Y":
```
kmax12: should we update this to make it not case sensitive?

featuretools/primitives/install.py
```diff
@@ -31,45 +33,73 @@ def install_primitives(directory_or_archive, prompt=True):
...
+        all_primitives = set()
+        for filepath in files:
+            primitive_name, primitive_obj = load_primitive_from_file(filepath)
+            if primitive_obj.name not in info['primitives']:
+                raise RuntimeError("Primitive %s not listed in "
+                                   "info.json" % (primitive_obj.name))
+            files_to_copy.append(filepath)
+            all_primitives.add(primitive_obj.name)
+
+        if all_primitives != set(info['primitives']):
```
kmax12: just need to add a test for this case

featuretools/primitives/install.py
```diff
@@ -31,45 +33,73 @@ def install_primitives(directory_or_archive, prompt=True):
...
+        # copy the files
+        installation_dir = get_installation_dir()
+
+        files = list_primitive_files(directory)
+        files_to_copy = []
+        all_primitives = set()
+        for filepath in files:
+            primitive_name, primitive_obj = load_primitive_from_file(filepath)
+            if primitive_obj.name not in info['primitives']:
+                raise RuntimeError("Primitive %s not listed in "
```
kmax12: just need to add a test for this case

Issue #340, opened 2018-12-07
Make s3fs and distributed optional dependencies


KOLANICH:

KOLANICH:
>Can you elaborate why you picked these two libraries in particular?

Because they are obviously unneeded for tinkering on a local machine.

kmax12:
Got it. You're right about s3fs.  We should do something like Pandas when we require it i.e 

```python
try:
    import s3fs
    from botocore.exceptions import NoCredentialsError
except ImportError:
    raise ImportError("The s3fs library is required to handle s3 files")
```
Right now, I think we only require it for the demo datasets download functions (`featuretools/demo/flight.py` and `featuretools/demo/flight.py`)

Distributed, on the other hand, is used by to parallelize locally over multiple processes. 

Leaving this issue open, so we at least fix it for s3. This is a good first issue if you, or anyone else, is interested in helping us out.  Thanks!

kmax12:
sure, we can add it directly into the `load_retail` or `load_flight` functions.

PR #404, opened 2019-02-05
Make S3 dependencies optional

floscha:
Based on the discussion in #340, this PR makes all S3-related dependencies optional, considering that S3-usage is presumably limited to a smaller range of users. However, the dependencies were moved to *test-requirements.txt* so that unit tests still run.

One corner case I left out is [S3 buckets](https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#access-bucket-intro) which are not supported right now. Adding such support would (IMHO) require adding some regular expressions which would unnecessarily clutter the code. If preferred, I can add those nonetheless.


featuretools/primitives/install.py
```diff
@@ -100,12 +98,20 @@ def download_archive(uri):
     local_archive = os.path.join(get_installation_temp_dir(), filename)
 
     with open(local_archive, 'wb') as f:
-        try:
+        if uri.startswith('s3://') or uri.startswith('https://s3'):
+            try:
+                import s3fs
+                from botocore.exceptions import NoCredentialsError
+            except ImportError:
+                raise ImportError("The s3fs library is required to handle s3 files")
```
kmax12: s3fs is actually only used as a fallback since `smart_open` cannot handle the anonymous clients which leads to the `NoCredentialsError` case. smart open can open s3 files in other cases.

see this issue: https://github.com/RaRe-Technologies/smart_open/issues/250

perhaps move this try/except into the `NoCredentialsError` block which is the only place s3fs is used.

kmax12:
do we need to add this check to the `load_retail` or `load_flight` functions as well?

floscha:
Both `load_retail` and `load_flight` work without having `s3fs` installed. However, I noticed that `smart_open` crashes when `boto3` isn't installed (which already happens when calling `import featuretools`), so I added it back to *requirements.txt* with 401f4d0.

featuretools/primitives/install.py
floscha: I see. I was assuming that `smart_open` also uses `s3fs` under the hood. I've now placed the `s3fs` import as you suggested.

kmax12:
Looks good merging

PR #432, opened 2019-02-13
Generate controllable name

CharlesBradshaw:
For controllable primitives, the name of the primitive needs to be unique, even if the primitive is instantiated twice. aka (ExamplePrimitive(a=5), ExamplePrimitive(a=6)). In those cases, the names for the primitive should include "a=5" and "a=6" respectively

### DIFF
--- a/featuretools/primitives/base/primitive_base.py
+++ b/featuretools/primitives/base/primitive_base.py
@@ -45,7 +45,9 @@ class PrimitiveBase(object):
         return os.path.join(config.get("primitive_data_folder"), filename)
 
     def get_args_string(self):
-        arguments, *_, defaults = inspect.getargspec(self.__init__)
+        temp = inspect.getargspec(self.__init__)
+        arguments = temp[0]
+        defaults = temp[:-1]
         if defaults is None:
             defaults = []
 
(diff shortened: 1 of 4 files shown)


featuretools/tests/primitive_tests/test_primitive_base.py
```diff
@@ -0,0 +1,38 @@
...
+class PrimitiveBaseTest(PrimitiveBase):
+
+    def __init__(self, a, b, c=5, d=np.nan):
+        self.a = a
+        self.b = b
+        self.c = c
+        self.d = d
+
+
+def test_get_args_string():
```
kmax12: let's also add a test for `generate_name` where `get_args_string` is called

featuretools/primitives/base/primitive_base.py
```diff
@@ -42,3 +43,31 @@ def get_function(self):
...
+        for arg in args:
+            val = variables[arg]
+            controllable_name += controllable_format.format(arg, val)
+
+        for kwarg, default in zip(kwargs, defaults):
+            val = variables[kwarg]
+            if val != default and (not np.isnan(val) or not np.isnan(default)):
+                controllable_name += controllable_format.format(kwarg, val)
+
+        return controllable_name[:-1]
```
kmax12: handle the comma like using something like `",".join(["a", "b"])`

kmax12:
continuing work on this in #481

PR #434, opened 2019-02-14
replace NUnique get_function

rwedge:


kmax12:
Actually, looks let's test are failing in 2.7. might have to use a different function

PR #439, opened 2019-02-19
Added colab notebook walkthroughs

gsheni:
- Added colab notebooks
- Didn't make a notebook for the [Handling Time](https://docs.featuretools.com/automated_feature_engineering/handling_time.html) page since it has a ton of links and code blocks


kmax12:
what about using https://mybinder.org instead? i think install the requirements is easier with that

gsheni:
Closed via #644

PR #455, opened 2019-03-06
Grouby transform feature

rwedge:
This PR creates a new feature class: `GroupByTransformFeature`

The idea behind this new class of features is to simplify writing primitives like `CumulativeSum` by handling the grouping necessary for the calculation elsewhere.  In order to calculate the cumulative sum of a customer's purchases, you would first need to group the purchases by customer.  With this new `GroupByTransformFeature` class, the primitive function no longer needs to account for this.  It can assume the data passed to it has already been grouped.


featuretools/computational_backends/feature_tree.py
```diff
@@ -9,6 +9,7 @@
 from featuretools import variable_types
 from featuretools.feature_base import (
     AggregationFeature,
+    GroupByTransformFeature,
```
kmax12: should we add a handler so `GroupByTransformFeature` get split by the `feature.groupby` value?

featuretools/computational_backends/pandas_backend.py
```diff
@@ -331,6 +331,46 @@ def strip_values_if_series(values):
...
+
+        frame = entity_frames[entity_id]
+
+        for f in features:
+            # handle when no data
+            if frame.shape[0] == 0:
+                set_default_column(frame, f)
+                continue
+
+            groupby = f.groupby.get_name()
```
kmax12: if we make the change to feature tree, we can ensure that all features have the same groupby once they get to this step. this means we can do the iterations of `for index, group in frame_data.groupby(groupby)` only once per group and calculate all the features at once

featuretools/computational_backends/pandas_backend.py
```diff
@@ -277,7 +280,9 @@ def generate_default_df(self, instance_ids, extra_columns=None):
         return default_df
 
     def _feature_type_handler(self, f):
-        if isinstance(f, TransformFeature):
+        if isinstance(f, GroupByTransformFeature):
```
kmax12: might be worth leaving a comment saying the `GroupByTransformFeature` has to come first since it is a subclass of TransformFeature, unless we change it as suggested above

featuretools/feature_base/feature_base.py
```diff
@@ -462,13 +463,34 @@ def generate_name(self):
...
 
+class GroupByTransformFeature(TransformFeature):
+    def __init__(self, base_features, primitive, groupby):
+        if not isinstance(groupby, FeatureBase):
+            groupby = IdentityFeature(groupby)
+        assert issubclass(groupby.variable_type, Discrete)
+        self.groupby = groupby
+        super(GroupByTransformFeature, self).__init__(base_features,
+                                                      primitive=primitive)
+        self.base_features.append(groupby)
```
kmax12: why not add this to `base_features` before calling the super method?

kmax12: also, not sure it is worth subclassing `TransformFeature`. Doesn't seem like they actually share much.

featuretools/primitives/standard/cum_transform_feature.py
```diff
@@ -31,67 +29,51 @@ class CumCount(TransformPrimitive):
 
     def get_function(self):
         def cum_count(values):
-            return values.groupby(values).cumcount() + 1
+            return pd.Series(range(1, len(values) + 1), index=values.index)
```
kmax12: another option that is easier to read would be. not sure which is more performant though

```
values.iloc[:] = range(1, len(values) + 1)
```

featuretools/feature_base/feature_base.py
```diff
@@ -462,13 +463,39 @@ def generate_name(self):
...
+        if hasattr(base_features, '__iter__'):
+            base_features.append(groupby)
+        else:
+            base_features = [base_features, groupby]
+
+        super(GroupByTransformFeature, self).__init__(base_features,
+                                                      primitive=primitive)
+
+    def copy(self):
+        return GroupByTransformFeature(self.base_features[:-1],
```
kmax12: add comment about excluding groupby

featuretools/feature_base/feature_base.py
```diff
@@ -462,13 +463,39 @@ def generate_name(self):
...
+                                       self.primitive,
+                                       self.groupby)
+
+    def generate_name(self):
+        base_names = [bf.get_name() for bf in self.base_features[:-1]]
```
kmax12: add comment about excluding groupby

PR #457, opened 2019-03-06
Add commutative argument to SubtractNumeric and DivideNumeric primitives

grayskripko:
Add commutative argument to SubtractNumeric and DivideNumeric primitives

### DIFF
--- a/featuretools/primitives/standard/binary_transform.py
+++ b/featuretools/primitives/standard/binary_transform.py
@@ -230,7 +230,9 @@ class SubtractNumeric(TransformPrimitive):
     name = "subtract_numeric"
     input_types = [Numeric, Numeric]
     return_type = Numeric
-    commutative = True
+
+    def __init__(self, commutative=True):
+        self.commutative = commutative
 
     def get_function(self):
         return np.subtract
@@ -307,6 +309,9 @@ class DivideNumeric(TransformPrimitive):
     name = "divide_numeric"
     input_types = [Numeric, Numeric]
     return_type = Numeric
+
+    def __init__(self, commutative=False):
+        self.commutative = commutative
 
     def get_function(self):
         return np.divide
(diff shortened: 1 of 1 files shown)


featuretools/primitives/standard/binary_transform.py
```diff
@@ -230,7 +230,9 @@ class SubtractNumeric(TransformPrimitive):
     name = "subtract_numeric"
```
kmax12: can we add a docstring to this class? Something like this

```
"""Subtracts one feature from another and returns the difference
    Args:
        commutative (bool): determines if Deep Feature Synthesis should generate
            both a - b and b - a, or just one.  If commutative is False, there is no
            guarantee which of the two will be generated. Defaults to True.
"""
```

featuretools/primitives/standard/binary_transform.py
```diff
@@ -308,6 +310,9 @@ class DivideNumeric(TransformPrimitive):
     input_types = [Numeric, Numeric]
```
kmax12: Please add a docstring for this primitive class too. It can be siimilar to the subtract example I gave

featuretools/primitives/standard/binary_transform.py
grayskripko: it looks strange to add `docstring` to `SubtractNumeric` and `DivideNumeric` classes only. All other primitives don't have docstrings. What do you think?

kmax12: it's only in binary_transform.py that don't have docstrings. the other primitive files do. 

we didn't put them here since most people aren't importing and using the binary primitives directly. however, now that you added a parameter, someone will need to know how to use it. 

if you have time, you're more than welcome to add docstrings to the other methods in binary_transform.py!

kmax12:
@grayskripko can you sign the CLA so I can merge this in?

grayskripko:
> @grayskripko can you sign the CLA so I can merge this in? I think you can do it by clicking here: https://cla-assistant.io/Featuretools/featuretools?pullRequest=457

I've done it now but, actually, it was the second time

kmax12:
You're right. I see you also did sign it on March 6th. Not sure why it didn't update it here. Will go ahead and merge now

PR #459, opened 2019-03-11
Test featuretools on CI using packaged source files

### LINKED ISSUE #442: Use packaged files to run tests on circle ci
Our CI tests runs the test cases with access to all the files in the repo, but when conda builds the featuretools package it only uses the files uploaded to pypi.  This can cause conda builds to fail if the hosted package on pypi is missing files.

We can catch this issue earlier in the process if we change the CI tests to also only use the packaged files. One approach might be to build the source distribution that would be uploaded to pypi and use those files to install featuretools and run the tests.

-----

rwedge:
Fixes #442

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,9 +8,22 @@ jobs:
       - run: sudo apt update && sudo apt install -y graphviz
       - checkout
       - run: virtualenv venv
-      - run: source venv/bin/activate && make installdeps
+      - run: |
+          source venv/bin/activate
+          python setup.py sdist
+          FEATURETOOLS_VERSION=$(python setup.py --version)
+          tar -zxvf "dist/featuretools-${FEATURETOOLS_VERSION}.tar.gz"
+      - run: |
+          source venv/bin/activate
+          FEATURETOOLS_VERSION=$(python setup.py --version)
+          pip install "featuretools-${FEATURETOOLS_VERSION}/"
+          pip install -r "featuretools-${FEATURETOOLS_VERSION}/test-requirements.txt"
       - run: source venv/bin/activate && make lint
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -20,9 +32,21 @@ jobs:
...
+      - run: |
+          source venv/bin/activate
+          python setup.py sdist
+          tar -zxvf "dist/featuretools-$(python setup.py --version).tar.gz"
+      - run: |
+          source venv/bin/activate
+          FEATURETOOLS_VERSION=$(python setup.py --version)
+          pip install -e "featuretools-${FEATURETOOLS_VERSION}/"
+          pip install -r "featuretools-${FEATURETOOLS_VERSION}/test-requirements.txt"
+          pip install -r dev-requirements.txt
```
gsheni: Dev requirements installs test-requirements.
So it seems like the above 2 lines are redundant.
https://github.com/Featuretools/featuretools/blob/master/dev-requirements.txt#L1

.circleci/config.yml
```diff
@@ -7,10 +7,22 @@ jobs:
     steps:
       - run: sudo apt update && sudo apt install -y graphviz
       - checkout
-      - run: virtualenv venv
-      - run: source venv/bin/activate && make installdeps
-      - run: source venv/bin/activate && make lint
-      - run: source venv/bin/activate && make test
+      - run: virtualenv test_env && virtualenv dev_env
```
kmax12: can we add any of these repeated steps to the make file? under a command like `build-circle-ci`

rwedge: We can try using commands instead
https://circleci.com/docs/2.0/configuration-reference/#commands-requires-version-21

PR #461, opened 2019-03-12
Primitive call

### LINKED ISSUE #460: Improve API for calling primitive functions
We should make it easier for a user to call primitive functions directly.

Currently, a user has to initialize the primitive and then call `primitve.get_function()`. 

```
In [1]: from featuretools.primitives import NMostCommon 
   ...: import pandas as pd 
   ...: most_common_3 = NMostCommon(n=3) 
   ...: values = pd.Series(["a", "b", "a", "b", "c", "c", "d"]) 
   ...: func = most_common_3.get_function()                                
   ...: func(values)                                                                                                                                                                                             
Out[1]: array(['c', 'b', 'a'], dtype=object)
```
The proposal is to make it like this

```
In [1]: from featuretools.primitives import NMostCommon 
   ...: import pandas as pd 
   ...: most_common_3 = NMostCommon(n=3) 
   ...: values = pd.Series(["a", "b", "a", "b", "c", "c", "d"]) 
   ...: most_common_3(values)                                                                                                                                                                                             
Out[1]: array(['c', 'b', 'a'], dtype=object)
```

CharlesBradshaw:
We don't want to call get_function in the init.

We could store the output of self.get_function() and check if that exists every time in the call function. 
``` python
def __call__(self,data):
    if not hasattr(self,method):
        self.method = self.get_function()
    return self.method(data)
```
That goes against the pythonic "it's Easier to Ask for Forgiveness than Permission" (EAFP) *in this case we're using "Look Before You  Leap" (LBYL)*. To fix this we can use a try except instead.

```python
def __call__(self,data):
    try:
        return self.method(data)
    except AttributeError:
        self.method = self.get_function()
        return self.method(data)
```
We already know when we need to call `get_function` , and it's on the first time` __call__` is run. So instead of using the try except function we can call `get_function` in the `__call__` function, and then redefine `__call__` to be the output of `get_function` This way there are no unnecessary checks

```python
def __call__(self,data):
    self.__call__ = self.get_function()
    return self.__call__(data)
```

This approach doesn't work because `__call__` is a special  method. When a class instance, for example `primitive` gets called like `primitive(data)`, python under the hood runs `type(primitive).__call__(primitive,data)` instead of `primitive.__call__data`

We can get around this by using a non special method as an intermediate.

```python
def __call__(self,*args,**kwargs):
    self._temp_call(*args,**kwargs)

def _temp_call(self, *args, **kwargs):
    self._temp_call = self.get_function().__get__(self)
    return self._temp_call(*args,**kwargs)
```
This way the second time `_temp_call` has been called, it's set to the output of `get_function`. 

Note that the `self.get_function()` has `__get__(self)` also. This is because to add a function to an instance of a class, and not the class itself, you have to bind the instance to the function, aka add `self` as the first parameter always. I don't fully understand how this works, but it's done using `__get__(x)`

Given that we can't overwrite __call__, and that it isn't standard to add/update the methods of an instance, it feels like the if statement, or the try except would be the better solution

Fixes #460

### DIFF
--- a/featuretools/primitives/base/primitive_base.py
+++ b/featuretools/primitives/base/primitive_base.py
@@ -34,6 +34,15 @@ class PrimitiveBase(object):
     # (bool) If True will only make one feature per unique set of base features
     commutative = False
 
+    def __call__(self,data):
+        # We only want to call get_function once, when __call__ is run for the
+        # first time. Since we know that we don't need to call get_function after
+        # that, we can redefine the __call__ function instead of checking for if
+        # it is the first time __call__ has been called 
+        self.__call__ = self.get_function()
+        return self.__call__(data)
+
+
     def generate_name(self):
         raise NotImplementedError("Subclass must implement")
 
--- a/featuretools/primitives/base/primitive_base.py
+++ b/featuretools/primitives/base/primitive_base.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import
 import os
 
 import numpy as np
+import pandas as pd
 
 from featuretools import config
 
@@ -39,7 +40,8 @@ class PrimitiveBase(object):
         # first time. Since we know that we don't need to call get_function after
         # that, we can redefine the __call__ function instead of checking for if
         # it is the first time __call__ has been called
-        self.__call__ = self.get_function()
+        method = self.get_function()
+        self.__call__ = lambda data: method(pd.Series(data))
         return self.__call__(data)
 
     def generate_name(self):
--- a/featuretools/primitives/base/primitive_base.py
+++ b/featuretools/primitives/base/primitive_base.py
@@ -35,14 +35,13 @@ class PrimitiveBase(object):
     # (bool) If True will only make one feature per unique set of base features
     commutative = False
 
-    def __call__(self, data):
-        # We only want to call get_function once, when __call__ is run for the
-        # first time. Since we know that we don't need to call get_function after
-        # that, we can redefine the __call__ function instead of checking for if
-        # it is the first time __call__ has been called
-        method = self.get_function()
-        self.__call__ = lambda data: method(pd.Series(data))
-        return self.__call__(data)
+    def __call__(self, *args, **kwargs):
+        series_args = [pd.Series(arg) for arg in args]
+        try:
+            return self.method(*series_args, **kwargs)
+        except AttributeError:
+            self.method = self.get_function()
+            return self.method(*series_args, **kwargs)
 
     def generate_name(self):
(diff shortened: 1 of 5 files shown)


featuretools/tests/primitive_tests/test_primitive_base.py
```diff
@@ -0,0 +1,9 @@
+from featuretools.primitives import Max
+
+
+def test_call():
```
kmax12: can you add a test where the primitive takes multiple arguments?

featuretools/primitives/base/primitive_base.py
```diff
@@ -34,6 +34,14 @@ class PrimitiveBase(object):
     # (bool) If True will only make one feature per unique set of base features
     commutative = False
 
+    def __call__(self, data):
+        # We only want to call get_function once, when __call__ is run for the
+        # first time. Since we know that we don't need to call get_function after
+        # that, we can redefine the __call__ function instead of checking for if
+        # it is the first time __call__ has been called
+        self.__call__ = self.get_function()
```
kmax12: it's a bit weird to overwrite the method at run time. we can just do 

```
if self.primitive_function is None:
    self.primitive_function = self.get_function()

return  self.primitive_function(data)
```

featuretools/primitives/base/primitive_base.py
```diff
@@ -34,6 +34,14 @@ class PrimitiveBase(object):
     # (bool) If True will only make one feature per unique set of base features
     commutative = False
 
+    def __call__(self, data):
```
kmax12: this won't handle primitives that requires to input variables. We should use `*args` and `**kwargs`

featuretools/tests/primitive_tests/test_primitive_base.py
kmax12: add a test for both an aggregation and transform primitive

kmax12: add a test for primitives that have `uses_calc_time=True`  and get a `time` argument

featuretools/tests/primitive_tests/test_primitive_base.py
```diff
@@ -0,0 +1,69 @@
...
+    answer = 86400.0
+    assert answer == primitive(datetimes, time=datetime(2015, 6, 8))
+
+
+def test_call_multiple_args():
+    class TestPrimitive(PrimitiveBase):
+        def get_function(self):
+            def test(x, y):
+                print("x", x)
+                print("y", y)
```
CharlesBradshaw: Remove prints

PR #469, opened 2019-03-21
Update ecosystem docs

jrkinley:
Add new use case to ecosystem docs.


kmax12:
Looks great. Thank you @jrkinley

PR #470, opened 2019-03-21
FilePath

glentennis:
Adding new variable type called `FilePath`

Am I correct to place this under `sessions` in the entity set tests? I wasn't sure where it should go.


gsheni:
@glentennis `sessions` is the correct entity to place this in.

PR #452, opened 2019-02-28
Add entry points

gsheni:
- Added an entrypoint for featuretools primitives

PR #326, opened 2018-11-25
Allow installation of additional primitives

kmax12:
This PR introduces a method to install additional primitives into an existing Featuretools installation.

Installation from command line

```bash
$ featuretools install s3://featuretools-static/primitives_to_install.tar.gz
Install primitives: CustomSum, CustomMean, CustomMax? (Y/n) Y
100%|███████████████| 3/3 [00:00<00:00, 1187.63it/s]
```

Now a user can access the primitives either by importing or by string

```python
import featuretools as ft
from featuretools.primitives import CustomSum

es = ft.demo.load_mock_customer(return_entityset=True)
feature_matrix, feature_defs = ft.dfs(entityset=es,
                                      target_entity="customers",
                                      agg_primitives=[CustomSum, "custommax"],
                                      trans_primitives=[],
                                      max_depth=2)
```
This returns

```
            zip_code  CUSTOMSUM(transactions.amount)                         ...                          CUSTOMMAX(sessions.CUSTOMSUM(transactions.amount))  CUSTOMMAX(sessions.CUSTOMMAX(transactions.amount))
customer_id                                                                  ...
1              60091                         9025.62                         ...                                                                    1613.93                                              139.43
2              13244                         7200.28                         ...                                                                    1320.64                                              146.81
3              13244                         6236.62                         ...                                                                    1477.97                                              149.15
4              60091                         8727.68                         ...                                                                    1351.46                                              149.95
5              60091                         6349.66                         ...                                                                    1700.67                                              149.02

[5 rows x 7 columns]
```

## How it works

The installation command is provided a directory or tar archive of primitive source files. If it is an archive, the installation script extracts it to directory. The archive can also be remote as in the example above, in which case it is downloaded to a temporary directory first. We use [smart_open](https://github.com/RaRe-Technologies/smart_open) to handle the downloading, so it supports downloading primitives from S3, HDFS, and HTTP/HTTPS.

Primitives are installed by copying their source files into a new directory `installed` within the `featuretools.primitives` submodule. The installation script only considers files with a `.py` extension. The installation script detects if a primitive is in the file by looking for an object that is a subclass of `PrimitiveBase`. Each of these files must only exactly one primitive, otherwise the installation will through an error. When `featuretools.primitives` is a loaded, primitives source files in `installed/` are automatically imported.

To support the CLI, there is new configuration for `entry_points` in the setup.py based on the instruction here: https://chriswarrick.com/blog/2014/09/15/python-apps-the-right-way-entry_points-and-scripts/

[...]

PR #475, opened 2019-03-28
Remove featuretools install command

kmax12:
We added a `featuretools install` command in #326 to allow installation of additional primitives. However, we now have a better way to do it using entry points implemented in #452. This PR removes the first way we implemented and keeps the new way only.


rwedge:
Looks good to me

PR #480, opened 2019-04-05
Generate controllable name

jeff-hernandez:


featuretools/primitives/base/primitive_base.py
```diff
@@ -45,32 +55,26 @@ def get_filepath(self, filename):
...
 
-        for kwarg, default in zip(kwargs, default_values):
-            val = variables[kwarg]
-            # Only add string where val != default
-            # Handles case where val = np.nan and default = np.nan
-            if val != default and not (np.isnan(val) and np.isnan(default)):
-                arg_strings.append(parameter_format.format(kwarg, val))
+        def valid(arg):
+            error = '"{}" must be attribute of {}'
+            assert hasattr(self, arg.name), error.format(arg.name, self.__class__.__name__)
```
jeff-hernandez: I made this assertion because arguments must be an attribute of the instance in order to check whether it has been changed from its default value.

featuretools/primitives/base/primitive_base.py
```diff
@@ -45,32 +56,26 @@ def get_filepath(self, filename):
...
+            is_positional_or_keyword = arg.kind == arg.POSITIONAL_OR_KEYWORD
```
jeff-hernandez: Creating primitives using `make_agg_primitive` and `make_trans_primitive` will return an empty `args_string` because arguments are passed into `__init__` as `**kwargs`.

PR #481, opened 2019-04-05
Automatically generate name for controllable primitives

jeff-hernandez:


featuretools/tests/primitive_tests/test_primitive_base.py
```diff
@@ -58,9 +58,50 @@ def get_function(self):
...
+
+
+def test_args_string_default():
+    class Primitive(PrimitiveBase):
+        def __init__(self, bool=True, int=0, float=None):
+            self.bool = bool
+            self.int = int
+            self.float = float
+
+    string = Primitive().get_args_string()
```
kmax12: let's also add a test case where a user does a mix of all the cases we tested individually. 

* change value from default
* uses the default
* doesn't specify an argument 

e.g `Primitive(bool=False, int=0)` in this test case

featuretools/primitives/base/primitive_base.py
```diff
@@ -51,3 +54,28 @@ def get_function(self):
 
     def get_filepath(self, filename):
         return os.path.join(config.get("primitive_data_folder"), filename)
+
+    def get_args_string(self):
+        if not isinstance(self.__init__, types.MethodType):  # __init__ must be defined
```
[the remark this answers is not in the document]
jeff-hernandez: In this example, the `__init__` is not defined. I included a [test case](https://github.com/Featuretools/featuretools/pull/481/files#diff-2665f8b4ba2ab8f3c27d6454a875fa7eR93).

```
    class Primitive(PrimitiveBase):
        pass
```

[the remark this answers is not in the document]
kmax12: so, I tested in python 3 and it worked without this, but just tested in python 2 and it doesn't so we need this. 

what if we just added this to `PrimitiveBase`?

```
def __init__(self):
        pass
```

I think that might be a cleaner solution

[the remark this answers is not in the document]
jeff-hernandez: Yes, I think that should work.

featuretools/primitives/base/primitive_base.py
```diff
@@ -51,3 +54,28 @@ def get_function(self):
...
+            return ''
+
+        v2 = version_info.major == 2
```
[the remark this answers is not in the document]
jeff-hernandez: Instead of two functions, could we add an optional argument on `inspect_function_args` to restrict keywords like `time`?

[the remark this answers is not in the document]
jeff-hernandez: Never mind, it might be better to split into two functions.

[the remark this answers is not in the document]
kmax12: let's try it with two functions and see how it looks

featuretools/primitives/base/aggregation_primitive_base.py
```diff
@@ -13,13 +13,16 @@ class AggregationPrimitive(PrimitiveBase):
     stack_on_self = True  # whether or not it can be in input_types of self
     allow_where = True  # whether DFS can apply where clause to this primitive
 
-    def generate_name(self, base_feature_names, child_entity_id,
-                      parent_entity_id, where_str, use_prev_str):
+    def generate_name(self, base_feature_names, child_entity_id, parent_entity_id, where_str, use_prev_str):
```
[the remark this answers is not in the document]
jeff-hernandez: Change applied in [this commit.](https://github.com/Featuretools/featuretools/pull/481/commits/42e9679dc1f8dc808ae225fb5740b3d79f7da663)

featuretools/primitives/base/primitive_base.py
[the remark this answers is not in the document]
jeff-hernandez: Change applied in [this commit.](https://github.com/Featuretools/featuretools/pull/481/commits/1c9b4e59ef9157b00ce014e27f49367246dacd89)

featuretools/primitives/base/primitive_base.py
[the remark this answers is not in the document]
jeff-hernandez: Change applied in [this commit.](https://github.com/Featuretools/featuretools/pull/481/commits/887e408ec6c909ee8677b9dd2cf6e3556f562d36)

featuretools/tests/primitive_tests/test_primitive_base.py
jeff-hernandez: Change applied in [this commit.](https://github.com/Featuretools/featuretools/pull/481/commits/bfb9d689acb3f1bbd6a62354687b8784beae6d64)

featuretools/primitives/base/primitive_base.py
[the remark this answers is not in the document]
jeff-hernandez: Should the test case in the link above be deleted?

featuretools/primitives/standard/aggregation_primitives.py
```diff
@@ -171,6 +171,7 @@ class NMostCommon(AggregationPrimitive):
     return_type = Discrete
 
     def __init__(self, n=3):
+        self.n = n
         self.number_output_features = n
```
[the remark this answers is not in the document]
jeff-hernandez: I got these errors after removing `number_output_features`.

- `test_pandas_backend.py::test_topn`
```
>               assert set(true_results.loc[i].values) == set(df.loc[i].values)
E               AssertionError: assert {'coke zero', 'toothpaste'} == {'toothpaste'}
E                 Extra items in the left set:
E                 'coke zero'
E                 Use -v to get the full diff

featuretools\tests\computational_backend\test_pandas_backend.py:457: AssertionError
```
- `test_deep_feature_synthesis.py::test_transform_no_stack_agg`
```
>       assert not feature_with_name(feature_defs, 'id != N_MOST_COMMON(sessions.device_type)')
E       AssertionError: assert not True
E        +  where True = feature_with_name([<Feature: cohort>, <Feature: age>, <Feature: région_id>, <Feature: loves_ice_cream>, <Feature: cancel_reason>, <Feature: engagement_level>, ...], 'id != N_MOST_COMMON(sessions.device_type)')
```
- `test_encode_features.py::test_encode_features_topn`
```
>               raise Exception('Must produce aggregated value')
E               Exception: Must produce aggregated value

..\..\..\..\miniconda3\lib\site-packages\pandas\core\groupby\groupby.py:3627: Exception
```

jeff-hernandez:
Creating primitives using `make_agg_primitive` and `make_trans_primitive` will return an empty `args_string` because arguments are not assigned as attributes in the class and passed into `__init__ ` as `**kwargs`.

jeff-hernandez:
If a primitive argument is stored as a pandas series, then the generated name won't be very readable. Is there a preferred method on working with these types of arguments?
```
primitive = Primitve(argument=pd.Series(range(4)))
string = primitve.get_args_string()
```
```
PRIMITIVE(argument='0    1\n1    2\ndtype: int64', ...)
```

featuretools/primitives/standard/aggregation_primitives.py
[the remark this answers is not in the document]
kmax12: ahh, number_output_features is required for multi-output features. let's leave it in

featuretools/primitives/base/primitive_base.py
```diff
@@ -51,3 +57,23 @@ def get_function(self):
...
+    def get_args_string(self):
+        def valid(arg):
+            error = '"{}" must be attribute of {}'
+            assert hasattr(self, arg.name), error.format(arg.name, self.__class__.__name__)
+            is_positional_or_keyword = arg.kind == arg.POSITIONAL_OR_KEYWORD
+            is_same_type = isinstance(getattr(self, arg.name), type(arg.default))
+            default = is_same_type and arg.default == getattr(self, arg.name)
+            return is_positional_or_keyword and not default
+
+        string = OrderedDict()
```
kmax12: does this need to be an `OrderedDict`? why not just make it a list

featuretools/primitives/base/primitive_base.py
```diff
@@ -51,3 +57,23 @@ def get_function(self):
...
+        args = signature(self.__class__).parameters.values()
+        for arg in args:
+            if not valid(arg):
+                continue
+            string[arg.name] = str(getattr(self, arg.name))
+        if len(string) == 0:
+            return ''
+        string = ', ' + ', '.join(map('='.join, string.items()))
```
kmax12: can we break this into two lines to make it easier to read

featuretools/primitives/base/primitive_base.py
```diff
@@ -51,3 +57,23 @@ def get_function(self):
 
     def get_filepath(self, filename):
         return os.path.join(config.get("primitive_data_folder"), filename)
+
+    def get_args_string(self):
+        def valid(arg):
+            error = '"{}" must be attribute of {}'
```
kmax12: can you add a docstring and/or comments this function? there's a lot of logic here that's a bit hard to quickly grasp.

it may also be helpful to actually refactor have multiple returns statements, so it's easier to follow through the different cases

kmax12:
> If a primitive argument is stored as a pandas series, then the generated name won't be very readable. Is there a preferred method on working with these types of arguments?
> 
> ```
> primitive = Primitve(argument=pd.Series(range(4)))
> # PRIMITIVE(argument='0    0\n1    1\n2    2\n3    3\ndtype: int64', ...)
> ```

for now, let's not handle the series case specifically

featuretools/primitives/base/utils.py
```diff
@@ -1,25 +1,22 @@
 try:
-    # python 3.7 deprecated getargspec
-    from inspect import getfullargspec as getargspec
+    # python 3
+    from inspect import signature
 except ImportError:
-    # python 2.7 - 3.6 backwards compatibility import
-    from inspect import getargspec
+    # python 2
+    from funcsigs import signature
```
rwedge: We should add this module as a dependency for python2

featuretools/primitives/base/primitive_base.py
jeff-hernandez: No, doesn't need to be. I will make into list.

featuretools/primitives/base/utils.py
jeff-hernandez: Okay sounds good.

featuretools/primitives/base/primitive_base.py
jeff-hernandez: Yes, I refactored and added comments [in this commit.](https://github.com/Featuretools/featuretools/pull/481/commits/38c542d102973819a5a0ce22cbe692dad1c4a340#diff-65ffde62fc03046613de064c794ed250R61)

featuretools/primitives/base/primitive_base.py
jeff-hernandez: Change applied in [this commit.](https://github.com/Featuretools/featuretools/pull/481/commits/38c542d102973819a5a0ce22cbe692dad1c4a340#diff-65ffde62fc03046613de064c794ed250R86)

featuretools/primitives/base/utils.py
jeff-hernandez: Change applied in [this commit.](https://github.com/Featuretools/featuretools/pull/481/commits/fa7d29d44f24e6d298332aaa9fc385262c590e0f#diff-b4ef698db8ca845e5845c4618278f29aR13)

featuretools/primitives/base/primitive_base.py
```diff
@@ -51,3 +56,33 @@ def get_function(self):
...
+
+    def get_args_string(self):
+        strings = []
+        args = signature(self.__class__).parameters.items()
+        for name, arg in args:
+            # assert that arg is attribute of primitive
+            error = '"{}" must be attribute of {}'
+            assert hasattr(self, arg.name), error.format(arg.name, self.__class__.__name__)
+
+            # skip if *args or **kwargs
```
kmax12: should this be "skip if not"?

featuretools/primitives/base/primitive_base.py
kmax12: looks much better! easy to follow now

featuretools/primitives/base/primitive_base.py
jeff-hernandez: The `*args` and `**kwargs`  would not be considered `POSITIONAL_OR_KEYWORD`. This comment could work as well.
```
# skip if not a standard argument (e.g. excluding *args and **kwargs)
```

kmax12: looks good

PR #482, opened 2019-04-05
Update how standard primitives are imported internally

kmax12:
This PR updates how the standard primitives are imported to better support loading extra primitives from entry points. 

As things stand, if a primitive is loaded from an entry point it overwrites the standard primitive. 

With this PR, a user can still access the standard primitives using `from featuretools.primitives.standard import PrimitiveName`


rwedge:
Usually we use an `api.py` file to manage the individual imports for a module and have `__init__.py` import everything from `api.py`.

PR #490, opened 2019-04-11
Update Sphinx Version

gsheni:
- Update sphinx and associated libraries version
- Added Google docstring parsing settings of sphinx to `conf.py`

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -10,5 +10,6 @@ matplotlib==3.0.2; python_version>'3.4'
 matplotlib==2.2.3; python_version<'3'
 nbconvert==5.4.1
 nbsphinx==0.4.2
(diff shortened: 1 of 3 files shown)


dev-requirements.txt
```diff
@@ -8,6 +8,8 @@ isort==4.3.4
 jupyter==1.0.0
 matplotlib==3.0.2; python_version>'3.4'
 matplotlib==2.2.3; python_version<'3'
+nbconvert==5.4.1
```
rwedge: jupyter==1.0.0 installs nbconvert

gsheni: yes but nbsphinx does not want nbconvert==5.4.0
https://github.com/spatialaudio/nbsphinx/blob/master/setup.py#L18

PR #492, opened 2019-04-12
Transform primitives docstring updates

thehomebrewnerd:
Updated docstrings and added doctests for transform primitives.


featuretools/primitives/standard/transform_primitive.py
```diff
@@ -62,7 +94,17 @@ def pd_diff(base_array, group_array):
 
 
 class Day(TransformPrimitive):
-    """Transform a Datetime feature into the day."""
+    """Determines the day value of a datetime.
```
kmax12: maybe add "day of month" to make this more clear?

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -181,7 +319,14 @@ def get_function(self):
 
 
 class NumWords(TransformPrimitive):
-    """Returns the number of words in a given string by counting the spaces.
+    """Determines the number of words in a string by counting the spaces.
```
kmax12: maybe add a line and example with some punctuation so it's clear how that is handled

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -194,22 +339,42 @@ def word_counter(array):
 
 
 class TimeSince(TransformPrimitive):
-    """Calculates time since the cutoff time."""
+    """Calculates time from a value to a specified cutoff datetime.
```
kmax12: what unit?

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -298,8 +508,12 @@ def get_function(self):
 
 
 class Percentile(TransformPrimitive):
-    """For each value of the base feature, determines the percentile in
-        relation to the rest of the feature.
+    """Determines the percentile rank for each value of the base feature.
```
kmax12: write and show how nans are handled

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -336,19 +564,24 @@ def get_function(self):
 
 class Haversine(TransformPrimitive):
     """Calculate the approximate haversine distance between two LatLong
```
kmax12: to be consistent, this should be `Calculates`

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -298,8 +509,17 @@ def get_function(self):
 
 
 class Percentile(TransformPrimitive):
-    """For each value of the base feature, determines the percentile in
-        relation to the rest of the feature.
+    """Determines the percentile rank for each value of the base feature.
```
kmax12: rather than say base feature, we should say "Determines the percentile rank for each value in a list".

the usage of base feature is old and should be replaced everywhere in the file

thehomebrewnerd: Am I correct in assuming you only want this replaced in the user-facing documentation and not also in the code that uses the base_feature terminology?

kmax12: correct. let's just update the docstrings.

PR #493, opened 2019-04-12
Add CLI entrypoint

gsheni:
- Add entrypoint to add/override cli commands

### DIFF
--- a/featuretools/__main__.py
+++ b/featuretools/__main__.py
@@ -40,7 +40,6 @@ for entry_point in pkg_resources.iter_entry_points('featuretools_cli'):
             for name, cmd in loaded.commands.items():
                 cli.add_command(cmd=cmd, name=name)
     except Exception:
-        raise
         pass
 
 if __name__ == "__main__":
--- a/featuretools/__main__.py
+++ b/featuretools/__main__.py
@@ -26,15 +26,21 @@ def info():
     print("")
 
     installed_packages = get_installed_packages()
-
     deps = [
-        ("featuretools", installed_packages['featuretools']),
-        ("pandas", installed_packages['pandas']),
         ("numpy", installed_packages['numpy']),
-        ("scipy", installed_packages['scipy']),
+        ("pandas", installed_packages['pandas']),
+        ("tqdm", installed_packages['tqdm']),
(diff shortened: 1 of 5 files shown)


featuretools/utils/cli_utils.py
```diff
@@ -0,0 +1,42 @@
...
+import platform
+import struct
+import sys
+
+import pkg_resources
+
+
+# Modified from here
+# https://github.com/pandas-dev/pandas/blob/d9a037ec4ad0aab0f5bf2ad18a30554c38299e57/pandas/util/_print_versions.py#L11
+def get_sys_info():
```
kmax12: can we add a test case that calls `get_installed_packages` and `get_sys_info` even if it doesn’t check the outputs? that would at least catch any runtime errors

featuretools/__main__.py
```diff
@@ -1,8 +1,10 @@
 import click
 import pandas as pd
+import pkg_resources
 
 import featuretools
 from featuretools.primitives.utils import get_featuretools_root
+from featuretools.utils.cli_utils import get_installed_packages, get_sys_info
```
kmax12: let’s also make them importable from `featuretools.utils` and access them from there. that way if we change the `cli_utils.py` file name it doesnt cause problems

featuretools/utils/cli_utils.py
gsheni: Fixed.

featuretools/__main__.py
gsheni: Fixed.

featuretools/__main__.py
```diff
@@ -14,9 +16,36 @@ def cli():
...
+        ("scikit-learn", installed_packages['scikit-learn']),
+        ("pip", installed_packages['pip']),
+        ("setuptools", installed_packages['setuptools']),
+    ]
+    for k, stat in deps:
+        print("{k}: {stat}".format(k=k, stat=stat))
 
-@click.command()
+
+@click.command(name='list-primitives')
```
gsheni: Replace the underscore with a dash to be more consistent with other CLI tools. 
https://github.com/pallets/click/issues/1123#issuecomment-425167243

PR #494, opened 2019-04-13
cum_transform_feature Docstrings

glentennis:
Rewriting docstrings in `cum_transform_feature.py` to include a full description and doctest.

### DIFF
--- a/featuretools/primitives/standard/cum_transform_feature.py
+++ b/featuretools/primitives/standard/cum_transform_feature.py
@@ -96,8 +96,8 @@ class CumMin(TransformPrimitive):
 
     Examples:
         >>> cum_min = CumMin()
-        >>> cum_min([1, 2, 3, 4, None, 5]).tolist()
-        [1.0, 1.0, 1.0, 1.0, nan, 1.0]
+        >>> cum_min([1, 2, -3, 4, None, 5]).tolist()
+        [1.0, 1.0, -3.0, -3.0, nan, -3.0]
     """
     name = "cum_min"
     input_types = [Numeric]
--- a/featuretools/primitives/standard/cum_transform_feature.py
+++ b/featuretools/primitives/standard/cum_transform_feature.py
@@ -10,13 +10,14 @@ class CumSum(TransformPrimitive):
     Description:
         Given a list of values, return the cumulative sum
         (or running total). There is no set window, so the
-        sum at each point is calculated over all prior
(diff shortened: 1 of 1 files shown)


featuretools/primitives/standard/cum_transform_feature.py
```diff
@@ -5,8 +5,19 @@
...
+    Description:
+        Given a list of values, return the cumulative sum
+        (or running total). There is no set window, so the
+        sum at each point is calculated over all prior
+        values. Ignores `NaN` values.
+
+    Examples:
+        >>> cum_sum = CumSum()
+        >>> cum_sum([1, 2, 3, 4, 5, None]).tolist()
+        [1.0, 3.0, 6.0, 10.0, 15.0, nan]
```
kmax12: it doesn't look like nan's are getting ignored in this example

featuretools/primitives/standard/cum_transform_feature.py
```diff
@@ -65,8 +109,19 @@ def cum_min(values):
...
+    Description:
+        Given a list of values, return the cumulative max
+        (or running max). There is no set window, so the
+        max at each point is calculated over all prior
+        values. `NaN` values are ignored.
+
+    Examples:
+        >>> cum_max = CumMax()
+        >>> cum_max([1, 2, 3, 4, 5, None]).tolist()
+        [1.0, 2.0, 3.0, 4.0, 5.0, nan]
```
kmax12: same here, doesn't like look like nan is getting ignored

glentennis:
@kmax12 I edited the descriptions and doctests to be more clear about the way `NaN` values are handled, let me know what you think!

Also, I noticed the way `NaN` values are handled in `CumMean` might not be optimal. `NaN` values in a window are treated as zero, so when given the list `[1, None, 1]`, CumMean returns:

```
[1, nan, .66]
```

When perhaps we would expect it to return

```
[1, nan, 1]
```
should i open an issue to fix this?

featuretools/primitives/standard/cum_transform_feature.py
```diff
@@ -5,8 +5,20 @@
 
 
 class CumSum(TransformPrimitive):
-    """Returns the cumulative sum after grouping"""
-
+    """Calculates the cumulative sum over a grouping.
```
kmax12: no need to say "over a grouping" for all these methods. the "over a grouping" is from an older version of featuretools

PR #495, opened 2019-04-15
BinaryTransform docstrings

glentennis:
Adding docstrings to BinaryTransform primitives, with a full description, args, and doctests.


featuretools/primitives/standard/binary_transform.py
```diff
@@ -15,6 +15,18 @@
 
 
 class GreaterThan(TransformPrimitive):
+    """Determines if values X are greater than values Y.
```
kmax12: what are X and Y? this should be "Determines if values in one list are greater than another list". fine to keep x and y in extended description below though

let's apply this rule to all the primitives below too

PR #499, opened 2019-04-16
Remove primitives that return strings instead of functions

kmax12:
We were returning strings in some cases to improve performance. This breaks api with other primitives. We can simply return pandas objects instead add the string optimization into pandas backend.

### DIFF
--- a/featuretools/primitives/standard/aggregation_primitives.py
+++ b/featuretools/primitives/standard/aggregation_primitives.py
@@ -7,7 +7,7 @@ import pandas as pd
 
 from ..base.aggregation_primitive_base import AggregationPrimitive
 
-import featuretools as ft
+from featuretools.utils import is_python_2
 from featuretools.variable_types import (
     Boolean,
     DatetimeTimeIndex,
@@ -27,9 +27,9 @@ class Count(AggregationPrimitive):
     default_value = 0
 
     def get_function(self):
-        # note: returning pd.Series.nunique errors for python2,
+        # note: returning class instance method errors for python2,
         # so using this branching code path while we support python2
-        if ft.utils.is_python_2():
+        if is_python_2():
             return pd.Series.count.__func__
         else:
             return pd.Series.count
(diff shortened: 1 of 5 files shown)


featuretools/primitives/standard/aggregation_primitives.py
```diff
@@ -26,7 +26,7 @@ class Count(AggregationPrimitive):
     default_value = 0
 
     def get_function(self):
-        return 'count'
+        return pd.Series.count
```
kmax12: this actually isn't as performant. added question on stackoverflow here: https://stackoverflow.com/questions/55731149/use-a-function-instead-of-string-in-pandas-groupby-agg

featuretools/primitives/standard/aggregation_primitives.py
```diff
@@ -261,7 +261,7 @@ class Skew(AggregationPrimitive):
     stack_on_self = False
 
     def get_function(self):
-        return 'skew'
+        return pd.Series.skew
```
kmax12: skew doesn't seem to matter as much as count

Screen Shot 2019-04-17 at 11 36 02 AM

featuretools/primitives/standard/aggregation_primitives.py
```diff
@@ -26,7 +27,12 @@ class Count(AggregationPrimitive):
     default_value = 0
 
     def get_function(self):
-        return 'count'
+        # note: returning pd.Series.nunique errors for python2,
```
rwedge: this note is about `Count`, so "pd.Series.count"?

featuretools/primitives/standard/aggregation_primitives.py
```diff
@@ -261,7 +264,12 @@ class Skew(AggregationPrimitive):
     stack_on_self = False
 
     def get_function(self):
-        return 'skew'
+        # note: returning pd.Series.nunique errors for python2,
```
rwedge: skew for this note

featuretools/primitives/standard/aggregation_primitives.py
```diff
@@ -7,6 +7,7 @@
 
 from ..base.aggregation_primitive_base import AggregationPrimitive
 
+import featuretools as ft
```
rwedge: can we import `is_python_2` instead of the entire featuretools module?

rwedge:
Looks good

Issue #496, opened 2019-04-16
Improve error message when Categorical variable has NaNs


johnnyheineken:
### Bug Description
When categorical variable has `NaN`s, ValueError with following traceback is shown:

In categorical columns, replacing `np.NaN` with string `NaN` helped to remove this error.

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-6-8bb8b3216cc3> in <module>
      6     target_entity=MONTH_ENTITY_NAME,
      7     features_only=not RUN_DFS,
----> 8     **dfs_definition
      9 )
     10 

c:\users\jan.hynek\appdata\local\programs\python\python37-32\lib\site-packages\featuretools\synthesis\dfs.py in dfs(entities, relationships, entityset, target_entity, cutoff_time, instance_ids, agg_primitives, trans_primitives, groupby_trans_primitives, allowed_paths, max_depth, ignore_entities, ignore_variables, seed_features, drop_contains, drop_exact, where_primitives, max_features, cutoff_time_in_index, save_progress, features_only, training_window, approximate, chunk_size, n_jobs, dask_kwargs, verbose, return_variable_types)
    233                                                   n_jobs=n_jobs,
    234                                                   dask_kwargs=dask_kwargs,
--> 235                                                   verbose=verbose)
    236     return feature_matrix, features

c:\users\jan.hynek\appdata\local\programs\python\python37-32\lib\site-packages\featuretools\computational_backends\calculate_feature_matrix.py in calculate_feature_matrix(features, entityset, cutoff_time, instance_ids, entities, relationships, cutoff_time_in_index, training_window, approximate, save_progress, verbose, chunk_size, n_jobs, dask_kwargs, profile)
    252                                                    target_time=target_time,
    253                                                    pass_columns=pass_columns,
--> 254                                                    dask_kwargs=dask_kwargs or {})
    255     else:
    256         feature_matrix = linear_calculate_chunks(chunks=chunks,

c:\users\jan.hynek\appdata\local\programs\python\python37-32\lib\site-packages\featuretools\computational_backends\calculate_feature_matrix.py in parallel_calculate_chunks(chunks, features, approximate, training_window, verbose, save_progress, entityset, n_jobs, no_unapproximated_aggs, cutoff_df_time_var, target_time, pass_columns, dask_kwargs)
    600             pbar = make_tqdm_iterator(total=len(_chunks), bar_format=pbar_str)
    601         for batch in iterator:
[...]

PR #504, opened 2019-04-17
Removing type casting in Equals and NotEquals Primitives

### LINKED ISSUE #500: Loading previously saved features: ValueError: cannot convert float NaN to integer
Not sure whether this is a bug or I am doing something incorrectly, but:

assume the feature matrix is created as in all of the example documentation:

```
feature_matrix, feature_defs = ft.dfs(entityset = es,
                                      target_entity = "test",
                                      agg_primitives = ["mean","min","median"],
                                      verbose = True,
                                      n_jobs = 1,
                                      max_depth = 2)
```

this is working fine and I am able to save and export these features without any problems.

However, if we want to move this into production the following workflow breaks:

1. Storing created features:
```
feature_matrix_enc, features_enc = ft.encode_features(feature_matrix, feature_defs)
```
2. Creating file for the export:
```
ft.save_features(features_enc, "feature_definitions")
```
3. Loading the features (for the test case - nothing _actually_ changes, same live environment):
```
saved_features = ft.load_features("feature_definitions")
```
4. Transforming them back, using the same entity ``es`` that was used to create the features:
```
feature_matrix = ft.calculate_feature_matrix(saved_features, es)
```
However, the last step returns the following error:

```
...

  File "pandas\_libs\lib.pyx", line 456, in pandas._libs.lib.astype_intsafe

  File "pandas/_libs/src\util.pxd", line 142, in util.set_value_at_unsafe

ValueError: cannot convert float NaN to integer
```

Is there a way to fix this? If the original feature matrix was created despite those NaN values, can we ignore them on load? Or am I missing something?

Big thanks!

kmax12:
We were doing the casting to be careful, but there is no need because pandas will handle it internally

This should fix #500 and #496


kmax12:
This is appears to be the problem here: https://stackoverflow.com/questions/57601595/how-to-write-seed-features-that-work-with-cutoff-time?noredirect=1#comment101752737_57601595

levng:
This PR fixes the problem! The `ValueError: Cannot convert non-finite values (NA or inf) to integer` error no longer occurs and the features are built without a hitch.

apemann:
Can this fix make it into the next release please? It looks like it's almost done.

apemann:
Thank you @rwedge and @kmax12!

PR #511, opened 2019-04-25
More Informative N Jobs Print

### LINKED ISSUE #508: Print number of workers when running in parallel
When featuretools runs with `n_jobs` great than 1, we print out a message like `EntitySet scattered to workers in 10.685 seconds`. We should update it to say `EntitySet scattered to X workers in 10 seconds` (add # workers and only need to the second precision on scattering time).

Here is relevant code: 

https://github.com/Featuretools/featuretools/blob/master/featuretools/computational_backends/calculate_feature_matrix.py#L575

CharlesBradshaw:
- Fixes #508 (Print number of workers when running in parallel)

Also gives a warning if the user asks for more jobs than were created. For example if the user asks for 10 jobs and only 4 workers are created (either due to not enough cores or not enough tasks) a warning will show

Also gives a warning if the EntitySet is not scattered to all of the created workers.


featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -44,7 +48,17 @@ def int_es():
     return make_ecommerce_entityset(with_integer_time_index=True)
 
 
+def test_scatter_warning():
+    with warnings.catch_warnings(record=True) as w:
```
rwedge: `pytest.warns` could probably work here

featuretools/tests/utils_tests/test_computational_backend_utils.py
```diff
@@ -0,0 +1,12 @@
+import warnings
+
+from featuretools.computational_backends.utils import create_client_and_cluster
+
+
+def test_create_client_and_cluster():
+    with warnings.catch_warnings(record=True) as w:
```
rwedge: same comment as with `test_scatter_warning`

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -44,7 +48,17 @@ def int_es():
...
+        # Cause all warnings to always be triggered.
+        warnings.simplefilter("always")
+
+        scatter_warning(1, 2)
+        assert len(w) == 1
+
 # TODO test mean ignores nan values
+
```
rwedge: Can you remove this TODO, it's no longer a TODO

featuretools/computational_backends/utils.py
```diff
@@ -214,8 +214,17 @@ def create_client_and_cluster(n_jobs, num_tasks, dask_kwargs, entityset_size):
...
+        cpu_workers = n_jobs_to_workers(n_jobs)
+        workers = min(cpu_workers, num_tasks)
+        if n_jobs != -1 and workers < n_jobs:
+            warning_string = "{} workers requested, but only {} workers created."
+            warning_string = warning_string.format(n_jobs, workers)
+            if cpu_workers < n_jobs:
+                warning_string += " Not enough cpu cores({}).".format(cpu_workers)
+
+            if num_tasks < n_jobs:
+                warning_string += " Not enough tasks({}).".format(num_tasks)
```
rwedge: Can you change this to " Not enough chunks (X), consider reducing the chunk size"

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -569,12 +576,15 @@ def parallel_calculate_chunks(chunks, features, approximate, training_window,
...
+        num_workers = len(client.scheduler_info()['workers'].values())
+
+        scatter_warning(num_scattered_workers, num_workers)
         if verbose:
             end = time.time()
             scatter_time = end - start
-            scatter_string = "EntitySet scattered to workers in {:.3f} seconds"
-            print(scatter_string.format(scatter_time))
-
+            scatter_string = "EntitySet scattered to {} workers in {:.3f} seconds"
```
rwedge: The issue wanted to remove the decimal part of the "in X seconds" message.  Maybe round to integer and if it took less than 1 second  display "<1" or "under 1"?

CharlesBradshaw: I added in the rounding but I decided to not add in the "<1" so that I don't have to deal with a complex code cov situation. In theory I could create a context manager to time it, and then run an on complete function which then prints out a special string but it feels too complex for a very minor piece of functionality.

kmax12: I think it'd be easiest to just round up to the near integer seconds. This is just diagnostic info, so +/- 1 second doesn't matter that much

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -569,12 +576,15 @@ def parallel_calculate_chunks(chunks, features, approximate, training_window,
...
+        num_workers = len(client.scheduler_info()['workers'].values())
+
+        scatter_warning(num_scattered_workers, num_workers)
         if verbose:
             end = time.time()
-            scatter_time = end - start
-            scatter_string = "EntitySet scattered to workers in {:.3f} seconds"
-            print(scatter_string.format(scatter_time))
-
+            scatter_time = round(end - start)
```
rwedge: instead of rounding up `round` rounds to the nearest integer.  We could use a function that rounds up or take the min of the rounded number and 1

CharlesBradshaw: What is the motivation for wanting to tell the user it took 1 second if it actually took 0.001 seconds

rwedge: I thought saying something took 0 seconds seem strange to the user since clearly it took at least some time to do, but it's probably fine so let's just leave it as is.

PR #523, opened 2019-04-30
Only calculate features for instances before cutoff

### LINKED ISSUE #437: Only calculate features for instances before cutoff time
Currently, if no instance ids are provided to `ft.dfs` or `ft.calculate_feature_matrix`, we calculate features for all instances in the target entity. 

However, if a single cutoff time is provided, we only select the instances where the time index is less than or equal to the cutoff time. 

The current logic that selects all instance ids is here: 
https://github.com/Featuretools/featuretools/blob/72ac0daa7bd60936eeb563b2c352c9a6631f50fc/featuretools/computational_backends/calculate_feature_matrix.py#L142

To fix this issue we have to add logic that checks if the target entity has a time index and removes instances from after the cutoff time.

CJStadler:
When a single cutoff_time is provided.

Resolves #437

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -142,7 +142,10 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
 
         if instance_ids is None:
             index_var = target_entity.index
-            instance_ids = target_entity.df[index_var].tolist()
+            df = target_entity._handle_time(target_entity.df,
+                                            time_last=cutoff_time,
+                                            training_window=training_window)
+            instance_ids = df[index_var].tolist()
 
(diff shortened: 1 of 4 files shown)


rwedge:
From the coverage report it looks like some code related to missing data relied on the old way of using all the instances in order to be tested.

https://github.com/Featuretools/featuretools/blob/269307e8b1e2214e8ad422d524e412ce0d47f4b3/featuretools/computational_backends/pandas_backend.py#L252-L255

https://github.com/Featuretools/featuretools/blob/269307e8b1e2214e8ad422d524e412ce0d47f4b3/featuretools/computational_backends/pandas_backend.py#L277-L280

We might be able to test this code with a test using a single cutoff time and a list of instance ids that include instances whose time index is after the cutoiff time.

CJStadler:
@rwedge I added a test and I think the coverage is good now. Back to you.

PR #525, opened 2019-05-01
Better cli info

gsheni:
- create a `show_info` function, which can be directly called `featuretools.show_info()` in python
- `featuretools info` CLI click command calls that  `show_info` function


featuretools/utils/cli_utils.py
```diff
@@ -40,3 +80,7 @@ def get_installed_packages():
     for d in pkg_resources.working_set:
         installed_packages[d.project_name] = d.version
     return installed_packages
+
+
+def get_featuretools_root():
```
kmax12: if we move this, we should also move the test to existing `test_cli_utils.py` file

gsheni: Fixed.

PR #533, opened 2019-05-07
Drop variables in a batch in `normalize_entity`

CJStadler:
Dropping the `additional_variables` from the original dataframe takes a significant portion of the time of `normalize_entity`, and this increases with the number of `additional_variables`. For example, with 10 variables about 40% of the time is spent in `df.drop`, compared to 10% for 1 variable.

Currently we call `df.drop` once for each variable. The number of variables passed to `df.drop` does not affect its running time (in my benchmarking), so dropping them all at once is O(1) instead of O(n).

`Entity.delete_variable` is no longer used, but I wasn't sure if it is considered public.


featuretools/entityset/entity.py
```diff
@@ -406,9 +406,18 @@ def delete_variable(self, variable_id):
         Remove variable from entity's dataframe and from
         self.variables
         """
-        self.df.drop(variable_id, axis=1, inplace=True)
-        v = self._get_variable(variable_id)
-        self.variables.remove(v)
+        self.delete_variables([variable_id])
```
rwedge: Can you add a test case for this method

CJStadler: 👍 And do you think we need to keep `delete_variable`?

rwedge: Let's just remove it. It's unused and `delete_variables([variable_id])` achieves the same result

kmax12:
> Another significant part of the run time is this sort call: https://github.com/Featuretools/featuretools/blob/master/featuretools/entityset/entityset.py#L744
> 
> It's unclear to me why this would be necessary, and removing it doesn't break any of the tests.

the reason we do the sort is because we have an assumption that dataframes are sorted by their time index. This assumption is import so a primitive like cumulative sum can assume it is getting values in sorted order.

That being said, the sorting here might just be defensive coding and unnecessary if we can verify the dataframe is sorted prior to this call or after this call.

PR #535, opened 2019-05-09
Remove unnecessary sorting from normalize_entity

CJStadler:
We can assume the base entity dataframe is sorted by time index, so if
we are using the same time index for the new entity then there is no
need to sort.

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -729,12 +729,17 @@ class EntitySet(object):
             make_time_index = True
 
         if isinstance(make_time_index, str):
+            # Set the new time index to make_time_index.
             base_time_index = make_time_index
             new_entity_time_index = make_time_index
+            already_sorted = (new_entity_time_index == base_entity.time_index)
         elif make_time_index:
+            # Create a new time index based on the base entity time index.
             base_time_index = base_entity.time_index
             if new_entity_time_index is None:
                 new_entity_time_index = "first_%s_time" % (base_entity.id)
+
+            already_sorted = True
 
             assert base_entity.time_index is not None, \
                 "Base entity doesn't have time_index defined"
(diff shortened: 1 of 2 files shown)


featuretools/entityset/entityset.py
```diff
@@ -730,7 +730,7 @@ def normalize_entity(self, base_entity_id, new_entity_id, index,
 
         if isinstance(make_time_index, str):
             base_time_index = make_time_index
-            new_entity_time_index = base_entity[make_time_index].id
```
CJStadler: The only reason I can see for this is to check that the `make_time_index` exists in the base entity (if it doesn't then this would raise). If that should be checked here then I think it would be better to add an explicit assertion.

rwedge: Yes let's add an explicit assertion

CJStadler: It looks like this is already checked when the entity is created:
https://github.com/Featuretools/featuretools/blob/4b6a5a4d38de94a64c08ad15b5b8cefa9954b6df/featuretools/entityset/entity.py#L574-L575
Is it worthwhile to add an earlier check?

rwedge: Let's split this off to a separate issue to be dealt with alter.  Looks like there are two situations where we could improve the error message:
* Specifying an existing column for `make_time_index` but not adding that column to `additional_variables` or `copy_variables`
* Setting `make_time_index` to a column that doesn't exist in the base frame

CJStadler: 👍 I can open an issue.

featuretools/entityset/entityset.py
```diff
@@ -790,6 +792,7 @@ def normalize_entity(self, base_entity_id, new_entity_id, index,
             new_entity_id,
             new_entity_df,
             index,
+            already_sorted=already_sorted,
```
rwedge: Can we update the `normalize_entityset` tests to confirm that created time indexes are sorted

PR #536, opened 2019-05-09
Cache feature names

CJStadler:
Computing the name of a feature can be expensive because primitives use
reflection to generate their names, and when features are stacked the
number of calls to `get_name()` grows. This commit caches the name when it
is computed the first time. Is it reasonable to assume that features are
immutable, or are there places where we should invalidate the cache?

With `max_depth=2` this sped up `ft.dfs` (just building the features) by
about 3x, and the improvement is greater for higher values of `max_depth`.

### DIFF
--- a/featuretools/feature_base/feature_base.py
+++ b/featuretools/feature_base/feature_base.py
@@ -24,8 +24,6 @@ from featuretools.variable_types import (
 
 
 class FeatureBase(object):
-    _name = None
-
     def __init__(self, entity, base_features, primitive):
         """Base class for all features
 
@@ -46,6 +44,8 @@ class FeatureBase(object):
         if not isinstance(primitive, PrimitiveBase):
             primitive = primitive()
         self.primitive = primitive
+
+        self._name = None
 
         assert self._check_input_types(), ("Provided inputs don't match input "
                                            "type requirements")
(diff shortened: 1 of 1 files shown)


kmax12:
I can't think of any reason the name would have to be invalidated since there are no public methods on features that allow you to change the underlying pieces that are used to generate a name. @rwedge can you think of anything?

rwedge:
> I can't think of any reason the name would have to be invalidated since there are no public methods on features that allow you to change the underlying pieces that are used to generate a name. @rwedge can you think of anything?

I can't think of anything

featuretools/feature_base/feature_base.py
```diff
@@ -64,9 +64,9 @@ def copy(self):
         raise NotImplementedError("Must define copy on FeatureBase subclass")
 
     def get_name(self):
-        if self._name is not None:
-            return self._name
-        return self.generate_name()
+        if not self._name:
```
CJStadler: @rwedge I realized that `_name` is defined as a class variable, but when we generate the name we then create an instance variable shadowing the class variable. This seems more confusing to me than putting `self._name = None` in the initializer, or is it a python idiom that I'm not familiar with?

kmax12: you're right. we should move to `self._name = None` in the init

PR #538, opened 2019-05-09
Improve duplicate features check in DFS

CJStadler:
The previous version becomes very slow as the number of features
increases. Because it looked up the count for every element its run time
was quadratic. The implementation in this commit should be linear.

Is this check even necessary though?

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -729,12 +729,17 @@ class EntitySet(object):
             make_time_index = True
 
         if isinstance(make_time_index, str):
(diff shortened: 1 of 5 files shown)


rwedge:
Logic seems good.

> Is this check even necessary though?

@kmax12 thoughts? Does the hash check in `DeepFeatureSynthesis._handle_feature` make this check redundant?

kmax12:
@rwedge i believe you're right. looks like` _handle_new_feature` won't add a duplicate feature and all features get added by calling that, so this check is unnecessary.

Issue #543, opened 2019-05-15
Support entitysets with multiple paths between entities


CJStadler:
Currently we cannot support entitysets where there are multiple paths between any two entities. The root cause is that aggregation and direct features do not specify a path to the entity of their base features, assuming that there can only be one. During feature matrix calculation when multiple paths are detected we therefore raise an error ("Diamond graph detected!").

By instead representing features in terms of relationships we can build multiple
features with different paths to the same entity.

## Example structures and options for naming features

The features discussed are aggregations, but the same logic should apply to
`DirectFeature`s, just in the opposite direction.

### 1. Two to One

Two relationships point to the same entity.

Example:
```
Entities:
  games
  teams
Relationships:
  games.home_team_id -> teams.id
  games.away_team_id -> teams.id
```

Current features generated by `dfs` (for `teams` target entity):
- `MEAN(games.home_score)`
- `MEAN(games.away_score)`

We should be able to represent
- `MEAN(home_games.home_score)`
- `MEAN(home_games.away_score)`
- `MEAN(away_games.home_score)`
- `MEAN(away_games.away_score)`

Naming:
1. Fully qualified: `MEAN(games[home_team_id].home_score)`
2. With user provided relationship names: `MEAN(home_games.home_score)`

### 2. Diamond

There are two paths (through different entities) to the same entity.

Example:
```
Entities:
  regions
  customers
  stores
  transactions
Relationships:
  transactions.store_id -> stores.id
  transactions.customer_id -> customers.id
  stores.region_id -> region.id
  customers.region_id -> region.id
```

Currently `dfs` generates `MEAN(transactions.amount)` (for `regions`).

We should be able to represent
- `MEAN(stores.transactions.amount)`
- `MEAN(customers.transactions.amount)`

Naming
1. Fully qualified: `MEAN(stores[region_id].transactions[store_id].amount)`
2. Because no relationship is ambiguous we can infer their names from the entity
   names: `MEAN(stores.transactions.amount)`

### 3. Triangle

Target entity has a direct path to base entity, and another through a third
entity.

Example:
```
Entities:
  users
  reports
  queries
Relationships:
  queries.creator_id -> users.id
  queries.report_id -> reports.id
  reports.creator_id -> users.id
```

Currently `dfs` generates `MEAN(queries.time)` (for `users`).

We should be able to represent
- `MEAN(queries.time)` (where queries is the relationship directly from queries
  to users).
- `MEAN(reports.queries.time)`

Naming
[...]

PR #544, opened 2019-05-16
Add relationship_path to agg and direct features

CJStadler:
This adds an optional constructor parameter, and stores the `relationship_path` on the feature. If no `relationship_path` is passed to the constructor we search for a path between the target entity and base entity. If none or more than one is found we raise an error.  When there are more than one possible paths between entities the feature name will include the full path.

Also add `parent_name` and `child_name` to `Relationship`.

This does not update any other logic to use the paths, as this will be added in later commits.

See #543 for context.

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -428,52 +428,64 @@ class EntitySet(object):
         e = "No path from {} to {}. Check that all entities are connected by relationships".format(start_entity_id, goal_entity_id)
         raise ValueError(e)
 
-    def find_forward_path(self, start_entity_id, goal_entity_id):
-        """Find a forward path between a start and goal entity
+    def find_forward_paths(self, start_entity_id, goal_entity_id):
+        """
+        Generator which yields all forward paths between a start and goal
+        entity. Does not include paths which contain cycles.
 
         Args:
             start_entity_id (str) : id of entity to start the search from
             goal_entity_id  (str) : if of entity to find forward path to
 
-        Returns:
-            List of relationships that go from start entity to goal
-                entity. None is return if no path exists
-
         See Also:
-            :func:`BaseEntitySet.find_backward_path`
+            :func:`BaseEntitySet.find_backward_paths`
             :func:`BaseEntitySet.find_path`
         """
+        for sub_entity_id, path in self._forward_entity_paths(start_entity_id):
+            if sub_entity_id == goal_entity_id:
+                yield path
 
-        if start_entity_id == goal_entity_id:
-            return []
-
-        for r in self.get_forward_relationships(start_entity_id):
-            new_path = self.find_forward_path(
-                r.parent_entity.id, goal_entity_id)
-            if new_path is not None:
-                return [r] + new_path
-
-        return None
-
-    def find_backward_path(self, start_entity_id, goal_entity_id):
-        """Find a backward path between a start and goal entity
+    def find_backward_paths(self, start_entity_id, goal_entity_id):
+        """
+        Generator which yields all backward paths between a start and goal
+        entity. Does not include paths which contain cycles.
 
         Args:
             start_entity_id (str) : Id of entity to start the search from.
             goal_entity_id  (str) : Id of entity to find backward path to.
 
         See Also:
-            :func:`BaseEntitySet.find_forward_path`
+            :func:`BaseEntitySet.find_forward_paths`
             :func:`BaseEntitySet.find_path`
-
-        Returns:
-            List of relationship that go from start entity to goal entity. None
-            is returned if no path exists.
         """
-        forward_path = self.find_forward_path(goal_entity_id, start_entity_id)
-        if forward_path is not None:
-            return forward_path[::-1]
-        return None
+        for path in self.find_forward_paths(goal_entity_id, start_entity_id):
+            # Reverse path
+            yield path[::-1]
+
+    def _forward_entity_paths(self, start_entity_id, seen_entities=None):
+        """
+        Generator which yields the ids of all entities connected through forward
+        relationships, and the path taken to each. An entity will be yielded
+        multiple times if there are multiple paths to it.
+
+        Implemented using depth first search.
+        """
+        if seen_entities is None:
+            seen_entities = set()
+
+        if start_entity_id in seen_entities:
+            return
+
+        seen_entities.add(start_entity_id)
+
+        yield start_entity_id, []
+
+        for relationship in self.get_forward_relationships(start_entity_id):
+            next = relationship.parent_entity.id
+            # Copy seen entities for each next node to allow multiple paths (but
+            # not cycles).
+            for sub_entity_id, sub_path in self._forward_entity_paths(next, seen_entities.copy()):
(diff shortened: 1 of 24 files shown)


featuretools/entityset/relationship.py
```diff
@@ -65,3 +73,35 @@ def parent_variable(self):
     def child_variable(self):
         """Instance of variable in child entity"""
         return self.child_entity[self._child_variable_id]
+
+    def parent_name(self):
```
kmax12: for discussion: perhaps this should be a `@property`, same with `child_name` to be consistent with `child_variable` and `parent_variable`

featuretools/entityset/relationship.py
```diff
@@ -65,3 +73,35 @@ def parent_variable(self):
...
+            return '%s[%s]' % (self._parent_entity_id, self._child_variable_id)
+
+    def child_name(self):
+        """The name of the child, relative to the parent."""
+        if self._is_unique():
+            return self._child_entity_id
+        else:
+            return '%s[%s]' % (self._child_entity_id, self._child_variable_id)
+
+    def get_arguments(self):
```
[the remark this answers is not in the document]
CJStadler: Ah, I wasn't aware of that. Yes we definitely only need one. I think there are three considerations:

1. Which dict/JSON structure to use? `get_arguments` is more explicit I think, but if we replace `relationship_to_description` with this it changes the entityset serialization format. Is compatibility a concern there?
2. Should the method live in `entityset/serialize.py`, or on `Relationship`? Since it's now being used in multiple places I would lean towards `Relationship`.
3. Which name to use? I chose `get_arguments` to match what we did for features. But I don't think it's that clear without context – "arguments" for what?

featuretools/entityset/relationship.py
```diff
@@ -26,6 +26,14 @@ def __init__(self, parent_variable, child_variable):
                 parent_variable.id != parent_variable.entity.index):
             raise AttributeError("Parent variable '%s' is not the index of entity %s" % (parent_variable, parent_variable.entity))
 
+    @classmethod
+    def from_dictionary(cls, arguments, es):
```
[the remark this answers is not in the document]
CJStadler: See above.

featuretools/entityset/relationship.py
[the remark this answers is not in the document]
kmax12: 1. let's use what you have currently for `get_arguments` and update the entityset serialization format. we don't have a schema version in the entityset serialization format yet, so no need to worry about that. can you put up an issue though that we should add a schema version to entityset like we do feature serialization? 

2. Let's move it to `Relationship`. If we do that, should we move the deserialization method to relationship too?

3. I like the `to_dictionary` / `from_dictionary` naming convention that we use elsewhere

featuretools/feature_base/feature_base.py
```diff
@@ -421,8 +457,9 @@ class AggregationFeature(FeatureBase):
     # each time point during calculation
     use_previous = None
 
-    def __init__(self, base_features, parent_entity, primitive, use_previous=None,
-                 where=None):
+    def __init__(self, base_features, primitive,
```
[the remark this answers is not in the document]
CJStadler: I made it optional because requiring it would force callers who only have a `relationship_path` to do `relationship_path[0].parent_entity`. In the constructor we then check that `parent_entity == relationship_path[0].parent_entity`, which in this case would always be True so its not adding any value.

But if we are more concerned with end users who are calling this directly then it makes sense to require it for extra safety.

featuretools/entityset/relationship.py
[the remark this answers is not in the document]
CJStadler: That all sounds good.

featuretools/entityset/relationship.py
CJStadler: That makes sense to me. I'm not that familiar with `@property`, but it seems to fit here.

kmax12: ya, let's do it

featuretools/feature_base/feature_base.py
```diff
@@ -626,3 +697,96 @@ def _check_feature(feature):
         return feature
 
     raise Exception("Not a feature")
+
+
+def _find_path(start_entity_id, end_entity_id, es, backward=False):
```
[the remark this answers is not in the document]
CJStadler: Since here we are enforcing that the path is unique the algorithm is different and more complicated than that in `EntitySet`. There should still be some logic that can be shared, but I thought it would be easier to look at this in isolation first. Then, once we are confident in the implementation here I can merge it into `EntitySet`.

featuretools/feature_base/feature_base.py
```diff
@@ -366,16 +371,45 @@ class DirectFeature(FeatureBase):
     input_types = [Variable]
     return_type = None
 
-    def __init__(self, base_feature, child_entity):
+    def __init__(self, base_feature, child_entity=None, relationship_path=None):
```
kmax12: for discussion: 

I think this extends the API for direct features. Previously the feature had to come from a parent, but with a relationship path it can no come from a grandparent, great grandparent, etc. 

I don't see anything wrong with this good change, but it does have implications on feature calculation in pandas backend

featuretools/entityset/entityset.py
```diff
@@ -410,52 +410,64 @@ def find_path(self, start_entity_id, goal_entity_id,
...
+
+        if start_entity_id in seen_entities:
+            return
+
+        seen_entities.add(start_entity_id)
+
+        yield start_entity_id, []
+
+        for relationship in self.get_forward_relationships(start_entity_id):
+            next = relationship.parent_entity.id
```
kmax12: stylistic change, let's avoid overwriting the builtin `next` function. perhaps name it `next_relationship`?

featuretools/feature_base/feature_base.py
CJStadler: Would it be better to keep the current behavior for now (only allow paths of length 1), and leave this enhancement for a future PR? Just to avoid the scope of the current project too much. On the other hand since we are refactoring so much it might be less overall work to build in support now.

kmax12: what if we leave support for creating such features now, but pandas backend need to get support for calculating them. that doesn't really change the status quo

featuretools/feature_base/feature_base.py
```diff
@@ -371,45 +371,47 @@ class DirectFeature(FeatureBase):
...
 
             assert self.parent_entity == relationship_path[-1].parent_entity, \
                 'Base feature must be defined on the entity at the end of relationship_path'
 
-            self._is_unique_path = _is_unique_forward_path(child_entity.id,
-                                                           self.parent_entity.id,
-                                                           child_entity.entityset)
-        if not relationship_path:
-            assert child_entity, 'child_entity or relationship_path must be provided'
+            path_is_unique = child_entity.entityset \
```
kmax12: now that i look at this, it seem like `has_unique_forward_path` should be defined on Entity. 

So this should would be `child_entity.has_unique_forward_path(self.parent_entity.id)`

kmax12: you're right. let's leave as is

PR #546, opened 2019-05-17
Remove unecessary time_last variable

CJStadler:
We just grouped on this column, so it should be equal to the group
identifier. This removes the variable which was called `time_last`, but
renames the group id to now be `time_last`.


kmax12:
@CJStadler let's wait to merge until after the next release

PR #547, opened 2019-05-17
Refactor get_pandas_data_slice to take single entity

CJStadler:
Instead of getting the data for all entities at once. This hypothetically could reduce memory usage because we only need to have the data for one entity at a time in memory. It also makes the code
somewhat clearer (no more nested dictionary).


featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -444,33 +444,31 @@ def approximate_features(features, cutoff_time, window, entityset, backend,
     for approx_entity_id, approx_features in to_approximate.items():
         # Gather associated instance_ids from the approximate entity
         cutoffs_with_approx_e_ids = approx_cutoffs.copy()
-        frames = entityset.get_pandas_data_slice([approx_entity_id, target_entity.id],
-                                                 target_entity.id,
-                                                 cutoffs_with_approx_e_ids[target_instance_colname])
-
-        if frames is not None:
```
CJStadler: I'm not sure what this is trying to check, and the `else` branch was untested, so I just got rid of it.

featuretools/computational_backends/pandas_backend.py
```diff
@@ -98,50 +98,11 @@ def calculate_all_features(self, instance_ids, time_last,
...
-                                                     index_eid=self.target_eid,
-                                                     instances=None,
-                                                     entity_columns=large_necessary_columns,
-                                                     time_last=time_last,
-                                                     training_window=training_window,
-                                                     verbose=verbose)
-
-        # Handle an empty time slice by returning a dataframe with defaults
-        if eframes_by_filter is None:
-            return self.generate_default_df(instance_ids=instance_ids)
```
CJStadler: We can't really check this condition anymore and I wasn't sure if this should be moved somewhere else so I just got rid of it. The tests still pass, so maybe it's not necessary?

featuretools/computational_backends/calculate_feature_matrix.py
rwedge: I added a test to master that hits the `else` branch

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -711,24 +711,22 @@ def test_related_instances_all_cutoff_time_same_entity(es):
 
 
 def test_get_pandas_slice(es):
-    filter_eids = ['products', u'régions', 'customers']
-    result = es.get_pandas_data_slice(filter_entity_ids=filter_eids,
-                                      index_eid='customers',
-                                      instances=[0])
+    def slice_for(filter_eid):
```
rwedge: Since this function is used in multiple tests we could define it once and use it across multiple tests

featuretools/entityset/entityset.py
```diff
@@ -268,84 +267,67 @@ def add_relationship(self, relationship):
         self.reset_data_description()
         return self
 
-    def get_pandas_data_slice(self, filter_entity_ids, index_eid,
+    def get_pandas_data_slice(self, filter_eid, index_eid,
                               instances, entity_columns=None,
                               time_last=None, training_window=None,
                               verbose=False):
```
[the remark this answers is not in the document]
CJStadler: I think we can get rid of `verbose` in `calculate_all_features` too? It is used in the function, but the only caller does not pass it so it will always be `False`.

featuretools/computational_backends/pandas_backend.py
```diff
@@ -84,9 +84,6 @@ def calculate_all_features(self, instance_ids, time_last,
 
         target_entity = self.entityset[self.target_eid]
 
-        if not target_entity.df.index.isin(instance_ids).any():
-            return self.generate_default_df(instance_ids=instance_ids)
```
CJStadler: Should we leave this as an optimization, since we can catch this case early?

rwedge: I think it depends on the time necessary to perform this check. How does the run time change as the index grows / we add more missing instances?  This code will be run for each unique time in `cutoff_time`.  Quickly catching if every instance is not present in the data might not be worth optimizing for.

CJStadler: Good point, I didn't think about the cost of the check itself. Since it should be a rare case it seems unlikely to be beneficial, and adds complexity.

PR #550, opened 2019-05-20
Specifies error when 'make_time_index' is an invalid string (#537)

### LINKED ISSUE #537: Improve error message when make_time_index is invalid
For `normalize_entity` function, the argument `make_time_index` is could be a string or a boolean.

There are two cases where `make_time_index` might not be valid (if its a string):
* Specifying an existing column for `make_time_index` but not adding that column to `additional_variables` or `copy_variables`.
* Setting `make_time_index` to a column that doesn't exist in the base frame.

See https://github.com/Featuretools/featuretools/pull/535#discussion_r282472988

ctduffy:
Fixes #537 
If the user's input into the "make_time_index" is invalid, gives more specific error as to why.


featuretools/entityset/entityset.py
```diff
@@ -715,6 +715,13 @@ def normalize_entity(self, base_entity_id, new_entity_id, index,
             if v == index:
                 raise ValueError("Not copying {} as both index and variable".format(v))
                 break
```
gsheni: We should put at the top of this file the following, so we correctly check that the type is a string in python 2 and 3.
```python
if is_python_2():
    string_types = basestring
else:
    string_types = str
```
Then below, we can check the following
```python
if isinstance(make_time_index, string_types):
```
This will help us deal with unicode issues with python 2 and python 3

featuretools/entityset/entityset.py
```diff
@@ -715,6 +715,13 @@ def normalize_entity(self, base_entity_id, new_entity_id, index,
...
+
+        if isinstance(make_time_index, str):
+            if make_time_index not in base_entity.df.columns:
+                raise ValueError("'make_time_index' must be a variable in the base frame")
+            elif make_time_index not in additional_variables + copy_variables + [index]:
+                raise ValueError("'make_time_index' must be a variable specified in 'additional_variables' or 'copy_variables'")
```
gsheni: Should we also say or `'additional_variables' or 'copy_variables' or 'index'` in the error message? Since we are checking in that field as well.

featuretools/entityset/entityset.py
ctduffy: checking the `string_types = basestring` comes up with an error when I make lint. Should I put some kind of lint ignore statement there? Or define basestring at some point earlier?

gsheni: Actually, we can do the following
```python
from featuretools.utils import is_string

if is_string(make_time_index):
```

PR #551, opened 2019-05-20
add extra requirements

gsheni:
- Add extra requirements for featuretools plugins

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,16 +18,22 @@ commands:
           source test_env/bin/activate
           cd "featuretools-$(python setup.py --version)/"
           pytest
-  install-ft-complete-and-run-import-tests:
-    description: "run unit tests on packaged testing files"
+  install-ft-complete:
+    description: "install featuretools[complete]"
     steps:
       - run: |
           virtualenv test_ft_complete_env
           source test_ft_complete_env/bin/activate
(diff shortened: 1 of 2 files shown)


rwedge:
Should we test installing these extras as part of our CI tests?

kmax12:
> Should we test installing these extras as part of our CI tests?

might as well

gsheni:
@rwedge PR now tests importing the extra packages.

.circleci/config.yml
```diff
@@ -18,6 +18,18 @@ commands:
...
+    description: "run unit tests on packaged testing files"
+    steps:
+      - run: |
+          virtualenv test_ft_complete_env
+          source test_ft_complete_env/bin/activate
+          python setup.py sdist
+          FEATURETOOLS_VERSION=$(python setup.py --version)
+          tar -zxvf "dist/featuretools-${FEATURETOOLS_VERSION}.tar.gz"
+          pip install -e "featuretools-${FEATURETOOLS_VERSION}/[complete]"
+          python -c "import featuretools_tsfresh_primitives"
```
rwedge: Let's make a second `-run` step in the command that tests the extra imports.  That way the imports don't share the same output log with the pip installation.

rwedge: Can we also test importing `featuretools.tsfresh`

gsheni: Fixed.

PR #559, opened 2019-05-22
test calculating features with instances that aren't in data

rwedge:
This PR adds some tests for the output of `calculate_feature_matrix` when some of the instances to calculate for aren't in the EntitySet's data


featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -1170,3 +1170,66 @@ def test_no_data_for_cutoff_time():
     # due to default values for each primitive
     # count will be 0, but max will nan
     np.testing.assert_array_equal(fm.values, [[0, np.nan]])
+
+
+def test_instances_not_in_data(es):
+    instances = list(range(20, 30))
```
CJStadler: To ensure that these are not in the data (and make it explicit to the reader) you could do something like take the max of the actual ids and then pick 10 higher numbers.

PR #569, opened 2019-05-29
Fixed testing issue with residual directories causing FileExistsError for makedirs()

### LINKED ISSUE #565: entityset_tests/test_serialization.py--makedir() throwing FileExistsError

Tests in tests/entityset_tests/test_serialization.py (starting @ line 82) throwing FileExistsError from makedirs(path) --> some test(s) leaving existing directories. 

-----

allisonportis:
Fix #565
Resolved by adding fixture that creates the path and makes the directory, yields the path, and removes the path folder in the finalization code.

### DIFF
--- a/featuretools/tests/entityset_tests/test_serialization.py
+++ b/featuretools/tests/entityset_tests/test_serialization.py
@@ -1,5 +1,5 @@
-import os
 import errno
+import os
 import shutil
 
 import pandas as pd
@@ -8,7 +8,6 @@ import pytest
 from featuretools.demo import load_mock_customer
 from featuretools.entityset import EntitySet, deserialize, serialize
 from featuretools.tests import integration_data
-
 
 CACHE = os.path.join(os.path.dirname(integration_data.__file__), '.cache')
 
--- a/featuretools/tests/entityset_tests/test_serialization.py
+++ b/featuretools/tests/entityset_tests/test_serialization.py
@@ -59,12 +59,19 @@ def test_entityset_description(es):
     assert es.metadata.__eq__(_es, deep=True)
 
 
-def test_invalid_formats(es):
+@pytest.fixture
+def path_management():
     path = os.path.join(CACHE, 'es')
+    os.makedirs(path)
+    yield path
+    shutil.rmtree(path)
+
+
+def test_invalid_formats(es, path_management):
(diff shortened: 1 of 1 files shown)


featuretools/tests/entityset_tests/test_serialization.py
```diff
@@ -59,12 +59,19 @@ def test_entityset_description(es):
     assert es.metadata.__eq__(_es, deep=True)
 
 
-def test_invalid_formats(es):
+@pytest.fixture
+def path_management():
     path = os.path.join(CACHE, 'es')
+    os.makedirs(path)
```
rwedge: we have logic here to handle if the path folder already exists

featuretools/tests/entityset_tests/test_serialization.py
```diff
@@ -59,12 +59,22 @@ def test_entityset_description(es):
     assert es.metadata.__eq__(_es, deep=True)
 
 
-def test_invalid_formats(es):
+@pytest.fixture
+def path_management():
     path = os.path.join(CACHE, 'es')
+    try:
+        os.makedirs(path)
+    except (OSError, IOError):
```
rwedge: We want to ignore the error raised by `os.makedirs` if the folder already exists but let any unexpected errors get raised.  `os.makedirs` raises an OSError in both python 2 and 3 so we should only catch OSError

allisonportis: I changed it to catch `OSError` and then check the error code of that error to be `errno.EEXIST` but I can't figure out how to order the imports correctly for `import errno`. I checked the pep8 guidelines and it's with the standard library imports but it still fails the make lint. I tried it in a bunch of different specific places as well.

featuretools/tests/entityset_tests/test_serialization.py
```diff
@@ -52,12 +53,23 @@ def test_entityset_description(es):
...
 
 
-def test_invalid_formats(es):
+@pytest.fixture
+def path_management():
     path = os.path.join(CACHE, 'es')
+    try:
+        os.makedirs(path)
+    except OSError as e:
+        if e.errno != errno.EEXIST:
```
rwedge: could you add a comment describing what `EEXIST` refers to?

Issue #560, opened 2019-05-22
Custom variable types (like custom primitives)


bschreck:
### Feature Request

Currently there's no functionality to add a new variable type like there is a new primitive. I thought I'd be able to just subclass `Categorical` (with something like `CatList` - to deal with list-type columns e.g. "item1,item2,item3"), and create a `TransformPrimitive` that requires the subclass as input variable type. 

However, an identity feature's `variable_type` depends on the `entityse`t's `metadata`, which relies on serialization/deserialization, and in the process only recognizes variables in the FT source `variable_types` submodule via `dir(variable_types)`. This is in `entityset/serialize.py` at the top.

-----
*Issues created here on Github are for bugs or feature requests. For usage questions and questions about errors, please ask on Stack Overflow with the [featuretools](https://stackoverflow.com/questions/tagged/featuretools) tag. Check the [documentation](https://docs.featuretools.com/#get-help) for further guidance on where to ask your question.*

kmax12:
Thanks @bschreck. We will look into fixing this for the next release

PR #571, opened 2019-05-30
Custom Variable Types

allisonportis:
Issue #560 

Added to entityset/deserialize.py so that description_to_variable() looks at all variables that are (recursively) a subclass of Variable, so that it will recognize custom variable types.


featuretools/entityset/serialize.py
```diff
@@ -2,13 +2,23 @@
...
+
+def find_subclasses(cls):
+    yield cls
+    for sub in cls.__subclasses__():
+        for c in find_subclasses(sub):
+            yield c
+
+
+def find_variable_types():
+    return {str(type.type_string): type for type in find_subclasses(Variable) if hasattr(type, 'type_string')}
```
rwedge: let's use `vtype` as the variable here

featuretools/entityset/serialize.py
```diff
@@ -2,13 +2,23 @@
 import os
 import shutil
 
-from featuretools import variable_types
+# from featuretools import variable_types
+from featuretools.variable_types import Variable
+
+
+def find_subclasses(cls):
```
rwedge: This is the same code as `_descendants` in `featuretools/primitives/utils.py`.  Let's add this as a function in `featuretools/gen_utils.py` and have both files import it from there

featuretools/entityset/deserialize.py
```diff
@@ -17,9 +17,10 @@ def description_to_variable(description, entity=None):
     Returns:
         variable (Variable) : Returns :class:`.Variable`.
     '''
+    variable_types = find_variable_types()
     is_type_string = isinstance(description['type'], str)
     type = description['type'] if is_type_string else description['type'].pop('value')
-    variable = VARIABLE_TYPES.get(type, VARIABLE_TYPES.get('None'))
+    variable = variable_types.get(type, variable_types.get('None'))
```
rwedge: can you add a comment here that "None" will return the Unknown variable type

featuretools/entityset/serialize.py
```diff
@@ -2,13 +2,23 @@
 import os
 import shutil
 
-from featuretools import variable_types
+# from featuretools import variable_types
```
rwedge: commented out code

featuretools/entityset/serialize.py
```diff
@@ -2,13 +2,16 @@
 import os
 import shutil
 
-from featuretools import variable_types
+from featuretools.utils.gen_utils import find_descendents
+from featuretools.variable_types import Variable
+
+
+def find_variable_types():
```
rwedge: With this new function some variables are becoming redundant.  In this file we have the `VARIABLE_TYPES` variable, which it not referenced in the file, only used in `test_serialization.py`.  In `featuretools/variable_types/variable.py` there's an `ALL_VARIABLE_TYPES` variable which is hard-coded and isn't used at all.

I think we should remove these two variables.  The test cases that depended on `VARIABLE_TYPES` can use this function instead.  I also think it would make more sense for this function to be in `variable.py` instead of `serialize.py`

allisonportis: agreed, I'll change all of that

allisonportis:
Hi @bschreck, we implemented a solution for custom variable types. Would you mind testing out this branch and see if it's working for you? Thanks!

bschreck:
@allisonportis This works great, thanks for the change!

PR #574, opened 2019-05-31
Remove empty directory tests/synthesis

CJStadler:
Alternatively, we could rename `tests/dfs_tests` to `tests/synthesis` to match `featuretools/synthesis`.


rwedge:
I like the `tests/synthesis` idea.  Can you move `test_encode_features.py` into that folder as well, since `encode_features` is also part of the synthesis module?

PR #575, opened 2019-06-03
Stop isort from changing __init__.py files

gsheni:
- Stop isort from changing __init__.py files to prevent circular imports


gsheni:
Might not be needed, so closing

PR #576, opened 2019-06-04
renamed features have name correctly serialized

### LINKED ISSUE #568: Renamed features do not have their name serialized
```py
import featuretools as ft
from featuretools.feature_base.features_serializer import FeaturesSerializer
from featuretools.feature_base.features_deserializer import FeaturesDeserializer

original = ft.IdentityFeature(es['log']['value'])
print(original)

renamed = original.rename('MyFeature')
print(renamed)

serializer = FeaturesSerializer([renamed])
serialized = serializer.to_dict()

deserializer = FeaturesDeserializer(serialized)
deserialized = deserializer.to_list()[0]
print(deserialized)
```
Outputs
```
<Feature: value>
<Feature: MyFeature>
<Feature: value>
```
Instead of
```
<Feature: value>
<Feature: MyFeature>
<Feature: MyFeature>
```

alexjwang:
Fixes #568


featuretools/tests/primitive_tests/test_identity_features.py
```diff
@@ -11,6 +11,7 @@ def test_serialization(es):
     value = ft.IdentityFeature(es['log']['value'])
 
     dictionary = {
+        'name': None,
```
gsheni: So I suppose that with this new PR, old features generated using the previous feature serializer will now break because they don't contain the name field, right?
Would it be possible to make this backwards compatible? Could we do a `argments.get('name', None)` at this line
https://github.com/Featuretools/featuretools/pull/576/files#diff-a6f70d99f6dca96d42267829c30c764cR346

gsheni (commented):
other than the 1 comment, looks good.

featuretools/tests/primitive_tests/test_identity_features.py
CJStadler: There is now `SCHEMA_VERSION` (in `features_serializer.py`) which should be updated when the serialization format changes.

featuretools/feature_base/feature_base.py
```diff
@@ -342,7 +342,9 @@ def from_dictionary(cls, arguments, entityset, dependencies, primitives_deserial
         entity_id = arguments['entity_id']
         variable_id = arguments['variable_id']
         variable = entityset[entity_id][variable_id]
-        return cls(variable)
+        cls_variable = cls(variable)
+        cls_variable._name = arguments['name']
```
CJStadler: Since `_name` is intended to be private, I wonder if there is a better way to accomplish this. We could use `rename`, since it is the public API, but that would unnecessarily create a copy of the feature. Would it be worth adding an optional `name` keyword argument to the constructor?

CJStadler (commented):
Sorry that this wasn't clear in the issue, but I think this needs to be fixed for all types of features, not just `IdentityFeature` (since any type of feature can be renamed).

featuretools/feature_base/feature_base.py
alexjwang: I added an optional `name` keyword argument to the constructor.

PR #580, opened 2019-06-04
renamed features have names correctly serialized

### LINKED ISSUE #568: Renamed features do not have their name serialized

alexjwang:
fixed #568


featuretools/feature_base/feature_base.py
```diff
@@ -328,21 +328,23 @@ def unique_name(self):
 class IdentityFeature(FeatureBase):
     """Feature for entity that is equivalent to underlying variable"""
 
-    def __init__(self, variable):
+    def __init__(self, variable, name=None):
```
gsheni: If we are going to add name=None to every single child class of `FeatureBase`, then doesn't make sense to add it directly to the `FeatureBase` class?

featuretools/feature_base/feature_base.py
```diff
@@ -328,21 +328,23 @@ def unique_name(self):
...
         entity_id = variable.entity_id
         self.variable = variable.entityset.metadata[entity_id][variable.id]
         self.return_type = type(variable)
         super(IdentityFeature, self).__init__(entity=variable.entity,
                                               base_features=[],
                                               relationship_path=[],
                                               primitive=PrimitiveBase)
+        if name is not None:
```
rwedge: Since each feature subclass has this same logic in their respective init we could pass `name` to the FeatureBase `__init__` call and handle this there

featuretools/tests/primitive_tests/test_feature_serialization.py
```diff
@@ -51,3 +55,24 @@ def test_pickle_features_with_custom_primitive(es):
 
     assert any([isinstance(feat.primitive, NewMax) for feat in features_original])
     pickle_features_test_helper(asizeof(es), features_original)
+
+
+def test_serialized_renamed_features(es):
```
gsheni: We should put all the types of features in `Identity feature, Transform feature, Aggregation feature, and Direct feature` in 1 feature list, and then rename each feature, and see if they were correctly serialized, and deserialized.  Let's not separate them into `test_agg_feats.py`, `test_direct_features`, `test_identify_features.py`, and `test_transform_features.py`

featuretools/feature_base/feature_base.py
```diff
@@ -328,21 +328,23 @@ def unique_name(self):
...
+            self._name = name
 
     @classmethod
     def from_dictionary(cls, arguments, entityset, dependencies, primitives_deserializer):
         entity_id = arguments['entity_id']
         variable_id = arguments['variable_id']
         variable = entityset[entity_id][variable_id]
-        return cls(variable)
+        return cls(variable, arguments['name'])
```
gsheni: Let's not rely on a positional call.
Lets do 
`return cls(variable=variable, name=arguments['name'])` so we specify what the arguments are.

featuretools/feature_base/feature_base.py
```diff
@@ -602,12 +612,14 @@ def __init__(self, base_features, primitive):
...
                                                primitive=primitive)
+        if name is not None:
+            self._name = name
 
     @classmethod
     def from_dictionary(cls, arguments, entityset, dependencies, primitives_deserializer):
         base_features = [dependencies[name] for name in arguments['base_features']]
         primitive = primitives_deserializer.deserialize_primitive(arguments['primitive'])
-        return cls(base_features, primitive)
+        return cls(base_features, primitive, arguments['name'])
```
gsheni: `return cls(base_features=base_features, primitive=primitive, name=arguments['name'])` so we specify what the arguments are.

featuretools/feature_base/feature_base.py
```diff
@@ -637,12 +650,15 @@ def __init__(self, base_features, primitive, groupby):
...
+        if name is not None:
+            self._name = name
+
     @classmethod
     def from_dictionary(cls, arguments, entityset, dependencies, primitives_deserializer):
         base_features = [dependencies[name] for name in arguments['base_features']]
         primitive = primitives_deserializer.deserialize_primitive(arguments['primitive'])
         groupby = dependencies[arguments['groupby']]
-        return cls(base_features, primitive, groupby)
+        return cls(base_features, primitive, groupby, arguments['name'])
```
gsheni: Let's be explicit in this call as well.

PR #581, opened 2019-06-04
Schema version added for entityset serialization

### LINKED ISSUE #556: Add a schema version for entityset serialization
This would allow us to check (during deserialization) that the format of the serialized entityset is compatible with the running version of featuretools.

This is already done for feature serialization: #532

alexjwang:
fixes #556

### DIFF
--- a/featuretools/entityset/deserialize.py
+++ b/featuretools/entityset/deserialize.py
@@ -3,8 +3,16 @@ import os
 
 import pandas as pd
 
+from .serialize import SCHEMA_VERSION
+
 from featuretools.entityset.relationship import Relationship
 from featuretools.entityset.serialize import FORMATS, VARIABLE_TYPES
+from featuretools.utils.gen_utils import is_python_2
+
+if is_python_2():
+    from itertools import izip_longest as zip_longest
+else:
+    from itertools import zip_longest
 
 
 def description_to_variable(description, entity=None):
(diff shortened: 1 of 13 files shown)


featuretools/entityset/deserialize.py
```diff
@@ -59,6 +67,19 @@ def description_to_entityset(description, **kwargs):
     Returns:
         entityset (EntitySet) : Instance of :class:`.EntitySet`.
     '''
+    current = SCHEMA_VERSION.split('.')
+    saved = description.get('schema_version').split('.')
+    error_text = ('Unable to load features. The schema version of the saved '
```
gsheni: This error message is wrong. It says `features`. It should say `Unable to load entityset`.
Also, if we are going to copy a method from the feature list serialization check, we should make a modular function, and call that. That way there is no copied code.

featuretools/feature_base/feature_base.py
```diff
@@ -328,21 +328,23 @@ def unique_name(self):
 class IdentityFeature(FeatureBase):
     """Feature for entity that is equivalent to underlying variable"""
 
-    def __init__(self, variable):
+    def __init__(self, variable, name=None):
```
gsheni: This is code from PR #580 
We should make sure that the PR stands by itself, and doesn't contain commits from other PRs

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -961,3 +967,37 @@ def _slice_for(es, filter_eid, time_last=None):
...
+        entities = {entity.id: serialize.entity_to_description(entity) for entity in es.entities}
+        relationships = [relationship.to_dictionary() for relationship in es.relationships]
+        dictionary = {
+            'schema_version': version,
+            'id': es.id,
+            'entities': entities,
+            'relationships': relationships,
+        }
+
+        error_text = ('Unable to load features. The schema version of the saved '
```
gsheni: Wrong error message.

PR #586, opened 2019-06-05
Schema version added for entityset serialization 

### LINKED ISSUE #556: Add a schema version for entityset serialization

alexjwang:
fixes #556


featuretools/entityset/deserialize.py
```diff
@@ -155,3 +165,18 @@ def read_entityset(path, **kwargs):
     '''
     data_description = read_data_description(path)
     return description_to_entityset(data_description, **kwargs)
+
+
+def _check_schema_version(description):
```
gsheni: So I noticed that this is similar to the function here
https://github.com/Featuretools/featuretools/blob/d4f5d5f13f25a8beda054f8b2e065d9002309828/featuretools/feature_base/features_deserializer.py#L119

If we are going to use shared functionality, we should use it put it in a `utils` file. 
https://github.com/Featuretools/featuretools/tree/master/featuretools/utils
Then both functions can call the utils file.

PR #588, opened 2019-06-06
Removed test case for allow_where

alexjwang:
follow up to #587


docs/source/changelog.rst
```diff
@@ -11,6 +11,7 @@ Changelog
     * Fixes
         * Normalize_entity specifies error when 'make_time_index' is an invalid string (:pr:`550`)
         * Removed all mentions of allow_where (:pr:`587`)
+        * Removed test case for allow_where (:pr:`588`)
```
rwedge: We can add this as a second PR number for the `Remove all mentions of allow_where` entry

PR #598, opened 2019-06-13
Store full entity dfs in separate trie

CJStadler:
@kmax12 and @rwedge this resolved the performance issues in a benchmark I wrote, but we should definitely run it against the EntitySets tests as well. I put this in a separate PR to make it easier to see the changes, but it should probably be merged into #572. This makes the logic a bit more complicated, so let me know if anything needs to be clarified.

If there is an aggregation of a `uses_full_entity` feature the aggregation
should not be calculated on the full entity. Previously because we only
had one set of dfs this was not possible.

Now, we always store the "filtered" df in `df_trie`, and if necessary also
store a full df in `large_df_trie`.

If any features at a given node require a full df we first calculate
them. Then we extract the filtered df from the result and use this to
calculate any remaining features. Because `FeatureSet.uses_full_entity` is
`True` for features which have a dependent `uses_full_entity` feature the
large df does not need to contain any of the `not_full_entity_features`.

### DIFF
--- a/featuretools/computational_backends/pandas_backend.py
+++ b/featuretools/computational_backends/pandas_backend.py
@@ -56,7 +56,7 @@ class PandasBackend(ComputationalBackend):
             training_window (Timedelta, optional): Window defining how much time before the cutoff time data
                 can be used when calculating features. If None, all data before cutoff time is used.
 
-            ignored (set[int]): Unique names of precalculated features.
+            ignored (set[str]): Unique names of precalculated features.
 
             precalculated_features (Trie[RelationshipPath -> pd.DataFrame]):
                 Maps RelationshipPaths to dataframes of precalculated_features
@@ -117,13 +117,13 @@ class _FeaturesCalculator(object):
         feature_trie = self.feature_set.feature_trie
 
         df_trie = Trie(path_constructor=RelationshipPath)
-        large_df_trie = Trie(path_constructor=RelationshipPath)
+        full_entity_df_trie = Trie(path_constructor=RelationshipPath)
 
         target_entity = self.entityset[self.target_eid]
         self._calculate_features_for_entity(entity_id=self.target_eid,
                                             feature_trie=feature_trie,
                                             df_trie=df_trie,
-                                            large_df_trie=large_df_trie,
+                                            full_entity_df_trie=full_entity_df_trie,
                                             precalculated_trie=self.precalculated_features,
                                             filter_variable=target_entity.index,
                                             filter_values=instance_ids)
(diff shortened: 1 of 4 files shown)


featuretools/computational_backends/pandas_backend.py
```diff
@@ -245,15 +260,26 @@ def _calculate_features_for_entity(self, entity_id, feature_trie, df_trie,
...
+                # Pass filtered values, even if we are using a full df.
+                if full_entity_features:
+                    filtered_df = df[df[filter_variable].isin(filter_values)]
+                else:
+                    filtered_df = df
+                sub_filter_values = filtered_df[relationship.parent_variable.id]
+
+                parent_data = (relationship,
+                               new_ancestor_relationship_variables,
+                               df)
```
kmax12: can parent data get away with just the `filtered_df`? i'd think this okay because of `sub_filter_values `is based on the `filtered_df`?

CJStadler: My thinking was that if the child node needs the full entity then it will need the full `parent_data` to get the relationship variables. But no tests fail if we pass `filtered_df` instead, so either I'm wrong or we don't have a test for such a case.

kmax12: ah, i see. even though we pass the `filtered_df` through, if could be ignored if a child feature uses full entity. let's leave as-is. ideally, write a test

kmax12: actually, the more i think about it, I can't think of what feature the child entity could define on the full entity that would rely on a ancestor relationship variable. maybe see if you can write a test and if not, just use the filtered df

CJStadler: Here is the test: https://github.com/Featuretools/featuretools/pull/598/files#diff-bf66dc93fe293a61ccc3fbd3c92eec31R76

The ancestor relationship variable is necessary when we calculate the agg. If this is changed to `filtered_df` then the test fails because the aggregation is `0` for the rows of the target other than the `instance_ids`. The rows of `log` that should be linked to those rows of the target entity have `NaN` for the relationship variable. Does that make sense?

CJStadler:
@rwedge and @kmax12 this required a more substantial change than I expected. The values stored in the feature trie are now 3-tuples of:
1. A boolean representing whether the full entity is needed for this node. Previously I was inferring this based on whether there are any features which need a full entity, but the full entity might be needed even if this is not the case. For example: if the only target feature is `PERCENTILE(MEAN(customers.transactions))` then there are no features on `customers`, but we still need the full entity so that the every row in `transactions` can get its relationship variable to the target entity.
2. A set of feature names representing the features which need to be calculated on the full entity.
3. A set of feature names which do not need to be calculated on the full entity.

Previously the features were partitioned into those two groups by checking whether any dependents of each feature are `uses_full_entity`. This is not correct because it does not take into account the path to each feature. For example, if we are calculating two features `MEAN(log.value)` and `CUM_SUM(customers.SUM(sessions.MEAN(log.value)))` (see [this test](https://github.com/Featuretools/featuretools/pull/598/files#diff-313936a1d6de52bacc492547ececd51bR80)), `MEAN(log.value)` should only be calculated on the full entity when it is on the path through `customers`. When it is calculated for the target it should only be calculated for the given `instance_ids`.

featuretools/computational_backends/pandas_backend.py
kmax12: i see. thanks

PR #600, opened 2019-06-17
Update DeepFeatureSynthesis to support multiple paths

CJStadler:
For entitysets with multiple paths between two entities DFS will now
create features through all paths. Entitysets with cycles in the
directed graph are not supported.

- Use feature.unique_name instead of hash in dicts
- Change EntitySet.get_backward_entities to yield paths.

Once #572 is merged the base branch should be changed to master.


featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -299,7 +300,7 @@ def _run_dfs(self, entity, entity_path, all_features, max_depth):
         """
 
         backward_entities = self.es.get_backward_entities(entity.id)
-        backward_entities = [b_id for b_id in backward_entities
+        backward_entities = [b_id for b_id, _ in backward_entities
```
kmax12: If an entity has two backward paths to a child (e.g teams -> games.home_team, teams -> games.away_team), would the child show up twice in this list? 

if it does, we'd end up making the same call to `_run_dfs` twice. I imagine this doesn't cause problems since it just wouldn't add same feature twice, but probably not ideal.

kmax12: Now that we have the concept of `RelationshipPath`, perhaps it makes sense to keep the relationship path here, and refactor the entity_path argument to use it rather than just a list of entities. 

the entity path is basically used to keep track of how we've travserved to a given entity, so the relationship path would be a natural thing to use instead.

featuretools/entityset/entityset.py
```diff
@@ -439,26 +439,24 @@ def get_forward_entities(self, entity_id, deep=False):
...
-                children_deep = children_deep.union(to_add)
-
-            children = children_deep
-        return set(children)
+        for relationship in self.get_backward_relationships(entity_id):
+            child_eid = relationship.child_entity.id
+            direct_path = RelationshipPath([(False, relationship)])
+            yield child_eid, direct_path
+
+            sub_entities = self.get_backward_entities(child_eid, deep=deep)
```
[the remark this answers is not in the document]
CJStadler: Yep! `deep` was not even being used... I'll update the test to catch this.

featuretools/synthesis/deep_feature_synthesis.py
CJStadler: You are right. And there is even a check for this, but it never gets run: https://github.com/Featuretools/featuretools/blob/1af738fdc77d65e177b66412cfe4a577c509d0df/featuretools/synthesis/deep_feature_synthesis.py#L409-L412

This is tangential to this PR, but should we remove the `return` to enforce this? There are several tests which would hit that exception. This is because the `variable_type` of `LAST(log.id)` is `Index`, so DFS on `customers` attempts to create two `COUNT(sessions)` features, one where the base feature is `LAST(log.id)` and the other where the base feature is `sessions.id`. Should the `variable_type` of `LAST(log.id)` be `Numeric` instead?

kmax12: Yeah, we should handle this outside of this PR. Some notes below

In fact, `LAST(log.id)` shouldn't be `Index` -- that is a bug. The `return_type` of `Last` is `None`, it should trigger the logic here

https://github.com/Featuretools/featuretools/blob/1af738fdc77d65e177b66412cfe4a577c509d0df/featuretools/feature_base/feature_base.py#L173

We already handle the timeindex case, but we should also handle the index case and just return that it is `Categorical`. 

Can you put up an issue to handle later, or PR and test case?

We can continue discuss on the new Issue/PR, but if we fix that, does any other test case raise the error?

CJStadler: I created #602 to fix that.

There is one other case where this is hit: when there is a `seed_feature` that is also generated by `dfs`.

featuretools/entityset/entityset.py
```diff
@@ -439,26 +439,25 @@ def get_forward_entities(self, entity_id, deep=False):
...
     def get_backward_entities(self, entity_id, deep=False):
-        """Get entities that are in a backward relationship with entity
+        """
+        Get entities that are in a backward relationship with entity
 
         Args:
-            entity_id (str) - Id entity of entity to search from.
-            deep (bool) - If True, recursively find backward entities.
+            entity_id (str): Id entity of entity to search from.
+            depth (int): The depth to search to. 1 means only get the direct
```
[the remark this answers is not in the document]
CJStadler: Yes, I forgot to change it back.

CJStadler:
Back to you @kmax12. I made once change in 3634849 that you might want to look at.

docs/source/changelog.rst
```diff
@@ -34,6 +34,7 @@ Changelog
         * Refactor get_pandas_data_slice to take single entity (:pr:`547`)
         * Updates TimeSincePrevious and Diff Primitives (:pr:`561`)
         * Remove unecessary time_last variable (:pr:`546`)
+        * Remove max_hlevel from DeepFeatureSynthesis (:pr:`608`)
```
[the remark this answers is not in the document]
CJStadler: Since this was merged into this PR it doesn't need to be in the changelog at all, right?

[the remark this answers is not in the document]
kmax12: We're not strict about it. This more about communicating what change rather than a one to one match to what merged in master. Perhaps add the PR as an additional PR number under changelog line for this PR.

PR #602, opened 2019-06-18
Change type of features applied to Index features

CJStadler:
If the primitive which does not specify a return type is applied to an
Index feature the resulting feature will now have type Categorical, not
Index. There should only be one Index feature per entity.

See https://github.com/Featuretools/featuretools/pull/600#discussion_r294910255


CJStadler:
@kmax12 is this ready now? (needs to be re-approved)

docs/source/changelog.rst
```diff
@@ -5,6 +5,7 @@ Changelog
 **Future Release**
     * Enhancements
     * Fixes
+      * Change type of features calculated on Index features to Categorical (:pr:`602`)
```
rwedge: This needs to be indented more to register as a sublist

PR #608, opened 2019-06-19
Remove max_hlevel from DFS

CJStadler:
This was not exposed in the `dfs` method so it always used the default
value of 2, it was undocumented, and it was unclear if it provided any
value.

Also remove `EntitySet.find_path` as it is no longer used.

The base branch should be changed to `master` after #600 is merged.

### DIFF
--- a/featuretools/synthesis/deep_feature_synthesis.py
+++ b/featuretools/synthesis/deep_feature_synthesis.py
@@ -637,13 +637,10 @@ class DeepFeatureSynthesis(object):
             if (variable_type == variable_types.PandasTypes._all or
                     f.variable_type == variable_type or
                     any(issubclass(f.variable_type, vt) for vt in variable_type)):
-                if max_depth is None or self._get_depth(f) <= max_depth:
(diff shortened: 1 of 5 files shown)


featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -648,9 +637,7 @@ def _features_by_type(self, all_features, entity, max_depth,
             if (variable_type == variable_types.PandasTypes._all or
                     f.variable_type == variable_type or
                     any(issubclass(f.variable_type, vt) for vt in variable_type)):
-                if ((max_depth is None or self._get_depth(f) <= max_depth) and
-                        (self.max_hlevel is None or
-                         self._max_hlevel(f) <= self.max_hlevel)):
+                if max_depth is None or self._get_depth(f) <= max_depth:
```
kmax12: I think we can replace `self._get_depth(f)` with `f.get_depth(stop_at=self.seed_features)` and remove `self._get_depth`

PR #609, opened 2019-06-19
Speedup groupby transform calculations

kmax12:
This PR updates the logic for how we calculate group by transform features.

Previously we iterated over the groups first and then the features, updating the resulting frame greedily as we went along.

Now we iterate over the features, then groups. For each feature, we accumate all the values across all groups and then update the frame just once. 

Basically this improves the number of update calls from `num features x num groups` to just `num features`

Also added a benchmarks folder to hold the code I used to test these changes


kmax12:
For now, what if we save notebooks with a cleared output? That will make them easier to diff. Just updated the current benchmark notebook with a cleared output

rwedge:
That sounds reasonble

PR #613, opened 2019-06-19
Remove variance_selection.py

CJStadler:
This is not referenced anywhere else.

The commit that added it (99a80d7) seems unrelated, was it added accidentally?


kmax12:
It's just legacy stuff from the early days. We can remove it now

PR #620, opened 2019-06-21
Add FAQ

gsheni:
- Added a notebook that goes through solutions for FAQs


ctduffy:
2 comments:

1. Near code snippet 32/in section "How do I prevent label leakage with DFS":
What about if the user uses train test split to split the training and testing data? This might be a better example, because there are a few things that must be done to ensure accuracy with this method (ie matching the labels after making the features). I also don't really understand fully how the training and testing data is generated/divided in this example.
Also, how are categorical variables encoded when splitting training and testing data? What happens if the categorical variable has a value in the testing data but doesn't have it in the training data (this might be an uninformed question)

2. Below code snippet 64/in section "Why am I getting...?"
Would it be possible to further clarify the difference between last time index and time index/clarify what it does with the excluded rows? Are these rows used in testing, or how are they factored in (if at all)? Again, not sure how well informed this question is.

These two questions might not be valid questions, in which case they can be ignored.

docs/source/index.rst
```diff
@@ -167,12 +167,12 @@ Table of contents
    guides/deployment
    guides/exporting_feature_matrix
    guides/advanced_custom_primitives
-   guides/frequently_asked_questions
 
 .. toctree::
    :maxdepth: 1
    :caption: Resources and References
 
+   guides/frequently_asked_questions
```
kmax12: let's put it at the root of the docs

gsheni: Moved

Issue #621, opened 2019-06-21
encode_features incorrectly modifies encoded feature_matrix for features with multiple columns


alexjwang:
https://github.com/Featuretools/featuretools/blob/f80b5ae1f55a979e24fd8afcea2c1fab280b0c70/featuretools/synthesis/encode_features.py#L95

For features with multiple columns, the following warning is given: "Feature has multiple columns and will not be encoded.  This may result in a matrix with non-numeric values."

Categorical variables are indeed not encoded; however, encode_features is still modifying the resulting encoded feature matrix. Specifically, it is creating new columns with the same feature name and values.

```py
import featuretools as ft
from featuretools.tests.testing_utils import make_ecommerce_entityset
import pandas as pd

es = make_ecommerce_entityset()
feature_matrix, feature_defs = ft.dfs(entityset=es,
                                target_entity="customers",
                                agg_primitives=['n_most_common'])

not_encoded = [x for x in feature_matrix.columns.tolist() if 'cohorts.N_MOST_COMMON(log.zipcode)' in x]
feature_matrix_enc, feature_defs_enc = ft.encode_features(feature_matrix, feature_defs)

encoded = [x for x in feature_matrix_enc.columns.tolist() if 'cohorts.N_MOST_COMMON(log.zipcode)' in x]

print(not_encoded)
print(encoded)
```
#### Output:
```
['cohorts.N_MOST_COMMON(log.zipcode)__0', 
 'cohorts.N_MOST_COMMON(log.zipcode)__1', 
 'cohorts.N_MOST_COMMON(log.zipcode)__2']
```
```
['cohorts.N_MOST_COMMON(log.zipcode)__0', 
 'cohorts.N_MOST_COMMON(log.zipcode)__1', 
 'cohorts.N_MOST_COMMON(log.zipcode)__2', 
 'cohorts.N_MOST_COMMON(log.zipcode)__0', 
 'cohorts.N_MOST_COMMON(log.zipcode)__1']
```

alexjwang:
## Current workaround
```py
import featuretools as ft
from featuretools.tests.testing_utils import make_ecommerce_entityset
import pandas as pd

es = make_ecommerce_entityset()
feature_matrix, feature_defs = ft.dfs(entityset=es,
                                target_entity="customers",
                                agg_primitives=['n_most_common'])

not_encoded = [x for x in feature_matrix.columns.tolist() if 'cohorts.N_MOST_COMMON(log.zipcode)' in x]
[...]

PR #622, opened 2019-06-21
Fix bug with encode_features and features that create multiple columns

rwedge:
### Pull Request Description
Fixes issue #621 

Features create multiple columns in the feature matrix aren't handled by `encode_features`.  This PR fixes a bug that would cause some columns of a feature with multiple outputs to be added to the column list twice, resulting in duplicate columns in the resulting matrix.

-----


featuretools/synthesis/encode_features.py
```diff
@@ -74,7 +74,7 @@ def encode_features(feature_matrix, features, top_n=10, include_unknown=True,
             assert fname in X.columns, (
                 "Feature %s not found in feature matrix" % (fname)
             )
-        feature_names.append(fname)
+            feature_names.append(fname)
```
CJStadler: 🐍 + ⬜️ 👾 = 🐛

PR #623, opened 2019-06-24
Generate transform features of direct features

### LINKED ISSUE #81: Primitive stacking on direct features
Suppose we have an entityset with a parent entity `E1` and a child entity `E2` and we're building features on `E2` with Deep Feature Synthesis. If `E1` has a categorical variable, it seems that the direct feature of that categorical will be automatically generated and used in resulting feature matrix. However, that feature won't be used for any stacked features.

This becomes a problem when we want a primitive of multiple variables from different tables. The following example shows an entityset where a user might expect the feature `CAT_PRIMITIVE(values_row_1, transactions.categorical)` to be generated. 

```
import pandas as pd
import featuretools as ft
import featuretools.variable_types as vtypes
datadict = {'values_row_1': [1, 1, 2],
            'transaction_id': [1, 2, 3],
            'categorical': ['cat', 'cat', 'lion']}

data = pd.DataFrame(datadict)

variable_types = {'values_row_1': vtypes.Numeric,
                  'transaction_id': vtypes.Categorical,
                  'categorical': vtypes.Categorical
                  }

es = ft.EntitySet('mock_entityset')

es.entity_from_dataframe(entity_id='values',
                         dataframe=data,
                         index='my_index',
                         variable_types=variable_types,
                         )

es.normalize_entity(base_entity_id='values',
                    new_entity_id='transactions',
                    index='transaction_id',
                    additional_variables=['categorical']
                    )

from featuretools.primitives import make_trans_primitive

def cat_primitive(value, categorical):
    return [x for x in categorical]

Prim = make_trans_primitive(cat_primitive,
                            input_types=[vtypes.Numeric, vtypes.Categorical],
                            return_type=vtypes.Numeric)

                      target_entity='values',
                      agg_primitives=[],
                      trans_primitives=[Prim])

features
```

### LINKED ISSUE #119: Why not build Entity features after Direct features？

CJStadler:
But we do not create transforms of single direct features, or when all inputs are direct features with the same relationship path. This would be redundant as `WEEKDAY(customers.signup_date)` is equivalent to `customers.WEEKDAY(signup_date)`, which should already have been calculated.

This is based on #123 

Resolves #119 
Resolves #81


featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -582,18 +583,18 @@ def test_commutative(es):
...
-    assert num_add_as_base_feat == 9
+    # Check that there are no two AddNumeric features with the same base
+    # features.
+    unordered_args = set()
+    for f in add_feats:
+        arg1, arg2 = f.base_features
+        args_set = frozenset({arg1.unique_name(), arg2.unique_name()})
+        unordered_args.add(args_set)
+
+    assert len(add_feats) == len(unordered_args)
```
CJStadler: Checking the count doesn't communicate why it should have that value. I changed this to instead test that no features are created which have the same two inputs.

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -328,14 +329,14 @@ def _run_dfs(self, entity, relationship_path, all_features, max_depth):
                                      max_depth=max_depth,
                                      relationship_path=sub_relationship_path)
 
-        """
-        Step 3 - Create Transform features
-        """
-        self._build_transform_features(
-            all_features, entity, max_depth=max_depth)
+        # add seed features, if any, for dfs to build on top of
+        for f in self.seed_features:
```
kmax12: any reason this can't stay inside of `_add_identity_features`?

CJStadler: I think I extracted this because if seed features are added before building aggregations then this test fails:
https://github.com/Featuretools/featuretools/blob/4a620527c34a5f7a7484ed9f3a2f954eb7d1c1d8/featuretools/tests/synthesis/test_deep_feature_synthesis.py#L367-L378

But I'm realizing now that you could write a similar test without seed features and it would fail with the current implementation because I moved `_add_identity_features` before building aggregations:
```py
# We don't want this feature because it will always be equal to 'age'
assert not feature_with_name(features, 'LAST(sessions.customers.age)')
```

I need to think some more about how to solve this.

CJStadler: I resolved this by adding identity and seed features first, and then making sure to not build aggregations of direct features of the target. For example, if we are building features for `customers` we do not use direct features such as `customers.age` as the base for aggregations.

CJStadler:
@kmax12 the entityset tests revealed that this was no longer building direct features of aggregations of transform. For example, `log: sessions.MEAN(log.ABSOLUTE(value))` would not have been built because the `ABSOLUTE` transform feature was not built until after we recursed into `sessions`.

To solve this I made two calls to `build_transform_features`: one in the original location to build on identity and aggregation features, and one at the end to build on direct features.  To avoid making duplicate features in the second call only sets of inputs which contain at least one direct feature will be used.

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -651,6 +654,28 @@ def _feature_in_relationship_path(self, relationship_path, feature):
 
         return False
 
+    def _get_matching_inputs(self, all_features, entity, max_depth, input_types,
```
kmax12: creating aggregation features has a similar `_features_by_type` followed by `match` structure. should we try to get it to call this method too?

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -651,6 +654,28 @@ def _feature_in_relationship_path(self, relationship_path, feature):
...
+                             primitive, direct_only):
+        features = self._features_by_type(all_features=all_features,
+                                          entity=entity,
+                                          max_depth=max_depth,
+                                          variable_type=set(input_types))
+
+        matching_inputs = match(input_types, features,
+                                commutative=primitive.commutative)
```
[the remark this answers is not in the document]
CJStadler: Good suggestion, I was able to make it much more efficient this way.

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -651,6 +660,28 @@ def _feature_in_relationship_path(self, relationship_path, feature):
 
         return False
 
+    def _get_matching_inputs(self, all_features, entity, max_depth, input_types,
+                             primitive, direct_only=False, feature_filter=None):
```
kmax12: is there a more descriptive name we can use for `direct_only`?

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -558,16 +565,18 @@ def _build_agg_features(self, all_features, parent_entity, child_entity,
             if type(input_types[0]) == list:
                 input_types = input_types[0]
 
-            features = self._features_by_type(all_features=all_features,
-                                              entity=child_entity,
-                                              max_depth=new_max_depth,
-                                              variable_type=set(input_types))
+            def feature_filter(f):
+                # Remove direct features of parent entity and features in relationship path.
+                return (not _direct_of_entity(f, parent_entity)) \
```
kmax12: wondering are both of these checks necessary? 

wouldn't `self._feature_in_relationship_path(relationship_path, f)` check if the feature is a direct of the entity?

featuretools/synthesis/deep_feature_synthesis.py
CJStadler: What do you think about `require_direct` or `require_direct_input`?

kmax12: I like `require_direct_input`. should we change `need_direct` later to that too?

featuretools/synthesis/deep_feature_synthesis.py
CJStadler: Maybe `_feature_in_relationship_path` should be renamed? It looks like it checks if the feature is an identity feature of a child or parent variable in the relationship path. For example, making an aggregation like `sessions: MEAN(log.session_id)` or a direct feature like `log: sessions.id` is pointless because that information is already on the target entity. That's my guess at least, although all the tests still pass if it is removed.

kmax12: ahh, you're right. let leave it as-is.

featuretools/synthesis/deep_feature_synthesis.py
CJStadler: Yeah I think that's the most clear.

PR #626, opened 2019-06-25
Remove reindex in Entity.query_by_values

CJStadler:
`reindex` may change the order of the dataframe, but the dataframe
should keep the existing ordering by time index.

After calculating features call `reindex` to order by `instance_ids`.


featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -118,6 +118,10 @@ def run(self, instance_ids):
 
         df.index.name = self.entityset[self.feature_set.target_eid].index
         column_list = []
+
+        # Order by instance_ids
+        df = df.reindex(_unique_list(instance_ids))
```
kmax12: to get the unique values in order of appearance, I think `pandas.unique` will be faster 

https://pandas.pydata.org/pandas-docs/version/0.24.2/reference/api/pandas.unique.html

featuretools/entityset/entity.py
```diff
@@ -249,10 +252,6 @@ def query_by_values(self, instance_vals, variable_id=None, columns=None,
         elif instance_vals.shape[0] == 0:
             df = self.df.head(0)
 
-        elif variable_id is None or variable_id == self.index:
-            df = self.df.reindex(instance_vals)
-            df.dropna(subset=[self.index], inplace=True)
-
         else:
             df = self.df[self.df[variable_id].isin(instance_vals)]
```
kmax12: I suspect the reason, we used reindex previously was because it is to select that way then `self.df[self.df[variable_id].isin(instance_vals)]`.

if that is the case, im wondering if there is a way to maintain the behavior, but get the speeds ups of using the index.

for example, there is the the isin method defined for pandas indices: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Index.isin.html

kmax12: here are some timings

```Python 3.6.4 (default, Dec 30 2017, 13:28:16) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.2.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import pandas as pd 
   ...: import numpy as np 
   ...:  
   ...: nrows = 100000 
   ...: sample = 1000 
   ...: numeric_series = pd.Series(range(0, nrows)) 
   ...: numeric_index = pd.Index(numeric_series) 
   ...: numeric_sample = numeric_series.sample(sample).values 
   ...:  
   ...: str_series = numeric_series.astype(str) 
   ...: str_index = pd.Index(str_series) 
   ...: str_sample = str_series.sample(sample).values                                                                                                                                                                                         

In [2]: %timeit numeric_series[numeric_series.isin(numeric_sample)]                                                                                                                                                                           
2.18 ms ± 114 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [3]: %timeit numeric_index.reindex(numeric_sample)                                                                                                                                                                                         
87.2 µs ± 1.65 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [4]: %timeit numeric_index[numeric_index.isin(numeric_sample)]                                                                                                                                                                             
1.64 ms ± 13 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [5]: %timeit str_series[str_series.isin(str_sample)]                                                                                                                                                                                       
6.22 ms ± 126 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [6]: %timeit str_index.reindex(str_sample)                                                                                                                                                                                                 
241 µs ± 8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [7]: %timeit str_index[str_index.isin(str_sample)]                                                                                                                                                                                         
5.73 ms ± 221 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

featuretools/tests/computational_backend/test_feature_set_calculator.py
```diff
@@ -746,3 +749,17 @@ def get_function(self):
     assert all(fm[f5.get_name()].sort_index() == value_sum)
     assert all(fm[f6.get_name()].sort_index() == value_sum)
     assert all(fm[f7.get_name()].sort_index() == value_sum)
+
+
+def test_returns_order_of_instance_ids(es):
+    feature_set = FeatureSet([ft.Feature(es['customers']['age'])])
+    calculator = FeatureSetCalculator(es,
```
kmax12: this works to make sure `FeatureSetCalculator` returns the input order, but I don't believe we have a test to confirm `calculate_feature_matrix` does that. 

we don't need to address that in this PR, just noting it for future reference

featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -118,6 +118,10 @@ def run(self, instance_ids):
 
         df.index.name = self.entityset[self.feature_set.target_eid].index
         column_list = []
+
+        # Order by instance_ids
+        df = df.reindex(list(pd.unique(instance_ids)))
```
CJStadler: I am turning this into a `list` because `instance_ids` may be `pd.Categorical`, and `pd.unique` not only removes duplicate values but also changes the categories. The groups then can't be concatenated later because the categories are different.
```
>>> import pandas as pd
>>> cats = pd.Categorical([1, 2, 3, 4])
>>> cats
[1, 2, 3, 4]
Categories (4, int64): [1, 2, 3, 4]
>>> sub_cats = cats[:2]
>>> sub_cats
[1, 2]
Categories (4, int64): [1, 2, 3, 4]
>>> pd.unique(sub_cats)
[1, 2]
Categories (2, int64): [1, 2]
```
Turning it into a `list` does have the side-effect of changing the index from categorical to int, but I don't know if that matters.

kmax12: would `pd.unique(instance_ids).values` work?

CJStadler: That gives
```
AttributeError: 'Categorical' object has no attribute 'values'
```

kmax12: weird, it's documented here: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Categorical.html#pandas.Categorical

kmax12: using `astype` might be better for the casting.

```
In [14]: l = np.random.choice(["a", "b", "d"], 100000)                                                                                                                                                                                        

In [15]: c = pd.Categorical(l)                                                                                                                                                                                                                

In [16]: %timeit c.astype(pd.Series)                                                                                                                                                                                                          
1.04 ms ± 6.14 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [17]: %timeit c.astype(list)                                                                                                                                                                                                               
1.03 ms ± 9.43 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [18]: %timeit list(c)                                                                                                                                                                                                                      
2.02 ms ± 26.3 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [24]: all(list(c) == c.astype(list))                                                                                                                                                                                                       
Out[24]: True

In [25]: all(list(c) == c.astype(pd.Series))                                                                                                                                                                                                  
Out[25]: True
```

CJStadler: We could do `pd.unique(instance_ids).astype(instance_ids.dtype)` which would keep it as `Categorical`. This is also the fastest I think:
```
In [1]: import numpy as np
   ...: import pandas as pd

In [2]: l = np.random.choice(["a", "b", "d"], 100000)
   ...: c = pd.Categorical(l)

In [3]: l2 = np.random.choice(["a", "b"], 100000)
   ...: c2 = pd.Categorical(l2)

In [4]: %timeit c2.astype(list)
1.21 ms ± 81 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [5]: %timeit c2.astype(c.dtype)
529 µs ± 13.4 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -120,7 +121,10 @@ def run(self, instance_ids):
         column_list = []
 
         # Order by instance_ids
-        df = df.reindex(list(pd.unique(instance_ids)))
+        unique_instance_ids = pd.unique(instance_ids)
```
kmax12: looks like before we call `FeatureSet.run()` some sorting of instance ids occurs. does it make sense to just remove that? 

https://github.com/Featuretools/featuretools/blob/5eef35b63823a257f04253130162f1338df5edee/featuretools/computational_backends/calculate_feature_matrix.py#L333

the other things, is we could do the conversion in the call there, as well. i'm don't have strong feelings either way though

CJStadler: Some tests depend on that sorting being done. Should we guarantee the order?

Why would we need another type conversion?

kmax12: let's leave that sorting for now.

i should have said "type conversion in the call there, instead"

PR #632, opened 2019-07-01
Add cumulative primitives to docs

CJStadler:
These seem to be missing.

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -156,6 +156,11 @@ Cumulative Transform Primitives
 
     Diff
     TimeSincePrevious
+    CumCount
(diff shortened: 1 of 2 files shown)


docs/source/api_reference.rst
```diff
@@ -156,6 +156,11 @@ Cumulative Transform Primitives
 
     Diff
     TimeSincePrevious
```
CJStadler: Do the above two belong under "Cumulative Transform Primitives"?

kmax12: that's the best place to put them for now since they do use multiple rows of the dataframe, although in this case just the previous

PR #634, opened 2019-07-01
Moved dask, distributed imports

gsheni:
- Moved dask, distributed imports so they are inside the function. This is because dask/distributed is an optional library for featuretools


gsheni:
> Can we move dask and distributed from `requirements.txt` to `test-requirements.txt` since they are now optional

Let's keep it in requirements.txt so that if users run DFS with `n_jobs > 1`, it won't give an error.

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -917,13 +918,12 @@ def test_dask_persisted_es(es, capsys):
 
 class TestCreateClientAndCluster(object):
     def test_user_cluster_as_string(self, monkeypatch):
-        monkeypatch.setitem(create_client_and_cluster.__globals__,
-                            'LocalCluster',
-                            mock_cluster)
-        monkeypatch.setitem(create_client_and_cluster.__globals__,
-                            'Client',
-                            MockClient)
+        from featuretools.computational_backends import utils
```
rwedge: I think we can just import this once at the top of the file.  It's ok to import even if distributed isn't installed since the distributed import happens inside the `get_client_cluster` function.  The monkeypatch changes are fixed between tests so we don't need to reimport.

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -917,13 +918,12 @@ def test_dask_persisted_es(es, capsys):
...
 
+        def mock_get_client_cluster():
```
rwedge: MockClient and mock_cluster are only used for these tets so we can move this function to that utils file and import it instead.

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
gsheni: Fixed.

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
gsheni: Fixed.

docs/source/changelog.rst
```diff
@@ -30,6 +30,7 @@ Changelog
         * Remove unused variance_selection.py file (:pr:`613`)
         * Remove Timedelta data param (:pr:`619`)
         * Remove DaysSince primitive (:pr:`628`)
+        * Moved dask, distributed imports (:pr:`634`)
```
kmax12: move change long entry to Future Release

featuretools/computational_backends/utils.py
```diff
@@ -278,3 +279,8 @@ def create_client_and_cluster(n_jobs, num_tasks, dask_kwargs, entityset_size):
             warned_of_memory = True
 
     return client, cluster
+
+
+def get_client_cluster():
```
kmax12: maybe just add a comment that you separated this out, to make it easier to mock during testing?

docs/source/changelog.rst

featuretools/computational_backends/utils.py
gsheni: Added comment.

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -859,6 +858,7 @@ def test_verbose_cutoff_time_chunks(es):
 
 
 def test_dask_kwargs(es):
+    from distributed.utils_test import cluster
```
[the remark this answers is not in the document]
gsheni: Moved back to top

PR #637, opened 2019-07-03
Compare entities by ID in DFS

CJStadler:
This is much faster than fully comparing the entities.


featuretools/entityset/relationship.py
```diff
@@ -108,7 +108,8 @@ def _is_unique(self):
         """Is there any other relationship with same parent and child entities?"""
         es = self.child_entity.entityset
         relationships = es.get_forward_relationships(self._child_entity_id)
-        n = len([r for r in relationships if r.parent_entity == self.parent_entity])
+        n = len([r for r in relationships
+                 if r._parent_entity_id == self._parent_entity_id])
```
CJStadler: This case wasn't causing the observed performance drop, but I also don't see any reason to compare the actual entities here.

CJStadler:
Even though `__eq__` wasn't being called with `deep=True` before, so it wasn't comparing dataframes, the comparison of variables is non-trivial (quadratic in the number of variables).

https://github.com/Featuretools/featuretools/blob/e67fe588ba3636a5b602b9c55229fe7da1214e2c/featuretools/entityset/entity.py#L120-L122

featuretools/entityset/entity.py
```diff
@@ -117,9 +117,8 @@ def __eq__(self, other, deep=False):
             return False
         if len(self.variables) != len(other.variables):
             return False
-        for v in self.variables:
-            if v not in other.variables:
-                return False
+        if set(self.variables) != set(other.variables):
```
CJStadler: I removed most of the places where this would be hit, but I thought I might as well make this comparison faster while we're looking at it.

docs/source/changelog.rst
```diff
@@ -14,6 +14,7 @@ Changelog
         * Keep dataframe sorted by time during feature calculation (:pr:`626`)
         * Fix bug in encode_features that created duplicate columns of
           features with multiple outputs (:pr:`622`)
+        * Fix performance regression in DFS (:pr:`637`)
```
kmax12: move to Future Release

PR #660, opened 2019-07-11
Add Trigger Test

gsheni:


docs/source/changelog.rst
```diff
@@ -14,6 +14,7 @@ Changelog
         * Hyperlinked ComposeML (:pr:`653`)
         * Added FAQ (:pr:`620`)
     * Testing Changes
+        * Add trigger test for PP (:pr:`680`)
```
kmax12: we can just write out premium primitives

.circleci/config.yml
```diff
@@ -43,6 +43,20 @@ jobs:
...
+        type: string
+        default: "3.7.3"
+    executor:
+      name: python
+      image_tag: << parameters.image_tag >>
+    steps:
+      - checkout
+      - run:
+          name: "Triger PP tests"
+          command: "curl -u ${PP_K}: -d build_parameters[CIRCLE_JOB]=python-35 https://circleci.com/api/v1.1/project/github/FeatureLabs/premium-primitives/tree/master"
```
rwedge: This job config can be simplified a lot.  Look at the changelog_updated job as a reference.  The image can just be busybox and it might not even require running the checkout command

rwedge: Actually, busybox doesn't have curl, only wget

gsheni: Using circleci-buildpack, curl image

PR #665, opened 2019-07-12
Fix relationship path deserialization

### LINKED ISSUE #657: Error running DFS after filtering feature definitions
Hello,

I am using the below code to filter out some features that i want to exclude when running DFS:

features_to_keep = []

for x in features:
    if 'Index' not in x.get_name():
        features_to_keep.append(x)
       
ft.save_features(features_to_keep, 'feature_definitions.json')

However, when i use the filtered feature definitions (features_to_keep) in DFS, I get the following error: 
'list' object has no attribute 'name'

    feature_matrix = ft.calculate_feature_matrix(features_to_keep, 
                                                 entityset = es, 
                                                 n_jobs = 1, 
                                                 verbose = 0,
                                                 chunk_size = 1000)

Can you please have a look and let me know any solution please?

Thanks

CJStadler:
We were deserializing the `AggregationFeatures.relationship_path` as a `list` instead of a `RelationshipPath`.

Fixes #657

### DIFF
--- a/featuretools/tests/primitive_tests/test_agg_feats.py
+++ b/featuretools/tests/primitive_tests/test_agg_feats.py
@@ -349,10 +349,11 @@ def test_serialization(es):
     }
 
     assert dictionary == max1.get_arguments()
-    assert max1 == \
-        ft.AggregationFeature.from_dictionary(dictionary, es,
-                                              {value.unique_name(): value},
-                                              primitives_deserializer)
+    deserialized = ft.AggregationFeature.from_dictionary(dictionary,
+                                                         es,
+                                                         {value.unique_name(): value},
+                                                         primitives_deserializer)
+    _assert_agg_feats_equal(max1, deserialized)
 
     is_purchased = ft.IdentityFeature(es['log']['purchased'])
     use_previous = ft.Timedelta(3, 'd')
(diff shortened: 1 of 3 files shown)


featuretools/tests/primitive_tests/test_agg_feats.py
```diff
@@ -605,3 +608,11 @@ def pd_top3(x):
         else:
             for i1, i2 in zip(true_results.iloc[i], df.iloc[i]):
                 assert (pd.isnull(i1) and pd.isnull(i2)) or (i1 == i2)
+
+
+def _assert_agg_feats_equal(f1, f2):
```
kmax12: would just checking `f1.unique_name() == f2.unique_name()` be enough given the what the method is supposed to do? i don't see any harm in adding the other stuff but unique name should have all this info below included if it is working properly, i think.

CJStadler: Generally that should be enough, but in this case that would be `True` even if one of `relationship_path`s is a `list` and the other is a `RelationshipPath`. `RelationshipPath` mostly behaves like a list so most of the logic still works, and the line which triggered the error is only hit under specific conditions:
https://github.com/Featuretools/featuretools/blob/a3966fa3b3f078d237f0f27465d56c5f27778395/featuretools/feature_base/feature_base.py#L617-L621

I could just compare `unique_name` and `relationship_path` – I added the other comparisons to be careful but they don't seem necessary.

kmax12: I see. we can leave it as is for now

PR #667, opened 2019-07-12
Do not calculate dependencies of precalculated features 

CJStadler:
Currently if there is a precalculated feature "A(B)" (and no other feature depends on B) we will still re-calculate B even though the resulting feature vector will never be used.

This commit solves this by making FeatureSet aware of which features will be precalculated, so that it can exclude them and their dependencies from the feature trie.

Script showing this:
<details>

```py
import pandas as pd
import featuretools as ft

from datetime import datetime

class Hello(ft.primitives.AggregationPrimitive):
    name = "hello"
    input_types = [ft.variable_types.Numeric]
    return_type = ft.variable_types.Numeric

    def get_function(self):
        def hello(s):
            print('Hello')
            return 0
        return hello

class World(ft.primitives.AggregationPrimitive):
    name = "world"
    input_types = [ft.variable_types.Numeric]
    return_type = ft.variable_types.Numeric

    def get_function(self):
        def hello(s):
            print('World')
            return 0
        return hello

agg_feat = ft.Feature(es['log']['value'], parent_entity=es['sessions'], primitive=Hello)
agg_feat2 = ft.Feature(agg_feat, parent_entity=es['customers'], primitive=World)
dfeat = ft.DirectFeature(agg_feat2, es['sessions'])
cutoff_time = pd.DataFrame({
    'time': [datetime(2013, 4, 9, 10, 31, 19), datetime(2013, 4, 9, 11, 0, 0)],
    'instance_id': [0, 2]
})
feature_matrix = ft.calculate_feature_matrix([dfeat],
                                             es,
                                             cutoff_time_in_index=False,
                                             approximate=ft.Timedelta(12, 's'),
                                             cutoff_time=cutoff_time,
                                             chunk_size='cutoff time')
```

On master this outputs

```
Hello
Hello
Hello
World
Hello
Hello
Hello
Hello
Hello
Hello
World
Hello
Hello
Hello
```

The 3 `Hello`s after each `World` are generating feature vectors which are never used (this is only obvious for the last 3 though).

On this branch we instead get

```
Hello
Hello
Hello
World
Hello
Hello
Hello
World
```

</details>

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -189,12 +189,11 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
 
     # Get features to approximate
     if approximate is not None:
-        approximate_feature_trie, all_approx_feature_set = gather_approximate_features(feature_set)
+        approximate_feature_trie = gather_approximate_features(feature_set)
         # Make a new FeatureSet that ignores approximated features
         feature_set = FeatureSet(features, ignored_feature_trie=approximate_feature_trie)
     else:
         approximate_feature_trie = None
-        all_approx_feature_set = None
 
     # Check if there are any non-approximated aggregation features
     no_unapproximated_aggs = True
@@ -205,7 +204,11 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
             no_unapproximated_aggs = False
             break
 
-        deps = feature.get_dependencies(deep=True, ignored=all_approx_feature_set)
+        if approximate_feature_trie:
+            all_approx_features = {f for _, feats in approximate_feature_trie for f in feats}
+        else:
+            all_approx_features = set()
+        deps = feature.get_dependencies(deep=True, ignored=all_approx_features)
         for dependency in deps:
             if isinstance(dependency, AggregationFeature):
                 no_unapproximated_aggs = False
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -191,9 +191,7 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
     if approximate is not None:
         approximate_feature_trie = gather_approximate_features(feature_set)
         # Make a new FeatureSet that ignores approximated features
-        feature_set = FeatureSet(features, ignored_feature_trie=approximate_feature_trie)
-    else:
-        approximate_feature_trie = None
+        feature_set = FeatureSet(features, approximate_feature_trie=approximate_feature_trie)
 
     # Check if there are any non-approximated aggregation features
     no_unapproximated_aggs = True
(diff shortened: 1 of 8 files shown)


featuretools/computational_backends/feature_set.py
```diff
@@ -15,11 +15,16 @@
 
 
 class FeatureSet(object):
-    def __init__(self, features):
+    def __init__(self, features, ignored_feature_trie=None):
```
kmax12: add doc about `ignored_feature_trie`

featuretools/computational_backends/feature_set.py
```diff
@@ -41,7 +46,14 @@ def __init__(self, features):
             fname: [self.features_by_name[dname] for dname in feature_dependents[fname]]
             for fname, f in self.features_by_name.items()}
 
-        self.feature_trie = self._build_feature_trie()
+        self._feature_trie = None
+
+    @property
+    def feature_trie(self):
```
kmax12: add note that this is optimization

featuretools/entityset/relationship.py
```diff
@@ -161,6 +161,13 @@ def __eq__(self, other):
         return isinstance(other, RelationshipPath) and \
             self._relationships_with_direction == other._relationships_with_direction
 
+    def __repr__(self):
+        if self._relationships_with_direction:
+            path = '%s.%s' % (next(self.entities()), self.name)
+        else:
+            path = 'empty'
```
kmax12: to discuss: is this the best string to use?

featuretools/computational_backends/utils.py
```diff
@@ -86,8 +86,8 @@ def gather_approximate_features(feature_set):
                 base_feature = base_feature.base_features[0]
 
             if isinstance(base_feature, AggregationFeature):
-                feature_list = approximate_feature_trie.get_node(path).value
-                feature_list.append(base_feature)
+                node_feature_set = approximate_feature_trie.get_node(path).value
+                node_feature_set.add(base_feature.unique_name())
                 approximate_feature_set.add(base_feature.unique_name())
 
     return approximate_feature_trie, approximate_feature_set
```
kmax12: refactor to not return `approximate_feature_set`

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -375,6 +379,7 @@ def calc_results(time_last, ids, precalculated_features=None, training_window=No
 
 
 def approximate_features(feature_set, cutoff_time, window, entityset,
+                         approximate_feature_trie,
```
kmax12: get this value from `feature_set`. likely renamed `ignored_feature_trie` in feature_set as well

featuretools/entityset/relationship.py
CJStadler: Options:
1. `<RelationshipPath empty>` (current)
2. `<RelationshipPath>`
3. `<RelationshipPath None>`
4. `<RelationshipPath []>`

I'm leaning towards number 4 because `RelationshipPath` generally acts like a list so `[]` makes it clear that this path is like an empty list. Since the normal format is `<RelationshipPath entity1.entity2>`, `<RelationshipPath empty>` could potentially look like `empty` is the name of an entity.

kmax12: i like `[]` too

PR #669, opened 2019-07-15
Add releases tests for premium primitives

gsheni:
- Add release tests for premium primitives


.circleci/config.yml
```diff
@@ -48,8 +48,17 @@ jobs:
       - image: circleci/buildpack-deps:curl
     steps:
       - run:
-          name: "Triger PP tests"
+          name: "Triger premium primitives tests for python 3.5"
```
rwedge: Trigger (same for all run descriptions)

.circleci/config.yml
```diff
@@ -227,9 +240,16 @@ workflows:
...
-                - master
\ No newline at end of file
+                - master
+      - run_premium_primitives_tests_on_release:
+          name: "run premium primitives tests on release"
+          filters:
+            branches:
+              ignore: /.*/
+            tags:
+              only: ^v\d+\.\d+\.\d+$
```
rwedge: does this need to be surrounded by forward slashes?

gsheni: Yes you do. If you don't put the slash (`\`) it means that it will look for the letter `d` rather than the digit between 0-9.

.circleci/config.yml

.circleci/config.yml
rwedge: No I mean shouldn't it be `/^v\d+\.\d+\.\d+$/`

docs/source/changelog.rst
```diff
@@ -17,6 +17,7 @@ Changelog
         * Added FAQ (:pr:`620`)
     * Testing Changes
         * Add trigger test for premium primitives (:pr:`660`)
+        * Add release tests premium primitives (:pr:`669`)
```
rwedge: Let's combine this line with the above and say "Add triggers for premium primitive tests"

gsheni: Fixed

PR #676, opened 2019-07-17
Change feature calculation to return in order of instance ids provided

kmax12:
This PR updates `ft.calculate_feature_matrix` to return rows in the order of the instance ids provided.

It also updates entity loading to only sort the input dataframe if a `time_index` is present. This ensures rows will be returned in the same order as the input if no time index is provided and no instance ids are provided


featuretools/tests/entityset_tests/test_es.py
```diff
@@ -480,7 +481,7 @@ def test_concat_entitysets(es):
     assert old_es_1.__eq__(es_1, deep=True)
     assert old_es_2.__eq__(es_2, deep=True)
 
-    assert es_3.__eq__(es, deep=True)
+    assert es_3.__eq__(es)
```
rwedge: If this is due to no longer sorting data for entities without time indexes, we could potentially not split up the data of non-time entities and then we could still do a deep comparison

kmax12: below, we check for the fact that the dataframes have the same data after concatenation, so I think it's fine to just remove the deep check here and not mess with the test

PR #679, opened 2019-07-22
Multi output primitive stackability

### LINKED ISSUE #593: Allow users to specify names of multioutput custom primitives
Currently, multi output features get named like `feature_name__1`, `feature_name__2`, ... `feature_name__n`.

It would be nice if there was a way for a primitive to override the `get_feature_names` method to give each output feature a specific name. 

This was brought up in #591.

ctduffy:
### Adding support for multi output primitive stacking
Closes #593


featuretools/feature_base/feature_base.py
```diff
@@ -436,7 +444,10 @@ def variable(self):
 
     @property
     def number_output_features(self):
-        return self.base_features[0].primitive.number_output_features
+        if isinstance(self.base_features[0], MultiOutputFeature):
```
rwedge: We might be able to just return `self.base_features[0].number_output_features` in all cases here

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -746,6 +747,37 @@ def test_makes_direct_features_through_multiple_relationships(games_es):
...
+    for i in range(3):
+        f = 'NUM_UNIQUE(sessions.N_MOST_COMMON(log.countrycode)[%d])' % i
+        assert feature_with_name(feat, f)
+
+
+def test_seed_multi_output_feature_stacking(es):
+    threecommon = NMostCommon(3)
+    tc = ft.Feature(es['log']['product_id'], parent_entity=es["sessions"], primitive=threecommon)
+
+    stacked = ft.Feature(tc, parent_entity=es['customers'], primitive=NumUnique)
```
rwedge: This feature is stacking on the multi-output feature without slicing it into MultiOutputFeatures, which I don't think we should allow

The "new_seeds" logic in DFS wouldn't be necessary since the users would have constructed the seed features using the MultiOutputFeature class

We should add checks to AggregationFeature and TransformFeature to catch stacking on a multi-output feature like this.

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -746,6 +747,37 @@ def test_makes_direct_features_through_multiple_relationships(games_es):
...
+
+    fm, feat = ft.dfs(entityset=es,
+                      target_entity="customers",
+                      seed_features=[stacked],
```
rwedge: This test will also work if we make tc the seed feature and include NumUniqe as an agg_primitive.  DFS will stack on the seed feature to create the features we are testing for

featuretools/tests/synthesis/test_deep_feature_synthesis.py
ctduffy: I think this is ok-It does that slicing under the hood, and the output still is the same (i.e. the columns are named correctly). I think it is easier for the user to construct the whole feature, which makes sure that that whole feature gets calculated, and then the program stacks it and creates the MultiOutputFeature class for them, since it is not fully a class.

My main point is that I think that allowing this is beneficial, and having users build multi output features rather than having a multi output feature that, when stacked on, creates the MultiOutputFeatures, would be complicated and harder for them. Also, it might cause some confusion, because the MultiOutputFeature is only created when there is stacking going on.

featuretools/feature_base/feature_base.py
```diff
@@ -521,6 +533,13 @@ def __init__(self, base_features, parent_entity, primitive,
                                                  primitive=primitive,
                                                  name=name)
 
+    def __getitem__(self, key):
```
rwedge: does it make sense to implement here or in base class?

rwedge: does it make sense to implement here or in base class

featuretools/tests/primitive_tests/test_feature_base.py
```diff
@@ -143,3 +152,59 @@ def test_to_dictionary(es):
         'arguments': direct_feature.get_arguments()
     }
     assert expected == direct_feature.to_dictionary()
+
+
+def test_multi_output_base_error_agg(es):
+    threecommon = NMostCommon(3)
+    tc = ft.Feature(es['log']['product_id'], parent_entity=es["sessions"], primitive=threecommon)
+    EText = "Cannot stack on whole multi-output feature."
```
rwedge: let's use snake case for the variable name

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -788,12 +790,33 @@ def gen_feat_names(self):
             assert base_feature.unique_name() != join_time_split.unique_name()
 
 
-def test_tranform_stack_agg(es):
-    topn = ft.Feature(es['log']['product_id'],
-                      parent_entity=es['customers'],
-                      primitive=NMostCommon(n=3))
-    with pytest.raises(AssertionError):
-        ft.Feature(topn, primitive=Percentile)
+def test_stacking_of_multi_output_transform_feat(es):
```
rwedge: Instead of adding new test case I think we can update the "test_make_transform_multiple_output_features" test to check for stacking

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -746,6 +749,53 @@ def test_makes_direct_features_through_multiple_relationships(games_es):
...
+        number_output_features = 6
+
+        def get_function(self):
+            def test_f(x):
+                times = pd.Series(x)
+                units = ["year", "month", "day", "hour", "minute", "second"]
+                return [times.apply(lambda x: getattr(x, unit)) for unit in units]
+            return test_f
+
+    fm, feat = ft.dfs(entityset=es,
```
rwedge: we can skip calculating the feature matrix by using `features_only` option since we don't actually use matrix

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -746,6 +749,53 @@ def test_makes_direct_features_through_multiple_relationships(games_es):
...
+    for i in range(6):
+        f = 'DIFF(TEST_TIME(date_of_birth)[%d])' % i
+        assert feature_with_name(feat, f)
+
+
+def test_seed_multi_output_feature_stacking(es):
+    threecommon = NMostCommon(3)
+    tc = ft.Feature(es['log']['product_id'], parent_entity=es["sessions"], primitive=threecommon)
+
+    fm, feat = ft.dfs(entityset=es,
```
rwedge: same thing here with skipping the matrix

featuretools/tests/primitive_tests/test_agg_feats.py
```diff
@@ -610,6 +612,26 @@ def pd_top3(x):
...
 
+def test_stacking_multi(es):
+    threecommon = NMostCommon(3)
+    tc = ft.Feature(es['log']['product_id'], parent_entity=es["sessions"], primitive=threecommon)
+
+    stacked = []
+    for i in range(3):
+        stacked.append(ft.Feature(tc[i], parent_entity=es['customers'], primitive=NumUnique))
+
+    fm, feat = ft.dfs(entityset=es,
```
rwedge: instead of running DFS let's just use calculate_feature_matirx

featuretools/tests/primitive_tests/test_agg_feats.py
```diff
@@ -610,6 +612,26 @@ def pd_top3(x):
...
+                      target_entity="customers",
+                      seed_features=stacked,
+                      agg_primitives=[],
+                      trans_primitives=[]
+                      )
+
+    for i in range(3):
+        f = 'NUM_UNIQUE(sessions.N_MOST_COMMON(log.product_id)[%d])' % i
```
rwedge: we should test that the actual values are what we expect

featuretools/tests/primitive_tests/test_groupby_transform_primitives.py
```diff
@@ -352,6 +355,31 @@ def get_function(self):
...
+        input_types=[Datetime],
+        return_type=Numeric,
+        number_output_features=6,
+    )
+
+    fl = dfs(
+        entityset=es,
+        target_entity="sessions",
+        agg_primitives=[],
+        trans_primitives=[TestTime, CumSum],
```
rwedge: `CumSum` should be in the `groupby_trans_primitives` list instead of `trans_primitives` so that we are testing a GroupByTransform feature instead of a Transform feature

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -789,34 +788,18 @@ def gen_feat_names(self):
         for base_feature in feature.base_features:
             assert base_feature.unique_name() != join_time_split.unique_name()
```
rwedge: I think we can delete this for loop checking that we don't stack on top of multi-output features, combine the dfs call above with the new one written below,  and use the new for loop checking that we do stack on multi-output features

featuretools/tests/primitive_tests/test_groupby_transform_primitives.py
```diff
@@ -352,6 +355,32 @@ def get_function(self):
         assert ((pd.isnull(x) and pd.isnull(y)) or x == y)
 
 
+def test_groupby_multi_output_stacking(es):
+    def test_time(x):
+        return x
```
rwedge: this is a miss on codecov because we don't actually calculate the feature values

if we do something like this:
```python
TestTime = make_trans_primitive(
    function=lambda x: x,
    name="test_time",
    input_types=[Datetime],
``` 

codecov won't complain and the test will still run

ctduffy: the issue is that the name of the primitive then becomes "lambda" instead of test_time in the name of the primitive
Screen Shot 2019-08-08 at 4 51 43 PM

rwedge: I believe if you add `name="test_time` as a parameter it won't use the lambda to come up with the feature name

ctduffy: oh missed that part, thanks!

featuretools/computational_backends/feature_set.py
```diff
@@ -7,6 +7,7 @@
 from featuretools.feature_base import (
     AggregationFeature,
     GroupByTransformFeature,
+    MultiOutputFeature,
```
kmax12: maybe we can name this better.  Something that indicates it just one of the outputs of a multi output feature

what about `FeatureOutputDimension`?

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -635,8 +640,17 @@ def _features_by_type(self, all_features, entity, max_depth,
...
 
-        for feat in all_features[entity.id]:
-            f = all_features[entity.id][feat]
+        entity_features = all_features[entity.id].copy()
+        for fname, feature in all_features[entity.id].items():
+            outputs = feature.number_output_features
+            if outputs > 1:
+                del(entity_features[fname])
+                for i in range(outputs):
+                    new_feat = MultiOutputFeature(feature, i)
```
kmax12: should we use the `[]` syntax here?

featuretools/tests/primitive_tests/test_features_serializer.py
```diff
@@ -42,6 +42,36 @@ def test_base_features_in_list(es):
...
+
+    serializer = FeaturesSerializer(features)
+
+    flist = [feat.unique_name() for feat in features]
+    fd = [feat.to_dictionary() for feat in features]
+    fdict = dict(zip(flist, fd))
+
+    expected = {
+        'ft_version': ft.__version__,
+        'schema_version': SCHEMA_VERSION,
```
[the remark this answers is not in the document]
ctduffy: What do you mean by increment? Like hard code in a schema version or just put in ft.__version__ there too?

featuretools/computational_backends/feature_set.py
```diff
@@ -135,6 +136,8 @@ def _add_feature_to_trie(self, trie, feature, approximate_feature_trie,
         sub_ignored_trie = approximate_feature_trie.get_node(feature.relationship_path)
 
         for dep_feat in feature.get_dependencies():
+            if isinstance(dep_feat, MultiOutputFeature):
```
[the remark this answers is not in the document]
ctduffy: that would break other usages of `get_dependencies`, particularly in the `from_dictionary` usages, for that function should return a specific slice or multi output feature with the associated column number, not just the base feature of the multi output without the column

Issue #670, opened 2019-07-15
EntitySet.normalize_entity allows using the child entity's index as the child variable in the relationship


rwedge:
Currently `Entityset.normalize_entity` will accept the base entity's index column as the index argument.  This creates a one-to-one relationship between the entities, which Featuretools does not support.  It also results in the index column in the base entity being converted from `Index` variable type to `Id` variable type.

We should add a check to `normalize_entity` that raises an error if the `index` argument to the method is already the index column of the base entity.

rwedge:
Fixed in #681

PR #681, opened 2019-07-22
Normalize entity same index as base entity

jeremyliweishih:
This request addresses #670 .

Adds a check and raises an exception if the index of the base entity is equal to the index argument. This request adds an accompanying test case.

As stated in #670, Feature tools does not handle one-to-one relationships in entitysets. Therefore, this check must be made to stop creating a one-to-one relationship with the same index.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -11,7 +11,7 @@ Changelog
         * Set index after adding ancestor relationship variables (:pr:`668`)
         * Fix user-supplied variable_types modification in Entity init (:pr:`675`)
         * Don't calculate dependencies of unnecessary features (:pr:`667`)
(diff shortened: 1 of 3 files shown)


docs/source/changelog.rst
```diff
@@ -11,6 +11,7 @@ Changelog
         * Set index after adding ancestor relationship variables (:pr:`668`)
         * Fix user-supplied variable_types modification in Entity init (:pr:`675`)
         * Don't calculate dependencies of unnecessary features (:pr:`667`)
+        * Normalize entity same index as base entity #681 (:pr:`681`)
```
rwedge: The ":pr:681" syntax will create a link to this pr, you can remove the "#681".  Also can you add a little more to the description so someone unfamiliar could grasp the issue?

PR #688, opened 2019-07-31
Adding logic for the default value when the parent is missing

### LINKED ISSUE #682: When calculating direct features use default value if parent missing
For example, if there is a relationship `transaction.session_id -> sessions.id` and we are calculating a feature `transactions: sessions.SUM(transactions.value)` any rows for which there is no corresponding session should be given the default value of `0` instead of `NaN`.

Of course this should not normally occur, but when it does it seems more reasonable to use the `default_value`.

`DirectFeature.default_value` is already implemented. We should be able to use the same logic that we do for aggregation features.
https://github.com/Featuretools/featuretools/blob/6f4ffd7ef7ea42f95dbaf3892615717a521299db/featuretools/computational_backends/feature_set_calculator.py#L611-L618

smartdev020:
### Pull Request Description

Adding logic for the default value in _calculate_direct_features. Now, if the parent missing, the calculated value will be the default value (not a 'nan').

Fixes #682

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,6 +6,7 @@ Changelog
     * Enhancements
         * Added drop_first as param in encode_features (:pr:`647`)
         * Generate transform features of direct features (:pr:`623`)
+        * When calculating direct features use default value if parent missing (:pr:`688`)
     * Fixes
         * Fix performance regression in DFS (:pr:`637`)
         * Fix deserialization of feature relationship path (:pr:`665`)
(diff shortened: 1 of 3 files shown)


iaroslav0:
As for this Pull Request. I run 'make test' and 'make lint'. It shows error as below.

flake8 featuretools && isort --check-only --recursive featuretools
ERROR: /work/featuretools/featuretools/__init__.py Imports are incorrectly sorted.

After using 'make lint-fix', the error is fixed, but the code cannot be run as some modules cannot be imported. I need to check out the code to be the original one.

rwedge:
stale pr, closing for now.  could move forward if it was updated to match current code

Issue #453, opened 2019-03-05
haversine error


jeff-hernandez:
### Bug/Feature Request Description
Haversine primitive raises an error in the following use case:
```python   
latlong1 = pd.Series({71: nan, 72: nan, 73: nan, 74: nan, 75: nan})
latlong2 = pd.Series({71: (40.798587799072266, -73.9694061279297), 72: (40.8021354675293, -73.96797943115234), 73: (40.77003860473633, -73.96251678466798), 74: (40.77336883544922, -73.96420288085938), 75: (40.73043823242188, -73.98625183105469)})
```
```
TypeError: 'float' object is not subscriptable
```

-----
*Issues created here on Github are for bugs or feature requests. For usage questions and questions about errors, please ask on Stack Overflow with the [featuretools](https://stackoverflow.com/questions/tagged/featuretools) tag. Check the [documentation](https://docs.featuretools.com/#get-help) for further guidance on where to ask your question.*

vnarayan13:
Thinking of something like this to handle NaNs and scalars in the [haversine function](https://github.com/Featuretools/featuretools/blob/master/featuretools/primitives/standard/transform_primitive.py#L367):

`
def haversine(latlong1, latlong2):
    try:
        lat_1s, lon_1s = map(np.array, zip(*latlong1))
    except TypeError:
        lat_1s, lon_1s = [np.tile(np.nan, len(latlong1))]*2
    try:
        lat_2s, lon_2s = map(np.array, zip(*latlong2))
    except TypeError:
        lat_2s, lon_2s = [np.tile(np.nan, len(latlong2))]*2
`

kmax12:
is there a way to vectorize the whole haversine primitive while also not erroring out on nans? any help on this is appreciated!

ekerazha:
It looks like I got it working by representing LatLong missing values as `(np.nan, np.nan)`, however I still get this error when I use the `first` or `last` agg_primitives.

ekerazha:
There's a pending PR about this issue: https://github.com/FeatureLabs/featuretools/pull/693

I don't know if it also fixes the problem with the `first` and `last` primitives.

rwedge:
Thanks for bringing this up again @ekerazha!

Some extra info on how this happens with `first` and `last` primitives would be helpful.  Does this only happen with LatLong data?  Do you have an example of when this occurs?

ekerazha:
@rwedge 
> Does this only happen with LatLong data?

[...]

PR #693, opened 2019-08-05
Fix for haversine error (#453) Handle NaNs in haversine transform_primitive 

Alex-Monahan:
Prior to handling NaN's, the code first checks if any are present (to save calculation time). Time with old method that failed on NaNs was 650 ms per 1 million rows. Time for data with no NaNs is now 750ms per 1 million rows. When NaNs are present, time is 1.5 seconds per 1 million rows, even using np.where to fix NaNs in a vectorized way.

### Pull Request Description
This allows the Haversine function to handle NaN values as inputs. It does so in a vectorized way so that performance is affected as little as possible.

-----


kmax12:
@Alex-Monahan thank you for the contribution. we will review shortly and hopefully get it into the next release

rwedge:
Closing, work continuing in #1007

PR #699, opened 2019-08-06
adding autonormalize to extras_require

allisonportis:
adding autonormalize to extras_require in setup.py

-----


kmax12:
let's also make sure to test it gets installed correctly. see here for example

https://github.com/Featuretools/featuretools/blob/master/.circleci/config.yml#L159

docs/source/changelog.rst
```diff
@@ -18,6 +18,7 @@ Changelog
         * Update variable type inference to better check for string values (:pr:`683`)
     * Changes
         * Moved dask, distributed imports (:pr:`634`)
+	* Added to Featuretools plugins (:pr:`699`)
```
kmax12: might as well mention that it's auto normalize specifically. also, this should be under "Enhancements" since it's new functionality!

kmax12: also add yourself to contributors

docs/source/frequently_asked_questions.ipynb
```diff
@@ -752,6 +752,18 @@
     "feature_matrix.head()"
    ]
   },
+  {
```
kmax12: can you add a separate question that demonstrates autonormalize on the same dataset as above?

allisonportis: In the example above, they use a normalized part of the customer dataset (just the transaction dataframe). Should I just demonstrate it on the whole customer table as a single table? And should I change the example above to use the whole dataset as a single table?

docs/source/frequently_asked_questions.ipynb
```diff
@@ -743,15 +885,763 @@
...
+       "<graphviz.dot.Digraph at 0x120fbb850>"
+      ]
+     },
+     "execution_count": 6,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "import autonormalize as an\n",
```
kmax12: can you update this to import from featuretools?

PR #700, opened 2019-08-07
adding categorical-encoding to setup.py

alexjwang:
### Pull Request Description
Incorporating categorical-encoding PyPi library release into Featuretools.


kmax12:
@alexjwang can you also add to docs/source/getting_started/install.rst? look at how we do it in #699

PR #704, opened 2019-08-12
Adding support for nlp-primitives

ctduffy:
### Pull Request Description
Incorporates the nlp-primitives PyPI library into Featuretools.


kmax12:
@ctduffy can you also add to docs/source/getting_started/install.rst? look at how we do it in #699

ctduffy:
Add the nlp-primitives API doc to the Featuretools documentation, which will require putting nlp_primitives as a requirement in FT dev reqs

PR #707, opened 2019-08-15
Created test methods that would create serialization files

jeremyliweishih:
### Pull Request Description
Created test methods for ES serialization and feature serialization that would create files that could be uploaded to s3 for testing purposes.

### DIFF
--- a/featuretools/tests/entityset_tests/test_serialization.py
+++ b/featuretools/tests/entityset_tests/test_serialization.py
@@ -28,7 +28,7 @@ TEST_KEY = "test_access_key_es"
 
 '''
 `test_create_serialization` creates a test tar file based off of the current schema version.
-This can be uploaded to S3 for deserialization tests.
+This file will be located in tests/integration_data/.cache and can be uploaded to S3 for deserialization tests.
 '''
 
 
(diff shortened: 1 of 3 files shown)


featuretools/tests/entityset_tests/test_serialization.py
```diff
@@ -16,14 +16,21 @@
...
 
+SCHEMA_VER = serialize.SCHEMA_VERSION
 CACHE = os.path.join(os.path.dirname(integration_data.__file__), '.cache')
 BUCKET_NAME = "test-bucket"
 WRITE_KEY_NAME = "test-key"
 TEST_S3_URL = "s3://{}/{}".format(BUCKET_NAME, WRITE_KEY_NAME)
-S3_URL = "s3://featuretools-static/test_serialization_data_1.0.0.tar"
-URL = 'https://featuretools-static.s3.amazonaws.com/test_serialization_data_1.0.0.tar'
+
+S3_URL = "s3://featuretools-static/test_serialization_data_{}.tar".format(SCHEMA_VER)
```
kmax12: this file shouldn't be based off of the schema version since it might change with the schema change. if anything it should be based on the featuretools version. but really, i would just say, let's manually increment it as needed

featuretools/tests/entityset_tests/test_serialization.py
```diff
@@ -191,6 +198,13 @@ def test_serialize_s3_parquet(es, s3_client, s3_bucket):
     assert es.__eq__(new_es, deep=True)
 
 
+def test_create_serialization(es, s3_client, s3_bucket):
```
kmax12: is this a test cases that supposed to run?

rwedge:
We probably want to document the process for uploading new testing files somewhere but adding new methods to automate this is probably unnecessary

PR #712, opened 2019-08-22
Change LatLong columns back into tuples when deserializing from csv or parquet

### LINKED ISSUE #691: LatLong variables not deserialized from parquet
When we serialize entity dataframes to parquet we turn all `object` types into strings: https://github.com/Featuretools/featuretools/blob/c583ff130f7c37ad2e4077e82d7d61bf0d2ef6cf/featuretools/entityset/serialize.py#L84-L89

This includes `LatLong` variables which are represented as tuples. When we deserialize the entityset the column now contains strings, and operations expecting a tuple will fail.

In addition, the conversion to strings is done in-place, mutating the entityset being serialized. Calculating features which expect a `LatLong` tuple will then fail.

## Test case

<details>

```py
import featuretools as ft

features = ft.dfs(entityset=es,
                  target_entity='log',
                  agg_primitives=[],
                  trans_primitives=['haversine'],
                  features_only=True)

target_entity = es['log']
cutoff_time = target_entity.df[[target_entity.index, target_entity.time_index]]

print('Running CFM before serialization')
ft.calculate_feature_matrix(features, entityset=es, cutoff_time=cutoff_time)
print('success')

print('Seralization to parquet')
filename = '/tmp/test_entityset_parquet'
es.to_parquet(filename)

print('Running CFM on same in-memory entityset, after serialization')
ft.calculate_feature_matrix(features, entityset=es, cutoff_time=cutoff_time)
print('success')

print('Deserializing parquet')
es = ft.read_entityset(filename)

print('Running CFM on deserialized entityset')
ft.calculate_feature_matrix(features, entityset=es, cutoff_time=cutoff_time)
print('success')
```

Currently, the second call to `calculate_feature_matrix` will fail with
```
File ".../featuretools/primitives/standard/transform_primitive.py", line 578, in haversine
    np.radians, [lon_1s, lat_1s, lon_2s, lat_2s])
TypeError: ufunc 'radians' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''
```

</details>

## Solution

Since we serialize the variable types we should be able to use that information to properly deserialize the columns from strings.

rwedge:
### Pull Request Description
Fixes #691 

Also fixes `Dataframe.__eq__` converting all object columns in the dataframe to unicode strings.

@CJStadler did you encounter this with csv as well or just parquet?

-----


featuretools/entityset/deserialize.py
```diff
@@ -127,22 +127,37 @@ def read_entity_data(description, path):
...
         error = 'must be one of the following formats: {}'
         raise ValueError(error.format(', '.join(FORMATS)))
     dtypes = description['loading_info']['properties']['dtypes']
-    return dataframe.astype(dtypes)
+    dataframe = dataframe.astype(dtypes)
+
+    if load_format in ['parquet', 'csv']:
+        latlongs = []
+        for var_description in description['variables']:
+            if var_description['type']['value'] == "latlong":
```
kmax12: perhaps use `LatLong.type_string` here instead of `"latlong`

PR #717, opened 2019-08-28
Don't delete the whole destination folder while saving entityset

### LINKED ISSUE #706: Saving entityset deletes whole destination folder
https://github.com/Featuretools/featuretools/blob/97eb274a48752985cd33d62a92eef2a16c874d5b/featuretools/entityset/serialize.py#L148

In the current implementation, when an entity set is serialized, the user must give a target folder for the entityset to go in. However, when Featuretools serializes the entityset, it deletes the whole folder if it exists and then copies the serialized data set into it. Instead, featuretools should only delete conflicting files (files that are the same as the ones it is trying to put in), and leave the rest of the files alone.

chidauri:
### Pull Request Description
Fixes #706  as the title suggests. The new functionality is to replace the conflicting files only and not the whole path.

-----

### DIFF
--- a/featuretools/tests/entityset_tests/test_serialization.py
+++ b/featuretools/tests/entityset_tests/test_serialization.py
@@ -1,4 +1,5 @@
 import errno
+import json
 import os
 import shutil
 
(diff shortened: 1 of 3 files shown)


rwedge:
Thanks again, @chidauri.  Could you add a test case confirming that this no longer deletes unrelated files in the destination folder?

featuretools/tests/entityset_tests/test_serialization.py
```diff
@@ -268,6 +268,13 @@ def test_serialize_url_csv(es):
         es.to_csv(URL, encoding='utf-8', engine='python')
 
 
+def test_serialize_data_description(es, tmpdir):
```
rwedge: this test name could be updated to provide more context about that it is testing extra files are not removed

rwedge:
@chidauri thanks for adding the test case!  I think it would also be good to test that old serialization files (e.g. data_description.json) get overwritten.

featuretools/tests/entityset_tests/test_serialization.py
chidauri: Is it okay now?

rwedge: Looks like there's an issue with py35 / py27 tests.

chidauri: Fixed :)

PR #718, opened 2019-08-28
Raise warning not error on schema version mismatch

### LINKED ISSUE #698: Raise warning not error on schema version mismatch
We currently raise an error if a schema version of a feature list or entityset is incorrect. Instead, we should raise a warning and then still attempt to load it in.

This is preferred in cases where the schema change may not actually impact a particular user from loading a from an old version if the change didn't affect them. The user could then reserialize after loading and stop getting the warning.

chidauri:
### Pull Request Description
Fixes #698 . 

-----

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -9,7 +9,7 @@ Changelog
         * Fixed entity set deserialization (:pr:`720`)
         * Added error message when DateTimeIndex is a variable but not set as the time_index (:pr:`723`)
     * Changes
-		* Raise warning and not error on schema version mismatch (:pr: `718`)
+	* Raise warning and not error on schema version mismatch (:pr: `718`)
     * Documentation Changes
         * Updated URL for Compose (:pr:`716`)
     * Testing Changes
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -9,12 +9,13 @@ Changelog
         * Fixed entity set deserialization (:pr:`720`)
         * Added error message when DateTimeIndex is a variable but not set as the time_index (:pr:`723`)
     * Changes
+		* Raise warning and not error on schema version mismatch (:pr: `718`)
     * Documentation Changes
         * Updated URL for Compose (:pr:`716`)
     * Testing Changes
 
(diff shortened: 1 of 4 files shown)


rwedge:
@chidauri would you be willing to modify the tests to check that the warning is raised? Here's an example of how we check for a warning in a different test:

https://github.com/Featuretools/featuretools/blob/7d9da1533a6cb7f81dc9feb521c44834c033c530/featuretools/tests/computational_backend/test_calculate_feature_matrix.py#L43-L47

chidauri:
@rwedge python2.7 test is failing, i am not sure how is that related to my change? and how to format changelog.rst? Thanks.

docs/source/changelog.rst
```diff
@@ -9,12 +9,13 @@ Changelog
         * Fixed entity set deserialization (:pr:`720`)
         * Added error message when DateTimeIndex is a variable but not set as the time_index (:pr:`723`)
     * Changes
+        * Raise warning and not error on schema version mismatch (:pr: `718`)
```
rwedge: Remove the space between ":pr:" and "`781`"

featuretools/utils/gen_utils.py
```diff
@@ -112,15 +113,15 @@ def check_schema_version(cls, cls_type):
             if c_num > s_num:
                 break
             elif c_num < s_num:
-                raise RuntimeError(error_text_upgrade)
+                warnings.warn(error_text_upgrade)
```
rwedge: We should add a `break` after this warning so that we do not continue to iterate after the warning

rwedge:
Also, let's rephrase these warnings.  Let's remove the "unable to load" sentence from both warnings and add "Attempting to load . . ." at the end of each warning.

rwedge:
@chidauri the python2.7 failure was unrelated.  I re-triggered the tests and it passed the second time.

chidauri:
@rwedge is it ok now?

rwedge:
@chidauri looks good!  I went ahead and changed the `error_text` variable names to `warning_text` for some added clarity.

PR #723, opened 2019-09-03
Raise an error when DatetimeTimeIndex is a variable but time_index is empty

### LINKED ISSUE #497: Handle case where DatetimeTimeIndex is set as variable type, but not specified as time index
Some users have been creating entities like this 

```
es = ft.EntitySet()

es.entity_from_dataframe(entity_id="my_entity",
                         index="my_index_col",
                         variable_type={
                            'my_time_index_col': ft.variable_types.DatetimeTimeIndex
                         }
```

when all they really have to do is

```
es.entity_from_dataframe(entity_id="my_entity",
                         index="my_index_col",
                         time_index="my_time_index_col")
```

I think to handle this we should raise an error if we detect the variable type of  `DatetimeTimeIndex`, but the actual time index property hasn't

christopherbunn:
Added a ValueError message when a DatetimeTimeIndex is set as a variable but is not specified as a time_index

fixes #497

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -486,7 +486,8 @@ class EntitySet(object):
             raise ValueError("time_index and index cannot be the same value, %s" % (time_index))
 
         if vtypes.DatetimeTimeIndex in variable_types.values() and time_index is None:
-            raise ValueError("'DatetimeTimeIndex' must be set using time_index parameter")
+            var_name = variable_types.keys()[variable_types.values().index(vtypes.DatetimeTimeIndex)]
+            raise ValueError("Variable %s must be set using time_index parameter" % (var_name))
 
         entity = Entity(
             entity_id,
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -485,9 +485,10 @@ class EntitySet(object):
         if time_index is not None and time_index == index:
             raise ValueError("time_index and index cannot be the same value, %s" % (time_index))
 
-        if vtypes.DatetimeTimeIndex in variable_types.values() and time_index is None:
-            var_name = variable_types.keys()[variable_types.values().index(vtypes.DatetimeTimeIndex)]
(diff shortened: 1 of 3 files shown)


featuretools/entityset/entityset.py
```diff
@@ -485,6 +485,9 @@ def entity_from_dataframe(self,
         if time_index is not None and time_index == index:
             raise ValueError("time_index and index cannot be the same value, %s" % (time_index))
 
+        if vtypes.DatetimeTimeIndex in variable_types.values() and time_index is None:
+            raise ValueError("'DatetimeTimeIndex' must be set using time_index parameter")
```
rwedge: We should include the variable name in the error message to help the user track down the source of the problem

featuretools/entityset/entityset.py
```diff
@@ -485,6 +485,10 @@ def entity_from_dataframe(self,
         if time_index is not None and time_index == index:
             raise ValueError("time_index and index cannot be the same value, %s" % (time_index))
 
+        if vtypes.DatetimeTimeIndex in variable_types.values() and time_index is None:
+            var_name = variable_types.keys()[variable_types.values().index(vtypes.DatetimeTimeIndex)]
```
rwedge: This works but I think we can clean it up a bit

We can loop through `variable_types.items()` once and raise an error the first time we encounter the problem.

featuretools/entityset/entityset.py
```diff
@@ -485,6 +485,11 @@ def entity_from_dataframe(self,
         if time_index is not None and time_index == index:
             raise ValueError("time_index and index cannot be the same value, %s" % (time_index))
 
+        if time_index is None:
+            for variable, variable_type in variable_types.items():
+                if variable_type == vtypes.DatetimeTimeIndex:
+                    raise ValueError("Variable %s must be set using time_index parameter" % (variable))
```
rwedge: Can you also include that the variable is DatetimeTimeIdex type, e.g. DatetimeTimeIndex variable %s.

Issue #719, opened 2019-08-29
Why does `training_window` in `dfs()` not support 'month'?


rightx2:
### Why does `training_window` not support month?
(replace this text with your issue)

 

#### Bug/Feature Request Description

`ft.dfs` is not support only `month`!!

I checked it out by printing some variables in `entityset/timedelta.py`

```
In [2]: self._readable_units
Out[2]: 
{'ms': 'Milliseconds',
 's': 'Seconds',
 'h': 'Hours',
 'm': 'Minutes',
 'd': 'Days',
 'o': 'Observations',
 'w': 'Weeks',
 'Y': 'Years',
 'u': 'Units'}

In [4]: self._readable_to_unit
Out[4]: 
{'milliseconds': 'ms',
 'seconds': 's',
 'hours': 'h',
 'minutes': 'm',
 'days': 'd',
 'observations': 'o',
 'weeks': 'w',
 'years': 'Y',
 'units': 'u'}
``` 

Any reason for that?

jeremyliweishih:
Hello @rightx2!

Could you please check what version of Featuretools you are on? Relative units such as `month` was added in Featuretools v0.10.0. If you're on a lower version you could upgrade with `pip install -U featuretools`. Thanks and let me know if you have any other questions.

rightx2:
@jeremyliweishih  I've upgraded lib to 0.10.1 but still doesn't work

What I've tried:

1. `training_window="3 mo", training_window="3 months", training_window="3 month"`

-> It gave me an error: `AssertionError: training window must be an absolute Timedelta`

2. `training_window=pd.Timedelta(3, "M")`

-> It runs ok but feature_matrix generated was wrong. It seemed like it didn't do any calculation....


3. `training_window=relativedelta(months=3)`

-> errors: `ValueError: Unable to parse timedelta: relativedelta(months=+3)`


Notice: If I removed `training_window` args, it worked like a charm based on `cutoff_time`.

jeremyliweishih:
@rightx2 it has been fixed in #728 and you can test it from master. The fixes will be going out on the next release. Thanks for bringing this up!

rwedge:
Fixed in version 0.11.0

Issue #618, opened 2019-06-21
ft.dfs training_window doesn't accept Timedelta


Lewington-pitsos:
### Bug/Feature Request Title
ft.dfs training_window doesn't accept Timedelta

-----

#### Bug/Feature Request Description

ft.dfs could not parse the timeDelta I passed in for `training_window` .

``` 
fm, features = ft.dfs(
    cutoff_time=cutoff,
    entityset=es,
    target_entity="transactions",
    agg_primitives=['mean', 'trend'],
    trans_primitives=["month", "time_since_previous"],
    ignore_variables={"transactions": ["year", "year_month"]},
    max_depth=3,
    verbose=True,
    training_window=pd.Timedelta(3, "M")
)

```
#### Expected Output

some features and a new dataframe

#### Output of ``featuretools.show_info()``

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-44-e968247ca5ac> in <module>
----> 1 ft.show_info()

AttributeError: module 'featuretools' has no attribute 'show_info'
```

</details>

The actual error was:

```
Built 57 features








Elapsed: 00:00 | Remaining: ? | Progress:   0%|          | Calculated: 0/11 chunks
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-41-3b1ada997f04> in <module>
      8     max_depth=3,
      9     verbose=True,
---> 10     training_window=pd.Timedelta(3, "M")
     11 )

/opt/anaconda3/lib/python3.7/site-packages/featuretools/utils/entry_point.py in function_wrapper(*args, **kwargs)
     44                     ep.on_error(error=e,
     45                                 runtime=runtime)
---> 46                 raise e
     47 
     48             # send return value

/opt/anaconda3/lib/python3.7/site-packages/featuretools/utils/entry_point.py in function_wrapper(*args, **kwargs)
     36                 # call function
     37                 start = time.time()
---> 38                 return_value = func(*args, **kwargs)
     39                 runtime = time.time() - start
     40             except Exception as e:

[...]

PR #728, opened 2019-09-04
Updated training window error assertion for relative units

christopherbunn:
- Updated the error assertion for query_by_values so that an error is asserted only when the training window is in observations #719
- Added a test case for when a relative unit is used for the training window #618 
- Updated test_calcuate_feature_matrix.py for new error assertion message

### DIFF
--- a/featuretools/tests/synthesis/test_dfs_method.py
+++ b/featuretools/tests/synthesis/test_dfs_method.py
@@ -140,3 +140,40 @@ def test_dask_kwargs(entities, relationships):
     for column in feature_matrix:
         for x, y in zip(feature_matrix[column], feature_matrix_2[column]):
             assert ((pd.isnull(x) and pd.isnull(y)) or (x == y))
+
+
+def test_accepts_pandas_training_window():
+    from featuretools.demo import load_mock_customer
+    es = load_mock_customer(return_entityset=True)
+    cutoff_times = pd.DataFrame()
+    cutoff_times['customer_id'] = [1, 2, 3, 4]
+    cutoff_times['time'] = pd.to_datetime(['2014-1-1 04:00', '2014-1-1 05:00',
+                                           '2014-1-1 06:00', '2014-1-1 08:00'])
+    cutoff_times['label'] = [True, True, False, True]
+
+    feature_matrix, features = dfs(
+        cutoff_time=cutoff_times,
+        entityset=es,
+        target_entity="customers",
+        agg_primitives=['mean', 'trend'],
+        trans_primitives=["month", "time_since_previous"],
+        ignore_variables={"transactions": ["year", "year_month"]},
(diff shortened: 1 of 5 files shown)


featuretools/tests/synthesis/test_dfs_method.py
```diff
@@ -140,3 +140,40 @@ def test_dask_kwargs(entities, relationships):
...
+    )
+
+    feature_matrix_2, features_2 = dfs(
+        cutoff_time=cutoff_times,
+        entityset=es,
+        target_entity="customers",
+        agg_primitives=['mean', 'trend'],
+        trans_primitives=["month", "time_since_previous"],
+        ignore_variables={"transactions": ["year", "year_month"]},
+        training_window="3 months",
```
rwedge: Can this test inspect feature values to confirm that the correct training window is being applied

christopherbunn: I haven't seen a dataset with a date range of > 1 month (since we are testing relative units). I would have to make a new test dataset to show the training window filtering out the data.

featuretools/tests/synthesis/test_dfs_method.py
```diff
@@ -140,3 +141,87 @@ def test_dask_kwargs(entities, relationships):
...
+                                  dataframe=transactions_df,
+                                  index="id",
+                                  time_index="transaction_time")
+
+    es = es.entity_from_dataframe(entity_id="cards",
+                                  dataframe=cards_df,
+                                  index="id")
+    relationship = Relationship(es["cards"]["id"], es["transactions"]["card_id"])
+    es = es.add_relationship(relationship)
+    es.add_last_time_indexes()
```
rwedge: Since these two tests repeat the same code for creating the entityset can we move into a fixture

featuretools/tests/synthesis/test_dfs_method.py
```diff
@@ -140,3 +167,43 @@ def test_dask_kwargs(entities, relationships):
...
+    feature_matrix_5, features_5 = dfs(entityset=datetime_es,
+                                       target_entity="transactions",
+                                       cutoff_time=pd.Timestamp("2013-2-28 04:00"),
+                                       training_window=Timedelta("1 year"))
+
+    assert feature_matrix.index.all([1, 2, 3, 4, 5])
+    assert feature_matrix_2.index.all([1, 2, 3, 4])
+    assert feature_matrix_3.index.all([2, 3, 4])
+    assert feature_matrix_4.index.all([2, 3, 4])
+    assert feature_matrix_4.index.all([2, 3])
```
rwedge: I don't think these assertions are working as intended.  `assert feature_matrix_4.index.all([2, 3, 4])` and `assert feature_matrix_4.index.all([2, 3])` are both evaluating as True

Issue #729, opened 2019-09-05
Allow DFS to take in pd.DateOffset objects


jeremyliweishih:
Currently DFS and TimeDelta does not take in pd.DateOffset objects. This could require more refactoring with our TimeDelta class. Certain areas to take into consideration:

- [ ] `training_window` must support this new functionality 
- [ ] `use_previous` must support this new functionality. Take into account that we need to be able to serialize features.

rwedge:
Addressed by #732

PR #732, opened 2019-09-06
Added support for pd.DateOffset in DFS

christopherbunn:
Adds support for pd.DateOffset in DFS parameters (#729)

### DIFF
--- a/featuretools/computational_backends/feature_set.py
+++ b/featuretools/computational_backends/feature_set.py
@@ -3,6 +3,8 @@ import logging
 from builtins import object
 from collections import defaultdict
 
+import pandas as pd
+
 from featuretools.entityset.relationship import RelationshipPath
 from featuretools.feature_base import (
     AggregationFeature,
@@ -11,8 +13,6 @@ from featuretools.feature_base import (
     TransformFeature
 )
 from featuretools.utils import Trie
-
-import pandas as pd
 
 logger = logging.getLogger('featuretools.computational_backend')
 
@@ -206,7 +206,7 @@ class FeatureSet(object):
 # These functions are used for sorting and grouping features
 
 
-def _get_use_previous(f): # TODO Sort and group features for DateOffset
+def _get_use_previous(f):  # TODO Sort and group features for DateOffset
     if isinstance(f, AggregationFeature) and f.use_previous is not None and not isinstance(f.use_previous, pd.DateOffset):
         return (f.use_previous.unit, f.use_previous.value)
     else:
--- a/featuretools/computational_backends/feature_set.py
+++ b/featuretools/computational_backends/feature_set.py
@@ -3,8 +3,6 @@ import logging
 from builtins import object
 from collections import defaultdict
 
-import pandas as pd
-
 from featuretools.entityset.relationship import RelationshipPath
 from featuretools.feature_base import (
     AggregationFeature,
@@ -206,8 +204,8 @@ class FeatureSet(object):
 # These functions are used for sorting and grouping features
 
 
-def _get_use_previous(f):  # TODO Sort and group features for DateOffset
-    if isinstance(f, AggregationFeature) and f.use_previous is not None and not isinstance(f.use_previous, pd.DateOffset):
+def _get_use_previous(f):  # TODO Sort and group features for DateOffset with two different temporal values
+    if isinstance(f, AggregationFeature) and f.use_previous is not None and f.use_previous.unit is not None:
         return (f.use_previous.unit, f.use_previous.value)
     else:
         return ("", -1)
--- a/featuretools/computational_backends/feature_set.py
+++ b/featuretools/computational_backends/feature_set.py
@@ -12,6 +12,8 @@ from featuretools.feature_base import (
 )
 from featuretools.utils import Trie
 
+import pandas as pd
+
 logger = logging.getLogger('featuretools.computational_backend')
 
 
@@ -204,8 +206,8 @@ class FeatureSet(object):
 # These functions are used for sorting and grouping features
 
 
-def _get_use_previous(f):
-    if isinstance(f, AggregationFeature) and f.use_previous is not None:
+def _get_use_previous(f): # TODO Sort and group features for DateOffset
+    if isinstance(f, AggregationFeature) and f.use_previous is not None and not isinstance(f.use_previous, pd.DateOffset):
         return (f.use_previous.unit, f.use_previous.value)
     else:
         return ("", -1)
--- a/featuretools/computational_backends/feature_set.py
+++ b/featuretools/computational_backends/feature_set.py
@@ -205,8 +205,13 @@ class FeatureSet(object):
 
 
 def _get_use_previous(f):  # TODO Sort and group features for DateOffset with two different temporal values
-    if isinstance(f, AggregationFeature) and f.use_previous is not None and f.use_previous.unit is not None:
-        return (f.use_previous.unit, f.use_previous.value)
+    if isinstance(f, AggregationFeature) and f.use_previous is not None:
+        if len(f.use_previous.times.keys()) > 1:
+            return ("", -1)
+        else:
+            unit = list(f.use_previous.times.keys())[0]
+            value = f.use_previous.times[unit]
+            return (unit, value)
     else:
         return ("", -1)
 
--- a/featuretools/computational_backends/utils.py
+++ b/featuretools/computational_backends/utils.py
@@ -55,13 +55,21 @@ def datetime_round(dt, freq):
     if not freq.is_absolute():
         raise ValueError("Unit is relative")
 
-    if freq.unit == 'm':
-        unit = 't'
+    # TODO: multitemporal units
+    all_units = list(freq.times.keys())
+    if len(all_units) == 1:
+        unit = all_units[0]
+        value = freq.times[unit]
+        if unit == 'm':
+            unit = 't'
+        # No support for weeks in datetime.datetime
+        if unit == 'w':
+            unit = 'd'
+            value = value * 7
+        freq = str(value) + unit
+        return dt.dt.floor(freq)
     else:
-        unit = freq.unit
-
-    freq = str(freq.value) + unit
-    return dt.dt.floor(freq)
+        assert "Frequency cannot have multiple temporal parameters"
 
 
 def gather_approximate_features(feature_set):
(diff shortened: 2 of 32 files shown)


featuretools/computational_backends/feature_set.py
```diff
@@ -204,8 +206,8 @@ def _dependent_uses_full_entity(self, feature):
 # These functions are used for sorting and grouping features
 
 
-def _get_use_previous(f):
-    if isinstance(f, AggregationFeature) and f.use_previous is not None:
+def _get_use_previous(f):  # TODO Sort and group features for DateOffset
```
jeremyliweishih: Can you update the TODO saying that its only necessary for DateOffset with two different temporal values?And also check if the DateOffset contains a unit before ignoring it?

featuretools/tests/primitive_tests/test_agg_feats.py
```diff
@@ -634,6 +634,35 @@ def test_stacking_multi(es):
         assert fm[cols[i]].tolist() == correct_vals[i] or fm[cols[i]].tolist() == correct_vals1[i]
 
 
+def test_use_previous_pd_dateoffset(es):
+    total_events_last_20_years = ft.Feature(es["log"]["id"],
```
jeremyliweishih: Was there a reason to use such a long time period for `use_previous`?

featuretools/tests/primitive_tests/test_agg_feats.py
```diff
@@ -634,6 +634,35 @@ def test_stacking_multi(es):
...
+
+    since_str = (pd.to_datetime('today') - pd.tseries.offsets.BDay(100))
+    total_events_last_100_bday = ft.Feature(es["log"]["id"],
+                                            parent_entity=es["customers"],
+                                            use_previous=pd.tseries.offsets.BDay(100),
+                                            primitive=Count)
+    since_str = since_str.strftime('%Y-%m-%d %H:%M:%S')
+    feature_matrix_3 = ft.calculate_feature_matrix([total_events_last_100_bday], es)
+
+    assert (feature_matrix_3["COUNT(log, Last 100 businessdays)".format(since_str)] == [0.0, 0.0, 0.0]).all()
```
jeremyliweishih: What exactly is `since_str` replacing in `"COUNT(log, Last 100 businessdays)"`?

jeremyliweishih: Would it be possible to choose a unit for business days that don't result in no results to confirm that it works?

featuretools/tests/synthesis/test_dfs_method.py
```diff
@@ -200,10 +200,32 @@ def test_accepts_relative_training_window(datetime_es):
...
+def test_accepts_pd_timedelta_training_window(datetime_es):
     feature_matrix, features = dfs(entityset=datetime_es,
                                    target_entity="transactions",
                                    cutoff_time=pd.Timestamp("2012-4-1 04:00"),
                                    training_window=pd.Timedelta(90, "D"))
 
     assert (feature_matrix.index == [2, 3, 4]).all()
+
+
+def test_accepts_pd_dateoffset_training_window(datetime_es):
```
jeremyliweishih: Since the feature matrix results are the same across the different cutoff_times is it possible to choose equivalent time periods across the different units? For example 60 days and 2 Months.

featuretools/utils/wrangle.py
```diff
@@ -46,6 +49,20 @@ def _check_timedelta(td, entity_id=None, related_entity_id=None):
     if isinstance(td, pd.Timedelta):
         unit = 's'
         value = td.total_seconds()
+    elif isinstance(td, pd.DateOffset):
```
jeremyliweishih: Maybe a comment here would clarify the logic.

featuretools/tests/primitive_tests/test_agg_feats.py
```diff
@@ -635,32 +635,15 @@ def test_stacking_multi(es):
...
-    assert (feature_matrix["COUNT(log, Last 20 Years)"] == feature_matrix_2["COUNT(log, Last 20 Years)"]).all()
-
-    since_str = (pd.to_datetime('today') - pd.tseries.offsets.BDay(100))
-    total_events_last_100_bday = ft.Feature(es["log"]["id"],
-                                            parent_entity=es["customers"],
-                                            use_previous=pd.tseries.offsets.BDay(100),
-                                            primitive=Count)
-    since_str = since_str.strftime('%Y-%m-%d %H:%M:%S')
-    feature_matrix_3 = ft.calculate_feature_matrix([total_events_last_100_bday], es)
+    time_diff = (pd.Timestamp('today') - pd.Timestamp('2011-04-09 10:32:00')).total_seconds()
```
jeremyliweishih: Might be better to set a constant time_diff instead of anchoring on the current date

featuretools/tests/entityset_tests/test_timedelta.py
```diff
@@ -184,3 +184,21 @@ def test_relative_month():
 
     time = pd.to_datetime('2020-01-31')
     assert time + td == pd.to_datetime('2020-07-31')
+
+
+def test_pd_dateoffset_to_timedelta():
```
jeremyliweishih: Could we also get tests ensuring addition and subtraction?

featuretools/tests/synthesis/test_dfs_method.py
```diff
@@ -212,20 +212,13 @@ def test_accepts_pd_timedelta_training_window(datetime_es):
 def test_accepts_pd_dateoffset_training_window(datetime_es):
     feature_matrix, features = dfs(entityset=datetime_es,
```
jeremyliweishih: Maybe we could do 90 Days == 3 Months == 1 Quarter End?

featuretools/tests/entityset_tests/test_timedelta.py
jeremyliweishih: Just want to make sure the math checks out.

featuretools/tests/synthesis/test_dfs_method.py
christopherbunn: For this timestamp (2012-4-1), a QuarterEnd offset sets it to 1-31-2012. I can change the pd timedelta test so that it's consistent with the other two tests.

featuretools/tests/primitive_tests/test_agg_feats.py
christopherbunn: It's currently anchored on this date as use_previous depends on the current date to run. If it was constant, then the result of the test could change to all zeros rather than the nonzero answer this test is trying to look for.

featuretools/computational_backends/feature_set.py
```diff
@@ -204,9 +204,14 @@ def _dependent_uses_full_entity(self, feature):
 # These functions are used for sorting and grouping features
 
 
-def _get_use_previous(f):
+def _get_use_previous(f):  # TODO Sort and group features for DateOffset with two different temporal values
```
jeremyliweishih: Can you put an issue up about sorting for DateOffset / How does sorting work with use_previous?

featuretools/computational_backends/utils.py
```diff
@@ -55,13 +55,21 @@ def datetime_round(dt, freq):
     if not freq.is_absolute():
         raise ValueError("Unit is relative")
 
-    if freq.unit == 'm':
-        unit = 't'
+    # TODO: multitemporal units
```
jeremyliweishih: Add exception for multi-temporal case

featuretools/entityset/timedelta.py
```diff
@@ -44,48 +44,51 @@ class Timedelta(object):
...
-
-        # weeks
-        if unit in self._convert_to_days:
-            self._original_unit = unit
-            self.value = self.value * self._convert_to_days[unit]
-            unit = 'd'
-
-        self.unit = unit
-        self.delta_obj = self.get_unit_type()
+            self.times = td.times
```
jeremyliweishih: Can we add helper functions to check value and fixed_units? This would make init more clear.

featuretools/entityset/timedelta.py
```diff
@@ -101,67 +104,100 @@ def _check_unit_plural(cls, s):
...
-
-    def get_name(self):
-        unit = self.readable_unit
-        if self.readable_unit == "Weeks":
-            # divide to convert back
-            return "{} {}".format(self.value / self._convert_to_days["w"], unit)
-        if self.value == 1:
-            unit = self.make_singular(unit)
+            readable_times = self.lower_readable_times()
+            return relativedelta(**readable_times)
```
jeremyliweishih: Could you add an issue about supporting relativedelta explicitly and what is needed to support it? It seems like we can parse out unit and values and delta_obj to it?

featuretools/entityset/serialize.py
```diff
@@ -19,7 +19,7 @@
     import tempfile
 
 FORMATS = ['csv', 'pickle', 'parquet']
-SCHEMA_VERSION = "1.0.0"
+SCHEMA_VERSION = "1.1.0"
```
jeremyliweishih: @rwedge Do we need to bump the ES schema or feature_base schema. Or Both?

rwedge: I think only feature_base schema.

featuretools/computational_backends/utils.py
christopherbunn: I think the assertion in line 72 should cover this case. Does it need to be changed to an exception?

featuretools/entityset/timedelta.py
christopherbunn: Created #744

featuretools/computational_backends/feature_set.py
christopherbunn: Created #746

featuretools/tests/primitive_tests/test_agg_feats.py
rwedge: If we changed `cutoff_time` to be a constant time as well instead of the current time than we could avoid anchoring on the current date.

featuretools/entityset/timedelta.py
```diff
@@ -44,48 +44,37 @@ class Timedelta(object):
...
-
-    _convert_to_days = {
-        "w": 7
+        "mo": "Months"
     }
 
     _readable_to_unit = {v.lower(): k for k, v in _readable_units.items()}
 
-    def __init__(self, value, unit=None):
+    def __init__(self, value, unit=None, delta_obj=None):
```
rwedge: Should we document the `delta_obj` parameter?

christopherbunn: I think we should so it's clear that it's more for _check_timedelta() and it's not necessary if a user only wants to use value and units to instantiate.

PR #733, opened 2019-09-09
Cum count fix

### LINKED ISSUE #607: CumCount doesn't get applied
The way we match inputs breaks in `ft.dfs` for `groupby_trans_primitives` breaks when one of the inputs to primitive is an `Id` variable like is the case in `CumCount`.

The issue is with how we append `Id` on this line when looking for matching inputs.

Here is an example the reproduces the issue

```import featuretools as ft

fl = ft.dfs(target_entity="sessions",
       entityset=es,
       agg_primitives=[],
       trans_primitives=[],
       groupby_trans_primitives=["cum_count"],
       features_only=True)
fl
```

produces 

```
[<Feature: customer_id>, <Feature: device>, <Feature: customers.zip_code>]
```

which is missing `<Feature: CUM_COUNT(sessions) by customer_id>`

frances-h:
### Pull Request Description
Fixes #607 by changing handling of groupby_trans_primitives that take ID as input

-----


featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -528,9 +530,15 @@ def _build_transform_features(self, all_features, entity, max_depth=0,
 
             for matching_input in matching_inputs:
                 if all(bf.number_output_features == 1 for bf in matching_input):
-                    new_f = GroupByTransformFeature(list(matching_input[:-1]),
-                                                    groupby=matching_input[-1],
-                                                    primitive=groupby_prim)
+                    # Need ID if it's part of the input, but not for other groupbys
+                    if id_is_input:
+                        new_f = GroupByTransformFeature(list(matching_input[:]),
+                                                        groupby=matching_input[-1],
```
rwedge: 1. What if a primitive has an input_types like `[Id, Numeric]`.

2. Say an entity had two Id columns and for CUM_COUNT the matching inputs found were [id_1] and [id_2].  This code will make `CUM_COUNT(id_1) by id_1` and `CUM_COUNT(id_2) by id_2` but it should also create `CUM_COUNT(id_1) by id_2` and `CUM_COUNT(id_2) by id_1`

featuretools/feature_base/feature_base.py
```diff
@@ -714,7 +714,10 @@ def copy(self):
     def generate_name(self):
         # exclude the groupby feature from base_names since it has a special
         # place in the feature name
-        base_names = [bf.get_name() for bf in self.base_features[:-1]]
+        if self.primitive.name == "cum_count":
+            base_names = [self.entity_id]
```
rwedge: The `Count` primitive uses this name scheme but I think there are complictions when this format is used with transform primitives. `Count` uses `COUNT(<path>)` instead of `COUNT(<entity.id>)`, and each path has a specific id variable associated with it.  

`CUM_COUNT(<entity.id>) by <groubpy_id>` could create multiple features with the same name if an entity has multiple id variables:
`CUM_COUNT(id_1.entity.id) by id_1` and `CUM_COUNT(id_2.entity.id) by id_1` will have identical feature names

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -271,6 +271,16 @@ def test_make_groupby_features(es):
...
 
 
+def test_make_groupby_features_with_id(es):
+    dfs_obj = DeepFeatureSynthesis(target_entity_id='sessions',
+                                   entityset=es,
+                                   agg_primitives=[],
+                                   trans_primitives=[],
+                                   groupby_trans_primitives=['cum_count'])
+    features = dfs_obj.build_features()
+    assert (feature_with_name(features, "CUM_COUNT(customer_id) by customer_id"))
```
rwedge: can we also test for the `CUM_COUNT(id_1) by id_2` case

I think "customers.CUM_COUNT(cohort) by région_id" would satisfy this

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -516,15 +516,28 @@ def _build_transform_features(self, all_features, entity, max_depth=0,
...
             if type(input_types[0]) == list:
-                input_types = input_types[0]
-            input_types.append(Id)
-
+                input_types = input_types[0][:]
+            id_is_input = Id in input_types
+            # Add Id if not present for the matching function
+            if not id_is_input:
+                input_types.append(Id)
+            id_index = input_types.index(Id)
```
rwedge: I think instead of checking matching inputs to get `Id` features we can use `self._features_by_type`

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -516,16 +516,27 @@ def _build_transform_features(self, all_features, entity, max_depth=0,
             input_types = groupby_prim.input_types[:]
             # if multiple input_types, only use first one for DFS
             if type(input_types[0]) == list:
-                input_types = input_types[0]
-            input_types.append(Id)
-
+                input_types = input_types[0][:]
+            id_is_input = Id in input_types
```
rwedge: I think we can simplify this whole for loop considerably now that we find available `Id` features directly.  
* We don't need to add an `Id` variable to the `input_types` to find possible `groupby` features anymore, since we already have a list of `Id` features: `id_matches`. 
* This removes the need for logic to handle whether the `input_types` has an `Id` feature or not.  
* Since we aren't adding an `Id` feature to `input_types` anymore, we don't need to split out the groupby variable using `matching_input[:-1]` / `matching_input[-1]`.

Issue #736, opened 2019-09-10
Remove Time Remaining Estimate


kmax12:
We currently give a time remaining estimate if there is more than one chunk of instances to calculate. And we always give a percent complete and time elapsed. 

Based on our experience it is hard to give a good estimate. When it works it doesn't provide much more info that percent complete and time elapsed and when it doesn't work, it can be more confusing than help.

I propose we remove the time remaining from the progress bar print out

Issue #705, opened 2019-08-13
Allow user to provide progress callback to dfs and calculate_feature_matrix


kmax12:
Right now, a user can only get calculation progress by setting `verbose=True` and look at the output. It'd be great if a user could provide a function that would be called as progress is made

for example,

```python
def progress_callback(update, progress_percent, time_remaining):
    """
    update: change in progress since last call
    progress_percent: percentage of total computation completed
    time_remaining: featuretools estimate of time remain. None if no estimate
    """
    pass

ft.calculate_feature_matrix(..., progress_callback=progress_callback)
```

The exact parameters to the callback should be discussed before implementing, but some options are proposed above.

grayskripko:
I have thought recently about the lack of callbacks in `dfs()` that could solve a wide range of problems like https://github.com/Featuretools/featuretools/issues/656. For example, xgboost have a callback argument with a small predefined [set of ready ones](https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/callback.py).

kmax12:
@grayskripko thanks for the pointer to xgboost. do you know of other machine learning libraries with similar functionality we could look at for further inspiration?

grayskripko:
not actually. The initial google query gave [keras callbacks](https://www.google.com/search?q=machine+learning+callback&oq=machine+learning+callback&aqs=chrome..69i57j0.4726j0j7&sourceid=chrome&ie=UTF-8)

grayskripko:
@rwedge why did you close the issue? Did you add callbacks? Did you decide not to do that?

angela97lin:
We added a `progress_callback` parameter to `dfs(...)` and `calculate_feature_matrix(...)` so users can provide a callback function with specified parameters that is called as incremental progress is made.

grayskripko:
yes, I see. I misunderstood the point of this issue.
Why didn't you use a flexible xgboost-like approach as I mentioned in my most first comment here? I'm about the predefined callbacks?

rwedge:
I think there was some confusion about the scope of this issue. I opened a new issue (#773) discussing pre-defined callbacks, let's continue the discussion there.

PR #739, opened 2019-09-13
Adding progress_callback parameter to dfs() and calculate_feature_matrix(), removing time remaining

angela97lin:
Addresses #705 and #736 by adding `progress_callback` parameter to `dfs()` and `calculate_feature_matrix(`) and removing time remaining field from the progress bar in `verbose=True` mode.

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -238,10 +238,10 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
     chunk_size = _handle_chunk_size(chunk_size, cutoff_time.shape[0])
     tqdm_options = {'total': (cutoff_time.shape[0] / .95),  # make total 5% higher to allot time for wrapping up at end
                     'smoothing': .05,  # arbitrary selection close to 0, which would be no smoothing
-                    'bar_format': PBAR_FORMAT
-                    }
+                    'bar_format': PBAR_FORMAT}
     if progress_callback is not None:
         if not verbose:
+            # allows us to utilize progress_bar updates without printing to anywhere
             tqdm_options.update({'file': open(os.devnull, 'w')})
     else:
         tqdm_options.update({'disable': (not verbose)})
@@ -285,16 +285,17 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
     if save_progress and os.path.exists(os.path.join(save_progress, 'temp')):
         shutil.rmtree(os.path.join(save_progress, 'temp'))
 
-    # force to 100% since we saved last 5 percent
     if progress_callback is not None:
         update = (progress_bar.total - progress_bar.n) / progress_bar.total * 100
         progress_percent = 100.0
         time_elapsed = progress_bar.format_dict["elapsed"]
         progress_callback(update, progress_percent, time_elapsed)
-
+    
+    # force to 100% since we saved last 5 percent
     progress_bar.update(progress_bar.total - progress_bar.n)
     progress_bar.refresh()
     progress_bar.close()
+    
     return feature_matrix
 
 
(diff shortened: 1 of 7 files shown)


featuretools/synthesis/dfs.py
```diff
@@ -159,6 +160,13 @@ def dfs(entities=None,
                 Numeric, Discrete, and Boolean. If given as
                 the string 'all', use all available variable types.
 
+        progress_callback (callable): function to be called with incremental progress updates.
```
kmax12: can we somehow include the details in the doc string that the percentages will be represented values between 0 and 100? (versus 0 to 1)

featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, entityset, feature_set, time_last=None,
         # total number of features (including dependencies) to be calculate
         self.num_features = len(feature_set.features_by_name.values())
 
-    def run(self, instance_ids, progress_callback=None):
+    def run(self, instance_ids, update_progress_callback=None):
```
kmax12: do we need to update the name of the argument? I say this because I'm not sure the rename doesn't increase clarity and we don't have a naming conflict internal to this function

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -230,14 +238,16 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
...
-
-    # make total 5% higher to allot time for wrapping up at end
-    progress_bar = make_tqdm_iterator(
-        total=cutoff_time.shape[0] / .95,
-        smoothing=.05,  # arbitrary selection close to 0, which would be no smoothing
-        bar_format=PBAR_FORMAT,
-        disable=(not verbose)
-    )
+    tqdm_options = {'total': (cutoff_time.shape[0] / .95),  # make total 5% higher to allot time for wrapping up at end
+                    'smoothing': .05,  # arbitrary selection close to 0, which would be no smoothing
```
kmax12: I believe we can get rid of smoothing now that we don't show time remaining. that was only used to crete the estimate

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -230,14 +238,16 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
...
+                    'bar_format': PBAR_FORMAT}
+    if progress_callback is not None:
```
kmax12: would it be cleaner if we set `disable: True` by default in `tqdm_options` above? Then the logic here could just be 
```
if verbose:
    tqdm_options.update({'disable': False})
elif progress_callback is not None:
    # allows us to utilize progress_bar updates without printing to anywhere
    tqdm_options.update({'file': open(os.devnull, 'w')})
```

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -557,15 +573,22 @@ def parallel_calculate_chunks(cutoff_time, chunk_size, feature_set, approximate,
...
         feature_matrix = []
         iterator = as_completed(_chunks).batches()
         for batch in iterator:
             results = client.gather(batch)
             for result in results:
                 feature_matrix.append(result)
+                previous_progress = progress_bar.n
                 progress_bar.update(result.shape[0])
+                if progress_callback is not None:
+                    update = (progress_bar.n - previous_progress) / progress_bar.total * 100
```
kmax12: these 4 lines are repeated from above and contain a fair amount of logic. can we refactor to have the logic in only one place?

featuretools/computational_backends/calculate_feature_matrix.py
angela97lin: Updated! (I edited the suggestion slightly since above logic would disable `progress_bar` even if `progress_callback is not None`.

Issue #655, opened 2019-07-10
Specify primitives to apply per column or per entity


philipus:
#### Bug/Feature Request Description

I would like to have optionally an API for DFS to set the trans_primitives or agg_primitives on column level instead of dataframe level because.

Two benefits you gain

1) control the features
2) save resources / run time

#### Expected Input

agg_primitives = {'quantity': [sum, mean], 'category': [last, mode]}

instead of

agg_primitives = ['sum','mean','last','mode']

ayushpatidar:
@kmax12 
if someone is not working on this can i work on this functionality.

kmax12:
@ayushpatidar no one is working on this at the moment. The first step is to finalize API and the behavior. Would you like to put together a proposal?

ayushpatidar:
@kmax12 
sure max.
how should we get started?

kmax12:
I started a shared markdown document [here](https://hackmd.io/s/BkoF8Dq-S). Would you like to start filling out the API and adding other design considerations? 

if you would like to discuss more, please join our [Slack](https://join.slack.com/t/featuretools/shared_invite/enQtNTEwODEzOTEwMjg4LTZiZDdkYjZhZTVkMmVmZDIxNWZiNTVjNDQxYmZkMzI5NGRlOTg5YjcwYmJiNWE2YjIzZmFkMjc1NDZkNjBhZTQ) and message me in the #development channel.

PR #748, opened 2019-09-17
add ability to specify primitives per entity or variable

frances-h:
Adds the ability to specify which entities or columns a primitive should be applied to (or ignore) as requested in #655


docs/source/guides/specifying_primitive_options.rst
```diff
@@ -0,0 +1,75 @@
...
+However, both of these options can be overridden by individual primitive options in the ``primitive_options`` parameter.
+
+Specifying for Individual Primitives
+************************************
+Options for individual primitives or groups of primitives are set by the ``primitive_options`` parameter of DFS. This parameter
+maps any desired options to specific primitives. In the case of conflicting options, options set at this level will override
+options set at the entire DFS run level, and the include options will always take priority over any ignore options.
+
+Specifying Entities for Individual Primitives
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
kmax12: is it possible to to turn this into runnable examples like we? that way when we build the documentation, we would get notified if the examples ever break. it also makes it easier for a reader to see what is going on without running the code themselves

docs/source/guides/specifying_primitive_options.rst
```diff
@@ -0,0 +1,75 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+GroupBy Transform Primitives can also have additional options ``include_groupby_entities``, ``ignore_groupby_entities``,
+``include_groupby_variables``, and ``ignore_groupby_variables``. These options are used to specify entities and columns
+to include/ignore to use to group inputs. By default, DFS only groups by ID columns. Specifying ``include_groupby_variables``
+overrides this default, and will only group by variables given. On the other hand, ``ignore_groupby_variables`` will
+continue to use the ID columns, ignoring any variables specified that are also ID columns.
+
+
+Specifying for each Input for Multiple Input Primitives
+*******************************************************
```
kmax12: can we add an example of this?

docs/source/guides/specifying_primitive_options.rst
```diff
@@ -0,0 +1,75 @@
...
+    feature_matrix, features_list = ft.dfs(entityset=es,
+                                           agg_primitives=['count', 'mode']
+                                           primitive_options={
+                                                'mean': {'ignore_variables': {'transactions': ['amount']}},
+                                                'mode': {'include_variables': {'sessions': ['device']}}
+                                           })
+
+
+
+Specifying GroupBy Options
```
kmax12: can we add an example of this?

docs/source/frequently_asked_questions.ipynb
```diff
@@ -1558,6 +1558,57 @@
     "There are now features where time unit is different from the default of seconds, such as `TIME_SINCE_LAST(sessions.session_start, unit=days)`, and `TIME_SINCE_FIRST(sessions.session_start, unit=years)`."
    ]
   },
+  {
+   "cell_type": "markdown",
```
kmax12: I think this example might fit better under DFS section since it is more about calling the `ft.dfs` function

docs/source/frequently_asked_questions.ipynb
```diff
@@ -1558,6 +1558,57 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "es = ft.demo.load_mock_customer(return_entityset=True)\n",
+    "\n",
+    "feature_matrix, feature_defs = ft.dfs(entityset=es,\n",
```
kmax12: i see this warning when I build the docs running this code. is it intentional? 

Screen Shot 2019-09-26 at 1 29 53 PM

kmax12: if it is intentional, any idea why it's not doing the string formating with the variable and entity?

docs/source/index.rst
```diff
@@ -165,6 +165,7 @@ Table of contents
    guides/performance
    guides/parallel
    guides/deployment
+   guides/specifying_primitive_options
```
kmax12: let's actually add this under tuning_dfs. i think it's a better to be a bit higher in this list

featuretools/primitives/options_utils.py
```diff
@@ -0,0 +1,228 @@
...
+    elif 'ignore_groupby_variables' in options:
+        def groupby_filter(f):
+            return passes_ignore_groupby_filter(f)
+    else:
+        def groupby_filter(f):
+            return (isinstance(f, IdentityFeature))
+    return groupby_filter
+
+
+def ignore_entity(options, entity, groupby=False):
```
kmax12: I wonder if there is a better than for this method. perhaps `ignore_entity_for_primitive`, `ignore_entity_for_options`?

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -500,6 +552,8 @@ def _build_transform_features(self, all_features, entity, max_depth=0,
                                                         input_types,
                                                         trans_prim,
                                                         require_direct_input=require_direct_input)
+            matching_inputs = filter_matches_by_options(matching_inputs,
```
kmax12: is it possible to move the call to `filter_matches_by_options` into the `self._get_matching_inputs`? I think that would be more logical since the caller of `_get_matching_inputs` would like to assume it is already only getting valid matches

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -518,25 +575,35 @@ def _build_transform_features(self, all_features, entity, max_depth=0,
...
+                default_type = variable_types.PandasTypes._all
+            else:
+                default_type = set([Id])
+            groupby_matches = self._features_by_type(all_features=all_features,
+                                                     entity=entity,
+                                                     max_depth=new_max_depth,
+                                                     variable_type=default_type)
+            # Convert groupby matches to tuples to make them consistent with inputs
+            groupby_matches = [(groupby_match, ) for groupby_match in groupby_matches]
+            groupby_matches = filter_matches_by_options(groupby_matches, current_options, groupby=True)
```
kmax12: i think it makes more sense to have two methods `filter_groupby_matches_by_options` and `filter_matches_by_options`.

Knowing to use the `groupby` parameter is no different than knowing 2 different methods exist. Also, then we can move the `[(groupby_match, ) for groupby_match in groupby_matches]` into the function rather than outside it.

featuretools/primitives/options_utils.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def dict_to_list_variable_check(option, es):
+    if not (isinstance(option, dict) and
+            all([isinstance(option_val, list) for option_val in option.values()])):
+        return False
+    else:
+        for entity, variables in option.items():
+            if entity not in es:
+                warnings.warn("Entity \'%s\' not in entityset" % (entity))
```
kmax12: do you need to use the the escape `\`here? doesn't using double quotes on the outside and singel quotes on the side get around that?

docs/source/frequently_asked_questions.ipynb
```diff
@@ -620,6 +620,66 @@
...
+    "                                                         })\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "**For a more examples of specifying options for DFS, please visit:**\n",
+    "\n",
+    "- [Specifying Primitive Options](https://docs.featuretools.com/guides/specifying_primitive_options.html)"
```
kmax12: does it work to make this just a relative link? `/guides/specifying_primitive_options.html`?

featuretools/primitives/options_utils.py
```diff
@@ -0,0 +1,234 @@
...
+                                     options,
+                                     groupby=True)
+
+
+def filter_matches_by_options(matches, options, groupby=False):
+    generator = _groupby_filter_generator if groupby else _variable_filter_generator
+    if len(options) > 1:
+        def is_valid_match(match):
+            variable_filter = [generator(option) for option in options]
+            if all([variable_filter[i](match[i]) for i in range(len(variable_filter))]):
```
kmax12: would it be cleaner to use zip here? 

`all([vf(m) for vf, m in zip(variable_filter, match))`

featuretools/primitives/options_utils.py
```diff
@@ -0,0 +1,234 @@
...
+
+def filter_groupby_matches_by_options(groupby_matches, options):
+    return filter_matches_by_options([(groupby_match, ) for groupby_match in groupby_matches],
+                                     options,
+                                     groupby=True)
+
+
+def filter_matches_by_options(matches, options, groupby=False):
+    generator = _groupby_filter_generator if groupby else _variable_filter_generator
+    if len(options) > 1:
```
kmax12: might be helpful to just leave a comment this handling the case where a primitive takes in multiple inputs, assuing that is what is going on

docs/source/frequently_asked_questions.ipynb
frances-h: It does. I was following the same format as the "Improve Computational Performance" link in the "How Do I Speed Up Runtime of DFS?" section, but can change it if the relative link is preferred.

kmax12: ah, we should probably make them relative there. we also host docs at http://featuretools.featurelabs.com for example

PR #749, opened 2019-09-17
Added fix and testcase for normalizing with DatetimeTimeIndex entry

### LINKED ISSUE #740: normalize_entity errors when copying time index variable to new, non-time-indexed entity
Using `copy_variables` to copy the time index column of the base entity to the new entity errors now due to how we addressed #497.

Reproducible example:

```python

es.normalize_entity(base_entity_id="customers",
                    new_entity_id="cancel_reason",
                    index="cancel_reason",
                    make_time_index=False,
                    copy_variables=["signup_date"])
```

```python-trace
ValueError: DatetimeTimeIndex variable signup_date must be set using time_index parameter
```

We should probably change the variable type from DatetimeTimeIndex to Datetime when it is copied over (same for NumericTimeIndex / Numeric)

christopherbunn:
Fixes #740 by setting all DatetimeTimeIndex variables in `additional_variable` and `copy_variables` to Datetime

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -598,6 +598,9 @@ class EntitySet(object):
         elif make_time_index:
             # Create a new time index based on the base entity time index.
             base_time_index = base_entity.time_index
+            t_index = base_entity[base_time_index]
+            if not isinstance(t_index, (vtypes.NumericTimeIndex, vtypes.DatetimeTimeIndex)):
+                raise TypeError("{0} is not a NumericTimeIndex or DatetimeTimeIndex, but type {1}".format(base_time_index, type(t_index)))
             if new_entity_time_index is None:
                 new_entity_time_index = "first_%s_time" % (base_entity.id)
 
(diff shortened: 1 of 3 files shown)


featuretools/entityset/entityset.py
```diff
@@ -577,7 +577,10 @@ def normalize_entity(self, base_entity_id, new_entity_id, index,
         transfer_types = {}
         transfer_types[index] = type(base_entity[index])
         for v in additional_variables + copy_variables:
-            transfer_types[v] = type(base_entity[v])
+            if type(base_entity[v]) == vtypes.DatetimeTimeIndex:
```
rwedge: we should also handle and copying `NumericTimeIndex` variable needing to convert to `Numeric` case

featuretools/entityset/entityset.py
```diff
@@ -593,6 +598,9 @@ def normalize_entity(self, base_entity_id, new_entity_id, index,
         elif make_time_index:
             # Create a new time index based on the base entity time index.
             base_time_index = base_entity.time_index
+            t_index = base_entity[base_time_index]
+            if not isinstance(t_index, (vtypes.NumericTimeIndex, vtypes.DatetimeTimeIndex)):
+                raise TypeError("{0} is not a NumericTimeIndex or DatetimeTimeIndex, but type {1}".format(base_time_index, type(t_index)))
```
rwedge: Let's start the error message with "Time index {0}" so there's more context about the error.  And at the end of the error message we can add a suggestion for fixing it by running `set_time_index` on the entity.

I think we should make this a sanity check we run any time `normalize_entity` is called since if this gets triggered the entity is definitely misconfigured

PR #754, opened 2019-09-25
Fix missing groupby features

frances-h:
Fix issue in #733 that removed some groupby features

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -11,9 +11,8 @@ Changelog
     * Fixes
         * Fixed entity set deserialization (:pr:`720`)
         * Added error message when DateTimeIndex is a variable but not set as the time_index (:pr:`723`)
-        * Fixed CumCount and other group-by transform primitives that take ID as input (:pr:`733`)
+        * Fixed CumCount and other group-by transform primitives that take ID as input (:pr:`733`, :pr:`754`)
         * Fix progress bar undercounting (:pr:`743`)
-        * Fixed issue where some group-by features weren't created (:pr:`754`)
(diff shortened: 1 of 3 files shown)


rwedge:
Can you add a regression test?

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -517,21 +517,29 @@ def _build_transform_features(self, all_features, entity, max_depth=0,
...
-            for matching_input in matching_inputs:
+            # If require_direct_input, require a DirectFeature in input or as a
+            # groupby, and don't create features of inputs/groupbys which are
+            # all direct features with the same relationship path
+            matches_with_groupby = set()
+            for id in id_matches:
+                for match in matching_inputs:
+                    if not require_direct_input or (
+                            not _all_direct_and_same_path(match + (id,)) and
+                            any([isinstance(m, DirectFeature) for m in (match + (id,))])):
```
rwedge: I think we instead creating another matches variable we could handle the require_direct_input check and creating a feature in the same loop and just move to the next iteration if the check fails

docs/source/changelog.rst
```diff
@@ -13,6 +13,7 @@ Changelog
         * Added error message when DateTimeIndex is a variable but not set as the time_index (:pr:`723`)
         * Fixed CumCount and other group-by transform primitives that take ID as input (:pr:`733`)
         * Fix progress bar undercounting (:pr:`743`)
+        * Fixed issue where some group-by features weren't created (:pr:`754`)
```
rwedge: Since this is fixing an unreleased bug we can add it as a second pr number on as the same line as pr 733

PR #772, opened 2019-10-14
Entropy aggregation primitive

twdobson:
### Add entropy aggregation primitive

For further details into how entropy is calculated see: https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.entropy.html

### DIFF
--- a/featuretools/primitives/options_utils.py
+++ b/featuretools/primitives/options_utils.py
@@ -80,7 +80,7 @@ def generate_all_primitive_options(all_primitives,
         else:
             # no user specified options, just use global defaults
             primitive_options[primitive] = [{'ignore_entities': ignore_entities,
-                                            'ignore_variables': ignore_variables}]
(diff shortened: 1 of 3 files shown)


twdobson:
Still need to document the addition. Require guidance as to where additions need to be made

rwedge:
Hi @twdobson, thanks for submitting a PR!

Some things to do:

* Right now the commits have a generic github icon next to them instead of your user icon. This is most likely the email on the commit does not match an email connected github account.  This will confuse the CLA assistant.

* For your question about documenting the docs page should build pages for each primitive based on their docstrings (see https://docs.featuretools.com/api_reference.html#aggregation-primitives).  You could try building the docs locally if you wanted to confirm that this happened.

rwedge:
Another thing we could do is add a 'base' parameter to the primitive to allow users to modify the base value used in the underlying scipy function

PR #779, opened 2019-10-18
Entropy aggregation primitive

twdobson:
### Entropy aggregation primitive

For details on entropy please see:
https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.entropy.html

### DIFF
--- a/featuretools/primitives/options_utils.py
+++ b/featuretools/primitives/options_utils.py
@@ -80,7 +80,7 @@ def generate_all_primitive_options(all_primitives,
         else:
             # no user specified options, just use global defaults
             primitive_options[primitive] = [{'ignore_entities': ignore_entities,
-                                            'ignore_variables': ignore_variables}]
+                                             'ignore_variables': ignore_variables}]
     return primitive_options, global_ignore_entities
 
 
@@ -143,7 +143,7 @@ def _variable_filter_generator(options):
     def passes_include_filter(f):
         return (not isinstance(f, IdentityFeature) or
                 (f.entity.id in options['include_variables'] and
-                f.variable.id in options['include_variables'][f.entity.id]))
+                 f.variable.id in options['include_variables'][f.entity.id]))
 
     if 'include_variables' in options:
         def variable_filter(f):
@@ -161,7 +161,7 @@ def _groupby_filter_generator(options):
     def passes_include_groupby_filter(f):
         return (isinstance(f, IdentityFeature) and
                 (f.entity.id in options['include_groupby_variables'] and
-                f.variable.id in options['include_groupby_variables'][f.entity.id]))
+                 f.variable.id in options['include_groupby_variables'][f.entity.id]))
 
     def passes_ignore_groupby_filter(f):
         return (isinstance(f, IdentityFeature) and (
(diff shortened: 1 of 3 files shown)


twdobson:
Hi @rwedge, had issues with assigning github account to my commits with my previous PR, so I redid the PR.

Previously when running 'make html' the Entropy aggregation primitive was not automatically generating documentation. Is there any extra step that I am missing? In addition, currently I am getting the following error when I run 'make html' now:

rwedge:
My mistake, you should add entropy to the Aggregation Primitives section of `docs/source/api_reference.rst` so it gets included in the docs

That docs error is strange, does you environment warn you when running this code:

```
import featuretools as ft
import pandas as pd
from sklearn.pipeline import Pipeline
from sklearn.ensemble import ExtraTreesClassifier
```

twdobson:
There were some issues with the packages in the environment. Have a dedicated featuretools env now, so that should be sorted. 

Apologies for the silly questions, still getting the hang of the contributing and getting everything setup. I am getting the following error when trying to make the docs: 
"
Configuration error:
There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "/usr/local/Caskroom/miniconda/base/envs/featuretools_development/lib/python3.7/site-packages/sphinx/config.py", line 361, in eval_config_file
    execfile_(filename, namespace)
  File "/usr/local/Caskroom/miniconda/base/envs/featuretools_development/lib/python3.7/site-packages/sphinx/util/pycompat.py", line 86, in execfile_
    exec(code, _globals)
  File "/Users/timothydobson/work/repos/featuretools_dev/docs/source/conf.py", line 87, in <module>
    version = featuretools.__version__
AttributeError: module 'featuretools' has no attribute '__version__'
"

twdobson:
I also fail the "featuretools/tests/cli_tests/test_cli.py " test locally - however, when I install featuretools into my env this issue (and configuration error) disappear but then the documentation does not build properly. I think it's because everything is imported from the version installed in the env and not from the source code i'm working off.

rwedge:
If you run 

```
pip install -e /path/to/featuretools/repo
```

featuretools will get installed in "editable" mode and changes you make to the source code will be reflected when you import featuretools again

twdobson:
@rwedge, thanks for the help - editable mode did the trick.

twdobson:
Hey @kmax12, @rwedge, I think the PR is ready for review! Thanks for all the help, let me know of any other changes.

twdobson:
Would the oldest version of scipy that has entropy, with the required parameters, work?

Alternatively, we could implement the function via numpy, the formula is simple: Entropy = -sum(pk * log(pk), axis=0), where pk = probability of categorical, which is the output from value_counts(normalize =True)

rwedge:
@twodobson it looks like scipy 0.11.0 is when the `base` parameter was added to the entropy function but our minimum scikit-learn version requires scipy 0.13.3 or higher so that may be a more practical requirement

twdobson:
@rwedge, sounds good. I have updated the PR to reflect this.

Quick two questions:
1. how did you go about figuring the which version to use? Did you use pipdeptree?
2. Once new enhancements are made for feature generation, is there a process for testing uplift or importance across different response sets?

twdobson:
Let me know if there is anything else that needs to be updated.

Issue #784, opened 2019-10-24
TypeError: Object of type 'int64' is not JSON serializable


RomaKoks:
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-232-e948c25a1cb0> in <module>
----> 1 es.to_parquet('incidents_es_parquet')

~/anaconda3/envs/eq-fault/lib/python3.6/site-packages/featuretools/entityset/entityset.py in to_parquet(self, path, engine, compression, profile_name)
    168                 profile_name (str) : Name of AWS profile to use, False to use an anonymous profile, or None.
    169         '''
--> 170         serialize.write_data_description(self, path, format='parquet', engine=engine, compression=compression, profile_name=profile_name)
    171         return self
    172 

~/anaconda3/envs/eq-fault/lib/python3.6/site-packages/featuretools/entityset/serialize.py in write_data_description(entityset, path, profile_name, **kwargs)
    152                 raise
    153 
--> 154         dump_data_description(entityset, path, **kwargs)
    155 
    156 

~/anaconda3/envs/eq-fault/lib/python3.6/site-packages/featuretools/entityset/serialize.py in dump_data_description(entityset, path, **kwargs)
    162     file = os.path.join(path, 'data_description.json')
    163     with open(file, 'w') as file:
--> 164         json.dump(description, file)
    165 
    166 

~/anaconda3/envs/eq-fault/lib/python3.6/json/__init__.py in dump(obj, fp, skipkeys, ensure_ascii, check_circular, allow_nan, cls, indent, separators, default, sort_keys, **kw)
    177     # could accelerate with writelines in some versions of Python, at
    178     # a debuggability cost
--> 179     for chunk in iterable:
    180         fp.write(chunk)
    181 

~/anaconda3/envs/eq-fault/lib/python3.6/json/encoder.py in _iterencode(o, _current_indent_level)
    428             yield from _iterencode_list(o, _current_indent_level)
    429         elif isinstance(o, dict):
--> 430             yield from _iterencode_dict(o, _current_indent_level)
    431         else:
    432             if markers is not None:

~/anaconda3/envs/eq-fault/lib/python3.6/json/encoder.py in _iterencode_dict(dct, _current_indent_level)
    402                 else:
    403                     chunks = _iterencode(value, _current_indent_level)
--> 404                 yield from chunks
    405         if newline_indent is not None:
    406             _current_indent_level -= 1

[...]

PR #798, opened 2019-11-05
Fix serialization after adding interesting values

frances-h:
Resolves issue #784 and adds a deep equality check to variables to verify `interesting_values` match


featuretools/entityset/serialize.py
```diff
@@ -140,14 +141,28 @@ def write_data_description(entityset, path, profile_name=None, **kwargs):
         dump_data_description(entityset, path, **kwargs)
 
 
+class NpJsonEncoder(json.JSONEncoder):
```
rwedge: It looks like the issue is that `interesting_values` is a list with numpy values in it.  What if instead of creating this custom encoder, we force `interesting_values` to be a pandas series?

Users can set interesting_values manually so we'd need to coerce to a series if they passed a different type.

featuretools/entityset/deserialize.py
```diff
@@ -36,7 +36,8 @@ def description_to_variable(description, entity=None):
         kwargs = {} if is_type_string else description['type']
         variable = variable(description['id'], entity, **kwargs)
         variable.interesting_values = description['properties']['interesting_values']
-    return variable
+        return variable
+    return (variable, description['properties']['interesting_values'])
```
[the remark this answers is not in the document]
frances-h: Makes sense. Before, interesting values were being lost and it didn't seem to make sense to loop through all the variables again when we didn't need to because we did it here.

featuretools/entityset/deserialize.py
```diff
@@ -51,14 +58,25 @@ def description_to_entity(description, entityset, path=None):
...
         index=description.get('index'),
         time_index=description.get('time_index'),
         secondary_time_index=description['properties'].get('secondary_time_index'),
-        variable_types=variable_types)
+        variable_types={variable: variable_types[variable][0] for variable in variable_types})
+    for variable in es[description['id']].variables:
+        interesting_values = variable_types[variable.id][1]['properties']['interesting_values']
+        try:
+            interesting_values = pd.read_json(interesting_values, typ="series")
+        except ValueError:
```
rwedge: It looks like this code block gets hit only when trying to load old entitysets that have non-series type interesting values

If we increment the metadata version to no longer support loading those in, we won't need to handle this case

featuretools/tests/primitive_tests/test_feature_serialization.py
```diff
@@ -17,8 +17,8 @@
 BUCKET_NAME = "test-bucket"
 WRITE_KEY_NAME = "test-key"
 TEST_S3_URL = "s3://{}/{}".format(BUCKET_NAME, WRITE_KEY_NAME)
-S3_URL = "s3://featuretools-static/test_feature_serialization_1.0.0"
```
rwedge: What happened when trying to load features that used an older entityset schema?

frances-h: It fails when it tries to deserialize the entityset from the features dictionary

featuretools/tests/primitive_tests/test_feature_serialization.py
```diff
@@ -17,8 +17,8 @@
...
-URL = "https://featuretools-static.s3.amazonaws.com/test_feature_serialization_1.0.0"
+S3_URL = "s3://featuretools-static/test_feature_serialization_feature_schema_3.2.0_entityset_schema_2.0.0.json"
+URL = "https://featuretools-static.s3.amazonaws.com/test_feature_serialization_feature_schema_3.2.0_entityset_schema_2.0.0.json"
```
rwedge: these urls are pretty long, maybe we should move the file name to a separate variable and construct the urls using the variable

PR #820, opened 2019-11-19
remove get_values() from test_es.py

systemshift:
### Pull Request Description
Replace deprecating get_values with to_numpy() to remove warnings when running tests

-----


featuretools/tests/entityset_tests/test_es.py
```diff
@@ -128,7 +128,7 @@ def test_query_by_id_with_time(es):
         instance_vals=[0, 1, 2, 3, 4],
         time_last=datetime(2011, 4, 9, 10, 30, 2 * 6))
 
-    assert df['id'].get_values().tolist() == [0, 1, 2]
```
rwedge: does `df['id'].tolist()` work as well?  It would be simpler

Issue #818, opened 2019-11-18
DFS does not generate error when `ignore_variables` values are specified incorrectly


thehomebrewnerd:
#### Bug Request Description

If the user specifies the values for the `ignore_variables` parameter of `ft.dfs` in the format of `dict[str -> str]` instead of the expected format of `dict[str -> list[str]]`, no error or warning is raised and DFS execution continues, but the output contains the data for the variables that were specified in the dictionary. The behavior is demonstrated in the attached notebook.

#### Expected Output
If the parameter is specified incorrectly an error should be raised, rather than failing silently.

#### Output of ``featuretools.show_info()``

<details>

Featuretools version: 0.12.0
Featuretools installation directory: /Users/nate.parsons/dev/notebooks/env/lib/python3.7/site-packages/featuretools

SYSTEM INFO
-----------
python: 3.7.4.final.0
python-bits: 64
OS: Darwin
OS-release: 18.7.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

INSTALLED VERSIONS
------------------
numpy: 1.17.2
pandas: 0.25.1
tqdm: 4.36.1
PyYAML: 5.1.2
cloudpickle: 1.2.2
dask: 2.5.2
distributed: 2.5.2
psutil: 5.6.3
Click: 7.0
scikit-learn: 0.21.3
pip: 19.2.3
setuptools: 40.8.0

</details>

[DFS_ignore_variables_incorrect_param_type.ipynb.zip](https://github.com/FeatureLabs/featuretools/files/3861029/DFS_ignore_variables_incorrect_param_type.ipynb.zip)

systemshift:
I can take a look at this.

PR #826, opened 2019-11-22
raise error for wrong input in ignore_variables

systemshift:
### Pull Request Description
As mentioned in #818, ignore_variables does not raise error when given a wrong input.
Currently it checks if the input is `{str: list}` only, would like to know if its needed to check the values of the list to be string or not.

-----


featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -165,6 +165,9 @@ def __init__(self,
 
         self.ignore_variables = defaultdict(set)
         if ignore_variables is not None:
+            # check if ignore_variables is not {str: list}
+            if not all(isinstance(i, str) for i in ignore_variables.keys()) or not all(isinstance(i, list) for i in ignore_variables.values()):
+                raise TypeError('ignore_variables should be dict[str -> list[str]]')
```
rwedge: Could you add a test case that triggers this error?

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -162,6 +162,18 @@ def test_ignores_variables(es):
         assert 'value' not in variables
 
 
+@pytest.fixture
```
rwedge: This shouldn't be a pytest fixture.  Fixtures are functions that get used by the test cases.  For example, this test case uses the `es` fixture, which returns an entityset.

systemshift:
Now it would be a good time to know if I should add the third check to see if the list is made of exclusively strings, and not say, ints or floats.

rwedge:
That would be good to check as well.

I think only the new "fixes" changelog entry is necessary.     The "testing changes" section is for PRs that only change the tests.

rwedge:
we should also test for the case where not all keys are string type

PR #834, opened 2019-12-05
Fix multi-output features not created when there is no child data

jeffzi:
### Fix multi-ouput features not created when there is no child data

When there is no child data, `calculate_feature_matrix` raises a `KeyError` because multi-output features are not created. The expected behaviour is to have those features represented by columns filled by numpy.nan, as it is the case with regular features.

Here is minimal reproducible example:

```python
import numpy as np
import pandas as pd

import featuretools as ft
from featuretools.primitives import NMostCommon

parent_df = pd.DataFrame({"id": [1]})
child_df = pd.DataFrame({"id": [1, 2, 3],
                         "parent_id": [1, 1, 1],
                         "time_index": pd.date_range(start='1/1/2018', periods=3),
                         "cat": ['a', 'a', 'b']})

es = ft.EntitySet(id="blah")
es.entity_from_dataframe(entity_id="parent", dataframe=parent_df, index="id")
es.entity_from_dataframe(entity_id="child", dataframe=child_df, index="id", time_index="time_index")
es.add_relationship(ft.Relationship(es["parent"]["id"], es["child"]["parent_id"]))

n_most_common = ft.Feature(es["child"]['cat'], parent_entity=es["parent"], primitive=NMostCommon)

# cutoff time before all rows
# We expect N_MOST_COMMON features to be np.nan
ft.calculate_feature_matrix(entityset=es, 
                            features=[n_most_common],
                            cutoff_time=pd.Timestamp("12/31/2017"))
#> [ ... ]
#> KeyError: "None of [Index(['N_MOST_COMMON(child.cat)[0]', 'N_MOST_COMMON(child.cat)[1]',\n       'N_MOST_COMMON(child.cat)[2]'],\n      dtype='object')] are in the [columns]"
```

<sup>Created on 2019-12-05 by the [reprexpy package](https://github.com/crew102/reprexpy)</sup>

This PR fixes the bug and adds a test for multi-output feature in `test_empty_child_dataframe`


docs/source/changelog.rst
```diff
@@ -6,14 +6,15 @@ Changelog
     * Enhancements
     * Fixes
         * Raise error when given wrong input for ignore_variables (:pr:`826`)
+        * Fix multi-ouput features not created when there is no child data (:pr:`#834`)
```
rwedge: the '#' character should be removed

jeffzi: My bad, I fixed it.

rwedge:
I think the PR looks good, once the PR fixing the issue with sklearn and the tests goes through I think this will be good to go.

PR #835, opened 2019-12-05
Move sklearn wrapper to separate library

rwedge:
There's a bug (scikit-learn/scikit-learn/issues/15795) in the 0.22 release of scikit-learn that is breaking one of our test cases.  Telling pip to ignore specifically the 0.22 release to avoid this bug.


requirements.txt
```diff
@@ -8,4 +8,4 @@ distributed>=1.24.2
 dask>=1.1.0
 psutil>=5.4.8
 click>=7.0.0
-scikit-learn>=0.20.0
+scikit-learn>=0.20.0,!=0.22
```
kmax12: is sklearn only used by the `DFSWrapper`? if so, maybe we should make it an optional dependency since that functionality is "beta" anyways

rwedge: Yeah it's only used in that wrapper

docs/source/api_reference.rst
```diff
@@ -26,20 +26,6 @@ Deep Feature Synthesis
 
     dfs
 
-
```
kmax12: i think we can keep this in like we do for the nlp primitives. we just need to mention how to install

https://docs.featuretools.com/en/stable/api_reference.html#natural-language-processing-primitives

PR #850, opened 2020-02-20
Entry Point Warnings

### LINKED ISSUE #841: Add warning if entry point plugin fails to load
There is a 'featuretools_plugin' [entrypoint](https://packaging.python.org/specifications/entry-points/) that allows other libraries to be imported as if they were part of featuretools.  For example, functions from the autonormalize library can be imported from the `featuretools.autonormalize` module location.

When Featuretools is imported, it sets up these entrypoints.  However, if an exception occurs while loading an entrypoint, the exception is silenced.  As an example, if pip failed to install the other library cleanly due to conflicting dependency versions, the entrypoint would not work.

While reporting the full stack trace of the exception could be too much info, raising a warning like "Failed to load featuretools plugin from autonormalize library" would communicate that something is wrong.

jeff-hernandez:
Plugins can fail to load silently at entry points. This PR closes #841 by warning users when plugins do fail to load.


featuretools/__init__.py
```diff
@@ -31,4 +31,4 @@
     try:
         sys.modules["featuretools." + entry_point.name] = entry_point.load()
     except Exception:
-        pass
+        print("Failed to load featuretools plugin from %s library" % entry_point.name)
```
rwedge: This should be a warning, not a print statement

setup.cfg
```diff
@@ -1,7 +1,7 @@
 [metadata]
 description-file = README.md
 [tool:pytest]
-addopts = --doctest-modules
+addopts = --doctest-modules --ignore=featuretools/tests/plugin_tests/featuretools_plugin
```
jeff-hernandez: During the test collection in pytest, an error is raised if the test package (`featuretools_plugin`) is in the path.

featuretools/tests/plugin_tests/test_plugin.py
```diff
@@ -0,0 +1,16 @@
...
+    import_featuretools,
+    install_featuretools_plugin,
+    reinstall_pandas,
+    uninstall_featuretools_plugin
+)
+
+
+def test_plugin_warning():
+    install_featuretools_plugin()
+    reinstall_pandas()
```
rwedge: While package version compatibility is an actual way entrypoints have failed in the past, I think we can simplify this test if we avoid the reinstalling packages and instead have the featuretools_plugin package raise an error in it's __init__ .py file so it always errors.  That way we don't need to reinstall pandas

setup.cfg
rwedge: Can we similarly have coverage ignore the test package?

jeff-hernandez: Yes, updated config in code coverage.

docs/source/changelog.rst
```diff
@@ -7,11 +7,12 @@ Changelog
         * Fix a connection closed error when using n_jobs (:pr:`853`)
     * Changes
         * Pin msgpack dependency for Python 3.5; remove dataframe from Dask dependency (:pr:`851`)
+        * Added Plugin Warnings at Entry Point (:pr:`850`)
     * Documentation Changes
         * Update link to help documentation page in Github issue template (:pr:`855`)
 
     Thanks to the following people for contributing to this release:
-    :user:`frances-h`, :user:`rwedge`
+    :user:`frances-h`, :user:`rwedge`, :user:`jeff-hernandez`
```
rwedge: The changelog entry will need to be moved to a new "Future Release" section

featuretools/__init__.py
```diff
@@ -31,4 +35,6 @@
     try:
         sys.modules["featuretools." + entry_point.name] = entry_point.load()
     except Exception:
-        pass
+        message = "Featuretools failed to load plugin {} from library {}"
+        logger.info(message.format(entry_point.name, entry_point.module_name))
```
rwedge: I think this should be a warning level log entry

jeff-hernandez:
Yes, will add that test case.

featuretools/__init__.py
```diff
@@ -35,6 +35,7 @@
     try:
         sys.modules["featuretools." + entry_point.name] = entry_point.load()
     except Exception:
-        message = "Featuretools failed to load plugin {} from library {}"
+        message = "Featuretools failed to load plugin {} from library {}. "
+        message += "For more information, set logging to debug."
```
rwedge: Let's change the beginning of the second line to "For a full stack trace, . . ."

PR #851, opened 2020-02-24
Pin msgpack for Python 3.5

frances-h:
Dask `distributed` runs into issues with the `msgpack` 1.0.0 release on Python 3.5. Dask no longer supports Python 3.5, so their fix for the `msgpack` update does not get applied to Python 3.5. This pins `msgpack` to before the 1.0.0 release on Python 3.5 to avoid the issue. 

Dask `distributed` required dataframe dependencies for a check where the full dependency was not needed. Dask has updated, so the additional dataframe dependencies are no longer required for `distributed` to work correctly with Featuretools

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -2,6 +2,13 @@
 
 Changelog
 ---------
+**Future Release**
+    * Changes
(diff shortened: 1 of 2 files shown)


rwedge:
Can you add this to the changelog?

PR #869, opened 2020-03-27
Add pip to test-requirements.txt

frances-h:
Adds `pip` to `test-requirements.txt` for the plugin warning tests


test-requirements.txt
```diff
@@ -1,3 +1,4 @@
+pip
```
rwedge: let's do pip>=20.0.2

PR #871, opened 2020-04-02
Use logger.warning instead of logger.warn

gsheni:
### Pull Request Description
- Use logger.warning instead of logger.warn since `warn` is [deprecated](https://github.com/python/cpython/blob/master/Lib/logging/handlers.py#L771). In addition, we are already using [logger.warning](https://github.com/FeatureLabs/featuretools/blob/bd5973ff099c8183a885155562e770581aab3350/featuretools/entityset/entity.py#L512)


rwedge:
Cool.  Can you put the readme changes in a separate PR?

gsheni:
> Cool. Can you put the readme changes in a separate PR?

Fixed

Issue #888, opened 2020-04-13
Compose LabelTimes don't work with Dask implementation


kmax12:
@frances-h to add more details

frances-h:
Done, waiting on https://github.com/FeatureLabs/compose/issues/117 to merge into main Dask feature branch

PR #909, opened 2020-04-15
Compose compatability for Dask

### LINKED ISSUE #883: Update dask predict next purchase demo to use Compose to make label times
It's possible Compose will require changes to its code to support using Dask dataframes / series

frances-h:
- uses Compose to generate labels in `test_instacart.py`
- forces LabelTimes to Pandas DataFrame for usage with Dask 

Resolves #883 and #888

### DIFF
--- a/featuretools/primitives/standard/aggregation_primitives.py
+++ b/featuretools/primitives/standard/aggregation_primitives.py
@@ -289,8 +289,8 @@ class PercentTrue(AggregationPrimitive):
             def format_chunk(x):
                 return x[:].fillna(0)
 
-            return (s.agg(lambda x: format_chunk(x).sum()),
-                    s.agg(lambda x: len(format_chunk(x))))
+            return (s.agg(lambda x: format_chunk(x).sum()).astype('float64'),
+                    s.agg(lambda x: len(format_chunk(x))).astype('int64'))
 
         def agg(val, length):
             return (val.sum(), length.sum())
--- a/featuretools/primitives/standard/aggregation_primitives.py
+++ b/featuretools/primitives/standard/aggregation_primitives.py
@@ -249,7 +249,10 @@ class NumTrue(AggregationPrimitive):
 
     def get_dask_aggregation(self):
         def chunk(s):
-            return s.agg(np.sum)
+            chunk_sum = s.agg(np.sum)
+            if chunk_sum.dtype == 'bool':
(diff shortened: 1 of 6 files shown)


test-requirements.txt
```diff
@@ -7,3 +7,4 @@ graphviz>=0.8.4
 moto>=1.3.13
 smart-open>=1.8.4
 boto3>=1.10.45
+composeml>=0.1.8
```
rwedge, line 10: Will this work with the current release of composeml or does it require the additions from https://github.com/FeatureLabs/compose/pull/121?

frances-h: it does require https://github.com/FeatureLabs/compose/pull/121, but no tests actually depend on it so I've removed the requirement

kmax12: does that mean we should add a test?

rwedge: it gets tested by the notebook but there are no unit tests associated with it.  we could add tests for both pandas and dask approaches passing in compose label times as the cutoff dataframe?

kmax12: i think we should have a unit test

kmax12:
is this something we should file an issue on with dask? i'd think it should support the LabelTimes object since it is basically a pandas dataframe

featuretools/primitives/standard/aggregation_primitives.py
```diff
@@ -289,8 +289,8 @@ def chunk(s):
             def format_chunk(x):
                 return x[:].fillna(0)
 
-            return (s.agg(lambda x: format_chunk(x).sum()),
-                    s.agg(lambda x: len(format_chunk(x))))
+            return (s.agg(lambda x: format_chunk(x).sum()).astype('float64'),
+                    s.agg(lambda x: len(format_chunk(x))).astype('int64'))
```
frances-h, line 293: Ran into an issue where if the sum or the length was either 1 or 0 Pandas would convert the result to a boolean, which would cause errors down the line.

test-requirements.txt
```diff
@@ -7,3 +7,4 @@ graphviz>=0.8.4
 moto>=1.3.13
 smart-open>=1.8.4
 boto3>=1.10.45
+composeml>0.1.8
```
rwedge, line 10: let's use `>=0.2.0` for composeml

PR #910, opened 2020-04-15
Fix hyperlinks in docs and docstrings with updated address

gsheni:
### Pull Request Description
- The ecosystem is incorrect and doesn't currently work. This PR fixes that. 
- There are additional links which have the same problem, and this PR addresses that.


rwedge:
Can you do a sweep for any other links that references specific pages on docs.featuretools.com and are missing the `/en/stable/` part of the path?

gsheni:
> Can you do a sweep of the repo for any other links that references specific pages on docs.featuretools.com and are missing the `/en/stable/` part of the path?

There was 2 more in the docstrings that I fixed.

rwedge:
I think there's another link in `docs/source/templates/layout.html` that needs to be fixed

gsheni:
> I think there's another link in `docs/source/templates/layout.html` that needs to be fixed

Fixed. 

I thought changing the link would not work since it was a static image. So I skipped it on my review. But the new link worked.

PR #914, opened 2020-04-17
Update set_time_index code path for Dask dataframes and impacted test

### LINKED ISSUE #913: Dask Entity.set_time_index doesn't change dtype
The Dask implementation of `Entity.set_time_index` currently skips changing the dtype of the new time index to either a `DatetimeTimeIndex` or a `NumericTimeIndex`, resulting in dtype errors when trying to use the entity's time_index.

rwedge:
Fixes #913 

This PR updates `Entity.set_time_index` so that Dask dataframes do not exit the function early.  Specific points in the code where costly validation checks occur have different logic for Dask frames.

This is necessary to fix the `test_build_es_from_scratch_and_run_dfs` test.
For that test, I also changed the logic to get the two dataframe indexes to match and removed two primitives that have the `uses_full_entity` attribute.  Features using these primitives were returning different values, which I think is happening because the dask entity data is not sorted

### DIFF
--- a/featuretools/entityset/entity.py
+++ b/featuretools/entityset/entity.py
@@ -404,16 +404,21 @@ class Entity(object):
     def set_time_index(self, variable_id, already_sorted=False):
         # check time type
         if isinstance(self.df, dd.core.DataFrame):
-            self.time_index = variable_id
-            self.entityset.time_type = self.variable_types[variable_id]
-            return
-
-        if len(self.df) == 0:
-            time_to_check = vtypes.DEFAULT_DTYPE_VALUES[self[variable_id]._default_pandas_dtype]
+            already_sorted = True  # skip sorting
+            old_vtype = self.variable_types[variable_id]
+            if old_vtype == vtypes.Datetime:
+                time_type = vtypes.DatetimeTimeIndex
+            elif old_vtype in (vtypes.Numeric, vtypes.Ordinal):
+                time_type = vtypes.NumericTimeIndex
+            else:
+                time_type = old_vtype
         else:
-            time_to_check = self.df[variable_id].head(1).iloc[0]
+            if len(self.df) == 0:
+                time_to_check = vtypes.DEFAULT_DTYPE_VALUES[self[variable_id]._default_pandas_dtype]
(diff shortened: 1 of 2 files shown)


featuretools/entityset/entity.py
```diff
@@ -404,16 +404,21 @@ def delete_variables(self, variable_ids):
...
+            old_vtype = self.variable_types[variable_id]
+            if old_vtype == vtypes.Datetime:
+                time_type = vtypes.DatetimeTimeIndex
+            elif old_vtype in (vtypes.Numeric, vtypes.Ordinal):
+                time_type = vtypes.NumericTimeIndex
+            else:
+                time_type = old_vtype
         else:
-            time_to_check = self.df[variable_id].head(1).iloc[0]
+            if len(self.df) == 0:
```
rwedge, line 416: I think we may be able to simplify further by changing this line to 

```python
if isinstance(self.df, dd.core.DataFrame) or len(self.df) == 0:
```
and removing the conditional logic about `old_vtype`

thehomebrewnerd: If you do that, won't you always force a `compute()` to get the length if `self.df` is a Dask dataframe?

rwedge: I think if the first statement of the or is true, the second is not evaluated

https://docs.python.org/3/reference/expressions.html#boolean-operations

frances-h: Not sure if I'm understanding correctly, but we could get into trouble calling .head on a dask dataframe since it by default grabs the first partition which could be empty. Since we're requiring vtypes when creating an entity, using the vtype provided to figure out the index type seems like the safer route.

rwedge: I'm proposing the logic look like this

```python
        if isinstance(self.df, dd.core.DataFrame) or len(self.df) == 0:
            time_to_check = vtypes.DEFAULT_DTYPE_VALUES[self[variable_id]._default_pandas_dtype]
        else:
            time_to_check = self.df[variable_id].head(1).iloc[0]
```

PR #920, opened 2020-04-20
Update dask tests

frances-h:
Update failing Dask entityset tests
- remove unsupported primitives from tests
- don't check name when comparing Pandas Series

Note that `test_aggregation` sporadically fails due to ambiguity in `NMostCommon`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,7 +79,7 @@ jobs:
           command: |
               C:\Users\circleci\Miniconda3\shell\condabin\conda-hook.ps1
               conda activate featuretools
-              pytest featuretools\
+              pytest featuretools\ --ignore=featuretools\dask-tests-tmp\
   install_ft:
     working_directory: ~/featuretools
     parameters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,7 +79,7 @@ jobs:
           command: |
               C:\Users\circleci\Miniconda3\shell\condabin\conda-hook.ps1
               conda activate featuretools
-              pytest featuretools\ --ignore=featuretools\dask-tests-tmp\
+              pytest featuretools\ -n 2
   install_ft:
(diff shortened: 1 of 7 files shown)


featuretools/tests/utils_tests/test_cli_utils.py
```diff
@@ -28,7 +28,7 @@ def test_installed_packages():
     requirements = ["pandas", "numpy", "tqdm",
                     "PyYAML", "cloudpickle",
                     "dask", "distributed", "psutil",
-                    "Click"]
+                    "click"]
```
rwedge, line 31: this should already be fixed on master, if we merge master in it will be resolved

.circleci/config.yml
```diff
@@ -79,7 +79,7 @@ jobs:
           command: |
               C:\Users\circleci\Miniconda3\shell\condabin\conda-hook.ps1
               conda activate featuretools
-              pytest featuretools\ -n 2
+              pytest featuretools\ -n 2 --ignore=featuretools\dask-tests-tmp\
```
rwedge, line 82: Let's add this to the default pytest options in `setup.cfg`.  The reason the linux tests don't currently require this ignore flag is a side effect of the linux tests packaging the library before running the tests.  The folder isn't being included in the packaged library, so it's not necessary to ignore it.  Since the windows tests don't package the library, they do need to ignore the folder.  Since this can also be a problem when running the tests locally, let's make pytest ignore the folder by default.

rwedge, line 82: Let's add this to the default pytest options in `setup.cfg`.

The reason the linux tests don't currently require this ignore flag is a side effect of the linux tests packaging the library before running the tests.  The folder isn't being included in the packaged library, so it's not necessary to ignore it.  Since the windows tests don't package the library, they do need to ignore the folder.

This can also be a problem when running the tests locally, let's make pytest ignore the folder by default.

PR #930, opened 2020-04-28
Remove Overlap in Training Windows

### LINKED ISSUE #918: Lookahead(cheating) issue
I will quote my stack overflow question and answer: https://stackoverflow.com/a/61261300/3595632.
The problem is that feature generated by `dfs()` count the same value that `cutoff_time` Dataframe already count:  When Jeff calculating `lt`, he already counted the amount of "2012-03-01" of store 0 for the aggregated amount of `cutoff_time = "2012-03-01"(store 0)` (and I think this makes sense). So, 1 month amount aggregation period: 2012-03-01 ~ 2012-03-31, **March**, 2012.
Let's see `SUM(sales.amount)` column value of (`store_id = 0` & `2012-03-01` cutoff time). The value is 44 by adding up the amounts of 2012-02-05, 2012-02-10, and """2012-03-01"""!!!. It counts """2012-03-01""" here again, which is a sort of cheating and look-ahead bias.

2012-03-01 is March, not February. We must predict 1 month total amount of March, using ONLY February information.... 
 

#### Expected Output
I think that `SUM(sales.amount)` column value of (`store_id = 0` & `2012-03-01` cutoff time) must be 16 + 15 = 31, excluding 2012-03-01 amount.

jeff-hernandez:
This fixes #918 by making the left endpoint of a training window exclusive. This will avoid overlap in consecutive training windows.


docs/source/changelog.rst
```diff
@@ -18,7 +19,12 @@ Changelog
     * Testing Changes
 
     Thanks to the following people for contributing to this release:
-    :user:`gsheni`, :user:`rwedge`
+    :user:`gsheni`, :user:`rwedge`, :user:`jeff-hernandez`
+
+**Breaking Changes**
+
+* Using training windows in feature calculations can result in different values than previous versions.
+  This is a result from removing an overlap in consecutive training windows by making the left endpoint exclusive.
```
rwedge, line 27: Adding an example scenario might help explain this as well

rightx2:
Please review my pull request

PR #931, opened 2020-04-28
Error if dask dataframe used for cutoff_time

frances-h:
- Error if Dask dataframe used as `cutoff_time` in  `calculate_feature_matrix`
- Compute Dask dataframe with warning if passed as `cutoff_time` in DFS


featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -83,6 +84,14 @@ def test_calc_feature_matrix(es):
                                  instance_ids=range(17),
                                  cutoff_time=17)
 
+    error_text = "cannot use Dask DataFrame for cutoff_time: "\
```
rwedge, line 87: let's make this a separate test case so it runs independently of the other checks in this larger test case

Issue #792, opened 2019-10-30
Improve custom naming for multi-output groupby features


thehomebrewnerd:
### Improve custom naming for multi-output groupyby features
In PR #780, the ability to specify custom feature names for multi-output custom primitives was added. With this change, it is not currently possible to completely override the feature names for a multi-output groupby feature. Custom naming for a multi-output primitive relies on overriding the `get_names` method of the underlying primitive, but the names for groupby features comes partially from the primitives `get_names` method and partially from the get_names` implementation on the `GroupbyTransformFeature` object. 

The current implementation appends text to the end of the names returned from the primitive `get_names` method, which specifies what the output was grouped by. It is not possible to customize or remove this appended text at the end of the feature name. 

This implementation should be updated to allow for the entire feature name to be custom, including the text appended at the end.

thehomebrewnerd:
Closing this issue as it doesn't seem to be an enhancement that is needed. If other users need this functionality in the future, this issue can be reopened or a new issue can be created.

PR #934, opened 2020-04-29
Update get_names output

### LINKED ISSUE #874: get_names() and get_name() do not return the same value if num_of_output features is 1
#### Bug Request Description
- `get_names` and `get_name` are two functions we use to get the feature name (corresponding to the 
column names in the feature matrix).
- If the `number_output_features` is 1 for a feature, they should return the same thing (a list with 1 item, that is the same as the `get_name`. 
- Related [issue](792) **But not the same**

#### Current Behavior
```python
import featuretools as ft
_, feature_defs = ft.dfs(entityset=es,
                                      target_entity="customers",
                                      agg_primitives=[],
                                      trans_primitives=["month"],
                                      max_depth=1)
print(feature_defs[1].get_name())
print(feature_defs[1].get_names())
```

##### Current Output
```python
 'MONTH(join_date)'
['MONTH(join_date)[0]']
```
#### Expected Output
```python
 'MONTH(join_date)'
['MONTH(join_date)']
```

thehomebrewnerd:
### Update Feature.get_names() Output

Update output of `Feature.get_names()` so that if the feature creates only a single output feature, the single name in the list returned by `Feature.get_names()` matches the name returned by `Feature.get_name()`

Fixes #874

### DIFF
--- a/featuretools/primitives/base/aggregation_primitive_base.py
+++ b/featuretools/primitives/base/aggregation_primitive_base.py
@@ -33,6 +33,8 @@ class AggregationPrimitive(PrimitiveBase):
                                        parent_entity_id,
                                        where_str,
                                        use_prev_str)
+        if n == 1:
+            return [base_name]
         return [base_name + "[%s]" % i for i in range(n)]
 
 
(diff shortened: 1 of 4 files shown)


thehomebrewnerd:
Closing this PR while we review possible refactoring of the various Feature naming methods that are currently implemented to address issues raised in #874 and #792

PR #935, opened 2020-04-30
Fix #930

rightx2:
### Pull Request Description
Fix #930 . There might be a misunderstanding.

### DIFF
--- a/featuretools/entityset/entity.py
+++ b/featuretools/entityset/entity.py
@@ -500,13 +500,13 @@ class Entity(object):
         """
         if self.time_index:
             if time_last is not None and not df.empty:
-                df = df[df[self.time_index] <= time_last]
+                df = df[df[self.time_index] < time_last]
                 if training_window is not None:
                     training_window = _check_timedelta(training_window)
-                    mask = df[self.time_index] > time_last - training_window
+                    mask = df[self.time_index] >= time_last - training_window
                     if self.last_time_index is not None:
                         lti_slice = self.last_time_index.reindex(df.index)
-                        lti_mask = lti_slice > time_last - training_window
+                        lti_mask = lti_slice >= time_last - training_window
                         mask = mask | lti_mask
                     else:
                         logger.warning(
--- a/featuretools/tests/computational_backend/test_calculate_feature_matrix.py
+++ b/featuretools/tests/computational_backend/test_calculate_feature_matrix.py
@@ -308,7 +308,7 @@ def test_training_window(es):
                                               es,
                                               cutoff_time=cutoff_time,
                                               training_window='2 hours')
-    prop_values = [4, 5, 1]
+    prop_values = [5, 5, 2]
     dagg_values = [3, 2, 1]
     assert (feature_matrix[property_feature.get_name()] == prop_values).values.all()
     assert (feature_matrix[dagg.get_name()] == dagg_values).values.all()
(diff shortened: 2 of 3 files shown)


rightx2:
@jeff-hernandez  Hi, Thanks for trying fixing this up. However, you are missing a point. The feature value on the EXACT cutoff time point should not be included when calculating feature generation for that cutoff time. Let's go back to the issue that I raised: 

> So, 1 month amount aggregation period: 2012-03-01 ~ 2012-03-31, March, 2012. Let's see SUM(sales.amount) column value of (store_id = 0 & 2012-03-01 cutoff time). The value is 44 by adding up the amounts of 2012-02-05, 2012-02-10, and """2012-03-01"""!!!. It counts """2012-03-01"""  here again, which is a sort of cheating and look-ahead bias` 

As I said here, "2012-03-01" should be not counted when calculating features for the cutoff time of "2012-03-01".  But your pull request doesn't solve this problem. Therefore, I've fixed it with a new pull request

jeff-hernandez:
Hi @rightx2, thanks for reaching out. Featuretools always includes the feature value [on the exact cutoff time](https://docs.featuretools.com/en/stable/automated_feature_engineering/handling_time.html#what-is-the-cutoff-time). If you want to calculate a one month training window, you can adjust the cutoff times to be the last day of the month instead of the first day of the following month and use a `MonthEnd` training window.


```python
from pandas.tseries.offsets import MonthEnd

cutoff_time - MonthEnd()
```

```
id
0   2012-02-29
1   2012-03-31
2   2012-02-29
3   2012-03-31
4   2012-02-29
5   2012-03-31
6   2012-02-29
7   2012-03-31
Name: cutoff_time, dtype: datetime64[ns]
```

rightx2:
@jeff-hernandez  Hmm. I don't understand why do you guys make it "includes the feature value on the exact cutoff time"... It doesn't look logically make sense to me...

1. Let say I have to predict 1month revenue for each store. Then I have to predict the revenue from 2020-03-01 00:00:00 ~ 2020-03-31 23:59:59, using the data from the past ~ 2020-02-29 23:59:59 (This time split is not overlapped), right?

2. But your instruction is the way of predicting "2020-03-01 00:00:00 ~ 2020-03-31 23:59:59" using "past ~ 2020-03-01 00:00:00" feature data, which seems not logically making sense because it counts the time of future and make time overlapped.

3. Moreover, if I followed your instruction, using `MonthEnd()`, I would miss the feature data of "2020-02-29 15:23" or "2020-02-29 22:23", because the cutoff time is "2020-02-29", which is 00:00:00.

4. Consequently, what I have to do is `cutoff_time - pd.Timedelta(seconds=1)` to handle above situation, which is so bothering and confusing for the beginner.

Am I missing something? Could you tell me the situation that your instruction, logic works well more generally?

jeff-hernandez:
@rightx2 thanks for sharing your feedback. After considering this, we do think it makes sense to let users choose whether data at exact cutoff times should be included in feature calculations. We plan to provide this option as a parameter in DFS. You can find more information about this parameter at #942. Let me know if you have additional feedback. Thanks.

rightx2:
@jeff-hernandez That would be nice :)  Is there any way I can contribute?

rightx2:
I'd like to participate in solving this issue as well

jeff-hernandez:
@rightx2 yes, any contributions are helpful and greatly appreciated. [This page](https://github.com/FeatureLabs/featuretools/blob/master/contributing.md) has helpful information and instructions for contributing. You can participate to solve this issue as well. Would you like to submit a pull request for #942?

rightx2:
@jeff-hernandez  Ok, I will try

jeff-hernandez:
Okay, sounds good. I will update #942 with any details about the implementation or test cases as they come up.

rwedge:
Closing in favor of other pr #959

PR #943, opened 2020-05-05
Restore len() call for Pandas in EntitySets.add_relationships

### LINKED ISSUE #903: Review usage of len() call in Entityset.add_relationship()
A `len()` call is used in `Entityset.add_relationship()` in an `if` statement to check if the child dataframe is empty. In the current dask implementation this code is commented out, and this did not cause any existing tests to fail. This code should be reviewed and updated if this check is needed.

frances-h:
Resolves #903 

The `len` call in question covers an edge case relating to adding a relationship with an empty child dataframe. The default dtype of an empty Pandas DataFrame is `object`, so adding a relationship where they dtype of the linking variable is not `object` fails.

We are not supporting empty entity dataframes with Dask, so this check can be safely skipped.

### DIFF
--- a/featuretools/tests/entityset_tests/test_es.py
+++ b/featuretools/tests/entityset_tests/test_es.py
@@ -123,6 +123,7 @@ def test_add_relationship_empty_child_convert_dtype(es):
     relationship = ft.Relationship(es["sessions"]["id"], es["log"]["session_id"])
     es['log'].df = pd.DataFrame(columns=es['log'].df.columns)
     assert len(es['log'].df) == 0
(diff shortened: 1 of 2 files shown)


featuretools/tests/entityset_tests/test_es.py
```diff
@@ -119,6 +119,18 @@ def test_add_relationship_errors_on_dtype_mismatch(es):
         es.add_relationship(mismatch)
 
 
+def test_add_relationship_empty_child_convert_dtype(es):
```
rwedge, line 122: can this test also have an assert to confirm that the colum data type was not `int64` to begin with?

PR #945, opened 2020-05-07
Add functions to get default primitives

gsheni:
### Pull Request Description
- Add `get_default_aggregation_primitives` and `get_default_transform_primitives` to primitives.utils
- Have DFS run these functions to get the default primitives. In addition, users can call these functions to get the list of primitives.
- Rename test file from `test_list_primitives.py` to `test_primitives_utils.py` since the file tests functions in the utils of primitives


docs/source/changelog.rst
```diff
@@ -2,6 +2,15 @@
 
 Changelog
 ---------
+**Future Release**
+    * Enhancements
+    * Fixes
+    * Changes
+        * Add `get_default_aggregation_primitives` and `get_default_transform_primitives` (:pr:`945`)
```
rwedge, line 9: I'd classify this as an enhancement

gsheni:
> What do you think of adding these to the list of `ft.primitives.utils` functions callable from `ft.primitives` (currently `get_aggregation_primitives`, `get_transformation_primitives`, and `list_primitives` are like that)

Yep, sounds good. Added it to the `api.py`

Issue #942, opened 2020-05-04
Parameterize Inclusive Cutoff Times


jeff-hernandez:
Data at cutoff times are currently included in feature calculations. We can make this a parameter to provide users the option to exclude the data.

#### Software Design
The parameter `include_cutoff_time` would be added in `dfs` and `calculate_feature_matrix`  to control whether data at cutoff times are included in feature calculations.

```python
ft.dfs(
    ...
    include_cutoff_time=True,
)
```

When using cutoff times with training windows, the training window start time (left endpoint) should be excluded from feature calculations when the cutoff time (right endpoint) is included. This will avoid overlap from consecutive training windows. Similarly, the training window start time should be included when the cutoff time is excluded.

#### Test Cases

There should be test cases that check feature calculations when:

- A training window is not used and cutoff times are inclusive.
- A training window is not used and cutoff times are exclusive.
- A training window is used and cutoff times are inclusive.
- A training window is used and cutoff times are exclusive.

rwedge:
Fixed in #959

PR #958, opened 2020-05-09
Add `include_cutoff_time` arg 

rightx2:
Add `include_cutoff_time` arg to control whether data at cutoff times are included in feature calculations and prevent `traininig_window` overlapping

### Pull Request Description
There was a data overlapping problem when calculating the feature matrix: The data at cutoff time is used both in calculating features and in calculating target values(#918 ). This could be data cheating and affect the result as well. There was a trial to solve the issue (#930 ), but It still didn't solve the cheating problem. So, we decided to parameterize it to control whether data at cutoff times are included in feature calculations or not(#942 ) and this PR solves it.


rightx2:
I will try with other pr (Missing update origin/master before sending PR)

PR #959, opened 2020-05-09
Add include_cutoff_time arg to control whether data at cutoff times a…

rightx2:
Add `include_cutoff_time` arg to control whether data at cutoff times are included in feature calculations and prevent `traininig_window` overlapping

### Pull Request Description
There was a data overlapping problem when calculating the feature matrix: The data at cutoff time might be used both in calculating features and in calculating target values(#918 ). This could cause data cheating and affect the result as well. There was a trial to solve the issue (#930 ), but It still didn't solve the cheating problem. So, we decided to parameterize it to control whether data at cutoff times are included in feature calculations or not(#942 ) and this PR solves it.

### DIFF
--- a/docs/source/automated_feature_engineering/handling_time.rst
+++ b/docs/source/automated_feature_engineering/handling_time.rst
@@ -135,20 +135,48 @@ Featuretools can automatically add last time indexes to every :class:`Entity` in
 Excluding data at cutoff times
 -----------------------------------------------
 
-There are some situations where data is right just on the cutoff time. For example, let say you have to predict one month revenue for each store using sales data. One of the them is the revenue from ``2020-01-01`` to ``2020-01-31`` and there are bunch of sale history data before that time, including the one occured at ``2020-01-01 00:00:00``. You might want to include(or exclude) the data in feature calculation for this cutoff time. This can be controlled by using the ``include_cutoff_time`` parameter to :func:`featuretools.dfs` or :func:`featuretools.calculate_feature_matrix`::
+If you don't want to use the data at cutoff times in feature calculation, you can exclude them by setting ``include_cutoff_time`` to ``False``. Then, it will not use the data from these time points in time or any data newer than that. If you set it to ``True``, it will use the data from the older(or within the ``training_window`` if you set) and the one from the cutoff time as well:
 
-    fm = ft.calculate_feature_matrix(features=features,
-                                     entityset=es_transactions,
-                                     cutoff_time=ct_transactions,
-                                     include_cutoff_time=True)
+.. ipython:: python
 
-    fm, features = ft.dfs(entityset=es,
-                          target_entity='customers',
-                          cutoff_time=pd.Timestamp("2014-1-1 04:00"),
-                          instance_ids=[1,2,3],
-                          cutoff_time_in_index=True,
-                          include_cutoff_time=False)
+    from featuretools.primitives import Count
 
+    es.add_last_time_indexes()
+
+    es['transactions'].df.head()
+    es['sessions'].df.head(2)
+
+    count_log = ft.Feature(
+        base=es['transactions']['transaction_id'],
+        parent_entity=es['sessions'],
+        primitive=Count,
+    )
+    cutoff_time = pd.DataFrame({
+        'session_id': [1, 1],
+        'time': ['2014-01-01 00:02:10', '2014-01-01 00:04:20'],
+    }).astype({'time': 'datetime64[ns]'})
+
+    # Case1. include_cutoff_time = True
+    actual = ft.calculate_feature_matrix(
+        features=[count_log],
+        entityset=es,
+        cutoff_time=cutoff_time,
+        cutoff_time_in_index=True,
+        training_window='3 minutes',
+        include_cutoff_time=True,
+    )
+    actual
+
+    # Case2. include_cutoff_time = False
+    actual = ft.calculate_feature_matrix(
+        features=[count_log],
+        entityset=es,
+        cutoff_time=cutoff_time,
+        cutoff_time_in_index=True,
+        training_window='3 minutes',
+        include_cutoff_time=False,
+    )
+    actual
 
 .. _approximate:
 
--- a/docs/source/automated_feature_engineering/handling_time.rst
+++ b/docs/source/automated_feature_engineering/handling_time.rst
@@ -134,46 +134,53 @@ Featuretools can automatically add last time indexes to every :class:`Entity` in
 
 Excluding data at cutoff times
 -----------------------------------------------
-
-If you don't want to use the data at cutoff times in feature calculation, you can exclude them by setting ``include_cutoff_time`` to ``False``. Then, it will not use the data from these time points in time or any data newer than that. If you set it to ``True``, it will use the data from the older(or within the ``training_window`` if you set) and the one from the cutoff time as well:
+The ``cutoff_time`` is the last point in time where data can be used for feature calculation. If you don't want to use the data at the cutoff time in feature calculation, you can exclude that data by setting ``include_cutoff_time`` to ``False`` in :func:`featuretools.dfs` or :func:`featuretools.calculate_feature_matrix`. If you set it to ``True`` (the default behavior), data from the cutoff time point will be used. If used with ``training_window``, it will work like belows:
 
 .. ipython:: python
 
-    from featuretools.primitives import Count
+    from featuretools.primitives import Sum
 
     es.add_last_time_indexes()
 
     es['transactions'].df.head()
     es['sessions'].df.head(2)
 
-    count_log = ft.Feature(
-        base=es['transactions']['transaction_id'],
+    sum_log = ft.Feature(
+        base=es['transactions']['amount'],
         parent_entity=es['sessions'],
-        primitive=Count,
+        primitive=Sum,
     )
     cutoff_time = pd.DataFrame({
         'session_id': [1, 1],
         'time': ['2014-01-01 00:02:10', '2014-01-01 00:04:20'],
     }).astype({'time': 'datetime64[ns]'})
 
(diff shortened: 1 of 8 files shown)


rightx2:
@jeff-hernandez Hi, jeff. This is my first trial. Could you please review them?

jeff-hernandez:
Hi @rightx2, thanks for the follow-up with this PR. I will provide a code review and keep you updated. Thanks!

rwedge:
I think we should also update the [Handling time](https://docs.featuretools.com/en/stable/automated_feature_engineering/handling_time.html) docs page to mention this new option

rightx2:
@jeff-hernandez I reflected your suggestions. Thanks.  @rwedge I've tried to add `include_cutoff_time`  on the Handling time section. Could you review that, please?

docs/source/automated_feature_engineering/handling_time.rst
```diff
@@ -131,6 +131,25 @@ For example, a customer's session has multiple transactions which can happen at
 
 Featuretools can automatically add last time indexes to every :class:`Entity` in an :class:`Entityset` by running ``EntitySet.add_last_time_indexes()``. If a ``last_time_index`` has been set, Featuretools will check to see if the ``last_time_index`` is after the start of the training window. That, combined with the cutoff time, allows DFS to discover which data is relevant for a given training window.
 
+
+Include data at cutoff times
```
rwedge, line 135: Since `include_cutoff_time` is True by default, let's title this `Excluding data at cutoff times`.

docs/source/automated_feature_engineering/handling_time.rst
```diff
@@ -131,6 +131,25 @@ For example, a customer's session has multiple transactions which can happen at
...
+-----------------------------------------------
+
+There are some situations where data is right just on the cutoff time. For example, let say you have to predict one month revenue for each store using sales data. One of the them is the revenue from ``2020-01-01`` to ``2020-01-31`` and there are bunch of sale history data before that time, including the one occured at ``2020-01-01 00:00:00``. You might want to include(or exclude) the data in feature calculation for this cutoff time. This can be controlled by using the ``include_cutoff_time`` parameter to :func:`featuretools.dfs` or :func:`featuretools.calculate_feature_matrix`::
```
rwedge, line 138: I think the key difference to highlight is that `include_cutoff_time=True` means "use data from this time and older" and `include_cutoff_time=False` means "do not use data from this point in time or any data newer than it"

We should also explain how this impacts training window -- I think expanding the example to show how using a training window would work in both cases would be helpful to the reader

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -121,6 +122,7 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
                 progress_percent: percentage (float between 0 and 100) of total computation completed
                 time_elapsed: total time in seconds that has elapsed since start of call
 
+        include_cutoff_time (bool): If True, data at cutoff times are included in feature calculations.
```
rwedge, line 125: I think for these docstring definitions we should mention it defaults to true

docs/source/automated_feature_engineering/handling_time.rst
rightx2: That makes more sense!

rightx2:
@rwedge I've done some corrections. Please check them out

docs/source/automated_feature_engineering/handling_time.rst
```diff
@@ -131,6 +131,53 @@ For example, a customer's session has multiple transactions which can happen at
 
 Featuretools can automatically add last time indexes to every :class:`Entity` in an :class:`Entityset` by running ``EntitySet.add_last_time_indexes()``. If a ``last_time_index`` has been set, Featuretools will check to see if the ``last_time_index`` is after the start of the training window. That, combined with the cutoff time, allows DFS to discover which data is relevant for a given training window.
 
+
+Excluding data at cutoff times
+-----------------------------------------------
+
+If you don't want to use the data at cutoff times in feature calculation, you can exclude them by setting ``include_cutoff_time`` to ``False``. Then, it will not use the data from these time points in time or any data newer than that. If you set it to ``True``, it will use the data from the older(or within the ``training_window`` if you set) and the one from the cutoff time as well:
```
rwedge, line 138: How does this sound?

"The `cutoff_time` is the last point in time where data can be used for feature calculation., but you may want to specify the first time point that cannot be used instead.  If you don't want to use the data at the cutoff time in feature calculation, you can exclude that data by setting ``include_cutoff_time`` to ``False`` in `dfs` or `calculate_feature_matrix`.  If you set it to ``True`` (the default behavior), data from the cutoff time point will be used."

I do think there should be a second paragraph on how training window works (or maybe it can be part of the examples).  Since with `include_cutoff_time=True`, the oldest point in the training window is excluded an the cutoff time point is included.  Whereas with `include_cutoff_time=False` the oldest point the window is included and the cutoff time point is excluded.  I think the example should demonstrate this as well.  If the training window was 130 seconds long transaction 308 would be on the oldest point on the window to demonstrate that behavior

If you want to have regular text in between the ipython cells you can break the code into multiple ipython blocks.  They'll still get run in order.

rwedge: Thanks for these updates 😄

rightx2:
@rwedge Thanks for reviewing my poor work!

I thought your suggestion is more clear than mine. I reflected them all.

p.s. Since I'm not damn good at English, especially writing and grammar, so feel free to correct any mistake or wrong expression, please :)

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -328,16 +341,63 @@ def test_training_window_overlap(es):
...
+        entityset=es,
+        cutoff_time=cutoff_time,
+        cutoff_time_in_index=True,
+        training_window='10 minutes',
+        include_cutoff_time=False,
+    )['COUNT(log)']
+    np.testing.assert_array_equal(actual.values, [0, 9])
+
+
+def test_overlap_without_training_window(es):
```
rwedge, line 367: let's call this test `test_include_cutoff_time`

rightx2:
@rwedge I'll try that

rwedge:
I updated the training window example a bit, you might need to pull

rightx2:
@rwedge Hi, I've tried to find a good `approximate` example for this for a couple of days, but I'm still stuck at it (have some trouble in understanding how `approximate` works). If you help me out, I would really appreciate it. Could you please?

rwedge:
Sure, @rightx2, I'll try to explain.

The gist of what approximate does:
* Divide the cutoff time dataframe into buckets based on the approximate window size
* For an instance in a bucket, don't calculate it's aggregation features using it's cutoff time
* Instead, calculate the aggregation features for EVERY instance in the bucket using the oldest time point in the bucket 
* This means any data between the oldest time point in the bucket and the actual cutoff time for an instance won't be included in the aggregation feature values
* The idea is if there is enough older data available for instances, losing the data within the bucket won't change the aggregation values too much (not always the case)
* This can save computational time by only needing to perform the filtering necessary to calculatie aggregation features once per bucket instead of filtering at every unique cutoff time point in the bucket

So for how `include_cutoff_time` would impact approximate, I think it will change whether the oldest time point in the bucket is included or excluded.

I ended up writing a test case to double check my logic:

```python
def test_approximate_dfeat_of_agg_on_target_include_cutoff_time(es):
    agg_feat = ft.Feature(es['log']['id'], parent_entity=es['sessions'], primitive=Count)
    agg_feat2 = ft.Feature(agg_feat, parent_entity=es['customers'], primitive=Sum)
    dfeat = DirectFeature(agg_feat2, es['sessions'])
    cutoff_time = pd.DataFrame({'time': [datetime(2011, 4, 9, 10, 31, 19)], 'instance_id': [0]})
    feature_matrix = calculate_feature_matrix([dfeat, agg_feat],
                                              es,
                                              approximate=Timedelta(20, 's'),
                                              cutoff_time=cutoff_time,
                                              include_cutoff_time=False)
    # log event 5 excluded due to approximate cutoff time point
    assert feature_matrix[dfeat.get_name()].tolist() == [5]
    assert feature_matrix[agg_feat.get_name()].tolist() == [5]

    feature_matrix = calculate_feature_matrix([dfeat, agg_feat],
                                              es,
                                              approximate=Timedelta(20, 's'),
                                              cutoff_time=cutoff_time,
                                              include_cutoff_time=True)
    # log event 5 included due to approximate cutoff time point
    assert feature_matrix[dfeat.get_name()].tolist() == [6]
    assert feature_matrix[agg_feat.get_name()].tolist() == [5]
```
Based on this test it looks like approximate is working with `include_cutoff_time` as I expected

rightx2:
@rwedge Seeing your explanation and example, I think I was almost there except understanding `DriectFeature`(I was struggling with `test_approximate_multiple_instances_per_cutoff_time` to understand it and I'm still). I should have spent more time. Anyway, thanks for your help!. I've added a piece of comment so that it can help someone like me in the future. Anything else to do to be merged?

rightx2:
I've solved the conflict issue in `changelog.rst` with `master` branch. Should I move the "Breaking change" part from `v0.14.0 Apr 30, 2020` to the "Future Release"?

rwedge:
We don't need to move the the Breaking Change part.  This PR will add new functionality, if users don't use `include_cutoff_time=False` their experience will not change. 

I like the extra comments in the test case, these can be hard to puzzle out what they are doing.

docs/source/changelog.rst
```diff
@@ -6,6 +6,7 @@ Changelog
     * Enhancements
         * Add ``get_default_aggregation_primitives`` and ``get_default_transform_primitives`` (:pr:`945`)
         * Allow cutoff time dataframe columns to be in any order (:pr:`969`)
+        * Add `include_cutoff_time` arg - control whether data at cutoff times are included in feature calculations (:pr:`959`)
```
rwedge, line 9: use "\`\`" here instead of "\`", to get code style formatting

PR #964, opened 2020-05-13
Dask cleanup

frances-h:
Removes unused code paths and reverts changes made for Dask that are unused or unsupported.

### DIFF
--- a/featuretools/computational_backends/feature_set_calculator.py
+++ b/featuretools/computational_backends/feature_set_calculator.py
@@ -130,11 +130,11 @@ class FeatureSetCalculator(object):
 
             if missing_ids:
                 default_df = self.generate_default_df(instance_ids=missing_ids,
-                                                    extra_columns=df.columns)
+                                                      extra_columns=df.columns)
                 df = df.append(default_df, sort=True)
 
             df.index.name = self.entityset[self.feature_set.target_eid].index
-        
+
         column_list = []
 
         # Order by instance_ids
--- a/featuretools/computational_backends/feature_set_calculator.py
+++ b/featuretools/computational_backends/feature_set_calculator.py
@@ -117,33 +117,30 @@ class FeatureSetCalculator(object):
 
         if len(df.columns) == 0:
             default_df = self.generate_default_df(instance_ids=instance_ids)
-            if isinstance(df, dd.core.DataFrame):
+            if isinstance(df, dd.DataFrame):
                 cols = [col for col in default_df.columns] + [default_df.index.name]
                 return dd.from_pandas(default_df.reset_index(), npartitions=1)[cols]
             return default_df
 
         # Fill in empty rows with default values. This only works for pandas dataframes
         # and is not currently supported for Dask dataframes.
-        if isinstance(df, dd.core.DataFrame):
-            missing_ids = []
-        else:
+        if isinstance(df, pd.DataFrame):
             index_vals = df[target_entity.index].values
             missing_ids = [i for i in instance_ids if i not in index_vals]
 
-        if missing_ids:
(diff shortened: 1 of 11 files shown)


featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -117,33 +117,30 @@ def progress_callback(*args):
...
             return default_df
 
         # Fill in empty rows with default values. This only works for pandas dataframes
         # and is not currently supported for Dask dataframes.
-        if isinstance(df, dd.core.DataFrame):
-            missing_ids = []
-        else:
+        if isinstance(df, pd.DataFrame):
             index_vals = df[target_entity.index].values
             missing_ids = [i for i in instance_ids if i not in index_vals]
```
thehomebrewnerd, line 129: Can we also revert the code for creating `missing_ids` to match `master`?

`missing_ids = [i for i in instance_ids if i not in df[target_entity.index]]`

featuretools/entityset/entity.py
```diff
@@ -447,10 +448,7 @@ def set_index(self, variable_id, unique=True):
             self.df = self.df.set_index(self.df[variable_id], drop=False)
             self.df.index.name = None
             if unique:
-                if isinstance(self.df.index, dd.core.Index):
-                    index_is_unique = self.df[variable_id].compute().is_unique
-                else:
-                    index_is_unique = self.df[variable_id].is_unique
+                index_is_unique = self.df[variable_id].is_unique
                 assert index_is_unique, "Index is not unique on dataframe (Entity {})".format(self.id)
```
thehomebrewnerd, line 452: I think we can eliminate the creation of the `index_is_unique` variable and update the assert to match `master`:

`assert self.df.index.is_unique, "Index is not unique on dataframe (Entity {})".format(self.id)`

featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -117,33 +117,32 @@ def progress_callback(*args):
 
         if len(df.columns) == 0:
             default_df = self.generate_default_df(instance_ids=instance_ids)
-            if isinstance(df, dd.core.DataFrame):
+            if isinstance(df, dd.DataFrame):
                 cols = [col for col in default_df.columns] + [default_df.index.name]
                 return dd.from_pandas(default_df.reset_index(), npartitions=1)[cols]
             return default_df
```
thehomebrewnerd, line 123: It looks like none of these lines are hit by our current tests - including the pandas code on line 119. Not sure what scenario causes this to be necessary, but wondering if the Dask code block is necessary?

frances-h: Given that we don't allow empty Dask EntitySets anyways, I have a hard time imagining we could ever hit this point (a Dask dataframe with no columns) but I wasn't sure if it was a valid edge case

featuretools/entityset/serialize.py
```diff
@@ -78,7 +78,7 @@ def write_entity_data(entity, path, format='csv', **kwargs):
         loading_info (dict) : Information on storage location and format of entity data.
     '''
     format = format.lower()
-    if isinstance(entity.df, dd.core.DataFrame):
+    if isinstance(entity.df, dd.DataFrame):
         basename = "{}-*.{}".format(entity.id, format)
```
thehomebrewnerd, line 82: This line isn't hit by our tests - do we need to add a test for this?

featuretools/computational_backends/feature_set_calculator.py
thehomebrewnerd: Seems like it could be an edge case since the pandas code is there. My inclination would be to remove at least the Dask code until we know for sure we need it with an example that we can create a test with so we would have the edge case documented going forward.

PR #967, opened 2020-05-14
Add Dask support to EqualScalar and NotEqualScalar primitives

### LINKED ISSUE #966: Support EqualScalar and NotEqualScalar Primitives with Dask
The EqualScalar primitive currently has Dask related code, but the NotEqualScalar primitive does not. Both are currently marked as unsupported by Dask. Review these primitives to determine if they can be supported using code similar to that in EqualScalar.

thehomebrewnerd:
Update `EqualScalar` and `NotEqualScalar` primitives to work properly with Dask.

Fixes #966

### DIFF
--- a/featuretools/primitives/standard/binary_transform.py
+++ b/featuretools/primitives/standard/binary_transform.py
@@ -274,6 +274,7 @@ class EqualScalar(TransformPrimitive):
     name = "equal_scalar"
     input_types = [Variable]
     return_type = Boolean
+    dask_compatible = True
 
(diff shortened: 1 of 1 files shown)


rwedge:
Could we resolve this by removing the pd.Series cast from the primitive function?

thehomebrewnerd:
> Could we resolve this by removing the pd.Series cast from the primitive functions?

As far as I can tell we can remove the casting without issue, which also allows us to remove a couple import statements. Pushing an update now.

Issue #487, opened 2019-04-10
ft.dfs with cutoff dataframe with different order of columns fails


sebrahimi1988:
I was creating features for this entity set:

![Screenshot 2019-04-10 at 17 29 59](https://user-images.githubusercontent.com/25546598/55896486-4f0df100-5bb6-11e9-9050-2f43bda2babc.png)

 with the following cutoff parameter:

![Screenshot 2019-04-10 at 17 25 24](https://user-images.githubusercontent.com/25546598/55896281-ddce3e00-5bb5-11e9-80bb-e9640e835a36.png)

and this command:
```
feature_matrix_2018, feature_defs = ft.dfs(entityset=es,
                                           target_entity='users',
                                           cutoff_time=ct,
                                           cutoff_time_in_index=False)
```
I was getting this error: 

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/miniconda3/lib/python3.7/site-packages/pandas/core/arrays/categorical.py in __init__(self, values, categories, ordered, dtype, fastpath)
    357             try:
--> 358                 codes, categories = factorize(values, sort=True)
    359             except TypeError:

~/miniconda3/lib/python3.7/site-packages/pandas/util/_decorators.py in wrapper(*args, **kwargs)
    187                     kwargs[new_arg_name] = new_arg_value
--> 188             return func(*args, **kwargs)
    189         return wrapper

~/miniconda3/lib/python3.7/site-packages/pandas/core/algorithms.py in factorize(values, sort, order, na_sentinel, size_hint)
    612                                            size_hint=size_hint,
--> 613                                            na_value=na_value)
    614 

~/miniconda3/lib/python3.7/site-packages/pandas/core/algorithms.py in _factorize_array(values, na_sentinel, size_hint, na_value)
    459     uniques, labels = table.factorize(values, na_sentinel=na_sentinel,
--> 460                                       na_value=na_value)
    461 

pandas/_libs/hashtable_class_helper.pxi in pandas._libs.hashtable.Int64HashTable.factorize()

ValueError: Buffer has wrong number of dimensions (expected 1, got 2)

During handling of the above exception, another exception occurred:

NotImplementedError                       Traceback (most recent call last)
<ipython-input-67-36cde9f8fcf4> in <module>
      2                                            target_entity='users',
      3                                            cutoff_time=ct,
----> 4                                            cutoff_time_in_index=False)

[...]

PR #549, opened 2019-05-19
As described in issue 487, FT expects cutoff_time dataframe's columns…

sebrahimi1988:
- As described in issue 487, FT expects cutoff_time dataframe's columns to be in instance_id_time, ... order. This commit addresses this issue by first moving the expected columns to the first and second position before cracking on with the rest of the code 

- A unittest has been added to check the functionality of the added code

PR #969, opened 2020-05-15
Update handling of cutoff time dataframe columns

thehomebrewnerd:
This PR builds off of #549 and serves as a replacement for that PR. This PR also adds additional functionality in the handling of the cutoff time dataframe time columns so that both the time and instance id columns can be in any order in the cutoff time dataframe as long as they are named properly.

Acceptable instance id column names: target entity index variable name or `instance_id`
Acceptable time column names: target entity time_index variable name or `time`.

Also updated the tests and documentation impacted by this change.

Note, this update will be a breaking change for users who previously depended on column ordering to identify the time column in a cutoff time dataframe.

Closes Issue #487


docs/source/changelog.rst
```diff
@@ -14,7 +15,16 @@ Changelog
         * Update testing dependencies (:pr:`976`)
 
     Thanks to the following people for contributing to this release:
-    :user:`gsheni`, :user:`rwedge`
+    :user:`gsheni`, :user:`rwedge`, :user:`thehomebrewnerd`, :user:`sebrahimi1988`
+
+**Breaking Changes**
+
+* Calls to `featuretools.dfs` or `featuretools.calculate_feature_matrix` that use a cutoff time
```
rwedge, line 22: `featuretools.dfs`, etc., should be ``featuretools.dfs``

thehomebrewnerd: Fixed

Issue #905, opened 2020-04-15
Run full unit tests with dask


kmax12:
* How do we run the test suite twice (w/ dask and pandas) without duplicated code?
* We can mark certain tests as expect to fail
* We should also be able to repeat easily as we support other types of dataframes like Koalas or cuDF

kmax12:
from @frances-h: https://docs.pytest.org/en/latest/parametrize.html

rwedge:
This been achieved in #999, albeit with xfails for testing features unsupported with dask.  Some additional cleanup is needed (like #1023) but the unit tests run with dask now.

PR #981, opened 2020-05-19
Dask entityset serialization/deserialization

### LINKED ISSUE #970: Add support for serializing/deserializing Dask EntitySets
Support for serializing a Dask EntitySet has been added but not deserializing one. Support for both should be added and functionality should be tested

frances-h:
Add support for deserializing Dask EntitySets. Resolves #970

### DIFF
--- a/featuretools/entityset/serialize.py
+++ b/featuretools/entityset/serialize.py
@@ -83,7 +83,7 @@ def write_entity_data(entity, path, format='csv', **kwargs):
         loading_info (dict) : Information on storage location and format of entity data.
     '''
     format = format.lower()
-    if isinstance(entity.df, dd.core.DataFrame):
+    if isinstance(entity.df, dd.DataFrame) and format == 'csv':
         basename = "{}-*.{}".format(entity.id, format)
     else:
         basename = '.'.join([entity.id, format])
--- a/featuretools/entityset/serialize.py
+++ b/featuretools/entityset/serialize.py
@@ -10,7 +10,7 @@ from featuretools.utils.s3_utils import get_transport_params, use_smartopen_es
 from featuretools.utils.wrangle import _is_s3, _is_url
 
 FORMATS = ['csv', 'pickle', 'parquet']
-SCHEMA_VERSION = "2.0.0"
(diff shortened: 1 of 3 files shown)


featuretools/tests/entityset_tests/test_serialization.py
```diff
@@ -139,13 +163,28 @@ def test_to_parquet(es, tmpdir):
...
 def test_to_parquet_manual_interesting_values(es, tmpdir):
     es['log']['product_id'].interesting_values = ["coke_zero"]
     es.to_pickle(str(tmpdir))
     new_es = deserialize.read_entityset(str(tmpdir))
     assert es.__eq__(new_es, deep=True)
 
 
+def test_dask_to_parquet_manual_interesting_values(dask_es, tmpdir):
+    dask_es['log']['product_id'].interesting_values = ["coke_zero"]
+    dask_es.to_pickle(str(tmpdir))
```
thehomebrewnerd, line 183: This test is using `to_pickle`. Should this be `to_parquet` based on the test name? Also appears to be an issue with `test_to_parquet_manual_interesting_values` above. For the deserialize operation below, the resulting entityset contains pandas dataframes, but the equality assertion still works.

frances-h: Fixed both tests. It's interesting that a dask/pandas entityset would be considered deeply equal, that's probably an issue that should be addressed separately.

featuretools/entityset/serialize.py
```diff
@@ -104,9 +109,11 @@ def write_entity_data(entity, path, format='csv', **kwargs):
         df.to_parquet(file, **kwargs)
     elif format == 'pickle':
         # Dask currently does not support to_pickle
-        if isinstance(df, dd.core.DataFrame):
-            df = df.compute()
-        df.to_pickle(file, **kwargs)
+        if isinstance(df, dd.DataFrame):
+            msg = 'Cannot serialize Dask EntitySet to pickle'
+            raise ValueError(msg)
```
thehomebrewnerd, line 114: We should add a test case for this, but could be done as part of this PR or as part of the other test clean up going on for issue #905

Issue #947, opened 2020-05-07
Unknown variable type should have type_string


gsheni:
#### Bug Request Description
- Currently the `find_variable_types` function returns the following
```python
{'None': featuretools.variable_types.variable.Unknown,
 'discrete': featuretools.variable_types.variable.Discrete,
 'categorical': featuretools.variable_types.variable.Categorical,
 'id': featuretools.variable_types.variable.Id,
 'zipcode': featuretools.variable_types.variable.ZIPCode,
 'country_code': featuretools.variable_types.variable.CountryCode,
 'subregion_code': featuretools.variable_types.variable.SubRegionCode,
 'ordinal': featuretools.variable_types.variable.Ordinal,
 'boolean': featuretools.variable_types.variable.Boolean,
 'numeric': featuretools.variable_types.variable.Numeric,
 'numeric_time_index': featuretools.variable_types.variable.NumericTimeIndex,
 'index': featuretools.variable_types.variable.Index,
 'datetime': featuretools.variable_types.variable.Datetime,
 'datetime_time_index': featuretools.variable_types.variable.DatetimeTimeIndex,
 'date_of_birth': featuretools.variable_types.variable.DateOfBirth,
 'time_index': featuretools.variable_types.variable.TimeIndex,
 'timedelta': featuretools.variable_types.variable.Timedelta,
 'text': featuretools.variable_types.variable.Text,
 'latlong': featuretools.variable_types.variable.LatLong,
 'ip': featuretools.variable_types.variable.IPAddress,
 'full_name': featuretools.variable_types.variable.FullName,
 'email': featuretools.variable_types.variable.EmailAddress,
 'url': featuretools.variable_types.variable.URL,
 'phone_number': featuretools.variable_types.variable.PhoneNumber,
 'filepath': featuretools.variable_types.variable.FilePath}
```
- The keys of this dictionary should all be lowercase. To fix this, we should look at changing the Unknown variable type so it has correct name, like so:
```python
class Unknown(Variable):
    type_string = "unknown"
```
- In addition we **may** want to have a clear key to Name transition. The following need to fixed EmailAddress, FilePath, IPAddress, SubRegionCode because they don't fit the majority lowercase to CamelCase transition. 

#### Expected Output
```python
{'unknown': featuretools.variable_types.variable.Unknown}
```

PR #982, opened 2020-05-19
Fixed issue #947 by editing type strings

tuethan1999:
### Pull Request Description
- Edited the type strings of the Unknown, EmailAddress, FilePath, IPAddress, SubRegionCode

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -6,6 +6,8 @@ Changelog
     * Enhancements
         * Add ``get_default_aggregation_primitives`` and ``get_default_transform_primitives`` (:pr:`945`)
     * Fixes
+        * Normalized type_strings of entity_classes so that the find_variable_types function produces a 
+          dictionary with a clear key to name transition
     * Changes
     * Documentation Changes
     * Testing Changes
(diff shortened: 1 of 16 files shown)


rwedge:
After this change, the `Variable` class is appearing in the dictionary returned`find_variable_types`.    Since the `Variable` class isn't a variable type we expect users to use as an actual variable, let's exclude it from the `find_variable_types` dictionary.  I think `ft.primitives.get_aggregation_primitives` and `ft.primitives.get_transform_primitives` do something to ensure `TransformPrimitive` and `AggregationPrimitive` don't get returned with their subclasses.

rwedge:
https://github.com/FeatureLabs/featuretools/blob/e9f12ed71799ccad9e842cd7cf170d362407f056/featuretools/entityset/deserialize.py#L30

Since we've changed the type string for `Unknown`, we should update this line.

I also think this edge case -- being unable to find a variable that matches the type string in a variable json -- is untested.  We should add a test case.

rwedge:
Let's also add a line to the breaking changes section of the changelog, detailing which Variables had their type strings changed.

docs/source/changelog.rst
```diff
@@ -26,6 +28,10 @@ Changelog
   no longer used to determine the time column. Now, both instance id columns and time columns in a cutoff time
   dataframe can be in any order as long as they are named properly.
 
+* The type_string attributes of all entity classes are now a snake case conversion of their class names. This
```
rwedge, line 31: * variable classes, not entity classes
* let's use code formatting (\`\`text\`\`) around type_string and the formal class names (Unknown, IPAddress, etc)
* Instead of talking about schema versions, in the last sentence let's say that old saved entitysets that used these variable may load incorrectly

PR #987, opened 2020-05-21
Add Age Primitive, and Unit tests for Age Primitive

### LINKED ISSUE #604: dfs doesn't use TimeSince primitives on datetime features
### Bug/Feature Request Title

TimeSince primitives don't calculate on Datetime and DateOfBirth features (but calculate on DatetimeTimeIndex). That doesn't seem right since you can have a few columns describing different timestamps you may want to use.

Actually, is there any way to check which primitives apply to which variable_types?

My featuretools package is built from master branch.

BTW show_info() doesn't provide any output so I used `featuretools.utils.cli_utils.print_info()` instead.
<details>
Featuretools version: 0.8.0
Featuretools installation directory: y:\git\featuretools\featuretools

SYSTEM INFO
-----------
python: 3.7.3.final.0
python-bits: 64
OS: Windows
OS-release: 2008ServerR2
machine: AMD64
processor: Intel64 Family 6 Model 45 Stepping 2, GenuineIntel
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

INSTALLED VERSIONS
------------------
numpy: 1.16.4
pandas: 0.24.2
tqdm: 4.32.1
toolz: 0.9.0
PyYAML: 5.1.1
cloudpickle: 1.2.1
future: 0.17.1
dask: 1.2.2
distributed: 1.28.1
psutil: 5.6.3
Click: 7.0
scikit-learn: 0.21.2
pip: 19.1.1
setuptools: 41.0.1
</details>

gsheni:
### Pull Request Description
- Add age primitive, and unit tests for age primitive. 
- Fixes #604

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -5,6 +5,7 @@ Changelog
 **Future Release**
     * Enhancements
         * Add ``get_default_aggregation_primitives`` and ``get_default_transform_primitives`` (:pr:`945`)
+        * Allow cutoff time dataframe columns to be in any order (:pr:`969`)
     * Fixes
     * Changes
     * Documentation Changes
@@ -14,7 +15,16 @@ Changelog
         * Update testing dependencies (:pr:`976`)
 
     Thanks to the following people for contributing to this release:
-    :user:`gsheni`, :user:`rwedge`
+    :user:`gsheni`, :user:`rwedge`, :user:`thehomebrewnerd`, :user:`sebrahimi1988`
+
+**Breaking Changes**
+
+* Calls to ``featuretools.dfs`` or ``featuretools.calculate_feature_matrix`` that use a cutoff time
+  dataframe, but do not label the time column with either the target entity time index variable name or
+  as ``time``, will now result in an ``AttributeError``. Previously, the time column was selected to be the first
+  column that was not the instance id column. With this update, the position of the column in the dataframe is
+  no longer used to determine the time column. Now, both instance id columns and time columns in a cutoff time
+  dataframe can be in any order as long as they are named properly.
 
 **v0.14.0 Apr 30, 2020**
     * Enhancements
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -8,15 +8,19 @@ Changelog
         * Allow cutoff time dataframe columns to be in any order (:pr:`969`)
         * Add Age primitive, and make it a default transform primitive for DFS (:pr:`987`)
     * Fixes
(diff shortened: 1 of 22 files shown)


docs/source/changelog.rst
```diff
@@ -6,6 +6,7 @@ Changelog
     * Enhancements
         * Add ``get_default_aggregation_primitives`` and ``get_default_transform_primitives`` (:pr:`945`)
         * Allow cutoff time dataframe columns to be in any order (:pr:`969`)
+        * Add Age primitive (:pr:`987`)
```
kmax12, line 9: we should make this a default primitive in DFS too

gsheni: Fixed.

gsheni: Also updated changelog

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -591,3 +592,34 @@ def generate_name(self, base_feature_names):
             name += u", unit={}".format(self.unit)
         name += u")"
         return name
+
+
+class Age(TransformPrimitive):
+    """Calculates the age in years as a floating point number given a
```
ctduffy, line 598: My one comment is that it might be helpful to have an option to have ages return in integer form as well. However, this is pretty easy to do on the end of whoever is using it, but I think that is my only potential improvement I can think of right now.

gsheni: We could add it as an optional argument. Something like this work:
```python
class Age(TransformPrimitive):
    def __init__(self, decimals= None):
        self.decimals = decimals

    def get_function(self):
        def age(x, time=None):
            values_ = (time - x).dt.days / 365
            if self.decimals is not None:
                return np.around(values_, self.decimals)
        return age
```

However, I'm not the biggest fan. The end-user could round afterwards.

ctduffy: Up to you, I agree that the user could round, but its pretty easy to add the optional argument and the optionality doesn't change the experience users who don't want to round.

kmax12: i think i'd stay away from adding this from now, otherwise we might have to add it to other primitives where people want rounding. let's get this merged in as is and we can always add it later if the need becomes strong

rwedge:
Looks like adding an new default primitive changed the expected features for our deserialize from S3 tests

gsheni:
> Looks like adding an new default primitive changed the expected features for our deserialize from S3 tests

Fixed this by having the deserialize tests use a fixed list of agg and trans primitives

PR #988, opened 2020-05-21
Passing strings to variable_types

tuethan1999:
### Pull Request Description

Allows the passing of strings into the variable types dict. 
- Old format: dict[str -> Variable]
- New format: dict[str ->Variable/str]

-----

### DIFF
--- a/featuretools/entityset/entity.py
+++ b/featuretools/entityset/entity.py
@@ -283,8 +283,10 @@ class Entity(object):
         for vid in variable_types.copy():
             vtype = variable_types[vid]
             if isinstance(vtype, str):
-                # convert form str to vtype
-                variable_types[vid] = string_to_class_map[vtype]
+                if vtype in string_to_class_map:
+                    variable_types[vid] = string_to_class_map[vtype]
+                else:
+                    variable_types[vid] = string_to_class_map['unknown']
 
         if index not in variable_types:
             variable_types[index] = vtypes.Index
--- a/featuretools/entityset/entity.py
+++ b/featuretools/entityset/entity.py
@@ -213,7 +213,6 @@ class Entity(object):
 
     def query_by_values(self, instance_vals, variable_id=None, columns=None,
                         time_last=None, training_window=None, include_cutoff_time=True):
-
         """Query instances that have variable with given value
 
         Args:
--- a/featuretools/entityset/entity.py
+++ b/featuretools/entityset/entity.py
@@ -212,7 +212,8 @@ class Entity(object):
         self.variables[self.variables.index(variable)] = new_variable
 
     def query_by_values(self, instance_vals, variable_id=None, columns=None,
-                        time_last=None, training_window=None):
+                        time_last=None, training_window=None, include_cutoff_time=True):
+
         """Query instances that have variable with given value
 
         Args:
@@ -225,6 +226,8 @@ class Entity(object):
             training_window (Timedelta, optional):
                 Window defining how much time before the cutoff time data
                 can be used when calculating features. If None, all data before cutoff time is used.
+            include_cutoff_time (bool):
+                If True, data at cutoff time are included in calculating features
 
         Returns:
             pd.DataFrame : instances that match constraints with ids in order of underlying dataframe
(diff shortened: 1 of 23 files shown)


featuretools/tests/entityset_tests/test_entity.py
```diff
@@ -158,3 +160,42 @@ def test_variable_types_unmodified():
...
+    es.entity_from_dataframe(
+        entity_id="reversed_variable_types",
+        dataframe=dataframe,
+        index="<class 'featuretools.variable_types.variable.Index'>",
+        time_index="<class 'featuretools.variable_types.variable.DatetimeTimeIndex'>",
+        variable_types=reversed_variable_types)
+    
+    entity = es["reversed_variable_types"]
+    for variable in entity.variables:
+        description = variable.to_data_description()
```
rwedge, line 198: I think we can test the variable types were set correctly more directly by checking the class of each variable in the entity without serializing the variables

featuretools/entityset/entity.py
```diff
@@ -278,6 +279,15 @@ def _create_variables(self, variable_types, index, time_index, secondary_time_in
...
         variables = []
         variable_types = variable_types.copy() or {}
+        string_to_class_map = find_variable_types()
+        for vid in variable_types.copy():
+            vtype = variable_types[vid]
+            if isinstance(vtype, str):
+                if vtype in string_to_class_map:
+                    variable_types[vid] = string_to_class_map[vtype]
+                else:
+                    variable_types[vid] = string_to_class_map['unknown']
```
rwedge, line 289: Let's add a warning here so the user knows the string wasn't recognized and `Unknown` was used instead

featuretools/tests/entityset_tests/test_entity.py
```diff
@@ -158,3 +160,62 @@ def test_variable_types_unmodified():
...
+        entity_id="reversed_variable_types",
+        dataframe=dataframe,
+        index="<class 'featuretools.variable_types.variable.Index'>",
+        time_index="<class 'featuretools.variable_types.variable.DatetimeTimeIndex'>",
+        variable_types=reversed_variable_types)
+
+    entity = es["reversed_variable_types"]
+    for variable in entity.variables:
+        variable_class = variable.__class__
+        assert variable_class == variable_types[variable_class.type_string]
```
rwedge, line 200: I think we should compare the `variable_class` to the matching variable in the `reversed_variable_types` dictionary since that was the dictionary used for `variable_types`

featuretools/tests/entityset_tests/test_entity.py
```diff
@@ -158,3 +160,62 @@ def test_variable_types_unmodified():
...
+    dataframe = pd.DataFrame(columns=list(reversed_variable_types))
+    entity = Entity('reversed_variable_types', dataframe, es,
+                    variable_types=reversed_variable_types,
+                    index="<class 'featuretools.variable_types.variable.Index'>",
+                    time_index="<class 'featuretools.variable_types.variable.DatetimeTimeIndex'>",
+                    )
+    for variable in entity.variables:
+        description = variable.to_data_description()
+        _variable = deserialize.description_to_variable(description, entity=entity)
+        assert variable.__eq__(_variable)
```
rwedge, line 180: this test should use the same approach as `test_passing_strings_to_variable_types_from_dataframe`

featuretools/tests/entityset_tests/test_entity.py
```diff
@@ -158,3 +160,62 @@ def test_variable_types_unmodified():
...
+        'id': range(5),
+        'home_team_id': [2, 2, 1, 0, 1],
+        'away_team_id': [1, 0, 2, 1, 0],
+        'home_team_score': [3, 0, 1, 0, 4],
+        'away_team_score': [2, 1, 2, 0, 0]
+    })
+    entities = {'teams': (teams, 'id', None, {'name': 'text'}), 'games': (games, 'id')}
+    relationships = [('teams', 'id', 'games', 'home_team_id')]
+
+    _, features = ft.dfs(entities, relationships, target_entity="teams")
```
rwedge, line 219: adding the optional parameter `features_only=True` will make `dfs` only return the features list

featuretools/tests/entityset_tests/test_entity.py
```diff
@@ -158,3 +160,62 @@ def test_variable_types_unmodified():
...
+    name_class = features[0].entity.variables[1].__class__
```
rwedge, line 220: An entity's variables can be looked up by name directly from the entity object

```
features[0].entity['name'].__class__
```

Small change but I think it makes the test a little easier to follow

PR #992, opened 2020-05-22
Support numeric time index for Dask entityset

### LINKED ISSUE #991: DFS fails with Dask Entityset if a numeric time index is used
Need to update calculate feature matrix to not cast time column to datetime if a numeric time index is used with Dask.

thehomebrewnerd:
Previously time columns were always cast to datetime for Dask entities during DFS. This caused an issue with a numeric time index. This PR fixes that issue.

Fixes #991

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -429,25 +429,20 @@ def calculate_chunk(cutoff_time, chunk_size, feature_set, entityset, approximate
             else:
                 # all rows have same cutoff time. set time and add passed columns
                 num_rows = len(ids)
+                if len(pass_columns) > 0:
+                    pass_through = group[['instance_id', cutoff_df_time_var] + pass_columns]
+                    pass_through.rename(columns={'instance_id': id_name,
+                                                 cutoff_df_time_var: 'time'},
+                                        inplace=True)
                 if isinstance(_feature_matrix, pd.DataFrame):
                     time_index = pd.Index([time_last] * num_rows, name='time')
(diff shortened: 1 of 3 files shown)


featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -442,7 +442,8 @@ def update_progress_callback(done):
                             _feature_matrix[col] = pass_through[col]
                 elif isinstance(_feature_matrix, dd.core.DataFrame) and (len(pass_columns) > 0):
                     _feature_matrix['time'] = time_last
-                    _feature_matrix['time'] = dd.to_datetime(_feature_matrix['time'])
+                    if isinstance(time_last, pd.Timestamp):
```
rwedge, line 445: would it be sufficient to check the `time_type` of the entityset here, given that in `calculate_feature_matrix` we validate the the cutoff time time column's dtype is compatible with the time type of the entityset?

thehomebrewnerd: After additional review, it seems this casting operation is not necessary any longer. I removed this entirely and also refactored a bit to remove a duplicated code block for creating `pass_through`.

thehomebrewnerd:
The question above prompted a review of the test `test_single_table_dask_entityset_cutoff_time_df` which revealed that the entityset time indexes were never set in that test. Setting the time index caused the dask feature matrix to be different than the pandas feature matrix for the rows that had `NaN` values except for the pass through label and id value. This was due to a merge issue with the Dask implementation, which has been resolved with the latest commit `fix dask test with cutoff times`.

PR #996, opened 2020-05-26
Replaced type string with camel to snake-case function

tuethan1999:
### Pull Request Description

Replaced the `Variable` class and subclass's`type_string` with a function that does a camel to snake case conversion based on the class name. Updated the schema version to reflect this as the LatLong and ZIPcode variable types were affected

### DIFF
--- a/featuretools/variable_types/variable.py
+++ b/featuretools/variable_types/variable.py
@@ -1,7 +1,12 @@
 import numpy as np
 import pandas as pd
 
-from featuretools.utils.gen_utils import find_descendents
+from featuretools.utils.gen_utils import camel_to_snake, find_descendents
+
+
+class ClassNameDescriptor(object):
+    def __get__(self, obj, type_):
+        return camel_to_snake(type_.__name__)
 
 
 class Variable(object):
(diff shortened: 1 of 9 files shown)


featuretools/variable_types/variable.py
```diff
@@ -1,7 +1,12 @@
 import numpy as np
 import pandas as pd
 
-from featuretools.utils.gen_utils import find_descendents
+from featuretools.utils.gen_utils import camel_to_snake, find_descendents
+
+
+class ClassNameDescriptor(object):
+    def __get__(self, instance, class_):
```
rwedge, line 8: a brief docstring  mentioning this is a descriptor and summarizing the purpose of the class could be useful here

featuretools/variable_types/variable.py
```diff
@@ -1,7 +1,14 @@
 import numpy as np
 import pandas as pd
 
-from featuretools.utils.gen_utils import find_descendents
+from featuretools.utils.gen_utils import camel_to_snake, find_descendents
+
+
+class ClassNameDescriptor(object):
+    """Descriptor to derive the a variable's type_string from it's class name
```
rwedge, line 8: I think the description could be a little more general, we might use this descriptor for other classes besides `Variable` and it's subclasses

featuretools/variable_types/variable.py
```diff
@@ -5,7 +5,7 @@
 
 
 class ClassNameDescriptor(object):
-    """Descriptor to derive the a variable's type_string from it's class name
+    """Descriptor to convert a class's name from camelcase to snakcase
```
rwedge, line 8: typo in snakecase word

PR #1007, opened 2020-06-02
Deal with `NaN` values in LatLong columns

### LINKED ISSUE #453: haversine error

thehomebrewnerd:
Closes #453 

This PR serves as replacement for #693 

If a single `Nan` value was present instead of a tuple in a `LatLong` columns, this could result in indexing errors when using the `Haversine`, `Latitude` or `Longitude` primitives. This PR fixes that issue.

A check was put in place during entity creation to replace any `NaN` values with `(NaN, NaN)`. This check is also implemented in the `Haversine`, `Latitude` and `Longitude` primitives in case the primitive functions is called directly as in the doctests.


featuretools/entityset/entity.py
```diff
@@ -82,6 +83,11 @@ def __init__(self, id, df, entityset, variable_types=None,
 
         self.set_secondary_time_index(secondary_time_index)
 
+        # Fill in single `NaN` values in LatLong variables with a tuple
```
rwedge, line 86: perhaps this conversion should happen somewhere in the logic of `Entity._create_variables`

do we want to alert the user about their input data being less than ideal?

thehomebrewnerd: I wasn't sure where this should reside exactly, but moving to `_create_variables` seems fine.

I also think it would make sense to warn if we perform this operation on a column, especially since we are modifying the entity dataframe.

featuretools/utils/entity_utils.py
```diff
@@ -126,6 +127,11 @@ def convert_all_variable_data(df, variable_types):
                                        new_type=desired_type,
                                        **type_args)
 
+        # Fill in any single `NaN` values in LatLong variables with a tuple
+        if issubclass(desired_type, vtypes.LatLong) and isinstance(df[var_id], pd.Series) and df[var_id].hasnans:
+            df[var_id] = replace_latlong_nan(df[var_id])
+            warnings.warn("All single `NaN` values in column `{}` have been replaced with `(NaN, NaN)`".format(var_id))
```
rwedge, line 133: I think we could work in that this conversion is happening due to the column being LatLong type

thehomebrewnerd: Updated message.

PR #1021, opened 2020-06-08
Remove Feature.get_names()

tuethan1999:
### Pull Request Description
The get_names method is just a helper method for the get_feature_names method. I made it a private method so that get_featurenames has unique functionality in the public api.

-----


thehomebrewnerd:
Is there any reason we can't just remove `Feature.get_names()` completely and move the logic for setting `Feature._names` into `Feature.get_feature_names()`? 

That would leave us with two public methods for getting feature names:
- `Feature.get_name()` to get the single string name of the base feature (which will also be the name of the feature in the case of a single output feature)
- `Feature.get_feature_names()` to get a list of all the new feature names that will be generated. For single output features, this will be a single item list, and for multi output features all the feature names will be in the list.

docs/source/changelog.rst
```diff
@@ -6,10 +6,12 @@ Changelog
     * Enhancements
     * Fixes
     * Changes
+        * Deprecated feature's get_name function, get_feature_names should be used instead
```
thehomebrewnerd, line 9: This should be the `get_names` not `get_name` function. You can also enclose the function names in double back ticks for code formatting: ``get_feature_names``.

docs/source/changelog.rst
```diff
@@ -6,10 +6,12 @@ Changelog
     * Enhancements
```
thehomebrewnerd, line 6: Uncomment the Future Release section

featuretools/feature_base/feature_base.py
```diff
@@ -82,7 +82,7 @@ def get_name(self):
             self._name = self.generate_name()
         return self._name
 
-    def get_names(self):
+    def __get_names(self):
```
thehomebrewnerd, line 85: If we keep this method, we should use a single `_` in the name: `def _get_names(self):`

featuretools/feature_base/feature_base.py
```diff
@@ -82,18 +82,13 @@ def get_name(self):
...
         n = self.number_output_features
         if n == 1:
-            names = [self.get_name()]
-        else:
-            names = self.get_names()
-        return names
+            return [self.get_name()]
+        if not self._names:
+            self._names = self.generate_names()
+        return self._names
```
thehomebrewnerd, line 91: I think we can improve this a little more - probably no need to create `n`, and we can set `self._names` for single output features as well:
```
def get_feature_names(self):
    if not self._names:
        if self.number_output_features == 1:
            self._names = [self.get_name()]
        else:
            self._names = self.generate_names()
    return self._names
```

docs/source/changelog.rst
```diff
@@ -6,13 +6,14 @@ Changelog
...
     * Fixes
     * Changes
+        * Deprecated ``feature's get_names`` function, ``get_feature_names`` should be used instead
```
thehomebrewnerd, line 9: Maybe change these to `Feature.get_names()` and `Feature.get_feature_names()` instead.

docs/source/changelog.rst
```diff
@@ -7,13 +7,14 @@ Changelog
     * Fixes
         * Improve warnings when using a Dask dataframe for cutoff times (:pr:`1026`)
     * Changes
+        * Deprecated ``Feature.get_names()``, ``Feature.get_feature_names()`` should be used instead(:pr:`1021`)
```
kmax12, line 10: was `get_names()` removed? if so, we should say "removed" not "deprecated". Deprecate means that we're in the process of removing want to tell people to stop using it, rather than have actually finished removing. See definition [here](https://techterms.com/definition/deprecated#:~:text=In%20the%20world%20of%20software,means%20to%20disapprove%20of%20something.)

Since we are removing, we should also add this to a breaking changes section with the comment of what to do instead.

featuretools/feature_base/feature_base.py
```diff
@@ -83,6 +84,8 @@ def get_name(self):
         return self._name
 
     def get_feature_names(self):
+        if not hasattr(self, '_names'):
```
tuethan1999, line 87: FeatureOutputSlice does not have a _names attribute

featuretools/tests/primitive_tests/test_direct_features.py
```diff
@@ -52,6 +54,43 @@ def test_direct_from_variable(es):
...
+    copy_feat = feat.rename("session_test")
+    assert feat.unique_name() != copy_feat.unique_name()
+    assert feat.get_name() != copy_feat.get_name()
+    assert feat.base_features[0].generate_name() == copy_feat.base_features[0].generate_name()
+    assert feat.entity == copy_feat.entity
+    assert feat._names != copy_feat._names
+    assert feat.get_feature_names() != copy_feat.get_feature_names()
+
+
+def test_rename_multioutput(es):
```
tuethan1999, line 69: I think we should talk about what the expected output behavior is for a renamed multi-output feature. It seems to be the case that most generate_names() functions for multi-output features do not depend on the feature's _name attribute, but rather on the base_feature's name.

In this example

`feat.get_name()` returns: `'N_MOST_COMMON(log.product_id, n=2)'`
`feat.get_feature_names()` returns: `['N_MOST_COMMON(log.product_id, n=2)[0]', 'N_MOST_COMMON(log.product_id, n=2)[1]']`

Even though `copy_feat.get_name()` returns: `session_test`
`copy_feat.get_feature_names()` returns: `['N_MOST_COMMON(log.product_id, n=2)[0]', 'N_MOST_COMMON(log.product_id, n=2)[1]']`

Is this the desired behavior or are we expecting it to return `session_test[0]', 'session_test[1]']`?

rwedge:
perhaps when rename updates `_name` it could also update `_names`

do we think in general, it is sufficient when renaming a feature with multiple outputs to expect

feature name - "session_test"
feature's first output column  - "session_test[0]"
feature's second output column - "session_test[1]"

and so on, or should we provide a way to rename the individual outputs as well?

featuretools/feature_base/feature_base.py
```diff
@@ -82,19 +86,14 @@ def get_name(self):
             self._name = self.generate_name()
         return self._name
 
-    def get_names(self):
+    def get_feature_names(self):
```
rwedge, line 89: we should update `test_serialized_renamed_features` in `test_feature_serialization.py` to test `get_feature_names` as well

The test case also appears to be missing tests for a multi-output feature, and a FeatureOutputSlice feature

featuretools/tests/primitive_tests/test_direct_features.py
rwedge: we expect `'session_test[0]', 'session_test[1]'`

featuretools/tests/primitive_tests/test_feature_serialization.py
```diff
@@ -94,15 +95,23 @@ def test_pickle_features_with_custom_primitive(pd_es, tmpdir):
 
 def test_serialized_renamed_features(es):
     def serialize_name_unchanged(original):
-        renamed = original.rename('MyFeature')
-        assert renamed.get_name() == 'MyFeature'
+        def check_names(feature, name):
+            assert feature.get_name() == name
+            if len(names) == 1:
```
rwedge, line 100: this section should test that `feature.get_feature_names()` matches the expected `names` value

instead of generating the expected `names` value inside of `check_names`, we could probably do it outside like we do with `new_name`, and pass it in

featuretools/tests/primitive_tests/test_feature_base.py
```diff
@@ -208,3 +209,40 @@ def test_multi_output_index_error(es):
     error_text = 'index is higher than the number of outputs'
     with pytest.raises(AssertionError, match=error_text):
         three_common[10]
+
+
+def test_rename(es):
+    feat = ft.Feature(es['log']['id'], parent_entity=es['sessions'], primitive=Count)
+    feat.get_feature_names()
+    copy_feat = feat.rename("session_test")
+    assert feat.unique_name() != copy_feat.unique_name()
```
rwedge, line 218: for these checks, can we also check what we expect the name to be, along with the confirming the name changed?

can we turn these checks into a reusable function since all three tests share the same checks

featuretools/tests/primitive_tests/test_feature_serialization.py
tuethan1999: Is there a reason we would want to though? We have a clear naming schema where `_names` depends on `_name`. Passing `names` in would produce repeated code outside the function

featuretools/tests/primitive_tests/test_feature_base.py
tuethan1999: Yes

featuretools/tests/primitive_tests/test_feature_serialization.py
rwedge: That explanation was missing some points, let me rephrase

* Currently, `names` variable in the function does not depend on `feature`.  It is always equal to `renamed.get_feature_names()`, even if `check_names` is testing `deserialized`
* If we moved the if/else for determining the expected `names` value outside of the function and passed the the result in it wouldn't create repeated code but it would add an extra line.  Let's leave it as is.

featuretools/tests/primitive_tests/test_feature_base.py
```diff
@@ -208,3 +210,32 @@ def test_multi_output_index_error(es):
...
+
+
+def test_rename_featureoutputslice(es):
+    multi_output_feat = ft.Feature(es['log']['product_id'],
+                                   parent_entity=es['customers'],
+                                   primitive=NMostCommon(n=2))
+    feat = ft.feature_base.FeatureOutputSlice(multi_output_feat, 0)
+    num_feature_names = len(feat.get_feature_names())
+    new_name = 'session_test'
+    new_names = [new_name] if num_feature_names == 1 else [new_name + '[{}]'.format(i) for i in range(num_feature_names)]
```
rwedge, line 240: since we know which features are multi output we can manually set `new_names` to the right value on a test by test basis in this file instead of relying on the output of `get_feature_names` to determine the correct value.

PR #1026, opened 2020-06-09
Updates for Dask cutoff time dataframes

### LINKED ISSUE #1025: Inconsistencies when passing Dask dataframe for cutoff times
There are some inconsistencies between `dfs` and `calculate_feature_matrix` in what happens if a Dask dataframe is passed in for cutoff times. In `dfs` a warning is raised and the Dask dataframe is computed and turned into a pandas dataframe. If `calculate_feature_matrix` is called directly with a Dask dataframe for cutoff times a `TypeError` is raised.

Recommended Fix: Remove error from `calculate_feature_matrix` and move warning from `dfs` to `calculate_feature_matrix`, so no matter which call is made, a warning will be raised followed by a compute call to convert the dataframe from Dask to pandas. The documentation should also be updated to clarify this detail as currently the documentation for Dask EntitySets indicates Dask cutoff time dataframes cannot be used at all.

thehomebrewnerd:
Fixes #1025 

Warns the user if a Dask dataframe is supplied to `dfs` or `calculate_feature_matrix`. Updates documentation to indicate that a Dask dataframe will be accepted for cutoff times, but should not be used due to the `.compute()` call that is required.


docs/source/guides/using_dask_entitysets.rst
```diff
@@ -108,7 +108,7 @@ DFS Limitations
 ***************
 There are a few key limitations when generating a feature matrix from a Dask ``EntitySet``.
 
-If a ``cutoff_time`` parammeter is passed to ``featuretools.dfs()`` it must either be a single cutoff time value, or a pandas dataframe. The current implementation does not support the use of a Dask dataframe for cutoff time values.
+If a ``cutoff_time`` parameter is passed to ``featuretools.dfs()`` it should be a single cutoff time value, or a pandas dataframe. The current implementation will still work if a Dask dataframe is supplied for cutoff timess, but a ``.compute()`` call will be made on the dataframe to convert it into a pandas dataframe. This conversion will result in a warning, and the process could take a considerable amount of time to complete depending on the size of the supplied dataframe.
```
rwedge, line 111: > ... will still work if a Dask dataframe is supplied for cutoff timess

extra s in times

thehomebrewnerd: Fixed

PR #1028, opened 2020-06-12
Refactor cutoff times

### LINKED ISSUE #961: Refactor handling of cutoff times when a single value or no value is passed to DFS
If the user supplies a single cutoff time to DFS or if no cutoff time is supplied, the current code in `calculate_feature_matrix` generates a new cutoff_time dataframe to use in downstream processing.  This approach is inefficient for large datasets as a large dataframe is being generated and the cutoff time column contains only one unique value which is equal to the cutoff time value supplied by the user.

The code should be refactored to eliminate the need to generate this intermediate dataframe when cutoff values are identical for all instances. 

Refactoring will also resolve issue #907 which cases slow run times when using Dask entities for DFS.

thehomebrewnerd:
Refactor handling of cutoff times to avoid creating a large dataframe unnecessarily when using a single or `None` cutoff time value. Also, move cutoff time quality checks to `utils.py` instead of being included inside `calculate_feature_matrix.py`.

When cutoff time is a single value or `None` a cutoff time tuple is created that contains the cutoff time value along with the instance ids.

Fixes #961

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -171,6 +171,9 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
                                             include_cutoff_time=include_cutoff_time)
             instance_ids = df[index_var]
 
+        if isinstance(instance_ids, dd.Series):
+            instance_ids = instance_ids.compute()
+
         # convert list or range object into series
         if not isinstance(instance_ids, pd.Series) and not isinstance(instance_ids, dd.Series):
             instance_ids = pd.Series(instance_ids)
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -149,6 +149,9 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
     cutoff_time = _validate_cutoff_time(cutoff_time, target_entity)
 
     if isinstance(cutoff_time, pd.DataFrame):
+        if instance_ids:
+            msg = "Passing 'instance_ids' is valid only if 'cutoff_time' is a single value or None - ignoring"
+            warnings.warn(msg)
         pass_columns = [col for col in cutoff_time.columns if col not in ['instance_id', 'time']]
         # make sure dtype of instance_id in cutoff time
         # is same as column it references
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -16,6 +16,8 @@ from featuretools.computational_backends.feature_set_calculator import (
     FeatureSetCalculator
 )
 from featuretools.computational_backends.utils import (
+    _check_cutoff_time_type,
+    _validate_cutoff_time,
     bin_cutoff_times,
     create_client_and_cluster,
     gather_approximate_features,
@@ -26,12 +28,7 @@ from featuretools.entityset.relationship import RelationshipPath
 from featuretools.feature_base import AggregationFeature, FeatureBase
 from featuretools.utils import Trie
 from featuretools.utils.gen_utils import make_tqdm_iterator
-from featuretools.utils.wrangle import _check_time_type
-from featuretools.variable_types import (
-    DatetimeTimeIndex,
-    NumericTimeIndex,
-    PandasTypes
-)
+from featuretools.variable_types import NumericTimeIndex
 
 logger = logging.getLogger('featuretools.computational_backend')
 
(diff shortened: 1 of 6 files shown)


featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -499,6 +501,18 @@ def test_training_window(pd_es):
                                               include_cutoff_time=False)
     prop_values = [5, 5, 2]
     dagg_values = [3, 2, 1]
+
+    assert (feature_matrix[property_feature.get_name()] == prop_values).values.all()
+    assert (feature_matrix[dagg.get_name()] == dagg_values).values.all()
+
+    # Case3. include_cutoff_time = False with single cutoff time value
```
frances-h, line 508: Can we add a case 4 for if `include_cutoff_time` is `True`?

featuretools/synthesis/dfs.py
```diff
@@ -252,33 +250,18 @@ def dfs(entities=None,
...
-                                                  approximate=approximate,
-                                                  cutoff_time_in_index=cutoff_time_in_index,
-                                                  save_progress=save_progress,
-                                                  chunk_size=chunk_size,
-                                                  n_jobs=n_jobs,
-                                                  dask_kwargs=dask_kwargs,
-                                                  verbose=verbose,
-                                                  progress_callback=progress_callback,
-                                                  include_cutoff_time=include_cutoff_time)
+    feature_matrix = calculate_feature_matrix(features,
```
frances-h, line 253: Do we have a test that validates that this works? i.e. one that passes in both a pandas `cutoff_time` df as well as some a list of instance ids which seems to be what the old check was avoiding.

thehomebrewnerd: I don't see any existing tests that cover that exact situation, but I tested it out and if you supply both a cutoff time and df and a list of instance ids, the instance ids list is ignored as indicated in the docstring. 

I do think it would be good to add a warning though to notify the user that the instance ids are not being used, and I'll make that update.

thehomebrewnerd: Added a warning and new test case to cover passing both a cutoff time df and instance ids.

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
thehomebrewnerd: Added this additional test case.

featuretools/computational_backends/utils.py
```diff
@@ -198,3 +205,72 @@ def get_client_cluster():
...
+
+
+def _check_cutoff_time_type(cutoff_time, es_time_type):
+    """
+    Check that the cutoff time values are of the proper type given the entityset time type
+    """
+    # Check that cutoff_time time type matches entityset time type
+    if isinstance(cutoff_time, tuple):
+        cutoff_time_dtype = type(cutoff_time[0])
+        numeric_types = [int, float]
```
rwedge, line 264: We may want to add more numeric types here.  Things like `cutoff_time=pd.Series([1])[0]` will cause trouble here, and the error message and using pd.to_numeric would be unhelpful currently

rwedge: the `_check_time_type` logic checks for numpy variations

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -172,61 +172,25 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
...
-
-    # make sure dtype of instance_id in cutoff time
-    # is same as column it references
-    target_entity = features[0].entity
-    dtype = entityset[target_entity.id].df[target_entity.index].dtype
-    cutoff_time["instance_id"] = cutoff_time["instance_id"].astype(dtype)
+            instance_ids = df[index_var]
+
+        if isinstance(instance_ids, dd.Series):
+            instance_ids = instance_ids.compute()
```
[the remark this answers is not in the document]
thehomebrewnerd: Warning here could be confusing. This `.compute()` is needed in the case where the target entity is a Dask dataframe, but `instance_ids` was `None`. In that case, the instance ids are created from the dataframe index column, but we need to compute this because the length of `instance_ids` is used by `_handle_chunk_size` and for the progress bar. Was thinking it would be better to compute once here rather than computing multiple times whenever the length is needed.

I think this could be cleaned up in the future with some additional refactoring.

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -172,61 +172,25 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
...
+
+        # convert list or range object into series
+        if not isinstance(instance_ids, pd.Series) and not isinstance(instance_ids, dd.Series):
```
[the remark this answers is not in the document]
thehomebrewnerd: Removed as this extra check is unnecessary with the  prior `.compute()` in place.

featuretools/computational_backends/utils.py
thehomebrewnerd: Changed the strategy to check if the cutoff time is an instance of `numbers.Number` which should take care of the scenario you mentioned where the type is `numpy.int64` or something along those lines.

Do you see any issues with this new approach?

thehomebrewnerd: I missed your comment before on `_check_time_type`. I updated the code to use `_check_time_type` for single cutoff time values instead of the approach with `numbers.Number` - leveraging existing code and eliminating additional imports.

PR #1032, opened 2020-06-23
Feature lineage visualization

### LINKED ISSUE #937: Graph feature lineage
It can be difficult to interpret the meaning of feature from the name alone.  One way aid in this would be a visual representation of the how the base features were combined, tracing back from the final feature to the original variables via the primitives that were applied

frances-h:
Add `graph_feature` function which takes in a feature and returns a graph demonstrating how the feature was generated:

simple_transform

trans_agg_stacked

Closes #937

### DIFF
--- a/featuretools/feature_base/feature_visualizer.py
+++ b/featuretools/feature_base/feature_visualizer.py
@@ -26,6 +26,16 @@ TARGET_TEMPLATE = '''
 
 
 def graph_feature(feat, to_file=None):
+    '''Generates a feature lineage graph for the given feature
+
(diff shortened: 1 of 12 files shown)


featuretools/feature_base/feature_visualizer.py
```diff
@@ -26,6 +26,16 @@
 
 
 def graph_feature(feat, to_file=None):
```
kmax12, line 28: nitpick, but let's avoid shortening variable names unless it's really needed. i think you can call this input `feature`

PR #1036, opened 2020-06-28
use mock_customer fixture

### LINKED ISSUE #1001: Create a pytest fixture for load_mock_customer in tests
There are several tests that create an EntitySet within the test using `ft.demo.load_mock_customer`. I think it would be better to create a pytest fixture for this EntitySet and use the fixture as input into the test, rather than creating the EntitySet inside the test. 

Using a fixture would also allow for easy parameterization to also run these tests with a Dask version of the EntitySet after PR #783 is closed.

systemshift:
### Pull Request Description
closes #1001
Unless I misunderstood, the issue has been mostly solved with the exception of this single case.
There are only 3 times `ft.demo.load_mock_customer` is called in tests, 2 already have fixtures.

-----


thehomebrewnerd:
This data is also used in `test_demo_data.py` on [Line 26](https://github.com/FeatureLabs/featuretools/blob/95117a7da15decf993bde65479f67fe883072754/featuretools/tests/demo_tests/test_demo_data.py#L26).

I think we should also move this fixture definition to `conftest.py` along with the Dask version, and parameterize them like we have done with the other fixtures. See `test_calculate_feature_matrix.py` [Line 1458](https://github.com/FeatureLabs/featuretools/blob/95117a7da15decf993bde65479f67fe883072754/featuretools/tests/computational_backend/test_calculate_feature_matrix.py#L1458) for the fixture definitions. We should then make sure that there are no versions of this fixture defined in the individual tests files unless there is some specific need to do so.

systemshift:
@thehomebrewnerd please review and let me know if I should add anything.
I do not know what is causing the codecov issue, given that I haven't touched that file.

docs/source/changelog.rst
```diff
@@ -18,6 +18,7 @@ Changelog
         * Add development install instructions to contributing.md (:pr:`1030`)
     * Testing Changes
         * Add ``required`` flag to CircleCI codecov upload command (:pr:`1035`)
+        * Add fixture for ``ft.demo.load_mock_customer`` (:pr:`1036`)
 
     Thanks to the following people for contributing to this release:
     :user:`tuethan1999`, :user:`kmax12`,  :user:`thehomebrewnerd`,  :user:`gsheni`, :user:`frances-h`
```
thehomebrewnerd, line 24: Add your user name to the list of contributors.

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -1455,11 +1455,6 @@ def test_string_time_values_in_cutoff_time(es):
...
 
 
-@pytest.fixture
-def pd_mock_customer():
-    return ft.demo.load_mock_customer(return_entityset=True, random_seed=0)
-
-
 @pytest.fixture
 def dd_mock_customer(pd_mock_customer):
     dd_mock_customer = copy.deepcopy(pd_mock_customer)
```
thehomebrewnerd, line 1460: We should also move the `dd_mock_customer` and `mock_customer` fixtures to `conftest.py`.

featuretools/tests/demo_tests/test_demo_data.py
```diff
@@ -22,8 +22,8 @@ def test_load_retail_diff():
     assert es_second['order_products'].df.shape[0] == nrows_second
 
 
-def test_mock_customer():
-    es = load_mock_customer(return_entityset=True)
+def test_mock_customer(pd_mock_customer):
+    es = pd_mock_customer
```
thehomebrewnerd, line 26: Let's use the `mock_customer` fixture here so this test gets run with both the pandas and Dask versions of the fixture.

featuretools/tests/utils_tests/test_entity_utils.py
```diff
@@ -17,19 +16,14 @@
...
 @pytest.fixture
-def pd_mock_customer_es():
-    return ft.demo.load_mock_customer(return_entityset=True, random_seed=0)
-
-
-@pytest.fixture
-def dask_mock_customer_es(pd_mock_customer_es):
-    dask_es = copy.deepcopy(pd_mock_customer_es)
+def dask_mock_customer_es(pd_mock_customer):
+    dask_es = copy.deepcopy(pd_mock_customer)
```
thehomebrewnerd, line 20: We can delete this and use the `dd_mock_customer` fixture from `conftest.py` instead.

featuretools/tests/utils_tests/test_entity_utils.py
```diff
@@ -17,19 +16,14 @@
...
     for entity in dask_es.entities:
         entity.df = dd.from_pandas(entity.df.reset_index(drop=True), npartitions=2)
     return dask_es
 
 
-@pytest.fixture(params=['pd_mock_customer_es', 'dask_mock_customer_es'])
+@pytest.fixture(params=['pd_mock_customer', 'dask_mock_customer_es'])
```
thehomebrewnerd, line 26: We can delete this fixture and use the `mock_customer` fixture from `conftest.py` instead. Will also need to update `test_get_linked_vars` to use the `mock_customer` feature as well to test with both pandas and Dask.

thehomebrewnerd:
> @thehomebrewnerd please review and let me know if I should change anything.
> I do not know what is causing the codecov issue, given that I haven't touched that file.

The codecov issue is likely because you are reducing the number of total lines of code, so the percent covered is decreasing.

featuretools/tests/utils_tests/test_entity_utils.py
systemshift: so there should be 2 more `test_get_linked_vars`, such as `test_get_linked_vars_pd` and `test_get_linked vars_dd`?

thehomebrewnerd: No, you do not need to create any additional tests. Because `mock_customer` is a parameterized fixture, anywhere it is used, pytest will automatically test using the defined parameter values. So in the current `test_get_linked_vars` if you just use `mock_customer` as the fixture instead, pytest will execute that test with `pd_mock_customer` and then `dd_mock_customer`. The change you have made to this test already should be sufficient.

thehomebrewnerd:
@systemshift Please merge in the latest changes from `featuretools/master` and resolve the conflicts in `changelog.rst`.

docs/source/changelog.rst
```diff
@@ -28,10 +28,11 @@ Changelog
         * Add development install instructions to contributing.md (:pr:`1030`)
     * Testing Changes
         * Add ``required`` flag to CircleCI codecov upload command (:pr:`1035`)
+        * Add fixture for ``ft.demo.load_mock_customer`` (:pr:`1036`)
```
thehomebrewnerd, line 31: This item should be added to the `Future Release` section above. Be sure to uncomment that section by removing the two dots `..` before the `Future Release` test.

docs/source/changelog.rst
```diff
@@ -28,10 +28,11 @@ Changelog
...
 
     Thanks to the following people for contributing to this release:
     :user:`frances-h`, :user:`gsheni`, :user:`kmax12`, :user:`rwedge`,
-    :user:`thehomebrewnerd`, :user:`tuethan1999`
+    :user:`thehomebrewnerd`, :user:`tuethan1999`, :user:`systemshift`
```
thehomebrewnerd, line 35: Similar to the prior comment, remove your user name from here and move it up to the future release section above. This may be a little confusing, but this happened because version 0.17.0 was released as you were making your updates. These changes will be part of the next release.

Issue #1014, opened 2020-06-03
Verify proper usage of warnings.warn() and logger.warning() throughout code


thehomebrewnerd:
In the current code base there are a number of places where `warnings.warn()` is used to issue a warning and there are a number of other places where `logger.warning()` is used for a warning. 

These warnings should be reviewed and updated for consistent and proper usage, perhaps using the discussion in this post as a guide for when to use each: https://stackoverflow.com/questions/9595009/warnings-warn-vs-logging-warning/14762106#14762106

thehomebrewnerd:
Closed by #1040

PR #1040, opened 2020-06-30
replace logger.warning with warnings.warn

systemshift:
### Pull Request Description
Based on #1014 and the linked stack overflow thread discussion.
I started with the intent where `warnings.warn()` should be used for designing a library(which featuretools is) to warn the user when it is not used correctly. Where as `logger.warning()` is more suitable for apps where the wrong input is user data.

So most of `warnings.warn()` seem to be used as they should be, with a few cases where logger was used and I felt `warnings.warn()` made more sense.
Now there are a few cases of `logger.warning()` that looked off, but I did not want to change them, so I would like someone to review the remaining `logger.warning()` cases to see if it should still be there and not replaced. 

-----


systemshift:
I will consider this PR done and ready to merge unless someone points out to warnings worth changing.

thehomebrewnerd:
@systemshift Please pull in the latest changes from the `main` branch, and you may need to update `changelog.rst` after this to get your update in the right location.

featuretools/primitives/options_utils.py
```diff
@@ -58,9 +58,9 @@ def generate_all_primitive_options(all_primitives,
     for primitive in all_primitives:
         if primitive in primitive_options and primitive.name in primitive_options:
             msg = "Options present for primitive instance and generic " \
-                  "primitive class (%s), primitive instance will not use generic " \
+                  "primitive class %s, primitive instance will not use generic " \
```
systemshift, line 61: This one was really hard to work around, using `()` in a string seems to break something with `pytest.warn`. Tried using string literals but no luck.

Would like to know if someone knows how to work around this issue.

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -1199,10 +1200,14 @@ def _check_schema_version(version, es, warning_text):
...
-            deserialize.description_to_entityset(dictionary)
-        assert record[0].message.args[0] == warning_text
+    # TODO Refactor this function and both parent functions to run tests with a less complex implementation
+    earlier_schema_warning = 'is no longer supported by this version'
+    if earlier_schema_warning in str(warning_text):
+        logger = logging.getLogger('featuretools')
+        logger.propagate = True
+        deserialize.description_to_entityset(dictionary)
+        assert warning_text in caplog.text
+        logger.propagate = False
```
thehomebrewnerd, line 1210: Before we were checking that a `UserWarning` got raised for both earlier and later schema versions with the assert statement inside the `pytest.warns` context manager. Now we changed one of these to be logged instead.

With these code changes here, we are only testing that warning is logged for the earlier schema test. We are no longer testing that the `UserWarning` still gets raised for the later schema test. To test that we need to add a later scheme condition to the if-else block to verify the warning gets raised as it did before.

featuretools/tests/primitive_tests/test_features_deserializer.py
```diff
@@ -275,10 +277,13 @@ def _check_schema_version(version, es, warning_text):
...
-
-        assert record[0].message.args[0] == warning_text
+    # TODO Refactor this function and both parent functions to run tests with a less complex implementation
+    earlier_schema_warning = 'is no longer supported by this version'
+    if earlier_schema_warning in str(warning_text):
+        logger = logging.getLogger('featuretools')
+        logger.propagate = True
+        FeaturesDeserializer(dictionary)
+        assert warning_text in caplog.text
+        logger.propagate = False
```
thehomebrewnerd, line 287: Similar to the entityset test, we need to make sure we are testing for the `UserWarning` for the later schema version test.

featuretools/primitives/options_utils.py
thehomebrewnerd: I don't think it is critical to have the parentheses here, but I'm also not sure how to work around this exactly either.

featuretools/tests/entityset_tests/test_es.py
thehomebrewnerd: One way you could refactor this would be to split `_check_schema_version` into three different functions. You could do something like `_check_schema_raises_warning` to check that the `UserWarning` gets raised inside `test_later_schema_version`. Then have `_check_schema_logs_warning` to test logging for the `test_earlier_schema_version` test. In both of those new utility functions we could use a third `_get_schema_dictionary` function to build and return the dictionary that will be needed by both of the new check functions.

Another idea would be to create a new parameter that gets passed to _check_schema_version to indicate whether we should check for a `UserWarning` or a logged warning instead of relying on the value of the warning text to make that determination.

This seems like it is getting more complicated than I anticipated though. @rwedge Do you have any thoughts on the best approach here?

featuretools/primitives/options_utils.py
rwedge: using escapes on the parentheses around mean --  "\\(%s\\)" -- in the `warning_msg` string should work.  The string is interpreted as a regex so a regular parentheses won't get the match we want

featuretools/tests/entityset_tests/test_es.py
rwedge: I think keep `_check_schema_version` as a single function, add a new parameter to it to indicate if it is expected to warn / log / neither, and use `warning_text` for matching purposes

featuretools/entityset/entity.py
```diff
@@ -585,8 +585,8 @@ def _create_index(index, make_index, df):
         # Case 1: user wanted to make index but did not specify column name
         assert not make_index, "Must specify an index name if make_index is True"
         # Case 2: make_index not specified but no index supplied, use first column
-        logger.warning(("Using first column as index. ",
-                        "To change this, specify the index parameter"))
+        warnings.warn(("Using first column as index. "
```
rwedge, line 588: Let's test this warning in `test_none_index` in test_es.py

featuretools/entityset/entity.py
```diff
@@ -595,8 +595,8 @@ def _create_index(index, make_index, df):
         if not make_index:
             # Case 4: user names index, it is not in df. does not specify
             # make_index.  Make new index column and warn
-            logger.warning("index %s not found in dataframe, creating new "
-                           "integer column", index)
+            warnings.warn("index {} not found in dataframe, creating new "
```
rwedge, line 598: Let's test this warning in `test_unknown_index` in test_es.py

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -89,10 +89,12 @@ def test_reset_metadata(es):
 
 
 def test_cannot_re_add_relationships_that_already_exists(es):
-    before_len = len(es.relationships)
-    es.add_relationship(es.relationships[0])
-    after_len = len(es.relationships)
-    assert before_len == after_len
+    warn_text = "Not adding duplicate relationship: " + str(es.relationships[0])
+    with pytest.warns(UserWarning, match=warn_text):
+        before_len = len(es.relationships)
```
rwedge, line 94: I think we should limit the code run in the `with pytest.warns(..):` block to just the line we expect to trigger the warning.  In this case, `es.add_relationship`. 

We should also do this in `test_none_index` and `test_unknown_index`

PR #1048, opened 2020-07-03
Fix TimeSince primitive. Make it work with all Datetime subclasses

### LINKED ISSUE #1047: TimeSince primitive works only with DatetimeTimeIndex type

Transform primitive ``time_since`` should match all Datetime variables:
https://github.com/FeatureLabs/featuretools/blob/8693e5cd438a065fc64a1e1535e74ec54ab85cad/featuretools/primitives/standard/transform_primitive.py#L389-L390

However, it only matches those whose type is DatetimeTimeIndex as the following example reflects.

```python
import featuretools as ft
import featuretools.variable_types as vtypes

target_entity = es["customers"]
# Calculate all datetime-like vars
datetime_vars = [id for id,t in target_entity.variable_types.items() if issubclass(t, vtypes.Datetime)]
print(len(datetime_vars))  # should print 4

# Apply time_since only to 'customers' entity
primitive_options={
    'time_since': {
        'include_entities': ['customers'],
    }
}

fm, fl = ft.dfs(
    target_entity="customers",
    entityset=es,
    trans_primitives=["time_since"],
    agg_primitives=[],
    max_depth=1,
    primitive_options=primitive_options
)

matched_columns = [c for c in fm.columns if c.startswith('TIME_SINCE')]
# Should output 4 columns corresponding to all Datetime vars
# but only matches the one whose type is DatetimeTimeIndex instead
print(len(matched_columns))  

```

This happens because only the first set of input types is selected when there is more than one:

https://github.com/FeatureLabs/featuretools/blob/8693e5cd438a065fc64a1e1535e74ec54ab85cad/featuretools/synthesis/deep_feature_synthesis.py#L561-L562

Nonetheless, later in the process all those features whose type is a subclass of the input_type are matched

https://github.com/FeatureLabs/featuretools/blob/8693e5cd438a065fc64a1e1535e74ec54ab85cad/featuretools/synthesis/deep_feature_synthesis.py#L749-L755

Therefore, we can solve this issue by simply using ``Datetime`` as ``time_since``'s input_type, since ``DatetimeTimeIndex`` is a subclass of the former. 

2020-07-03 19:36:57,089 featuretools - WARNING    Featuretools failed to load plugin nlp_primitives from library nlp_primitives. For a full stack trace, set logging to debug.
Featuretools version: 0.17.0
Featuretools installation directory: /mnt/c/Users/Montana/PycharmProjects/featuretools/featuretools

SYSTEM INFO
-----------
python: 3.8.3.final.0
python-bits: 64
OS: Linux
OS-release: 4.19.121-microsoft-standard
machine: x86_64
processor: x86_64
byteorder: little
LC_ALL: None
LANG: C.UTF-8
LOCALE: en_US.UTF-8

INSTALLED VERSIONS
------------------
numpy: 1.19.0
pandas: 1.0.5
tqdm: 4.47.0
PyYAML: 5.3.1
cloudpickle: 1.5.0
dask: 2.19.0
distributed: 2.19.0
psutil: 5.7.0
pip: 20.1.1
setuptools: 47.3.1.post20200622

monti-python:
### Pull Request Description
fixes #1047


featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -1116,3 +1117,20 @@ def is_greater(x):
 
     for name in expected_names:
         assert name in fm.columns
+
+
+def test_time_since_primitive_matches_all_datetime_types(pd_es):
```
thehomebrewnerd, line 1122: Since the `TimeSince` primitive is compatible with Dask entitysets, we should use the `es` test fixture here instead of the `pd_es` test fixture so that this gets tested with both pandas and Dask entitysets.

monti-python: Done, thanks for the comment!

monti-python:
@thehomebrewnerd I've updated the PR, thanks for reviewing!

PR #1050, opened 2020-07-06
Fix issue with missing instances and categorical entity index

### LINKED ISSUE #1046: Invalid instance ID causes TypeError (non-category item)
### Invalid instance ID causes TypeError (non-category item)
- Before 0.17.0, if an invalid instance ID was passed to `calculate_feature_matrix`, there would no error and the feature matrix would calculate.
- However, some modification in https://github.com/FeatureLabs/featuretools/commit/f21606a6961e4b9ad2f50d9340d0de712a594e33 resulted in a change where this no longer occures. Now an error shows up.
- Installing any featuretools version before f21606a6961e4b9ad2f50d9340d0de712a594e33 will fix the error. 

#### Expected Output
- `calculate_feature_matrix` is able to run with an invalid instance ID, and calculate the feature matrix accordingly. 

#### Steps to Reproduce
- Run `calculate_feature_matrix` with an invalid instance ID. **(Instance ID of 3 is invalid for ecommerce & customers entity)**

```python
import featuretools as ft
from featuretools import dfs

target_entity = 'customers'
instance_ids = [0, 1, 2, 3]
max_depth = 3
max_features = -1

features = dfs(entityset=es, target_entity=target_entity,
               max_depth=max_depth, features_only=True)

df = ft.calculate_feature_matrix(entityset=es,
                                 features=features,
                                 instance_ids=instance_ids)
```

thehomebrewnerd:
Fixes #1046 

Changed order of operations in merging default dataframe into feature matrix to allow combination of categorical and integer indexes. Added test case to cover this situation.


featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -130,7 +130,8 @@ def progress_callback(*args):
             if missing_ids:
                 default_df = self.generate_default_df(instance_ids=missing_ids,
```
gsheni, line 131: Would it make sense to warn the user that there are invalid instance IDs?

featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -130,7 +130,8 @@ def progress_callback(*args):
...
                                                       extra_columns=df.columns)
-                df = df.append(default_df, sort=True)
+
+                df = default_df.append(df, sort=True)
```
gsheni, line 134: So this results in using the default feature values + missing instance IDs to create additional rows.
Would it make more sense to not create this? It seems incorrect to add this data...

thehomebrewnerd:
@rwedge Added checks to make sure that the feature matrix has a categorical index, if the target entity has a categorical index. If the user supplies an instance id that is not in the target entity, the index categories will be different in the feature matrix (to account for adding the missing ids), but the index dtype will be maintained.

This change required some updates to a few tests which had errors due to an invalid indexing approach being used for a series with a categorical index. For example, `df[col][0]` returns the first value in the series if the index is not categorical, but results in an error if the index is categorical. Changed to `df[col].values[0]` which works for both.

featuretools/computational_backends/feature_set_calculator.py
thehomebrewnerd: Since this isn't directly related to the initial but, will create a new issue to track.

featuretools/computational_backends/feature_set_calculator.py
thehomebrewnerd: Since this isn't directly related to the initial but, will create a new issue to track.

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -1615,12 +1615,13 @@ def test_some_instances_not_in_data(pd_es):
 
 
 def test_missing_instances_with_categorical_index(pd_es):
-    instance_ids = [0, 1, 2, 3]
+    instance_ids = [0, 1, 3, 2]
```
rwedge, line 1618: why swap the order?

featuretools/tests/computational_backend/test_feature_set_calculator.py
```diff
@@ -306,8 +306,7 @@ def test_make_agg_feat_of_grandchild_entity(es):
     df = calculator.run(np.array([0]))
     if isinstance(df, dd.DataFrame):
         df = df.compute().set_index('id')
-        df.index = pd.Int64Index(df.index)
-    v = df[agg_feat.get_name()][0]
+    v = df[agg_feat.get_name()].values[0]
```
rwedge, line 309: `loc` or `iloc` might work as well

thehomebrewnerd: Yes, correct. Both of those also work fine. Would you prefer one of those approaches instead?

rwedge: let's use `loc`

thehomebrewnerd: Updated all to use `.loc` instead of `.values`.

PR #1055, opened 2020-07-08
Fix feature visualization for features with '<' or '>' in name

frances-h:
Replace '>' and '<' in feature names with HTML escape characters to fix graphviz parsing error


docs/source/changelog.rst
```diff
@@ -6,6 +6,7 @@ Changelog
     * Enhancements
     * Fixes
         * Remove warnings.simplefilter in feature_set_calculator to un-silence warnings (:pr:`1053`)
+        * Fix feature visualzation for features with '>' or '<' in name (:pr:`1055`)
```
rwedge, line 9: > visualzation

should be visualization

PR #1061, opened 2020-07-10
Add Generate Statistics to Entity Utils

gsheni:
### Pull Request Description
- Add a function to generate statistics for a given dataframe and variable types
- Move `convert_vtypes` into a separate function so it can be used elsewhere


featuretools/tests/variables/test_variables_utils.py
```diff
@@ -64,3 +67,26 @@ def test_invalid_format():
...
+    }
+    converted = convert_vtypes(variable_types)
+    for ky in converted.keys():
+        assert ky in variable_types.keys()
+        assert converted[ky] == Numeric
+    for v in converted.values():
+        assert issubclass(v, Variable)
+
+
+def test_convert_vtypes_unknown():
```
gsheni, line 86: Possibly a repetitive test:
https://github.com/FeatureLabs/featuretools/blob/main/featuretools/tests/entityset_tests/test_entity.py#L190

gsheni:
@rwedge Added option for passing an Entity.

Any suggestions on where to add it to the docs? We could either add it to this [page](https://docs.featuretools.com/en/stable/guides/tuning_dfs.html
) (at the end) OR add it as a FAQ (`How do I generate statistics for my Entity/DataFrame?`)

gsheni:
We are reevaluating our approach to calculating statistics in a more generalizable way, and will be closing the PR in the meantime.

PR #1064, opened 2020-07-14
check input variable type for .delete_variables()

### LINKED ISSUE #1059: Warn user if supplying wrong input to Entity.delete_variables()
If a user calls `Entity.delete_variables()` and passes in a string variable name instead of a list of variable names, the method attempts to loop over all the characters in the supplied string and fails with an error that is not too helpful for the user to understand what happened. We should update this method to raise a `TypeError` with an appropriate message if a string is supplied instead of a list.

Example:
```

es['customers'].delete_variables('age')

Traceback (most recent call last):
KeyError: 'Variable: a not found in entity'
```

systemshift:
### Pull Request Description
close #1059

If someone has better warning string that is more informative let me know.

-----

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -10,7 +10,7 @@ Changelog
     * Changes
         * Make DFS match ``TimeSince`` primitive with all ``Datetime`` types (:pr:`1048`)
         * Change default branch to ``main`` (:pr:`1038`)
-        * Check input variable type for ``.delete_variables()`` (:pr:`1064`)
+        * Raise TypeError if improper input is supplied to ``Entity.delete_variables()`` (:pr:`1064`)
     * Documentation Changes
         * Remove benchmarks folder (:pr:`1049`)
     * Testing Changes
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -10,6 +10,7 @@ Changelog
     * Changes
         * Make DFS match ``TimeSince`` primitive with all ``Datetime`` types (:pr:`1048`)
         * Change default branch to ``main`` (:pr:`1038`)
+        * Check input variable type for ``.delete_variables()`` (:pr:`1064`)
     * Documentation Changes
         * Remove benchmarks folder (:pr:`1049`)
     * Testing Changes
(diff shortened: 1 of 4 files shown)


docs/source/changelog.rst
```diff
@@ -10,6 +10,7 @@ Changelog
     * Changes
         * Make DFS match ``TimeSince`` primitive with all ``Datetime`` types (:pr:`1048`)
         * Change default branch to ``main`` (:pr:`1038`)
+        * Check input variable type for ``.delete_variables()`` (:pr:`1064`)
```
thehomebrewnerd, line 13: Let's update this a bit to say:

Raise `TypeError` if improper input is supplied to ``Entity.delete_variables()`

featuretools/entityset/entity.py
```diff
@@ -414,6 +414,9 @@ def delete_variables(self, variable_ids):
         Remove variables from entity's dataframe and from
         self.variables
         """
+        # check variable is a list
+        if isinstance(variable_ids, str):
+            raise TypeError('strings not accepted, must be of type list')
```
thehomebrewnerd, line 419: Let's change the message to say:

"variable_ids must be a list of variable names"

featuretools/entityset/entity.py
```diff
@@ -414,6 +414,9 @@ def delete_variables(self, variable_ids):
         Remove variables from entity's dataframe and from
         self.variables
         """
+        # check variable is a list
+        if isinstance(variable_ids, str):
```
thehomebrewnerd, line 418: Maybe we should raise this error if the provided input is not a list. Otherwise, it wouldn't get raised if an integer or float was passed, for example. So, let's change the condition to `if not isinstance(variable_ids, list)` instead.

systemshift:
> Looking good, but please also update the docstring for `delete_variables()` to more clearly indicate what is expected for the `variable_ids` argument.

should I add anything else?

featuretools/entityset/entity.py
```diff
@@ -413,7 +413,17 @@ def delete_variables(self, variable_ids):
         """
         Remove variables from entity's dataframe and from
         self.variables
+
+        Args:
+            variable_ids (list): Variables to delete
```
thehomebrewnerd, line 418: Let's be just a little more specific and indicate the list should contain strings:
`variable_ids (list[str]): List of variable to delete`

systemshift: yes, you are correct

PR #1073, opened 2020-07-17
Warn user if primitives are unused during dfs

thehomebrewnerd:
This PR implements changes that will issue a warning if a user supplies primitives to `dfs` that do not get used in generating the feature matrix. If the default primitives are used, this check is not performed - the check only happens when the user supplies a list of primitives.

Added tests to check for warnings in a variety of situations including trans_primitives, agg_primitives, where_primitives, groupby_trans_primitives and custom primitives.

### DIFF
--- a/featuretools/exceptions.py
+++ b/featuretools/exceptions.py
@@ -2,3 +2,9 @@ class UnknownFeature(Exception):
 
     def __init__(self, *args, **kwargs):
         Exception.__init__(self, *args, **kwargs)
+
+
+class UnusedPrimitiveWarning(Warning):
+
+    def __init__(self, *args, **kwargs):
+        Warning.__init__(self, *args, **kwargs)
--- a/featuretools/exceptions.py
+++ b/featuretools/exceptions.py
@@ -5,6 +5,4 @@ class UnknownFeature(Exception):
 
 
 class UnusedPrimitiveWarning(UserWarning):
-
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    pass
--- a/featuretools/exceptions.py
+++ b/featuretools/exceptions.py
@@ -4,7 +4,7 @@ class UnknownFeature(Exception):
         Exception.__init__(self, *args, **kwargs)
 
 
-class UnusedPrimitiveWarning(Warning):
+class UnusedPrimitiveWarning(UserWarning):
 
     def __init__(self, *args, **kwargs):
-        Warning.__init__(self, *args, **kwargs)
+        super().__init__(*args, **kwargs)
--- a/featuretools/synthesis/dfs.py
+++ b/featuretools/synthesis/dfs.py
@@ -5,6 +5,7 @@ from featuretools.entityset import EntitySet
 from featuretools.feature_base import (
     AggregationFeature,
     DirectFeature,
+    FeatureOutputSlice,
     GroupByTransformFeature,
     TransformFeature
 )
@@ -332,13 +333,16 @@ def warn_unused_primitives(unused_primitives):
 
 def get_base_features(features):
     """Take a list of features and add all base features for any direct features"""
-    if any([isinstance(feature, DirectFeature) for feature in features]):
-        new_features = []
-        for feature in features:
-            if isinstance(feature, DirectFeature):
-                new_features.extend(feature.base_features)
-            else:
-                new_features.append(feature)
-        return get_base_features(new_features)
-    else:
+    updated = False
+    new_features = []
+    for feature in features:
+        if isinstance(feature, DirectFeature) or isinstance(feature, FeatureOutputSlice):
+            new_features.extend(feature.base_features)
+            updated = True
+        else:
+            new_features.append(feature)
+
+    if not updated:
         return features
+    else:
+        return get_base_features(new_features)
--- a/featuretools/synthesis/dfs.py
+++ b/featuretools/synthesis/dfs.py
@@ -2,6 +2,7 @@ import warnings
 
 from featuretools.computational_backends import calculate_feature_matrix
 from featuretools.entityset import EntitySet
+from featuretools.exceptions import UnusedPrimitiveWarning
 from featuretools.feature_base import (
     AggregationFeature,
     FeatureOutputSlice,
@@ -309,7 +310,7 @@ def warn_unused_primitives(unused_primitives):
         "This may be caused by a using a value of max_depth that is too small, not setting interesting values, " + \
         "or it may indicate no compatible variable types for the primitive were found in the data."
 
-    warnings.warn(warning_msg)
+    warnings.warn(warning_msg, UnusedPrimitiveWarning)
 
 
 def _categorize_features(features):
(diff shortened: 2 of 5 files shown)


featuretools/synthesis/dfs.py
```diff
@@ -247,6 +256,33 @@ def dfs(entities=None,
...
+        elif isinstance(feature, TransformFeature):
+            trans_features.append(feature)
+
+    trans_unused = get_unused_primitives(trans_primitives, trans_features)
+    agg_unused = get_unused_primitives(agg_primitives, agg_features)
+    groupby_unused = get_unused_primitives(groupby_trans_primitives, groupby_features)
+    where_unused = get_unused_primitives(where_primitives, where_features)
+
+    unused_primitives = [trans_unused, agg_unused, groupby_unused, where_unused]
+    if any(unused_primitives):
```
gsheni, line 283: Should there be a way to turn off the warnings if the user wanted to?

thehomebrewnerd: > Should there be a way to turn off the warnings if the user wanted to?

That thought crossed my mind at one point too. I don't think we have any other controls in place for users to turn off specific warnings in other places though, but this one does feel a bit more informational than some of the others do.

I don't have a strong opinion one way or the other at this point.

featuretools/synthesis/dfs.py
```diff
@@ -247,6 +256,33 @@ def dfs(entities=None,
     features = dfs_object.build_features(
         verbose=verbose, return_variable_types=return_variable_types)
 
+    trans_features = []
+    agg_features = []
+    groupby_features = []
+    where_features = []
+
+    base_features = get_base_features(features)
```
frances-h, line 264: I think this misses the situation where a primitive is used, but the feature that uses it is used as a stacking feature and not directly returned. In practice though, I'm not sure how frequently this edge case would show up.

I managed to force with this example, dfs warns that the `GreaterThanScalar` primitive isn't used:
```python
ft.dfs(
    entityset=es,
    target_entity='régions',
    agg_primitives=['percent_true'],
    trans_primitives=[ft.primitives.GreaterThanScalar(5)],
    primitive_options={'greater_than_scalar': {'include_entities': ['stores']}},
    features_only=True
)
```
The `GreaterThanScalar` primitive is used for one of the returned features, but not as the topmost primitive:
```python
[<Feature: language>,
<Feature: PERCENT_TRUE(customers.loves_ice_cream)>,
<Feature: PERCENT_TRUE(sessions.customers.loves_ice_cream)>,
<Feature: PERCENT_TRUE(stores.num_square_feet > 5)>]
```

featuretools/tests/synthesis/test_dfs_method.py
```diff
@@ -323,6 +309,135 @@ def test_accepts_pd_dateoffset_training_window(datetime_es):
...
+    with pytest.warns(None) as record:
+        dfs(entityset=pd_es,
+            target_entity='customers',
+            groupby_trans_primitives=['cum_sum'],
+            max_depth=1)
+
+    assert not record
+
+
+def test_warns_with_unused_custom_primitives(pd_es):
```
frances-h, line 381: does this test not work with the Dask entityset?

featuretools/synthesis/dfs.py
```diff
@@ -265,3 +301,48 @@ def dfs(entities=None,
...
+    if trans_unused:
+        unused_string += " trans_primitives: {}".format(trans_unused)
+    if agg_unused:
+        unused_string += " agg_primitives: {}".format(agg_unused)
+    if groupby_unused:
+        unused_string += " groupby_trans_primitives: {}".format(groupby_unused)
+    if where_unused:
+        unused_string += " where_primitives: {}".format(where_unused)
+
+    warning_msg = "Some specified primitives were not used during DFS.{}. ".format(unused_string) + \
```
frances-h, line 327: For readability, would it be better to put each unused group indented on their own line? Something like:
> Some specified primitives were not used during DFS:
> &ensp;&ensp;trans_primitives: ['greater_than_scalar']
> &ensp;&ensp;agg_primitives: ['mode', 'num_unique']
> This may be caused by a using a value of max_depth that is too small, not setting interesting values, or it may indicate no compatible variable types for the primitive were found in the data.

featuretools/tests/synthesis/test_dfs_method.py
thehomebrewnerd: As of now, `make_trans_primitive` and `make_agg_primitive` can be used to create a Dask compatible primitive. We could certainly define a custom primitive that would work without those functions, but I wasn't sure it would add much value beyond testing with pandas.

featuretools/synthesis/dfs.py
thehomebrewnerd: Good catch. I'll look into this.

featuretools/synthesis/dfs.py
thehomebrewnerd: Yes, I think this would be better. I'll make this update.

featuretools/synthesis/dfs.py
thehomebrewnerd: I think I was able to fix this by changing the logic I was using to parse through the features. I added a test to make sure no warning was raised in this scenario.

featuretools/synthesis/dfs.py
rwedge: Another possibility -- instead of issuing a warning, include this information in the verbose output?

featuretools/synthesis/dfs.py
```diff
@@ -265,3 +284,73 @@ def dfs(entities=None,
...
+    trans_unused, agg_unused, groupby_unused, where_unused = unused_primitives
+    unused_string = ""
+    if trans_unused:
+        unused_string += "  trans_primitives: {}\n".format(trans_unused)
+    if agg_unused:
+        unused_string += "  agg_primitives: {}\n".format(agg_unused)
+    if groupby_unused:
+        unused_string += "  groupby_trans_primitives: {}\n".format(groupby_unused)
+    if where_unused:
+        unused_string += "  where_primitives: {}\n".format(where_unused)
```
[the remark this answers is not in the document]
thehomebrewnerd: Updated this code block to use iteration instead.

featuretools/synthesis/dfs.py
thehomebrewnerd: I like the idea of including this in the verbose output, based on the nature of the information we are passing to the user.

thehomebrewnerd: A custom `UnusedPrimitiveWarning` class was created by subclassing `UserWarning`. This will allow users to suppress this warning if desired.

featuretools/exceptions.py
```diff
@@ -2,3 +2,9 @@ class UnknownFeature(Exception):
 
     def __init__(self, *args, **kwargs):
         Exception.__init__(self, *args, **kwargs)
+
+
+class UnusedPrimitiveWarning(UserWarning):
+
+    def __init__(self, *args, **kwargs):
```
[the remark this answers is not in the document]
thehomebrewnerd: Updated

PR #1079, opened 2020-07-21
Fixes for Pandas 1.1.0

### LINKED ISSUE #1072: Test featuretools compatibility with Pandas 1.1.0rc0
New Pandas release candidate.  Not up on PyPI at time of writing but we should run our unit tests using the release candidate when it becomes available.

thehomebrewnerd:
Fixes #1072 

- Update `wrangle.py` to change handling for special date offsets to support both older and newer versions of pandas
- Use dask single-threaded scheduler for compute calls to avoid issue with Dask tests failing with threaded scheduler - see [Dask Issue 6454](https://github.com/dask/dask/issues/6454)
- Bump pandas minimum version to `1.0.0` as changes to needed in `wrangle.py` to work with `1.1.0` will cause failures with older versions.
- Bump dask minimum version to `2.12.0` to avoid error that happens with version `2.11.0` which does not implement `dd.to_numeric()`


thehomebrewnerd:
The fixes implemented in this PR were verified to work with pandas versions as old as `v1.0.0`, but they fail on versions older than that.

thehomebrewnerd:
Based on limited testing, this code `chunk_sum = chunk_sum.astype('int64')` identified as a codecov issue appears to be hit with pandas 1.0.0 or 1.0.5, but not with pandas 1.1.0.

rwedge:
For all these `scheduler="single-threaded"` options, could we set that via an environmental variable or something so we don't need to declare it each `compute` call

If not, I think we should make a `SCHEDULER_TYPE` variable and set that to "singlet-threaded", then use `SCHEDULER-TYPE` everywhere instead

requirements.txt
```diff
@@ -1,6 +1,6 @@
 scipy>=0.13.3
 numpy>=1.13.3
-pandas>=0.24.1,<1.1.0
+pandas>=1.0.0
```
rwedge, line 3: are there any changes that require using 1.0 or higher?

thehomebrewnerd: Yes - the changes in `wrangle.py` don't work with older versions because of the way the date offsets are handled. There might be a way to come up with something that would work with both, but I'd have to spend some more time looking into it.

thehomebrewnerd: Might be able to use a different `if-else` construct to check for the two scenarios that were present in the previous code. I'll check.

PR #1094, opened 2020-07-31
Upgrade pandas to 1.1.0 and update .week usage

wsankey:
### Pull Request Description
Addresses resolving the deprecation warning in the primitive cited in this issue: https://github.com/FeatureLabs/featuretools/issues/1091. Tested via ipython (images below) and message is resolved.

### Before
deprecated

### After
nomoremessage

-----


docs/source/changelog.rst
```diff
@@ -17,6 +17,7 @@ Changelog
         * Change default branch to ``main`` (:pr:`1038`)
         * Raise TypeError if improper input is supplied to ``Entity.delete_variables()`` (:pr:`1064`)
         * Updates for compatibility with pandas 1.1.0 (:pr:`1079`, :pr:`1089`)
+        * Upgrade pandas to >=1.1.0 (:pr: `1094`)
```
thehomebrewnerd, line 20: Also, remove the space between `:pr:` and ``1094``

wsankey: Ah that's what it was, thank you!

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -387,7 +390,7 @@ class TimeSince(TransformPrimitive):
         [-1000, -1000000000, -120000000000]
     """
     name = 'time_since'
-    input_types = [Datetime]
+    input_types = [[DatetimeTimeIndex], [Datetime]]
```
rwedge, line 393: this change looks unrelated

wsankey: Yes, unrelated. I had checked out this file from master, not main. Reverted.

wsankey:
The message persists for me in the ipython terminal with this change. Changing the line to `warnings.filterwarnings("ignore", category=FutureWarning)` resolves it.

wsankey:
nomsg

thehomebrewnerd:
> The message persists for me in the ipython terminal with this change. Changing the line to `warnings.filterwarnings("ignore", category=FutureWarning)` resolves it within the ipython terminal. This may be filtered with that line within the module but I haven't checked in that context.

I think that change will cause any FutureWarning afterward to be suppressed - so if the user did something outside of featuretools that would generate a different FutureWarning, the would never see it. We were trying to limit scope to the specific message within featuretools. I tested from a command line script and it seemed to work, but didn't check in ipython.

rwedge:
@wsankey when you were testing before were you defining the Week primitive in the ipython or importing it from the featuretools module?   It's possible if you defined it in memory it did not get filtered due to the `module="featuretools"` part of the filter

wsankey:
Understood. Yes that was the issue. Setting the specific warning message and importing the class resolved this issue. Thank you for the guidance @thehomebrewnerd and @rwedge!

Updated PR to be limited in scope.

rwedge:
We should probably add a test case to confirm the warning is being filtered out

I believe this tests that a warning does not occur

```python
with pytest.warns(None) as record):
    function()

assert not record
```

@wsankey would you be willing to write a test case?

wsankey:
Yes, sure @rwedge I just added a unit test. Not certain this is the proper spot but I manually checked it failed, then passed.

docs/source/changelog.rst
```diff
@@ -17,6 +17,7 @@ Changelog
         * Change default branch to ``main`` (:pr:`1038`)
         * Raise TypeError if improper input is supplied to ``Entity.delete_variables()`` (:pr:`1064`)
         * Updates for compatibility with pandas 1.1.0 (:pr:`1079`, :pr:`1089`)
+        * Set pandas version to pandas>=0.24.1,<2.0.0. Filter week and weekofyear primitive warning. (:pr:`1094`)
```
rwedge, line 20: Let's make this "Set pandas version to pandas>=0.24.1,<2.0.0. Filter pandas deprecation warning in Week primitive."

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -212,6 +214,7 @@ class Week(TransformPrimitive):
 
     def get_function(self):
         def week(vals):
+            warnings.filterwarnings("ignore", message="Series.dt.weekofyear and Series.dt.week have been deprecated.", module="featuretools")
```
rwedge, line 217: this line is a little long (142), can we reduce line length

featuretools/tests/primitive_tests/test_transform_primitive.py
```diff
@@ -83,3 +83,14 @@ def test_age_nan():
     ages = age(dates, time=datetime(2020, 2, 26))
     correct_ages = [10.159, np.nan, 8.159]
     np.testing.assert_array_almost_equal(ages, correct_ages, decimal=3)
+
+
+def test_no_deprecation_message():
```
rwedge, line 88: let's call this "test_week_no_deprecation_message"

wsankey: :+1:

featuretools/primitives/standard/transform_primitive.py
wsankey: :+1:

wsankey:
Thanks! Anyway to rerun these tests, I think it was a coincidence they failed..

wsankey:
OK thank you!

rwedge:
> Thanks! Anyway to rerun these tests, I think it was a coincidence they failed..

Only the code coverage test failed and that was due to a known issue, wouldn't have been impacted by these code changes

PR #1095, opened 2020-07-31
WIP: Fix primitive ordering

tamargrey:
### Pull Request Description
We determined that `dfs` was producing different features depending on the order of the transform primitives list that was provided to it. This PR makes it such that the order of the input transform primitive list has no impact on the features that `dfs` produces. 

Since this PR changes the set of features produced by `dfs`, there were some changes to what features are possible. As of now, features are built up by depth, only using inputs from the previous depth. Self stacking for transform primitives was introduced, but it't not yet clear whether they'll be an option.


gsheni:
Few concerns:
- What determines if `stack_on_self` should be False or True? We should be clear on the definition, and where it applies. It seem the following is currently the case
    - Only 1 Input to Primitive -> stack_on_self is False
    - If the return type is an input type (ex - Numeric input & Numeric output) --> `stack_on_self` is False (see Negate)
- This MR will also put a burden on other primitives (custom and premium primitives), as they will now have to change to `stack_on_self` to be False because in most cases, it doesn't really apply. At the very least, the default `stack_on_self` should be False to minimize disruptions.

gsheni:
One thing you could do is set the PrimitiveBase to have `stack_on_self` be **True**, but then have `stack_on_self`  be **False** for TransformPrimitives.

tamargrey:
We have a couple of decisions to make here:

- Do we want to allow self stacking for transform primitives? (Impacts the number of features and could also touch @gsheni work as he mentioned above.
- Do we want to allow features to be built from multiple levels? (aka f(f(),base - impacts number of features)

Here's some data about the number of features produced:

trans prims used | entity set used | main (current implementation) | no self stacking (n-1 access) | no self stacking (all levels) | self stacking (n-1 access) | self stacking (all levels)
-- | -- | -- | -- | -- | -- | --
['add_numeric', 'negate'] | 5 col (used as example in design doc) | 12 / 21 | 15 | 24 | 27 | 44
['divide_numeric', 'negate'] | 5 col | 18 / 36 | 24 | 33 | 90 | 117
['equal', 'negate'] | 5 col | 34 / 16 | 19 | 34 | 94 | 149
['divide_numeric', 'equal'] | 5 col | 64/19 | 34 | 64 | 169 | 257
['add_numeric', 'negate'] | ecommerce | 762 | 725 | 754 | 733 | 762
['divide_numeric', 'negate'] | ecommerce | 1197 | 1131 | 1188 | 1139 | 1196
['equal', 'negate'] | ecommerce | 2041 | 1718 | 1768 | 7606 | 12451
['divide_numeric', 'equal'] | ecommerce | 2821 | 2498 | 2498 | 8281 | 13076
Half of primitives (28/57) | ecommerce | 11088 | 21852 | 28432 | 35108 | 46483
40/57 primitives | ecommerce | 28709 | 39795 | 51701 | 54419 | 71120
All | ecommerce | 100K | 190K | 210K | 250K | 290K

Some interesting things to note:
- the current implementation can have different results, so that's why it shows multiple numbers of features
- Some runs of the current implementation can produce the f(f(), base) case (otherwise known as the case where primitive being built at level n has access to all previously produced features) and this explains why some of the current implementations produce more features than the updated implementations.
- I don't fully trust the implementation of "all levels" that I used to get these numbers (some of the numbers seem too small and it set off many warnings), so I'd probably want to rethink the implementation if we go with it

tamargrey:
In the design doc I said that this wouldn't take away any features, but since the current implementation can get the `f(f(), base)` case, we actually will lose some features.

tamargrey:
Performance Testing results:

- Self stacking, previous depth:added ~1 hour 
- no self stacking, previous depth: added ~6 minutes 
- Self stacking, all depths: added ~1 hour
- No self stacking, all depths: added ~6 minutes 

Self stacking clearly contributes the most to time added.

PR #1105, opened 2020-08-07
Use CircleCI Matrix Jobs

### LINKED ISSUE #921: Investigate using CircleCI matrix jobs in our config file
Circle CI [added a new syntax](https://circleci.com/blog/circleci-matrix-jobs/) for triggering multiple runs of the same job with different parameters each time.  Our current file has 60 lines of code just for declaring which jobs to run in the workflow and what parameters to use for each job, hopefully this new feature will simplify things.

gsheni:
### Use CircleCI Matrix Jobs
- Closes #921 
- Use CircleCI matrix jobs in config to trigger multiple runs of same job with different parameters

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -125,7 +125,6 @@ jobs:
         type: string
       codecov:
         type: boolean
-        default: false
     executor:
       name: python
       image_tag: << parameters.image_tag >>
@@ -247,7 +246,7 @@ workflows:
           matrix:
             parameters:
               image_tag: ["3.8", "3.7", "3.6"]
-              codecov: [True, True, True]
+              codecov: [false, false, true]
           name: << matrix.image_tag >> unit tests
           requires:
             - << matrix.image_tag >> install featuretools
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -134,7 +134,10 @@ jobs:
       - attach_workspace:
           at: ~/featuretools
       - when:
-          condition: << parameters.codecov >>
+          condition:
+            and:
+              - << parameters.codecov >>
+              - equal: ["3.6", << pipeline.image_tag >>]
           steps:
             - run: |
                 source venv/bin/activate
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -241,73 +233,42 @@ workflows:
...
-          name: "py36 lint test"
-          image_tag: "3.6"
-          requires:
-            - "py36 install featuretools"
-      - lint_test:
-          name: "py37 lint test"
+          matrix:
+            parameters:
+              image_tag: ["3.8", "3.7", "3.6"]
+              codecov: [false, false, true]
```
thehomebrewnerd, line 249: I think this is going to cause 9 tests to be launched - one for each combination of image_tag and codecov.

thehomebrewnerd: Maybe we can remove the `codecov` parameter here and change the `when` condition in the `lint_test` job to use the python version instead, and launch when we are running 3.6.

docs/source/changelog.rst
```diff
@@ -38,6 +38,7 @@ Changelog
         * Implement automated process for checking critical dependencies (:pr:`1045`, :pr:`1054`, :pr:`1081`)
         * Don't run changelog check for release PRs or automated dependency PRs (:pr:`1057`)
         * Fix non-deterministic behavior in Dask test causing codecov issues (:pr:`1070`)
+        * Use CircleCI matrix jobs in config to trigger multiple runs of same job with different parameters (:pr:`1105`)
```
rwedge, line 41: move up to future release section and uncomment future release line

.circleci/config.yml

docs/source/changelog.rst
gsheni: Fixed

.circleci/config.yml
```diff
@@ -152,7 +146,10 @@ jobs:
                 cp unpacked_sdist/.coverage .coverage
                 codecov --required
       - unless:
-          condition: << parameters.codecov >>
+          condition:
+            or:
+              - equal: [ "3.7", << parameters.image_tag >> ]
+              - equal: [ "3.8", << parameters.image_tag >> ]
```
thehomebrewnerd, line 152: I think we want to use the same condition we use above for `when`. My understanding is that `when` gets run if the condition is true, and `unless` gets run when the condition is false: https://circleci.com/docs/2.0/reusing-config/#the-when-step

gsheni: Gotcha, fixed

PR #1110, opened 2020-08-14
Remove single-threaded

rwedge:
New dask release, checking if our problems persist


rwedge:
Looks like it's still there

Issue #1037, opened 2020-06-29
Add dask (and koalas in future) as optional requirement


candalfigomoro:
In the code I see statements such as `import dask.dataframe as dd`, how does it behave when Dask is not installed?

Maybe we should use something like this https://stackoverflow.com/a/52826085 to handle cases where Dask is not installed.

I think we should also add add "dask" as "extras_require" in the setup.py file (see https://github.com/FeatureLabs/featuretools/blob/master/setup.py#L9) and remove it from install requirements, since not everybody needs Dask support.

The same approach should probably be used when Koalas dataframes support will be merged.

thehomebrewnerd:
If Dask is not installed, this will generate a `ModuleNotFound` error when importing `featuretools`. However, starting with version 0.16.0 of Featuretools, Dask will be installed automatically when Featuretools is installed. In that case, in order for the error to be generated, a user would have to uninstall Dask after installing Featuretools.

rwedge:
Since Featuretools uses dask in its general `n_jobs` approach, dask has been a requirement for years before adding dask dataframe support.

Featuretools has an `import_or_raise` method used to handle importing optional requirements in the code.  We could use this for the koalas code and make koalas an extra.
https://github.com/FeatureLabs/featuretools/blob/95117a7da15decf993bde65479f67fe883072754/featuretools/utils/gen_utils.py#L76-L88

candalfigomoro:
> Since Featuretools uses dask in its general `n_jobs` approach, dask has been a requirement for years before adding dask dataframe support.

Ok, I thought it was only necessary for Dask dataframes support.

So this is only relevant when Koalas support will be merged to avoid having Koalas as a hard dependency.

PR #1111, opened 2020-08-14
[Koalas] Make Koalas optional dependency 

frances-h:
Allow for Koalas to be an optional dependency:
- instead of importing directly, try to import and return None if we fail
- custom `is_instance` function which returns False if the module passed in is None, otherwise proceeds with this `isinstance` check as normal

See issue #1037

### DIFF
--- a/featuretools/entityset/deserialize.py
+++ b/featuretools/entityset/deserialize.py
@@ -132,7 +132,9 @@ def read_entity_data(description, path):
     if entity_type == 'dask':
         lib = dd
     elif entity_type == 'koalas':
-        lib = import_or_raise('databricks.koalas', 'Koalas not found, cannot serialize Koalas entityset')
+        import_error = 'Cannot load Koalas entityset - unable to import Koalas. ' \
+                       'Consider doing a pip install with [koalas] extra to install Koalas with pip'
+        lib = import_or_raise('databricks.koalas', import_error)
         read_kwargs['multiline'] = True
         kwargs['compression'] = str(kwargs['compression'])
     else:
--- a/featuretools/entityset/deserialize.py
+++ b/featuretools/entityset/deserialize.py
@@ -4,14 +4,13 @@ import tarfile
 import tempfile
 from pathlib import Path
 
+import dask.dataframe as dd
 import numpy as np
 import pandas as pd
-from dask import dataframe as dd
-from databricks import koalas as ks
 
 from featuretools.entityset.relationship import Relationship
 from featuretools.entityset.serialize import FORMATS
-from featuretools.utils.gen_utils import check_schema_version
+from featuretools.utils.gen_utils import check_schema_version, import_or_raise
 from featuretools.utils.s3_utils import get_transport_params, use_smartopen_es
 from featuretools.utils.wrangle import _is_s3, _is_url
 from featuretools.variable_types import LatLong, find_variable_types
@@ -133,7 +132,7 @@ def read_entity_data(description, path):
     if entity_type == 'dask':
         lib = dd
     elif entity_type == 'koalas':
-        lib = ks
+        lib = import_or_raise('databricks.koalas', 'Koalas not found, cannot serialize Koalas entityset')
         read_kwargs['multiline'] = True
         kwargs['compression'] = str(kwargs['compression'])
     else:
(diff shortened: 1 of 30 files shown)


featuretools/utils/gen_utils.py
```diff
@@ -91,6 +91,43 @@ def import_or_raise(library, error_msg):
...
+    except ImportError:
+        return None
+
+
 def camel_to_snake(s):
     s = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', s)
     return re.sub('([a-z0-9])([A-Z])', r'\1_\2', s).lower()
+
+
+def is_instance(obj, modules, classnames):
```
thehomebrewnerd, line 113: As an alternate approach here would it make more sense to structure this to accept a list of strings of the form "module.classname" directly, instead of having to parse/combine two separate lists for modules and classnames? That way the signature would look a little more like the built-in `isinstance` function 

You could then split the string that is passed and get the module with `globals().get(module_name)`.

So, call the function like this instead: `is_instance(df, ("ks.DataFrame", "dd.DataFrame"))`

Not sure if that is better or not, but just throwing it out for thought.

candalfigomoro:
We could also add pyspark and koalas as extras_require in the setup.py file.

featuretools/utils/gen_utils.py
frances-h: I like that this approach mimics the built-in `isinstance`, but I think in order for it to work we'd need to import everything directly into the utils file (since `globals()` would return the function's global scope, not the callees'). To me, that's less than ideal since it removes some of the flexibility, and could possibly introduce namespace issues. We could get around this by adding the callees global dictionary as an argument to the function, so calling it might look something like this: `is_instance(df, ('ks.DataFrame', 'dd.DataFrame'), globals())`, but I'm not sure this is a great solution either.

thehomebrewnerd: I wasn't thinking through the scoping issues enough when I made the suggestion. I agree that passing in the `globals()` from the calling scope isn't a great solution. So far, I haven't thought of a way to make my proposal work cleanly, so with that sticking with your original implementation might be the best approach here.

featuretools/tests/primitive_tests/test_dask_primitives.py
```diff
@@ -26,13 +26,15 @@ def test_transform(pd_es, dask_es):
                           trans_primitives=trans_primitives,
                           agg_primitives=agg_primitives,
                           max_depth=2,
+                          max_features=20000,
```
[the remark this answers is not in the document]
frances-h: Not directly related, but the circleci tests will sometimes time out without it. I've removed it here since it's not directly relevant though.

setup.py
```diff
@@ -13,6 +13,7 @@
   'nlp_primitives': ['nlp-primitives >= 0.2.2'],
   'autonormalize': ['autonormalize >= 1.0.0'],
   'sklearn_transformer': ['featuretools-sklearn-transformer >= 0.1.0'],
+  'koalas': ['pyspark >= 3.0.0', 'koalas >= 1.0.1']
```
candalfigomoro, line 16: @frances-h  Koalas min version should be 1.1.0

featuretools/utils/gen_utils.py
rwedge: what if we leveraged the fact that `ks` will be `None` if koalas is not installed and that python stops evaluating an "and" statement after the first false to make checks like this:

```python
if ks and isinstance(df, ks.DataFrame):
````
if `ks` is `None` the `isinstance` code won't be run, so we don't need to worry about it

File: featuretools/entityset/deserialize.py [hunk 4]
```diff
@@ -132,7 +132,9 @@ def read_entity_data(description, path):
     if entity_type == 'dask':
         lib = dd
     elif entity_type == 'koalas':
-        lib = import_or_raise('databricks.koalas', 'Koalas not found, cannot serialize Koalas entityset')
+        import_error = 'Cannot load Koalas entityset - unable to import Koalas. ' \
+                       'Consider doing a pip install with [koalas] extra to install Koalas with pip'
```
rwedge, line 136: I think the more direct "featuretools[koalas]" will be clearer than mentioning that it's an extra

PR #1113, opened 2020-08-17
Better test case for normalizing from no time index to time index

tamargrey:
### Updated test case
Creates an entity set with one entity that does not have a `time_index` but does have a column of datetimes. That column, `time` is used via `make_time_index` as the time index of an entity built from normalization.

### DIFF
--- a/featuretools/tests/entityset_tests/test_es.py
+++ b/featuretools/tests/entityset_tests/test_es.py
@@ -967,17 +967,29 @@ def test_normalize_entity_new_time_index_additional_success_check(es):
                         copy_variables=[])
 
 
-def test_normalize_time_index_from_none(es):
-    es['customers'].time_index = None
-    es.normalize_entity('customers', 'birthdays', 'age',
-                        make_time_index='date_of_birth',
-                        copy_variables=['date_of_birth'])
-    assert es['birthdays'].time_index == 'date_of_birth'
-    df = es['birthdays'].df
+def test_normalize_time_index_from_none():
+    df = pd.DataFrame({
+        "id": [0, 1, 2, 3],
+        "A": [5, 4, 2, 3],
+        'time': [datetime(2020, 6, 3), (datetime(2020, 3, 12)), datetime(2020, 5, 1), datetime(2020, 4, 22)]
+    })
+    es = ft.EntitySet("es")
+    es = es.entity_from_dataframe(
+        entity_id="data",
+        dataframe=df,
+        index="id"
+    )
+    assert es['data'].time_index is None
+
+    es.normalize_entity('data', 'normalized', 'A',
+                        make_time_index='time',
+                        copy_variables=['time'])
+    assert es['normalized'].time_index == 'time'
+    df = es['normalized'].df
 
     # only pandas sorts by time index
     if isinstance(df, pd.DataFrame):
-        assert df['date_of_birth'].is_monotonic_increasing
+        assert df['time'].is_monotonic_increasing
 
 
 def test_raise_error_if_dupicate_additional_variables_passed(es):


tamargrey:
A couple things I'm unsure about:
* If there's a good way to use our demo or test data to do this, I'm open to it; I just couldn't find an entity that didn't have `time_index` set but still had a datetime column
* Along those lines, maybe I'm misunderstanding something, and what I described above isn't the test case we're looking for
* Should we also test the case where both `base_entity.time_index` is `None` and we don't set `make_time_index` to anything?

docs/source/changelog.rst
```diff
@@ -20,6 +20,7 @@ Changelog
         * Fix broken downloads badge in README.md (:pr:`1107`)
     * Testing Changes
         * Use CircleCI matrix jobs in config to trigger multiple runs of same job with different parameters (:pr:`1105`)
+        * Change to normalize from base entity with no time index for test_normalize_time_index_from_none (:pr:`1113`)
 
     Thanks to the following people for contributing to this release:
     :user:`gsheni`, :user:`systemshift`, :user:`thehomebrewnerd`
```
gsheni, line 26: @tamargrey put your name here.

tamargrey: changed--thanks for catching!

docs/source/changelog.rst
```diff
@@ -20,9 +20,10 @@ Changelog
         * Fix broken downloads badge in README.md (:pr:`1107`)
     * Testing Changes
         * Use CircleCI matrix jobs in config to trigger multiple runs of same job with different parameters (:pr:`1105`)
+        * Change to normalize from base entity with no time index for test_normalize_time_index_from_none (:pr:`1113`)
```
gsheni, line 23: actually this is in the wrong place. It needs to be above under Future Release. 0.18.1 is already released.

gsheni: Maybe change the description to `Better test case for normalizing from no time index to time index (:pr:`1113`)`

tamargrey: made both of these changes

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -967,17 +967,30 @@ def test_normalize_entity_new_time_index_additional_success_check(es):
...
+    df = pd.DataFrame({
+        "id": [0, 1, 2, 3],
+        "A": [5, 4, 2, 3],
+        'time': [datetime(2020, 6, 3), (datetime(2020, 3, 12)), datetime(2020, 5, 1), datetime(2020, 4, 22)]
+    })
+    es = ft.EntitySet("es")
+    es = es.entity_from_dataframe(
+        entity_id="data",
+        dataframe=df,
+        index="id")
```
rwedge, line 980: the old version of this test is testing both a dask entityset and a pandas one.  We could make a parametrized entityset fixture similar to `es` with this portion of the code, and keep the rest of the code to test the entityset

tamargrey: Added in the fixtures so we now have `params=['dd_normalize_es', 'pd_normalize_es']` for this test-- let me know if it looks ok

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -967,17 +967,50 @@ def test_normalize_entity_new_time_index_additional_success_check(es):
...
+    return es.entity_from_dataframe(
+        entity_id="data",
+        dataframe=df,
+        index="id")
+
+
+@pytest.fixture
+def dd_normalize_es(pd_normalize_es):
+    es = ft.EntitySet(id=pd_normalize_es.id)
+    for entity in pd_normalize_es.entities:
```
rwedge, line 987: we can remove this for loop since we know there is only one entity

tamargrey: Ended up doing `entity = pd_normalize_es.entities[0]`, but I played around with some other ways of doing this that just build the entity set from the dataframe directly, and none of them worked with the dask test. This might be in me not understanding something about either dask or fixtures, but I'm not sure why something like this doesn't work:

```
@pytest.fixture
def normalize_df():
    return pd.DataFrame({
        "id": [0, 1, 2, 3],
        "A": [5, 4, 2, 3],
        'time': [datetime(2020, 6, 3), (datetime(2020, 3, 12)), datetime(2020, 5, 1), datetime(2020, 4, 22)]
    })


@pytest.fixture
def pd_normalize_es(normalize_df):
    df = normalize_df
    es = ft.EntitySet("es")
    return es.entity_from_dataframe(
        entity_id="data",
        dataframe=df,
        index="id")


@pytest.fixture
def dd_normalize_es(normalize_df):
    es = ft.EntitySet("es")
    return es.entity_from_dataframe(
        entity_id="data",
        dataframe=dd.from_pandas(normalize_df, npartitions=2),
        index="id")

@pytest.fixture(params=['dd_normalize_es', 'pd_normalize_es'])
def normalize_es(request):
    return request.getfixturevalue(request.param)
```

rwedge: It might be that dask entities need the variable types for each column set in `entity_from_dataframe` to avoid the cost of trying to infer them.

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -967,17 +967,50 @@ def test_normalize_entity_new_time_index_additional_success_check(es):
...
+    return es.entity_from_dataframe(
+        entity_id="data",
+        dataframe=df,
+        index="id")
+
+
+@pytest.fixture
+def dd_normalize_es(pd_normalize_es):
+    es = ft.EntitySet(id=pd_normalize_es.id)
+    entity = pd_normalize_es.entities[0]
```
rwedge, line 987:
```
entity = pd_normalize_es['data'
```
would be another way to get the entity

tamargrey: I like that better--changing

PR #1121, opened 2020-08-25
Updates to improve Entity.query_by_values performance

thehomebrewnerd:
Updated `Entity.query_by_values` to use a `merge` operation instead of an `isin` operation for Koalas to improve performance. This change also caused row ordering to be lost in some cases, so multiple tests were updated as a result.

This change reduced the time spent in the `query_by_values` function from 235 seconds to 4 seconds for the instacart predict-next-purchase problem.


thehomebrewnerd:
> lgtm 👍

@frances-h I made a couple slight changes after your changes to make Koalas optional. Can you take another quick look and make sure you are still good with this before I merge? Thanks!

PR #1123, opened 2020-08-26
Fix to_encode option in encode_features

### LINKED ISSUE #1115: featuretools, encode_features, date type column, wrong output issue
###  encode_features, date type column, wrong output issue 

-----

I want to encode only Category type column. so I coded like below
`to_encode_column_list = list(result.select_dtypes(include=["category"]).columns)`
`encode_options = kwargs.get("encode_options", {"top_n": sys.maxsize, "verbose": True, "to_encode": to_encode_column_list})`
`result, features = ft.encode_features(result, features, **encode_options)`
It encoded fine for category column. but the problem is, this code changed date type column like below image.
BEFORE RUN ENCODE
image
AFTER RUN ENCODE
image

and date1, date2 column dtypes changed too.
datetime64[ns] --> int64

#### Expected Output
date type columns same as before. not changed.

<details>

Featuretools version: 0.16.0

SYSTEM INFO
-----------
python: 3.7.7.final.0
python-bits: 64
OS: Windows
OS-release: 10
machine: AMD64
processor: Intel64 Family 6 Model 63 Stepping 2, GenuineIntel
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.None

INSTALLED VERSIONS
------------------
numpy: 1.19.1
pandas: 1.0.4
tqdm: 4.46.1
PyYAML: 5.3.1
cloudpickle: 0.8.0
dask: 2.8.1
distributed: 2.18.0
psutil: 5.7.2
Click: 7.0
pip: 20.0.2
setuptools: 44.1.0

</details>

rwedge:
Fixes #1115 

`encode_features` has an option `to_encode` that allows users to specify a list of features they want encoded.  `encode_features` was only partially using the info provided from `to_encode` -- it was only creating encoded features from the features specified by `to_encode`, but it was converting all features to a numeric pandas dtype instead of just the encoded ones.


featuretools/synthesis/encode_features.py
```diff
@@ -159,11 +166,10 @@ def encode_features(feature_matrix, features, top_n=DEFAULT_TOP_N, include_unkno
...
-            continue
-        try:
-            new_X[c] = pd.to_numeric(new_X[c], errors='raise')
-        except (TypeError, ValueError):
-            pass
+        if c in encoded_columns:
+            try:
+                new_X[c] = pd.to_numeric(new_X[c], errors='raise')
+            except (TypeError, ValueError):
+                pass
```
tamargrey, line 173: If we can't figure out why this is here/if it's still necessary, maybe we should add a warning so that we can at least tell when it does end up throwing this error instead of quietly passing

jeff-hernandez: If we do add a warning, I think it'd be good to also test the warning to increase code coverage.

rwedge: Our current tests don't cover this except block - we would need to find an example where this could occur

rwedge: In the interest of getting this fix into today's release, I'm open to doing a separate issue / PR cycle on the usefulness of this block / how we want to handle the exceptions.

frances-h (approved):
lgtm 👍

PR #1126, opened 2020-08-27
Feature selection

tamargrey:
### Feature Selection
In order to provide some insight into the quality of features produced by `dfs`, we're adding in functions that can notify users which features might not be of much importance to a machine learning model. The feature selection tools we are adding at this stage are as follows:

- `find_highly_null_features` - which features have many null values 
- `find_single_value_features` - which features don't have much variance in their values
- `find_highly_correlated_features` - which features are highly correlated to one another 

We're leaving the potential for these to be integrated into `dfs` or to be built up to have a larger and more strict API like that of EvalML's data checks. For now, though, these will just be standalone functions that can be called on a feature matrix produced by `dfs`.

### DIFF
--- a/featuretools/selection/selection.py
+++ b/featuretools/selection/selection.py
@@ -25,64 +25,70 @@ The three functions below use logic from EvalML DataChecks
 '''
 
 
-def find_highly_null_features(feature_matrix, pct_null_threshold=0.95):
+def remove_highly_null_features(feature_matrix, features=None, pct_null_threshold=0.95):
     """
-    Determine features from a feature matrix that have higher than a set threshold
-    of null values.
+        Determine features from a feature matrix that have higher than a set threshold
+        of null values.
 
-    Args:
-        feature_matrix (:class:`pd.DataFrame`): DataFrame whose columns are feature names and rows are instances
-        pct_null_threshold (float): If the percentage of NaN values in an input feature exceeds this amount,
-                that feature will be considered highly-null. Defaults to 0.95.
+        Args:
+            feature_matrix (:class:`pd.DataFrame`): DataFrame whose columns are feature names and rows are instances
+            features (list[:class:`featuretools.FeatureBase`] or list[str], optional): List of features to select
+            pct_null_threshold (float): If the percentage of NaN values in an input feature exceeds this amount,
+                    that feature will be considered highly-null. Defaults to 0.95.
 
-    Returns:
-        List of featxure names that will match columns in the inputted feature matrix
-        where the null percentage was above the set threshold
-
+        Returns:
+            feature matrix (pd.DataFrame): The feature matrix generated. If `features_only` is ``True``,
+                the feature matrix will not be generated.
+            features (list[:class:`.FeatureBase`]): The list of generated feature defintions.
     """
     if pct_null_threshold < 0 or pct_null_threshold > 1:
         raise ValueError("pct_null_threshold must be a float between 0 and 1, inclusive.")
 
     percent_null_by_col = (feature_matrix.isnull().mean()).to_dict()
-    # --> Should we be returning the actual features here? something else?
-    if pct_null_threshold == 0.0:
-        return [f_name for f_name, pct_null in percent_null_by_col.items() if pct_null > 0.0]
+    keep = [f_name for f_name, pct_null in percent_null_by_col.items() if pct_null < pct_null_threshold]
 
-    return [f_name for f_name, pct_null in percent_null_by_col.items() if pct_null >= pct_null_threshold]
+    return _apply_feature_selection(keep, feature_matrix, features)
 
 
-def find_single_value_features(feature_matrix, count_nan_as_value=False):
+def remove_single_value_features(feature_matrix, features=None, count_nan_as_value=False):
     """
-    Determines columns in feature matrix where all the values are the same.
+        Determines columns in feature matrix where all the values are the same.
 
-    Args:
-        feature_matrix (:class:`pd.DataFrame`): DataFrame whose columns are feature names and rows are instances
-        count_nan_as_value (bool): If True, missing values will be counted as their own unique value.
-                    If set to True, a feature that has one unique value and all other data is missing will be
-                    counted as only having a single unique value. Defaults to False.
+        Args:
+            feature_matrix (:class:`pd.DataFrame`): DataFrame whose columns are feature names and rows are instances
+            features (list[:class:`featuretools.FeatureBase`] or list[str], optional): List of features to select
+            count_nan_as_value (bool): If True, missing values will be counted as their own unique value.
+                        If set to True, a feature that has one unique value and all other data is missing will be
+                        counted as only having a single unique value. Defaults to False.
 
-    Returns:
-        List of feature names of all the single value features - note None is not counted
-        as a value, so "single value" is either 0 or 1 unique values
+        Returns:
+            feature matrix (pd.DataFrame): The feature matrix generated. If `features_only` is ``True``,
+                the feature matrix will not be generated.
+            features (list[:class:`.FeatureBase`]): The list of generated feature defintions.
     """
     unique_counts_by_col = feature_matrix.nunique(dropna=not count_nan_as_value).to_dict()
-    return [f_name for f_name, unique_count in unique_counts_by_col.items() if unique_count <= 1]
+
+    keep = [f_name for f_name, unique_count in unique_counts_by_col.items() if unique_count > 1]
+    return _apply_feature_selection(keep, feature_matrix, features)
 
 
-def find_highly_correlated_features(feature_matrix, pct_corr_threshold=0.95, features_to_check=None, features_to_exclude=None):
+def remove_highly_correlated_features(feature_matrix, features=None, pct_corr_threshold=0.95, features_to_check=None, features_to_exclude=None):
     """
-    Determines whether any pairs of features are highly correlated with one another.
+        Determines whether any pairs of features are highly correlated with one another.
 
-    Args:
-        feature_matrix (:class:`pd.DataFrame`): DataFrame whose columns are feature names and rows are instances
-        pct_corr_threshold (float): The correlation threshold to be considered highly correlated. Defaults to 0.95.
-        cols_to_check (set{str}, optional): Set of column names to check whether any pairs are highly correlated.
-                    If null, defaults to checking all columns.
-        cols_to_exclude set[str], optional): Set of colum names to not check correlation between.
-                    If null, will not exclude any columns.
+        Args:
+            feature_matrix (:class:`pd.DataFrame`): DataFrame whose columns are feature names and rows are instances
+            features (list[:class:`featuretools.FeatureBase`] or list[str], optional): List of features to select
+            pct_corr_threshold (float): The correlation threshold to be considered highly correlated. Defaults to 0.95.
+            cols_to_check (set{str}, optional): Set of column names to check whether any pairs are highly correlated.
+                        If null, defaults to checking all columns.
+            cols_to_exclude set[str], optional): Set of colum names to not check correlation between.
+                        If null, will not exclude any columns.
 
-    Returns:
-        list[tuple(str, str)] of column pairs that is correlated above the set threshold.
+        Returns:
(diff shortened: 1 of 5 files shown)


candalfigomoro:
I'm not sure featuretools should contain a feature selection module, as feature selection is an extremely complex matter and it also depends on which model you will use afterwards (e.g. maybe a Gradient Boosting Machine could take advantage of information and interactions that your simple correlation-based method couldn't capture).

However, if you really want to add a selection module, maybe you could take a look at tsfresh's selection module: https://tsfresh.readthedocs.io/en/latest/text/feature_filtering.html

>  To do so, for every feature the influence on the target is evaluated by an univariate tests and the p-Value is calculated.

Afaik it doesn't compute correlation between features since, as you already said, it can blow up quickly.

featuretools/selection/selection.py
```diff
@@ -18,3 +18,102 @@ def remove_low_information_features(feature_matrix, features=None):
...
+
+    correlation_by_pairs = {}
+    # Get all pairs of columns and calculate their correlation
+    for f_name1, col1 in fm_to_check.iteritems():
+        for f_name2, col2 in fm_to_check.iteritems():
+            pair = tuple(sorted((f_name1, f_name2)))
+            if f_name1 == f_name2 or pair in correlation_by_pairs:
+                continue
+
+            correlation_by_pairs[pair] = abs(col1.corr(col2))
```
kmax12, line 114: we should move the threshold check higher up. that way if we drop a feature early, we dont continue to calculate correlations with it, which is an expensive operation.

tamargrey: I definitely agree that we should be more thoughtful here than just "grab every pair," but I'd be worried that this method would keep us from discovering other columns that are correlated to the ones we drop early.

I feel like it'd be really nice to say something like "if `A` and `B` are highly correlated then features built off of them are likely to be highly correlated as well so don't calculate them" but I think that's a different problem than dropping `A` and `B` from further checks.

Another thought I've had is maybe presenting the correlation check as something that should happen on an EntitySet (so basically just the identity features) before running dfs and then using that info to maybe exclude columns from dfs, but that's a whole other thing than running a correlation check after dfs is done.

kmax12: i see your point. it is possible that A/B and B/C are correlated pairs, but A/C aren't. So, if we prematurely drop B, we might miss B/C.

However, i dont think that's a problem. In my mind the goal of this function should be to return a list of features where any pair wont have a correlation about a threshold. There are multiple going to be many sets where that is the case, we just have to return one that is "good".

tamargrey: Oh, that's a good way of thinking about it. Would you expect all of these functions to be returning a list of good features  then, or just this one? Up until now, I'd been thinking of it as if we're telling you about which features might be problematic.

kmax12: ya, i think they should all return `(feature_matrix, feature_defs)` like `ft.encode_features`. that was a user can easily call them and continue the process with the good features

featuretools/selection/selection.py
```diff
@@ -18,3 +18,115 @@ def remove_low_information_features(feature_matrix, features=None):
...
+    return _apply_feature_selection(keep, feature_matrix, features)
+
+
+def remove_single_value_features(feature_matrix, features=None, count_nan_as_value=False):
+    """Removes columns in feature matrix where all the values are the same.
+
+        Args:
+            feature_matrix (:class:`pd.DataFrame`): DataFrame whose columns are feature names and rows are instances.
+            features (list[:class:`featuretools.FeatureBase`] or list[str], optional): List of features to select.
+            count_nan_as_value (bool): If True, missing values will be counted as their own unique value.
```
rwedge, line 58: > If True, missing values will be counted as their own unique value.  If set to True, a feature that has one unique value and all other data is missing will be counted as only having a single unique value.

These sentences seem at odds with each other

tamargrey: Hmm yeah, definitely contradictory. How about:

> If True, missing values will be counted as their own unique value. If set to True, a feature that has one unique value and all other data missing will be removed from the feature matrix.

rwedge: Yeah, that sounds good

featuretools/selection/selection.py
```diff
@@ -18,3 +18,115 @@ def remove_low_information_features(feature_matrix, features=None):
...
+    if features_to_check is None:
+        features_to_check = feature_matrix.columns
+    else:
+        for f_name in features_to_check:
+            assert f_name in feature_matrix.columns, "feature named {} is not in feature matrix".format(f_name)
+
+    if features_to_keep is None:
+        features_to_keep = []
+
+    numeric_dtypes = ['int16', 'int32', 'int64', 'float16', 'float32', 'float64']
```
rwedge, line 103: this overlaps nicely with `PandasTypes._pandas_numerics` in `variable_types/variable.py`, perhaps we should import and use here

featuretools/selection/selection.py
```diff
@@ -18,3 +18,115 @@ def remove_low_information_features(feature_matrix, features=None):
...
+
+    # Get all pairs of columns and calculate their correlation, dropping any columns that are highly correlated
+    dropped = set()
+    for f_name1, col1 in fm_to_check.iteritems():
+        for f_name2, col2 in fm_to_check.iteritems():
+            if f_name1 == f_name2 or f_name1 in dropped or f_name2 in dropped:
+                continue
+
+            if abs(col1.corr(col2)) >= pct_corr_threshold:
+                dropped.add(f_name1)
```
rwedge, line 117: We don't want to drop both features -- we want to keep one and drop the other.

featuretools/selection/selection.py
```diff
@@ -18,3 +18,115 @@ def remove_low_information_features(feature_matrix, features=None):
...
+    numeric_dtypes = ['int16', 'int32', 'int64', 'float16', 'float32', 'float64']
+    boolean = ['bool']
+    numeric_and_boolean_dtypes = numeric_dtypes + boolean
+
+    fm_to_check = (feature_matrix[features_to_check]).select_dtypes(include=numeric_and_boolean_dtypes)
+
+    # Get all pairs of columns and calculate their correlation, dropping any columns that are highly correlated
+    dropped = set()
+    for f_name1, col1 in fm_to_check.iteritems():
+        for f_name2, col2 in fm_to_check.iteritems():
```
rwedge, line 112: Checking if `f_name1` is in `dropped` here would let us skip iterating over the other columns

featuretools/selection/selection.py
tamargrey: Changed to keep the less complex of the two features, which we'll treat as the feature that comes earlier in the list

featuretools/selection/selection.py
tamargrey: The new implementation iterates from most complex to least complex, so when we run into a pair of columns that are highly correlated, we can drop the more complex one and don't need to keep checking its correlation to the rest of features. This also keeps us from dropping any columns that we'll run into later on in the iteration.

featuretools/selection/selection.py
```diff
@@ -18,3 +21,120 @@ def remove_low_information_features(feature_matrix, features=None):
...
+
+
+def remove_single_value_features(feature_matrix, features=None, count_nan_as_value=False):
+    """Removes columns in feature matrix where all the values are the same.
+
+        Args:
+            feature_matrix (:class:`pd.DataFrame`): DataFrame whose columns are feature names and rows are instances.
+            features (list[:class:`featuretools.FeatureBase`] or list[str], optional): List of features to select.
+            count_nan_as_value (bool): If True, missing values will be counted as their own unique value. 
+                        If set to True, a feature that has one unique value and all other
```
frances-h, line 62: Shouldn't this be:
> If set to False, a feature that has one unique value and all other data missing will be removed from the feature matrix.

Unless I'm misunderstanding, one unique value and the rest of the data missing should count as two unique values when NaNs are being counted as a unique value so the feature shouldn't get dropped.

tamargrey: No yeah, you're understanding correctly. It should be False here--thanks for pointing it out!

featuretools/tests/selection/test_selection.py
```diff
@@ -35,11 +36,118 @@ def test_remove_low_information_feature_names(feature_matrix):
...
+    no_tolerance = ft.selection.remove_highly_null_features(fm, pct_null_threshold=0)
+    no_tolerance_cols = set(no_tolerance.columns)
+    diff = set(fm.columns) - no_tolerance_cols
+    assert len(diff) == 3
+    assert 'all_nulls' not in no_tolerance_cols
+    assert 'half_nulls' not in no_tolerance_cols
+    assert 'quarter' not in no_tolerance_cols
+
+
+def test_remove_single_value_features():
```
rwedge, line 84: this test case should include a test without the features

featuretools/selection/selection.py
```diff
@@ -18,3 +21,120 @@ def remove_low_information_features(feature_matrix, features=None):
...
+                                                            f_name not in dropped)]
+    return _apply_feature_selection(keep, feature_matrix, features)
+
+
+def _apply_feature_selection(keep, feature_matrix, features=None):
+    new_matrix = feature_matrix[keep]
+    new_feature_names = set(new_matrix.columns)
+
+    if features is not None:
+        features = [f for f in features if f.get_name() in new_feature_names]
```
rwedge, line 138: have you tested this with a multi-output feature?  this might not handle the case where a feature has multiple columns associated with it

featuretools/selection/selection.py
```diff
@@ -18,3 +21,120 @@ def remove_low_information_features(feature_matrix, features=None):
...
+                If no feature list is provided as input, the feature list will not be returned.
+    """
+    unique_counts_by_col = feature_matrix.nunique(dropna=not count_nan_as_value).to_dict()
+
+    keep = [f_name for f_name, unique_count in unique_counts_by_col.items() if unique_count > 1]
+    return _apply_feature_selection(keep, feature_matrix, features)
+
+
+def remove_highly_correlated_features(feature_matrix, features=None, pct_corr_threshold=0.95, features_to_check=None, features_to_keep=None):
+    """Removes columns in feature matrix that are highly correlated with another column.
```
rwedge, line 77: The assumption that the right-most features are more complex is based around the user not re-ordering the feature list returned by DFS.  We might want to make note of that here.

tamargrey: Ok, adding this section

>         Note:
>             We make the assumption that, for a pair of features, the feature that is further right in the feature matrix
>             produced by ``dfs`` is the more complex one. The assumption does not hold if the order of columns
>             in the feature matrix has changed from what ``dfs`` produces.

featuretools/selection/selection.py
tamargrey: I added a `test_multi_output_selection` and it's performing like I'd expect it to--not sure if the extra test is necessary though. Could maybe just view it as a multiple entity test as well

rwedge: `test_multioutput_selection` would need to pass in the feature list as well

rwedge: We can divide up the feature into FeatureOutputSlice sub-features if some but not all features should be kept

tamargrey: Using `FeatureOutputSlice` to do this

featuretools/selection/selection.py
```diff
@@ -18,3 +21,132 @@ def remove_low_information_features(feature_matrix, features=None):
...
+
+    if features is not None:
+        sliced_features = []
+        for f in features:
+            if (f.number_output_features == 1 or
+                    all([slice_name in new_feature_names for slice_name in f.get_feature_names()])):
+                sliced_features.append(f)
+            else:
+                sliced_features.extend([f[i] for i in range(f.number_output_features)])
+        new_features = [f for f in sliced_features if f.get_name() in new_feature_names]
```
rwedge, line 150: if `new_feature_names` is a set of column names, wouldn't `f.get_name()` fail for a multi-output feature?

frances-h (dismissed):
lgtm 👍

rwedge (approved):
looks good!

PR #1140, opened 2020-09-04
Optimize Unused Primitive Check

tamargrey:
The check in `dfs` for unused primitives has been blowing up at the check for unused primitives when dfs produces many features. This PR makes two changes to help slow that down:

- Instead of returning all of the features split into their primitive-type categorizations, `_categorize_features` now returns just the primitives used categorized into primitive types, meaning that if `sum` was used 4000 times, it only needs one entry in the categorization. This makes the check for unused primitives dependent on the number of possible primitives and not the number of possible features.
- `explored` in `_categorize_features` is changed to be a set so that when we check `explored` to avoid repeating features, it's a constant time check

Note: This PR will go along with a change to the performance tests [FeatureLabs/EntitySets#115](https://github.com/FeatureLabs/EntitySets/pull/115) to include a test that produces more than 50K features so that we can see the impacts of this PR and future changes that might impact `dfs` at larger numbers of features

### DIFF
--- a/featuretools/synthesis/dfs.py
+++ b/featuretools/synthesis/dfs.py
@@ -11,7 +11,6 @@ from featuretools.feature_base import (
 )
 from featuretools.synthesis.deep_feature_synthesis import DeepFeatureSynthesis
 from featuretools.utils import entry_point
-import time
 
 
 @entry_point('featuretools_dfs')
@@ -241,7 +240,6 @@ def dfs(entities=None,
                            target_entity="transactions",
                            features_only=True)
     '''
-    start = time.time()
     if not isinstance(entityset, EntitySet):
         entityset = EntitySet("dfs", entities, relationships)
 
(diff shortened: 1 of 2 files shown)


featuretools/synthesis/dfs.py
```diff
@@ -337,19 +336,19 @@ def get_feature_data(feature):
...
         elif isinstance(feature, TransformFeature):
-            transform.append(feature)
+            transform.add(feature.primitive.name)
 
         feature_deps = feature.get_dependencies()
         if feature_deps:
             dependencies.extend(feature_deps)
 
-        explored.append(feature.get_name())
+        explored.add(feature.get_name())
```
tamargrey, line 351: This won't catch the case where there are multiple versions of a primitive set with different parameters but one was unused. It wasn't caught before, and the question of what are unique primitive and feature names is outside the scope of this PR, so we'll revisit this at a later date via #1141

rwedge (approved):
Looks good

PR #1145, opened 2020-09-08
Update sklearn transformer requirements

tamargrey:
Update to 0.1.1 to go with the remove transform stacking PR changes.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -5,7 +5,7 @@ pytest-xdist==1.26.1
 pytest-cov==2.6.1
 fastparquet>=0.1.6
 graphviz>=0.8.4
-moto[all]>=1.3.13
+moto[all]>=1.3.15
 smart-open>=1.8.4
 boto3>=1.10.45
 composeml>=0.2.0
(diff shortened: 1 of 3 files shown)


rwedge:
Unrelated to changed here but issue seem to be due to update of moto library:

https://github.com/spulec/moto/issues/3290

rwedge (approved):
looks good!

test-requirements.txt
```diff
@@ -5,7 +5,7 @@ pytest-xdist==1.26.1
 pytest-cov==2.6.1
 fastparquet>=0.1.6
 graphviz>=0.8.4
-moto>=1.3.13
+moto[all]>=1.3.13
```
rwedge, line 8: if we're going to set a minimum requirement let's use the latest (1.3.15).  It's the first version to introduce the optional requirements, and since this is the test requirements we can be more aggressive about requiring the latest versions

tamargrey: sounds good

rwedge (approved):

PR #1150, opened 2020-09-09
Serialize multi output features

tamargrey:
`FeatureOutputSlice` features were unable to be serialized, so this PR changes their `base_feature` field of the arguments dictionary to be the feature name instead of an actual feature object itself.

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -5,6 +5,7 @@ Changelog
 **Future Release**
     * Enhancements
     * Fixes
+        * Allow FeatureOutputSlice features to be serialized (:pr:`1150`)
     * Changes
     * Documentation Changes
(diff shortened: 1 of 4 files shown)


docs/source/changelog.rst
```diff
@@ -2,6 +2,18 @@
 
 Changelog
 ---------
+**Future Release**
+    * Enhancements
+    * Fixes
+        * Allow FeatureOutputSlice features to be serialized (:pr:`1050`)
```
rwedge, line 8: pr number should be 1150

rwedge (commented):
Looks like there are some conflicts with the main branch.

I like the new dictionary tests, let's also test an agg with a where feature, and a groupy transform feature

rwedge (commented):
looks good, but out of date with main again

rwedge (approved):

PR #1163, opened 2020-09-21
Remove single-threaded scheduler setting

### LINKED ISSUE #1093: Remove single threaded scheduler from dask_es test fixture after Dask issue is resolved
In PR #1079 a change was made to use the dask single threaded scheduler for tests involving the `dask_es` pytest fixture because of failures in the unit tests that were occurring when using the default threaded scheduler with pandas `1.1.0` installed. 

There is a related [Dask issue #6454](https://github.com/dask/dask/issues/6454) as a result of these failures. 

When the Dask issue is resolved, we should switch back to using the default threaded scheduler for the tests by removing the following line from the `dask_es` fixture in `conftest.py`:
```
dask.config.set(scheduler='single-threaded')
```

rwedge:
Fixes #1093 

Testing if newer version of dask/distributed seems to have resolved issue with scheduler

### DIFF
--- a/docs/source/changelog.rst
+++ b/docs/source/changelog.rst
@@ -10,9 +10,10 @@ Changelog
     * Documentation Changes
         * Update release doc for clarity and to add Future Release template (:pr:`1151`)
     * Testing Changes
+        * Stop requiring single-threaded dask scheduler in tests (:pr:`1163`)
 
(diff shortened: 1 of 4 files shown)


thehomebrewnerd (approved):
Looks good!

PR #1166, opened 2020-09-23
Fix approximate compose feature matrix type  closes #1165

### LINKED ISSUE #1165: Approximate Compose calculate_feature_matrix produces LabelTime object instead of Dataframe
### Approximate Compose calculate_feature_matrix produces LabelTime object instead of Dataframe
-----

When passing in a `composeml.label_times.object.LabelTimes` object as a cutoff-time df and using approximate, 
a LabelTimes object is returned instead of a `pandas.core.frame.DataFrame`. This is a problem because `calculate_feature_matrix` is expected to return a DataFrame. Some code might not support DataFrames and not LabelTimes. This bug is easily overlooked. Since the LabelTimes object is a subclass of DataFrame it has almost all the functionality of a pandas Dataframe and `isinstance(df, pandas.core.frame.DataFrame) == True`.

#### Expected Output
```
    feature_matrix = calculate_feature_matrix(features,
                                              entityset,
                                              cutoff_time=compose_labels,
                                              approximate='1s')
    assert(type(feature_matrix) == pd.core.frame.DataFrame)
```

python: 3.7.7.final.0
python-bits: 64
OS: Darwin
OS-release: 18.7.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: en_US.UTF-8
LOCALE: en_US.UTF-8

INSTALLED VERSIONS
------------------
numpy: 1.18.4
pandas: 1.0.3
tqdm: 4.46.0
PyYAML: 5.3.1
cloudpickle: 1.4.1
dask: 2.16.0
distributed: 2.16.0
psutil: 5.7.0
pip: 20.1.1
setuptools: 41.2.0

tuethan1999:
### Fix approximate compose feature matrix type 
Using a the feature matrix as the base frame instead of cutoff times preserves DataFrame type

-----


featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -184,6 +184,42 @@ def label_func(df):
...
+        window_size='1m'
+    )
+
+    df = es['log'].df
+    df = to_pandas(df)
+    labels = lm.search(
+        df,
+        num_examples_per_instance=-1
+    )
+    labels = labels.rename(columns={'cutoff_time': 'time'})
```
rwedge, line 207: These exact label times are used in separate test as well.  Perhaps we should convert them to a fixture?

rwedge (commented):
Can we add a check to the regular compose label times test confirming the feature matrix is not a compose labeltimes object?

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -184,6 +184,42 @@ def label_func(df):
...
+
+    property_feature = ft.Feature(es['log']['value']) > 10
+
+    feature_matrix = calculate_feature_matrix([property_feature],
+                                              es,
+                                              cutoff_time=labels,
+                                              approximate='1s',
+                                              verbose=True)
+    assert(type(feature_matrix) == pd.core.frame.DataFrame)
```
tamargrey, line 216: could do `isinstance(feature_matrix, pd.core.frame.DataFrame)` if you'd like

tamargrey (commented):
This might be outside of the scope of this PR, but the docstring for create_feature_matrix has the return type as `pd.DataFrame: The feature matrix.` which won't be true for dask or koalas dataframes. It might be worth adding checks like the one in this PR `assert(type(feature_matrix) == pd.core.frame.DataFrame)` for the dask and koalas feature matrices calculated from cfm in `test_cfm_compose` and `test_cfm_dask_compose`

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
rwedge: the compose LabelTimes object is a subclass of pd.DataFrame, so  an isinstance check would still return True if the feature matrix was a LabelTimes object

rwedge (commented):
In the contributors section of the changelog, can you remove the space between your name and the `:user:` tag

rwedge (approved):
Looks good!

PR #1184, opened 2020-10-09
Feature selection guide

### LINKED ISSUE #1167: Add docs page for the feature selection functions
Add a page under Guides on the featuretools docs that acs as as style guide for 
 - remove_highly_null_features
- remove_single_value_features
- remove_highly_correlated_features

tamargrey:
closes #1167 

Adds a guide for the feature selection functions:
- ft.selection.remove_highly_null_features
- ft.selection.remove_single_value_features
- ft.selection.remove_highly_correlated_features


docs/source/guides/feature_selection.ipynb
```diff
@@ -0,0 +1,305 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "import featuretools as ft\n",
+    "\n",
+    "from featuretools.selection import (\n",
+    "    remove_low_information_features,\n",
```
rwedge, line 32: we can remove this import

docs/source/guides/feature_selection.ipynb
```diff
@@ -0,0 +1,305 @@
...
+   "outputs": [],
+   "source": [
+    "remove_single_value_features(fm)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Notice that we've actually lost two of the three columns with null values because, with the function used as it is above, null values are not considered in whether a column has only one unique value. If we'd like to consider `NaN` its own value, we can set `count_nan_as_value` to `True` and we'll see `quarter_null` back in the matrix."
```
rwedge, line 168: > null values are not considered in whether a column has only one unique value

how about "null values are not considered when counting a column's unique values"

docs/source/guides/feature_selection.ipynb
```diff
@@ -0,0 +1,305 @@
...
+    "remove_single_value_features(fm, count_nan_as_value=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Remove Highly Correlated Features\n",
+    "\n",
+    "The last feature selection function we have allows us to remove columns that would likely be redundant to the model we're attempting to build by considering the correlation between pairs of calculated features.\n",
```
rwedge, line 186: To help avoid confusion, let's use columns to refer to columns in the entity/dataset and use features to for the columns of the feature matrix

docs/source/release_notes.rst
```diff
@@ -24,6 +24,7 @@ Release Notes
     * Documentation Changes
         * Update release doc for clarity and to add Future Release template (:pr:`1151`)
         * Use the PyData Sphinx theme (:pr:`1169`)
+        * Add guide for feature selection functions (:pr:`1184`)
```
rwedge, line 27: this should be in the future release section

docs/source/guides/feature_selection.ipynb
tamargrey: Yeah, that sounds better--changed

docs/source/guides/feature_selection.ipynb
tamargrey: Ah, that makes sense--we just had a similar conversation in woodwork about columns vs DataColumns. Changed a bunch of places where `columns` was being used

docs/source/guides/feature_selection.ipynb
```diff
@@ -0,0 +1,304 @@
...
+    "    'corr_1': [99, 100, 77, 33],\n",
+    "    'corr_2': [99, 100, 77, 33],\n",
+    "})\n",
+    "\n",
+    "entities = {\n",
+    "        \"first\": (df1, 'id'),\n",
+    "        \"second\": (df2, 'id',  None, {'words': NaturalLanguage}),\n",
+    "    }\n",
+    "\n",
+    "es = ft.EntitySet(\"data\", entities, )\n",
```
kmax12, line 63: i wonder if we should just use one of the demo datasets? perhaps we need to add a demo dataset with nulls. 

i think that it's better for a examples to use standardize datasets. mostly for the user to have consistency as they read the docs, but also to avoid potential mistakes like the example her missing a relationship between the entities

docs/source/guides/feature_selection.ipynb
```diff
@@ -0,0 +1,304 @@
+{
```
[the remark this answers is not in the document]
tyler3991: I'll create an issue in EvalML to add a guide, and we can link the guide in here later. @kmax12

tamargrey:
Switched to using the flight demo dataset.

Used the `cutoff_time` parameter to dfs to create some null values in the feature matrix.

Also added sections that list the features removed to hopefully make the results of running the functions more clear

rwedge:
@kmax12 thoughts on the dataset change / showing all dropped features?

docs/source/guides/feature_selection.ipynb
```diff
@@ -0,0 +1,375 @@
...
+    "We look at the above feature matrix and decide to remove the highly null features"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ft.selection.remove_highly_null_features(fm)"
```
kmax12, line 84: in the other examples, it looks like we pass in `features=features`, would it make sense to do the same here?

kmax12 (approved):

docs/source/guides/feature_selection.ipynb
tamargrey: We definitely can! The feature list is optional, and it just impacts whether or not we get an updated feature list back, which I'd added in for the others so that we could highlight which features were removed.

If it's better to be consistent here, happy to pass the feature list in here

kmax12: actually, i dont know if consistency is what is most important. rather i would make sure it reads as a clear narrative. so it's fine to start without using the parameter, but then add it in later, just make sure it is explained

so, i'd make sure to clearly call out and explain the usage of the `features` parameter in the guide. perhaps the best way to do that would be a note section like we use elsewhere

image

tamargrey: Got it--added a note when we first use features as a parameter and an extra line highlighting how we use the results

image

kmax12: looks good. thanks!

kmax12 (approved):

rwedge (approved):
Looks good

PR #1191, opened 2020-10-15
Include pyspark and koalas in checkdeps

thehomebrewnerd:
Update `checkdeps` in Makefile to include pyspark and koalas so that future dependency checks will also track the versions for these.


rwedge (approved):
Looks good

PR #1192, opened 2020-10-15
set max koalas version

rwedge:
Some tests failing with koalas 1.3.0, temporarily pinning below 1.3.0 while we investigate

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -2,14 +2,16 @@
 
 Release Notes
 -------------
-.. **Future Release**
+**Future Release**
     * Enhancements
     * Fixes
     * Changes
     * Documentation Changes
     * Testing Changes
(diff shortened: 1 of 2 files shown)


docs/source/release_notes.rst
```diff
@@ -2,14 +2,16 @@
...
 Release Notes
 -------------
-.. **Future Release**
+**Future Release**
     * Enhancements
     * Fixes
     * Changes
     * Documentation Changes
     * Testing Changes
+        * Add ``pyspark`` and ``koalas`` to automated dependency checks (:pr:`1192`)
```
thehomebrewnerd, line 11: Wrong comment?

thehomebrewnerd (approved):
LGTM!

docs/source/release_notes.rst
rwedge: Left an actual change log line.  Maybe when we add koalas 1.3.0 support back in, we can delete this log entry and add this pr number to the new log entry.

PR #1195, opened 2020-10-16
create koalas-requirements.txt

rwedge:
Move koalas-related requirements into a separate requirements file that both setup.py and dev-requirements.txt reference


tamargrey (approved):
lgtm

PR #1201, opened 2020-10-26
English sentence feature descriptions

### LINKED ISSUE #940: Generate human-readable feature definition
For a give feature, return a sentence that explains the feature.  For example, `SUM(transactions.amount)` might become the sentence "For every customer, sum the amount of all transactions for that customer"

frances-h:
Add `describe_feature` function which takes a feature and optionally custom, manually defined metadata and returns an English sentence description of what the feature is:

`<Feature: value>`: The "value".
`<Feature: ABSOLUTE(value)>`: The absolute value of the "value".
`<Feature: sessions.device_name>`: The "device_name" for the instance of "sessions" associated with this instance of "log".
`<Feature: MODE(log.zipcode)>`: The most frequently occurring value of the "zipcode" of all instances of "log" for each "id" in "customers".

Resolves #940


thehomebrewnerd (commented):
Looking good to me - I will take another look tomorrow and dive a little deeper into the logic, and leave any additional comments then.

featuretools/feature_base/feature_descriptions.py
```diff
@@ -0,0 +1,153 @@
...
+                                  feature.entity_id)
+    base_features = feature.base_features
+    # shortens stacked direct features to make it easier to understand
+    while isinstance(base_features[0], ft.DirectFeature):
+        base_feat = base_features[0]
+        base_feat_description = ' the instance of "{}" associated ' \
+            'with'.format(base_feat.relationship_path[-1][1].parent_entity.id)
+        direct_description = base_feat_description + direct_description
+        base_features = base_feat.base_features
+    direct_description = ' of' + direct_description
```
thehomebrewnerd, line 131: Also wondering if this should be `for` instead of `of`. If we changed this along with the change above the output of  the first test in `test_direct_description` would change from:
```
The "loves_ice_cream" of the instance of "customers" associated with this instance of "sessions".
```

to:
```
"loves_ice_cream" for the instance of "customers" associated with this instance of "sessions".
```

Again, not say we should change, just an idea.

thehomebrewnerd (approved):
Just a few more suggestions for consideration. I don't think any of these are blocking issues though, and overall this looks good!

rwedge (commented):
I think it'd be good to add some examples of how to use this functionality to the docs.

Perhaps a brief example, similar to what we did with feature graphs, along with a guide entry about how the user can customize the descriptions?

docs/source/guides/feature_descriptions.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Primitive Templates
+--------------------
+Primitives descriptions are generated using primitive templates. By default, these are defined using the ``description_template`` attribute on the primitive. Primitives without a template default to using the primitive name instead. Primitive description templates are string templates that take input feature descriptions as the positional arguments. These can be overwritten by mapping primitive instances or primitive names to custom templates and passing them into ``describe_feature`` through the ``primitive_templates`` argument.
```
rwedge, line 50: > Primitives without a template default to using the primitive name instead.

This is referring to primitive.name as opposed to the primitIve's class name.  Do you think clarifying that would be necessary?

frances-h: Actually it should probably have a fallback if `primitive.name` isn't defined. We could clarify that it uses `primitive.name` if it's defined, otherwise it just uses the class name.

featuretools/primitives/base/primitive_base.py
```diff
@@ -98,3 +105,33 @@ def get_arguments(self):
...
+                        if len(template) > 2:
+                            raise IndexError('Slice out of range of template')
+                        return template[1].format(*input_column_descriptions,
+                                                  nth_slice=convert_to_nth(slice_index))
+                else:
+                    template = template[0]
+            return template.format(*input_column_descriptions)
+
+        # generic case:
+        name = self.name.upper() if self.name is not None else type(self).__name__
```
rwedge, line 128: do we have a test case for the scenario where the primitive is missing the name attribute?

docs/source/index.rst
```diff
@@ -117,6 +117,35 @@ One of the reasons DFS is so powerful is that it can create a feature matrix for
     feature_matrix_sessions.head(5)
 
 
+Understanding Feature Output
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. ipython:: python
+    :suppress:
+
+    features = {f.get_name(): f for f in features_defs}
```
rwedge, line 125: for continuity reasons, what if we accessed `feature_defs` directly.  something like

```
>  example_feature = feature_defs[11]
> print(example_feature)
> graph_feature(example_feature)
```

docs/source/guides/feature_descriptions.rst
```diff
@@ -0,0 +1,116 @@
...
+Feature Descriptions
+---------------------
+Custom feature definitions will get used in the description in place of the automatically generated description. This can be used to better explain what a variable or feature is, or to provide descriptions that take advantage of a user's existing knowledge about the data or domain. 
+
+.. ipython:: python
+
+    feature_descriptions = {
+        'customers: join_date': 'the date the customer joined'}
+
+    ft.describe_feature(feature_defs[8], feature_descriptions=feature_descriptions)
```
rwedge, line 43: for the continuity angle, I don't think we've feature_defs[8] before this block so printing it out could be helpful

docs/source/guides/feature_descriptions.rst
```diff
@@ -0,0 +1,116 @@
...
+
+Primitive Templates
+--------------------
+Primitives descriptions are generated using primitive templates. By default, these are defined using the ``description_template`` attribute on the primitive. Primitives without a template default to using the ``name`` attribute of the primitive if it is defined, or the class name if it is not. Primitive description templates are string templates that take input feature descriptions as the positional arguments. These can be overwritten by mapping primitive instances or primitive names to custom templates and passing them into ``describe_feature`` through the ``primitive_templates`` argument. 
+
+.. ipython:: python
+
+    primitive_templates = {
+        'sum': 'the total of {}'}
```
rwedge, line 64: since there's only one element in the dictionary, and the text length is short, I think we can make the dictionary a single line

docs/source/guides/feature_descriptions.rst
```diff
@@ -0,0 +1,116 @@
...
+
+
+Here, we create and pass in a custom description of the intermediate feature ``SUM(transactions.amount)``. The description for ``MEAN(sessions.SUM(transactions.amount))``, which is built on top of ``SUM(transactions.amount)``, uses the custom description in place of the automatically generated one. Feature descriptions can be passed in as a dictionary that maps the custom descriptions to either the feature object itself or the unique feature name in the form ``"[entity_name]: [feature_name]"``, as shown above.
+
+Primitive Templates
+--------------------
+Primitives descriptions are generated using primitive templates. By default, these are defined using the ``description_template`` attribute on the primitive. Primitives without a template default to using the ``name`` attribute of the primitive if it is defined, or the class name if it is not. Primitive description templates are string templates that take input feature descriptions as the positional arguments. These can be overwritten by mapping primitive instances or primitive names to custom templates and passing them into ``describe_feature`` through the ``primitive_templates`` argument.
```
rwedge, line 59: Maybe we add a final sentence explaining the below example, mentioning that we're overriding the default "average of" template with our custom "total of" template

rwedge (approved):
Looks good!

PR #1202, opened 2020-10-27
Update last time index code to fix Koalas error

thehomebrewnerd:
### Update add last time index code to fix Koalas error

When Koalas 1.3.0 was released a new error occurred when setting last time indexes as a result. This PR implements changes to fix this error. The updated code was tested with Koalas 1.1.0, 1.2.0 and 1.3.0 and works fine for each. As a result, the upper version restriction on Koalas in `koalas-requirements.txt` was also removed.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -5,9 +5,8 @@ Release Notes
 **Future Release**
     * Enhancements
     * Fixes
-        * Update ``EntitySet.add_last_time_indexes`` to work with Koalas 1.3.0 (:pr:`1202`)
+        * Update ``EntitySet.add_last_time_indexes`` to work with Koalas 1.3.0 (:pr:`1192`, :pr:`1202`)
     * Changes
-        * Restrict koalas version to below 1.3.0 (:pr:`1192`)
         * Keep koalas requirements in separate file (:pr:`1195`)
     * Documentation Changes
         * Added footer to the documentation (:pr:`1189`)
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -5,6 +5,7 @@ Release Notes
 **Future Release**
     * Enhancements
     * Fixes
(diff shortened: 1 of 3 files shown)


docs/source/release_notes.rst
```diff
@@ -5,6 +5,7 @@ Release Notes
 **Future Release**
     * Enhancements
     * Fixes
+        * Update ``EntitySet.add_last_time_indexes`` to work with Koalas 1.3.0 (:pr:`1202`)
```
rwedge, line 8: Can we remove the "Restrict koalas version" entry in the release notes and and add its PR  number to this entry?

thehomebrewnerd: Done!

featuretools/entityset/entityset.py
```diff
@@ -819,7 +819,7 @@ def add_last_time_indexes(self, updated_entities=None):
                         lti.index = entity.df[entity.index].copy()
                         lti = lti.apply(lambda x: None)
                     elif is_instance(entity.df, ks, 'DataFrame'):
-                        lti = ks.Series(index=lti.to_list()).rename(lti.name)
```
rwedge, line 822: I see we're removing the rename section of this line - is it no longer necessary?

thehomebrewnerd: Good question - nothing in the tests seem to rely on the series being named, so either it isn't necessary or we aren't covering a situation in our tests where it is...

rwedge: instead of the rename call, can we use the `name` parameter of the ks.Series constructor?

thehomebrewnerd: Had to use it on the pd.Series instead - ks.Series won't accept a name if the data is a pd.Series, but it looks like it takes on the name of the pd.Series.

rwedge (approved):
Looks good

PR #1207, opened 2020-10-30
Custom variable descriptions

frances-h:
Allow custom descriptions to be set directly on a variable and used in feature descriptions.
- Descriptions default to `'the "[variable name]"'` and can be set with `variable.description`
- Descriptions must be set prior to feature generation in order for them to propagate


rwedge (commented):
do we directly test serializing a variable with a custom description?

rwedge (commented):
let's document the `description` argument in the `Variable` docstring

rwedge (approved):
Looks good!

PR #1217, opened 2020-11-07
Calculating direct features use default value if parent missing

seriallazer:
### Pull Request Description
(replace this text with your description)

-----


seriallazer:
Some of the windows-unittests are failling, but I am not sure if they are related to this PR. Can someone please chip in?

seriallazer:
Hi @kmax12 @rwedge 
All the tests are passing but the CIrcleCI reports the status as "Unauthorized".
Can someone please help me get around this?

rwedge:
@seriallazer we recently updated our CI and it may be having issues with forked PRs.

I think we should also test this behavior when using koalas and dask dataframes

seriallazer:
@rwedge Do you mind pointing out what do I need to do to get these changes through?

rwedge:
@seriallazer sure, take a look at [this code](https://github.com/alteryx/featuretools/blob/ddbba0257d7a5077d1d8f97dde1ff56564276c55/featuretools/tests/conftest.py#L76-L156) as an example.  It creates a `diamond_es` fixture that, when passed to a test case, results versions of the test case being run, one with each dataframe type.  The [to_pandas](https://github.com/alteryx/featuretools/blob/main/featuretools/tests/testing_utils/es_utils.py) method is useful for converting the result back to a pandas dataframe so the assertion check can be done in pandas.

seriallazer:
@rwedge I've made the required fixture changes for dask/koalas, could you please review now

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -1831,30 +1841,17 @@ def test_calc_feature_matrix_with_cutoff_df_and_instance_ids(es):
...
-    })
-
-    es = ft.EntitySet()
-    es.entity_from_dataframe(entity_id="transactions",
-                             dataframe=transactions,
-                             index="id")
-    es.entity_from_dataframe(entity_id="sessions",
-                             dataframe=sessions,
-                             index="id")
+def test_calculate_feature_matrix_returns_default_values(pd_transactions_defaultvalue_es):
```
rwedge, line 1844: the fixture should be `transactions_defaultvalue_es` so that all 3 versions get tested, not just pandas

rwedge (commented):
Thanks for making those changes!  Left a few comments

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
seriallazer: The dask dataframes looses index information when we convert from pandas to dask. I found this is a common problem outside as well: https://stackoverflow.com/questions/60070170/losing-index-information-when-using-dask-dataframe-to-parquet-with-partitionin

Do we have a way to resolve this internally?

thehomebrewnerd: @seriallazer Are you referring to the `to_pandas(feature_matrix)` call on line 1853? If so here is some info that might help. For Dask and Koalas inputs, Featuretools does not set the index of the returned feature matrix like it does for pandas. Also row order is not guaranteed with Dask and Koalas. The `to_pandas` helper function has a couple optional parameters that can help in this case. Since you are setting the index values to be `id` when creating the entityset, you can update your call to this instead to set the index and make sure it is sorted in the original order:
```python
feature_matrix = to_pandas(feature_matrix, index='id, sort_index=True)
```

featuretools/tests/conftest.py
```diff
@@ -156,6 +156,66 @@ def ks_diamond_es(pd_diamond_es):
...
+
+    es.add_relationship(ft.Relationship(es["sessions"]["id"], es["transactions"]["session_id"]))
+    return es
+
+
+@pytest.fixture
+def dask_transactions_defaultvalue_es(pd_transactions_defaultvalue_es):
+    entities = {}
+    for entity in pd_transactions_defaultvalue_es.entities:
+        entities[entity.id] = (dd.from_pandas(entity.df.reset_index(), npartitions=4), entity.index, None, entity.variable_types)
```
thehomebrewnerd, line 192: Also, I don't think you need the `.reset_index()` call here. The entity.df does not appear to have the index set, so calling reset_index here is giving an extra column named `index` that we don't need for anything.

seriallazer: The index information was lost even without the `.reset_index()` call.

thehomebrewnerd: I guess maybe I don't understand what you mean by the index information being lost - we typically do not rely on any information in the dataframe index for Dask or Koalas dataframes, but instead we use a column in the dataframe that is defined to be the index - in this case `id`. However, when we create a feature matrix from a pandas dataframe, the feature matrix dataframe has its index set to the index column. For Dask and Koalas the dataframe index is not set (so the index values are not meaningful), but rather there is a normal column in the dataframe that contains the index values.

thehomebrewnerd: Because of this difference, and the fact that we can't rely on the row order in our Dask or Koalas feature matrix, we have the ability to set the index and sort the index via our `to_pandas` test utility function.

As I was trying to understand this issue, I was able to get `test_calculate_feature_matrix_returns_default_values` to pass with pandas, Dask and Koalas inputs if I removed the `reset_index` call here along with making the other changes I mentioned in `test_calculate_feature_matrix.py`.

Are you still seeing problems after making those changes? Maybe I am missing something still.

seriallazer: This fixed the issue for me: `feature_matrix = to_pandas(feature_matrix, index='id', sort_index=True)`
@thehomebrewnerd thanks for your patience, its my first time working with dask/koalas

thehomebrewnerd: No problem - I know it can be a little confusing because of the different ways things are handled between pandas/Dask and Koalas. Glad to hear that fixed the issue!

seriallazer: @thehomebrewnerd Quick question a bit off-topic: why does CircleCI ask me access to my private repos even to check the status of my CI/CD jobs for this repo?

thehomebrewnerd: @seriallazer Not sure exactly - are you seeing that as part of the process of authenticating with Github? 

@rwedge might know a bit more of the details there than I do.

seriallazer: @thehomebrewnerd yes part of the Github authentication.

Also most of the tests have passed but the CircleCI status shows "Unauthorized". Not really sure whats the process for CircleCI honestly (couldnt find any documentation about it).

rwedge: @seriallazer I believe it's due to this being a forked PR.  We're planning to switch to a different CI process soon, so we could wait for that.  An alternative we could do is make a copy of your branch on this repo and trigger a test build on that version of the branch.

rwedge:
@seriallazer glad you were able to work through adding the dask / koalas tests!  You should also add an entry about this PR to the release notes file (docs/source/release_notes.rst) under the Fixes category, and add yourself to the list of contributors for the upcoming release

seriallazer:
@rwedge thanks for pointing out the release-notes, missed that.

I've updated the release-notes, but it seems to have merge conflicts right now

featuretools/tests/conftest.py
seriallazer: @rwedge I see, sure I can do that going forward.
Although for this PR, the status still says "Unauthorized" even though all tests pass now.
Maybe someone with appropriate perms can retrigger the full-set?

rwedge:
@seriallazer, I pushed the current state of your branch to the repo so we could run the CI.  One test case is failing when using the koalas dataframes, the `test_get_filepath` test in `test_transform_features.py`.  My hunch is one of the features created by DFS in the second half of the test is highlighting an issue with either the new code path or the default value of a primitive.  I was able to replicate this test failure locally as well.

seriallazer:
> @seriallazer, I pushed the current state of your branch to the repo so we could run the CI. One test case is failing when using the koalas dataframes, the `test_get_filepath` test in `test_transform_features.py`. My hunch is one of the features created by DFS in the second half of the test is highlighting an issue with either the new code path or the default value of a primitive. I was able to replicate this test failure locally as well.

Unfortunately I am unable to reproduce this on Mac OSX :|

seriallazer:
> @seriallazer, I pushed the current state of your branch to the repo so we could run the CI. One test case is failing when using the koalas dataframes, the `test_get_filepath` test in `test_transform_features.py`. My hunch is one of the features created by DFS in the second half of the test is highlighting an issue with either the new code path or the default value of a primitive. I was able to replicate this test failure locally as well.

@rwedge Roy, can u please push the branch to run the full CI on it again. Seems like I am not a contributor to the repo which is why I cant trigger it. 
Btw, how does one become a contributor to this repo?

rwedge:
@seriallazer hopefully today or tomorrow we'll have migrated to a new CI system that works better with forked PRs

I wonder if we could be more selective on what columns we need to call fillna on - if default value is `np.nan` and the Pandas datatype already set missing values to `np.nan`, we wouldn't need to check that column.

I was able to track down the issue with Koalas - the "sessions.customers.loves_ice_cream" feature, which is boolean type.

rwedge:
@seriallazer try updating your branch with the latest changes from main, we've switched CI

thehomebrewnerd:
@seriallazer I know this PR has been stalled for a bit, but are you interested in helping us finish this off? If you don't have time, or are not interested, I can work on it from here, but wanted to check first.

thehomebrewnerd:
Closing and replacing with #1312

PR #1218, opened 2020-11-09
DFS string support for custom primitives closes #1077

### LINKED ISSUE #1077: [Feature Request] Simplify variable type and custom primitive parameterization
It would be easier if the `variable_types` parameter in `entity_from_dataframe()` and if the custom primitives passed to `dfs()` is a string rather than the class itself

tuethan1999:
### Pull Request Description
Allows user to pass in the snakecase string of a custom primitive's class name into DFS instead of passing in the actual class. The implementation of this is similar to how variable typestrings are handled.
-----


rwedge:
Closing due to inactivity, but linking to #1077 for reference when this gets worked on again

PR #1222, opened 2020-11-09
Update PR url in release notes check-in test

rwedge:
Update the url so that the check for users updating the release notes works with the new org

Also sets a specific version range for urllib3 to deal with conflicting requirements from other dependencies

Not sure why the ci checks are on the job level now

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -271,7 +271,7 @@ jobs:
           password: $DOCKERHUB_PASSWORD
     steps:
       - checkout
-      - run: cat docs/source/release_notes.rst | grep ":pr:\`${CIRCLE_PULL_REQUEST##https://github.com/FeatureLabs/featuretools/pull/}\`"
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -271,7 +271,7 @@ jobs:
           password: $DOCKERHUB_PASSWORD
     steps:
       - checkout
-      - run: cat docs/source/release_notes.rst | grep ":pr:\`${CIRCLE_PULL_REQUEST##https://github.com/FeatureLabs/featuretools/pull/}\`"
+      - run: cat docs/source/release_notes.rst | grep ":pr:\`${CIRCLE_PULL_REQUEST##https://github.com/alteryx/featuretools/pull/}\`"
```
gsheni, line 274: Not sure if you want to fix other instances of this as well:
- https://github.com/alteryx/featuretools/search?q=https%3A%2F%2Fgithub.com%2FFeatureLabs

gsheni (approved):
looks good for fixing the CI

.circleci/config.yml
rwedge: I'll make an issue for it, we can do a separate PR

PR #1224, opened 2020-11-11
Replace FeatureLabs Github URLs with alteryx

### LINKED ISSUE #1223: Replace FeatureLabs github org urls with Alteryx org urls
Now that we've moved repos we should update the URLs in the documentation accordingly.

rwedge:
- Replaces URLs in various places in the documentation
- Fixes #1223


gsheni (approved):
lgtm

PR #1232, opened 2020-11-16
Fix for koalas 1.4.0

thehomebrewnerd:
### Fix for koalas 1.4.0
Update `astype` casting in entityset deserialize to work with koalas 1.4.0. Also was tested with koalas 1.1.0 and verified to work. Removed upper restriction on koalas version from koalas-requirements.txt


rwedge (approved):
Looks good!

PR #1234, opened 2020-11-16
Update dependency check with new version 

gsheni:
- The dependency action will start failing with the following message:
```text
The set-env command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by 
setting the ACTIONS_ALLOW_UNSECURE_COMMANDS environment variable to true. For more information see: 
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```
- The fix updates the version of the create-pull-request action we use. We updated our fork to include v3.


dsherry (approved):
👍 thanks!

rwedge (approved):

PR #1236, opened 2020-11-19
Upgrade pip at start of featuretools[complete] install test

rwedge:
Testing new pip resolver with CI


rwedge:
No issues I could see with using the new resolver, but I did notice we don't upgrade pip during the featuretools complete test, so I'll pivot this PR to add the pip upgrade step.

jeff-hernandez (approved):

PR #1251, opened 2020-12-08
Move query_by_values to EntitySet

### LINKED ISSUE #1244: Move query_by_values from Entity to EntitySet
- the `query_by_values` functions on Entity needs to moved to EntitySet with the Woodwork integration

thehomebrewnerd:
### Move query_by_values to EntitySet
Moves the `query_by_values` method from `Entity` to `EntitySet`. Also moves the method `_vals_to_series` to `EntitySet` as this method was only called from inside `query_by_values`.

Closes #1244

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -17,7 +17,7 @@ Release Notes
 
 **Breaking Changes**
 
-* ``Entity.query_by_values`` has been removed and replaced with ``EntitySet.query_by_values``, with an
+* ``Entity.query_by_values`` has been removed and replaced by ``EntitySet.query_by_values`` with an
     added ``entity_id`` parameter to specify which entity in the entityset should be used for the query.
 
 **v0.22.0 Nov 30, 2020**
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -15,6 +15,11 @@ Release Notes
     Thanks to the following people for contributing to this release:
     :user:`jeff-hernandez`, :user:`rwedge`, :user:`thehomebrewnerd`
 
+**Breaking Changes**
+
+* ``Entity.query_by_values`` has been removed and replaced with ``EntitySet.query_by_values``, with an
+    added ``entity_id`` parameter to specify which entity in the entityset should be used for the query.
+
 **v0.22.0 Nov 30, 2020**
     * Enhancements
         * Allow variable descriptions to be set directly on variable (:pr:`1207`)
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,12 +6,13 @@ Release Notes
     * Enhancements
     * Fixes
     * Changes
+        * Move ``query_by_values`` method from ``Entity`` to ``EntitySet`` (:pr:`1251`)
     * Documentation Changes
     * Testing Changes
         * Use repository-scoped token for dependency check (:pr:`1245`:)
 
     Thanks to the following people for contributing to this release:
-    :user:`jeff-hernandez`
+    :user:`jeff-hernandez`, :user:`thehomebrewnerd`
 
 **v0.22.0 Nov 30, 2020**
     * Enhancements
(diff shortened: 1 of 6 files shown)


rwedge (commented):
A couple if statements in `_vals_to_series` aren't getting hit.  Are there input combinations to featuretools that would result in these lines getting hit?

thehomebrewnerd:
> A couple if statements in `_vals_to_series` aren't getting hit. Are there input combinations to featuretools that would result in these lines getting hit?

I can take a closer look. This code was copied directly from the code that was in Entity, so I'm assuming it was uncovered there as well. I'll confirm.

If it was uncovered previously, do you want to try to add new tests to fix the coverage as part of this PR?

thehomebrewnerd:
> > A couple if statements in `_vals_to_series` aren't getting hit. Are there input combinations to featuretools that would result in these lines getting hit?
> 
> I can take a closer look. This code was copied directly from the code that was in Entity, so I'm assuming it was uncovered there as well. I'll confirm.
> 
> If it was uncovered previously, do you want to try to add new tests to fix the coverage as part of this PR?

After further review, these lines also are uncovered currently in main. Looking at the code flow, I don't think it is possible for these lines to ever get hit during normal circumstances. 

First of all, we specify that instance ids passed to `calculate_feature_matirx` should be of type `list` in the docstring. However, even if a single value were passed, this would get converted to a series here: https://github.com/alteryx/featuretools/blob/753615c89360910768811680a7a71e08ab83ad79/featuretools/computational_backends/calculate_feature_matrix.py#L191-L193

If the user does not specify instance ids, we grab a series from the dataframe here: https://github.com/alteryx/featuretools/blob/753615c89360910768811680a7a71e08ab83ad79/featuretools/computational_backends/calculate_feature_matrix.py#L178-L184

So, given those two scenarios I don't think it is possible for our normal code flow to ever call `query_by_values` and consequently `_vals_to_series` with instances that are anything but a `pd.Series`. We do have several tests that call `query_by_values` passing a list of instance values.

Unless I have missed something, I do not believe that the uncovered code is necessary. The documentation says that the user should pass a list for instance ids. Even if a single value is passed, this will get converted to a series by the time `_vals_to_series` gets called. I haven't tested what would happen if a DataFrame is passed, but since we are not telling users that is supported in the documentation, I'm not sure we need to handle that anyway.

With all this in mind we could:
1. Delete the uncovered code 
2. Add a couple new tests that use a single instance value and a DataFrame

I think the first options is better as it removes what appears to be dead code. Thoughts?

rwedge:
It's dead code as far as the featuretools code flow is concerned, but those input options are described in the `query_by_values` docstring as valid inputs.  Since `query_by_values` is a public method, it could be getting used.  I know someone on the slack channel ended up using query by values for their purposes.

What if we added some tests now, and made an issue about removing those input options in the future?  See if we get any feedback.

thehomebrewnerd:
> It's dead code as far as the featuretools code flow is concerned, but those input options are described in the `query_by_values` docstring as valid inputs. Since `query_by_values` is a public method, it could be getting used. I know someone on the slack channel ended up using query by values for their purposes.
> 
> What if we added some tests now, and made an issue about removing those input options in the future? See if we get any feedback.

The tests should be easy enough to add. As a side note to all this,  `query_by_values` is not currently listed in the API Reference. Do you want it added there as well?

rwedge:
Let's leave it out for now

docs/source/release_notes.rst
```diff
@@ -6,13 +6,14 @@ Release Notes
     * Enhancements
     * Fixes
     * Changes
+        * Move ``query_by_values`` method from ``Entity`` to ``EntitySet`` (:pr:`1251`)
```
rwedge, line 9: Let's also add a "Breaking Changes" section to highlight the move

thehomebrewnerd: Added

rwedge (approved):
Looks good once updated to even with main

PR #1269, opened 2020-12-09
Remove add interesting values from Entity

thehomebrewnerd:
### Remove add interesting values from Entity
This PR moves all functionality for adding interesting values to EntitySet. Since `add_interesting_values` was already a method on EntitySet, one change was made. An additional parameter was added to the method to specify an entity_id, if the user wants to add interesting values for a single Entity - equivalent to the previous call to `Entity.add_interesting_values`. If this parameter is not passed, `EntitySet.add_interesting_values` will add interesting values for all entities in the entityset.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,8 +6,8 @@ Release Notes
     * Enhancements
     * Fixes
     * Changes
-        * Remove ``add_interesting_values`` from ``Entity`` (:pr:`1269`)
         * Move ``query_by_values`` method from ``Entity`` to ``EntitySet`` (:pr:`1251`)
+        * Remove ``add_interesting_values`` from ``Entity`` (:pr:`1269`)
     * Documentation Changes
     * Testing Changes
         * Use repository-scoped token for dependency check (:pr:`1245`:, :pr:`1248`)
@@ -18,11 +18,11 @@ Release Notes
 
 **Breaking Changes**
 
+* ``Entity.query_by_values`` has been removed and replaced by ``EntitySet.query_by_values`` with an
+    added ``entity_id`` parameter to specify which entity in the entityset should be used for the query.
 * ``Entity.add_interesting_values`` has been removed. To add interesting values for a single
     entity, call ``EntitySet.add_interesting_values`` and pass the id of the entity for
     which to add interesting values in the ``entity_id`` parameter. 
-* ``Entity.query_by_values`` has been removed and replaced by ``EntitySet.query_by_values`` with an
-    added ``entity_id`` parameter to specify which entity in the entityset should be used for the query.
 
 **v0.22.0 Nov 30, 2020**
     * Enhancements
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,13 +6,20 @@ Release Notes
     * Enhancements
     * Fixes
     * Changes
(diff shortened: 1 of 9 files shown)


featuretools/tests/entityset_tests/test_es.py
```diff
@@ -1514,3 +1514,18 @@ def test_entityset_init():
...
     assert es['cards'].__eq__(es_copy['cards'], deep=True)
     assert es['transactions'].__eq__(es_copy['transactions'], deep=True)
+
+
+def test_add_interesting_values_verbose_output(caplog):
+    es = ft.demo.load_retail(nrows=200)
+    es['order_products'].convert_variable_type('quantity', ft.variable_types.Discrete)
+    logger = logging.getLogger('featuretools')
+    logger.propagate = True
+    logger = logging.getLogger('featuretools.entityset')
```
rwedge, line 1524: if we use a separate variable for the second logger we don't need to call getLogger again later

docs/source/release_notes.rst
```diff
@@ -6,13 +6,20 @@ Release Notes
...
         * Fix install error during docs CI test (:pr:`1250`)
 
     Thanks to the following people for contributing to this release:
-    :user:`jeff-hernandez`, :user:`rwedge`
+    :user:`jeff-hernandez`, :user:`rwedge`, :user:`thehomebrewnerd`
+
+**Breaking Changes**
+
+* ``Entity.add_interesting_values`` has been removed. To add interesting values for a single
+    entity, call `EntitySet.add_interesting_values` and pass the id of the entity for
```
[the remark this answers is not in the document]
thehomebrewnerd: Fixed

featuretools/tests/entityset_tests/test_es.py
thehomebrewnerd: Updated

featuretools/entityset/entityset.py
```diff
@@ -909,19 +909,65 @@ def add_last_time_indexes(self, updated_entities=None):
...
     ###########################################################################
 
-    def add_interesting_values(self, max_values=5, verbose=False):
+    def add_interesting_values(self, max_values=5, verbose=False, datatable_id=None):
         """Find interesting values for categorical variables, to be used to generate "where" clauses
 
         Args:
             max_values (int) : Maximum number of values per variable to add.
             verbose (bool) : If True, print summary of interesting values found.
+            datatable_id (str) : The DataTable for which to add interesting values. If not specified
```
rwedge, line 918: Maybe this should be a higher level discussion, but should this be `datatable_id` or `entity_id`?  Are we still keeping entities as a concept, and if so when should we refer to a thing as an entity or the datatable?

thehomebrewnerd: I've gone back and forth on this. Maybe the safe thing for now is to leave it as `entity_id` and if the concept of an `Entity` goes away we can rename to something more appropriate at that time.

gsheni: Keep `entity_id` as the parameter to maintain status quo? We did this in the other MR as well: 
- https://github.com/alteryx/featuretools/pull/1276#discussion_r541235051

thehomebrewnerd: Updated to use `entity_id` here.

rwedge (approved):
Looks good

PR #1273, opened 2020-12-10
Fix Percent Unique

### LINKED ISSUE #1271: Infer Variable Types Always Returns Numeric For Input Numeric Fields, Regardless of Number of Unique Values.
https://github.com/alteryx/featuretools/blob/753615c89360910768811680a7a71e08ab83ad79/featuretools/utils/entity_utils.py#L85

This bit of logic creates a sample based on the min of 10,000 and the number of unique values in a column. For example, in the titanic dataset the Pclass column has three unique values. Therefore the sample will have three rows in it. 

Then it gets the unique values in the sample set, which can be any combination of [1, 2, 3]. For example, when I ran this the sample had [2, 3, 3] in it. 

Then it calculates the percent_unique by taking the sample.size (three in this case) and dividing by the number of unique values in the sample set, which can be 1, 2 or 3

**percent_unique will never be less than .05 since the number of elements in the sample set is always greater than the number of unique values in it. The conditional if percent_unique < .05 will always be greater than or equal to 1.0 so the type will always be Numeric.**

This was identified by Roland in the code, and Alan has identified this behavior in the Feature Types tool as well.

jeff-hernandez:
Closes #1271 by dividing the unique value count by sample size.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -5,6 +5,7 @@ Release Notes
 **Future Release**
     * Enhancements
     * Fixes
+        * Fix percent unique for inferred variable types (:pr:`1273`)
     * Changes
     * Documentation Changes
     * Testing Changes
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -5,7 +5,7 @@ Release Notes
 **Future Release**
     * Enhancements
     * Fixes
-        * Fix percent unique for inferred variable types (:pr:`1273`)
+        * Fix logic for inferring variable type from unusual dtype (:pr:`1273`)
     * Changes
         * Move ``query_by_values`` method from ``Entity`` to ``EntitySet`` (:pr:`1251`)
     * Documentation Changes
--- a/featuretools/utils/entity_utils.py
+++ b/featuretools/utils/entity_utils.py
@@ -87,7 +87,7 @@ def infer_variable_types(df, link_vars, variable_types, time_index, secondary_ti
                 .sample(min(10000, df[variable].nunique(dropna=False)))
 
             unique = sample.unique()
-            percent_unique = sample.size / len(unique)
+            percent_unique = len(unique) / sample.size
 
             if percent_unique < .05:
                 inferred_type = vtypes.Categorical
--- a/featuretools/utils/entity_utils.py
+++ b/featuretools/utils/entity_utils.py
@@ -12,7 +12,7 @@ from featuretools.utils.gen_utils import import_or_none, is_instance
 ks = import_or_none('databricks.koalas')
 
 
-def infer_variable_types(df, link_vars, variable_types, time_index, secondary_time_index):
+def infer_variable_types(df, link_vars, variable_types, time_index, secondary_time_index, random_state=0):
     '''Infer variable types from dataframe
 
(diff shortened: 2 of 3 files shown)


featuretools/utils/entity_utils.py
```diff
@@ -12,7 +12,7 @@
 ks = import_or_none('databricks.koalas')
 
 
-def infer_variable_types(df, link_vars, variable_types, time_index, secondary_time_index):
+def infer_variable_types(df, link_vars, variable_types, time_index, secondary_time_index, random_state=0):
```
rwedge, line 15: instead of adding a `random_state` option, can we use two data columns that will have the desire distribution regardless of the random state used for sampling?

jeff-hernandez: We can make data columns that most likely yield the desired distribution, but the test would still be a bit random, so there's no guarantee that the sample would always yield the desired distribution.

rwedge: If we make a data column where 99% of the rows are one value and 1% a second value, any sample will report under 5% unique values.

Likewise if we make a data column where every row is unique, we're guaranteed the sample will be above 5% unique values

rwedge: Actually I see what you're saying, but the idea that we would only sample 2 rows if nunique is 2 is confusing me

thehomebrewnerd: I haven't looked at this in detail, but wouldn't we want `n` to be `10000` if the dataframe contains more than 10000 rows, or just the length of the dataframe if we have less than 10000? Basically if we have less than 10000 rows, use all the data, otherwise sample it?

thehomebrewnerd: That's the logic that gets used for Categorical inference. Would this be different for some reason?

jeff-hernandez: The current implementation uses the number of unique values as the sample size. 

```python
nunique = column.nunique(dropna=False))
sample = column.sample(n=min(10000, nunique)
percent_unique = len(sample.unique()) / sample.size
```

If there were only two unique values, the sample size would only be two and we couldn't get below 5%.

rwedge: I think we should use the same logic to get the sample size that we use earlier in this method

```python
min(10000, len(df[variable]))
```

thehomebrewnerd: Isn't using the number of unique values as the sample size the root of this issue? For columns that have a large number of rows and a small number of unique values, using only a small number of rows for the sample will be inadequate for getting a good estimation of the percent unique values in the column.

jeff-hernandez: Yeah, makes sense. I'll remove the random state.

docs/source/release_notes.rst
```diff
@@ -5,6 +5,7 @@ Release Notes
 **Future Release**
     * Enhancements
     * Fixes
+        * Fix percent unique for inferred variable types (:pr:`1273`)
```
rwedge, line 8: How about "Fix logic for inferring variable type from unusual dtype"

rwedge (approved):
Looks good

PR #1279, opened 2020-12-15
Fix install command for add-ons

### LINKED ISSUE #1198: Update Install documentation
The Featuretools [install documentation](https://featuretools.alteryx.com/en/stable/install.html) should be updated to include a section on installing the Koalas add-on along with all the other add-ons that are mentioned.

Also, code formatting for the sklearn transformer install command needs to be fixed to match the other sections:
image

gsheni:
- Fixes #1198
- Proof:
Screen Shot 2020-12-15 at 6 12 15 PM

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -16,37 +16,37 @@ Add-ons
 --------
 You can install add-ons individually or all at once by running::
 
-    python -m pip install featuretools[complete]
+    python -m pip install "featuretools[complete]"
 
 Update checker:
     Receive automatic notifications of new Featuretools releases::
 
-        python -m pip install featuretools[update_checker]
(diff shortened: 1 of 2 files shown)


docs/source/install.rst
```diff
@@ -16,44 +16,44 @@ Add-ons
 --------
 You can install add-ons individually or all at once by running::
 
-    python -m pip install featuretools[complete]
+    python -m pip install "featuretools[complete]"
```
gsheni, line 19: The reason for the quotes is because some shells (like zsh) use square brackets for pattern matching while other shells (like bash) don't. Adding quotes makes it work for all shells.

rwedge (approved):
Looks good!

PR #1286, opened 2020-12-28
Fix for pandas 1.2.0

thehomebrewnerd:
### Fix for pandas 1.2.0

pandas 1.2.0 included a new warning which caused unit test failures in Featuretools. This PR removes the usage of `.ravel` which was causing a warning to be raised, when none were expected in the tests.


rwedge (approved):
Looks good

PR #1290, opened 2020-12-30
Allow passing entities without relationships to CFM

### LINKED ISSUE #1287: relationships is not truly optional in ft.calculate_feature_matrix()
When I pass just a single entity in ft.calculate_feature_matrix() then I need to pass relationships=().

However, in ft.dfs() I don't have to pass relationships=().

It could be better to continue working with relationships=None to such cases.

Example:

```
new_data_set = ft.calculate_feature_matrix(feature_defs, entities={ "passengers" : (new_raw_data_set, id_column, None, variable_types) }, relationships=())
```

Could be better as:

```
new_data_set = ft.calculate_feature_matrix(feature_defs, entities={ "passengers" : (new_raw_data_set, id_column, None, variable_types) })
```

rwedge:
Fixes #1287

Before, `calculate_feature_matrix` would  check that both`entities` and `relationships` were not `None` before creating an entityset with them, which required passing an empty iterable of relationships in order to work with a single entity.  This PR changes that by removing that check and relying on the `EntitySet` init method to handle valdiating the `entities` and `relationships`

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -142,8 +142,7 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
     # handle loading entityset
     from featuretools.entityset.entityset import EntitySet
     if not isinstance(entityset, EntitySet):
-        if entities is not None and relationships is not None:
-            entityset = EntitySet("entityset", entities, relationships)
+        entityset = EntitySet("entityset", entities, relationships)
 
     if any(isinstance(es.df, dd.DataFrame) for es in entityset.entities):
         if approximate:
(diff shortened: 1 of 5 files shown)


featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -142,8 +142,7 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
     # handle loading entityset
     from featuretools.entityset.entityset import EntitySet
     if not isinstance(entityset, EntitySet):
-        if entities is not None and relationships is not None:
```
tamargrey, line 145: Just wondering--why not keep the check for whether entities is not None here? Is there a world in which having no entities wouldn't end up erroring? 

If not, it might be nice to catch this with its own error instead of `Entity transactions does not exist in entityset`, which is where it gets caught in the test below.

rwedge: Sure, if we catch there we can give a more descriptive error

tamargrey (approved):
lgtm!

PR #1293, opened 2021-01-02

machineFL:


thehomebrewnerd:
Something seems to have introduced non-deterministic behavior in one of the tests when using Dask. Probably not related to the tqdm update in this PR, but not sure what caused the issue. I am working on an update to fix this now, and will submit a PR to resolve soon.

thehomebrewnerd (approved):

PR #1294, opened 2021-01-04
Fix Non-Deterministic Dask Test

thehomebrewnerd:
### Fix Non-Deterministic Dask Test

A recent update has caused some non-deterministic behavior in `test_calculate_feature_matrix::test_entities_relationships` when running with Dask input. This PR adds a sort operation on the output feature matrices to make sure they are always ordered the same before the equality assertion.

The source of the non-determinism was not established, but was first discovered due to an automated dependency update check when tqdm was updated, although that does not seem to be the source of the problem.


docs/source/release_notes.rst
```diff
@@ -2,19 +2,21 @@
...
 Release Notes
 -------------
-.. **Future Release**
+**Future Release**
     * Enhancements
     * Fixes
     * Changes
     * Documentation Changes
     * Testing Changes
+        * Fix non-deterministic Dask test ``test_calculate_feature_matrix::test_entities_relationships`` (:pr:`1294`)
```
rwedge, line 11: Including the exact test name is maybe a bit too specific for a testing entry - if people want that info they can check the PR

thehomebrewnerd: Updated

rwedge (approved):
Looks good

PR #1303, opened 2021-01-12
Reorganize Tests

thehomebrewnerd:
### Reorganize Tests
As the test suite has continued to grow, several tests seemed to be located in directories/files that did not align with the specific item being covered by the test. This PR attempt to resolve some of these issue and better place some tests in the directory or file that most closely aligns with the objective of the test.


featuretools/tests/synthesis/test_dask_dfs.py
```diff
@@ -0,0 +1,369 @@
...
+
+    fm, _ = ft.dfs(entityset=pd_es,
+                   target_entity="data",
+                   trans_primitives=primitives_list,
+                   max_depth=1)
+
+    pd.testing.assert_frame_equal(fm, dask_fm.compute().set_index('id').loc[fm.index])
+
+
+def test_secondary_time_index():
```
rwedge, line 288: let's change this to `test_dask_entityset_secondary_time_index` to be consistent

Or we could drop the `dask_entityset` from all of the test case names and rely on the  test name + file name to infer, but I think that approach would be harder to quickly parse from pytest results

featuretools/tests/synthesis/test_koalas_dfs.py
```diff
@@ -0,0 +1,379 @@
...
+    fm, _ = ft.dfs(entityset=pd_es,
+                   target_entity="data",
+                   trans_primitives=primitives_list,
+                   max_depth=1)
+
+    pd.testing.assert_frame_equal(fm, ks_fm.to_pandas().set_index('id').loc[fm.index])
+
+
+@pytest.mark.skipif('not ks')
+def test_secondary_time_index():
```
rwedge, line 298: same as with dask test, let's make this `test_ks_entityset_secondary_time_index`

featuretools/tests/synthesis/test_dask_dfs.py
thehomebrewnerd: renamed

featuretools/tests/synthesis/test_koalas_dfs.py
thehomebrewnerd: renamed

rwedge (approved):
Looks good!

PR #1306, opened 2021-01-13
Update test_deep_feature_synthesis to remove ft.dfs calls

### LINKED ISSUE #1305: Update tests in test_deep_feature_synthesis.py
### Update tests in test_deep_feature_synthesis.py

The tests in `tests/synthesis/test_deep_feature_synthesis.py` appear to be testing various aspects of the DeepFeatureSynthesis object. There are a few tests in this file that generate a list of features using a call to `ft.dfs`. These tests could be updated to use a call to `DeepFeatureSynthesis.build_features()` instead. This would better align with the other tests in the file, and would improve test clarity as they are not testing any other aspects of the dfs method.

thehomebrewnerd:
### Update test_deep_feature_synthesis to remove ft.dfs calls

Update several tests in `test_deep_feature_synthesis.py` to use DeepFeatureSynthesis object to generate features instead of calling ft.dfs with features_only=True.

Closes #1305

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -8,10 +8,8 @@ Release Notes
     * Changes
     * Documentation Changes
     * Testing Changes
-        * Fix non-deterministic Dask test (:pr:`1294`)
         * Unpin python-graphviz package on Windows (:pr:`1296`)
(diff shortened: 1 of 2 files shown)


rwedge (approved):
Looks good

PR #1312, opened 2021-01-20
Fill default values when parent is missing

### LINKED ISSUE #682: When calculating direct features use default value if parent missing

thehomebrewnerd:
### Fill default values when parent is missing
Replaces PR #1217 

Closes #682


featuretools/tests/conftest.py
```diff
@@ -152,6 +152,64 @@ def ks_diamond_es(pd_diamond_es):
     return ft.EntitySet(id=pd_diamond_es.id, entities=entities, relationships=relationships)
 
 
+@pytest.fixture(params=['pd_transactions_defaultvalue_es', 'dask_transactions_defaultvalue_es', 'ks_transactions_defaultvalue_es'])
+def transactions_defaultvalue_es(request):
```
rwedge, line 156: thoughts on shortening this to `default_value_es` and adjusting the other fixtures to match?

thehomebrewnerd: Yeah, that would be cleaner for sure.

thehomebrewnerd: @rwedge Updated

rwedge (approved):
Looks good!

PR #1313, opened 2021-01-21
Fix broken docs links

### LINKED ISSUE #1299: Update URL for installing graphviz in error message
### Update URL for installing graphviz in error message

If a user tries to run `es.plot()` without graphviz installed an error is raised displaying a URL linking to instructions for installing graphviz. This URL is not valid and needs to be fixed.

https://github.com/alteryx/featuretools/blob/3da2a768705a823f54b9ed76851a787985c24ae7/featuretools/utils/plot_utils.py#L6

thehomebrewnerd:
### Fix broken docs links
Several links in the documentation and code were still linking to `docs.featuretools.com`. Updated these links to point to the corresponding file in the new documentation at `featuretools.alteryx.com`.

Closes #1299

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,9 +80,9 @@ customer_id                                                                     
 
 [5 rows x 69 columns]
 ```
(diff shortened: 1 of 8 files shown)


rwedge (approved):
Looks good

Issue #1176, opened 2020-10-05
Add Python 3.9 support


gsheni:
- With the release of Python 3.9, we should test our CI in an environment with Python 3.9, and make sure Featuretools supports it. 
- https://www.python.org/downloads/release/python-390/

gsheni:
Tracking dependencies that support 3.9
- [ ] scipy
- [ ] numpy
- [ ] pandas
- [ ] tqdm
- [ ] pyyaml
- [ ] cloudpickle
- [ ] distributed
- [ ] dask[dataframe]
- [ ] psutil
- [ ] click
- [ ] pyspark
- [ ] koalas
- [ ] codecov
- [ ] matplotlib
- [ ] nbconvert
- [ ] nbsphinx
- [ ] Sphinx

tyler3991:
#1331 will address 3.9 w/ koalas

PR #1318, opened 2021-01-25
Add python 3.9 support and CI test

gsheni:
- #1176

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -42,6 +42,9 @@ jobs:
     strategy:
       matrix:
         python_version: ["3.6", "3.7", "3.8", "3.9"]
+        exclude:
+          # excludes python 3.9 until Koalas support released
+          - python_version: "3.9"
     steps:
       - name: Set up python ${{ matrix.python_version }}
         uses: actions/setup-python@v2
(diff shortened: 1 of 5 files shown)


gsheni:
- I added a skip for installing Koalas on Python 3.9 (which means unit tests for Koalas as skipped).
- Waiting on numba to add support for Python 3.9
  - https://github.com/numba/numba/issues/6345

.github/workflows/tests.yml
```diff
@@ -128,8 +134,12 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        python_version: ["3.6", "3.7", "3.8"]
+        python_version: ["3.6", "3.7", "3.8", "3.9"]
         optional_libraries: ["minimal", "optional"]
+        exclude:
+          # excludes python 3.9 until Koalas support released
```
rwedge, line 140: if we exclude the optional tests entirely, we won't test dask on 3.9.  

we could try adding a step like this
```
pip uninstall -r koalas-requirements.txt -y
```

gsheni:
We will revisit testing on Python 3.9 once numba supports it. I am currently running into this [numba](https://github.com/numba/numba/issues/6345) issue.

PR #1320, opened 2021-01-27
Remove unnecessary `pytest.skip` calls

thehomebrewnerd:
### Remove unnecessary `pytest.skip` calls

Removes multiple unnecessary calls to pytest.skip which was being used to skip Koalas tests on Windows. However, the current test environment is not installing Windows, so the tests were being skipped anyway due to a `pytest.importorskip` call that is in place.


rwedge (approved):
Looks good!

PR #1323, opened 2021-01-28
Fix Entityset.__eq__ bug and test coverage

### LINKED ISSUE #1281: EntitySet.__eq__ bug fix and test coverage improvement
The `EntitySet.__eq__` method contains code that will never evaluate to False. This could cause two entitysets that have different relationships defined to evaluate as equal when they should be not equal.

This code block
```python
for r in other.relationships:
    if r not in other.relationships:
        return False
```

should be changed to:
```python
for r in self.relationships:
    if r not in other.relationships:
        return False
```

Also, there are other lines in the `__eq__` method that are not currently covered by the existing unit tests. New tests should be added to improve coverage for this method.

thehomebrewnerd:
### Fix `EntitySet.__eq__` bug and test coverage

Fixes #1281 

Fixes a bug in `EntitySet.__eq__` where relationships were not being compared properly between `self` and `other`. Also adds a new test to cover lines in `__eq__` and `__ne__` that were previously uncovered

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,7 +6,7 @@ Release Notes
     * Enhancements
     * Fixes
         * Calculate direct features uses default value if parent missing (:pr:`1312`)
-        * Fix bug in ``EntitySet.__eq__`` (:pr:`1323`)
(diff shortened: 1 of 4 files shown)


rwedge (approved):
Looks good!

PR #1332, opened 2021-02-03
Update docs for custom primitives

jeff-hernandez:
The documentation for defining custom primitives was updated to use the class-based approach instead of the function-based approach. The updated pages are listed below:
- [Feature Primitive](https://featuretools.alteryx.com/en/doc_custom_naming/getting_started/primitives.html)
- [Advanced Custom Primitives Guide](https://featuretools.alteryx.com/en/doc_custom_naming/guides/advanced_custom_primitives.html)


docs/source/guides/advanced_custom_primitives.ipynb
```diff
@@ -0,0 +1,233 @@
...
+    "Features with Multiple Outputs\n",
+    "==============================\n",
+    "\n",
+    "Some calculations can output multiple features. With custom primitives, you can stack and create new features based on more than one output from a single feature.\n",
+    "\n",
+    "\n",
+    "Case Count Example\n",
+    "~~~~~~~~~~~~~~~~~~\n",
+    "\n",
+    "Make a primitive to count the upper case and lower case letters of a text. First, derive a new transform primitive class from :class:`TransformPrimitive`. The primitive will take in a text column as the input and return two numeric columns as the output, so set the input type as :class:`NaturalLanguage`, the return type as :class:`Numeric`, and :code:`number_output_features` to two. Then, override :meth:`get_function` to return the implementation as a function and :meth:`generate_names` to define custom naming for the multiple outputs."
```
rwedge, line 116: Same as with above example, could use some more context about the usefulness of the example

docs/source/guides/advanced_custom_primitives.ipynb
```diff
@@ -0,0 +1,233 @@
...
+    "..             def predict(x):\n",
+    "..                 return model.predict(x)\n",
+    "..             return predict\n",
+    "\n",
+    "\n",
+    ".. The ``get_filepath`` method is used to find the location of the trained model.\n",
+    "\n",
+    ".. .. note::\n",
+    "\n",
+    "..     The primitive loads the model within the `get_function` method, but outside of the `score` function.  This way the model is loaded from disk only once when the Featuretools backend requests the primitive function instead of every time `score` is called."
```
rwedge, line 207: Let's remove this commented out example completely

docs/source/guides/advanced_custom_primitives.ipynb
```diff
@@ -0,0 +1,233 @@
...
+    "Primitives With Additional Arguments\n",
+    "====================================\n",
+    "\n",
+    "Some features require more advanced calculations than others. Advanced features usually entail additional arguments to help output the desired value.  With custom primitives, you can use primitive arguments to help you create advanced features.\n",
+    "\n",
+    "\n",
+    "String Count Example\n",
+    "~~~~~~~~~~~~~~~~~~~~\n",
+    "\n",
+    "Make a primitive to count the number of times a specific string value occurs inside a text. First, derive a new transform primitive class from :class:`TransformPrimitive`. The primitive will take in a text column as the input and return a numeric column as the output, so set the input type as :class:`NaturalLanguage` and the return type as :class:`Numeric`. The specific string value is the additional argument, so define it as a *keyword* argument inside :meth:`__init__`. Then, override :meth:`get_function` to return the implementation as a function."
```
rwedge, line 42: For this paragraph, I think it could use a little more explanation on why we're using this example / what we gain from the example before diving into the specific of how we create the class.

docs/source/guides/advanced_custom_primitives.ipynb
```diff
@@ -0,0 +1,233 @@
...
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "Features with Multiple Outputs\n",
+    "==============================\n",
+    "\n",
+    "Some calculations can output multiple features. With custom primitives, you can stack and create new features based on more than one output from a single feature.\n",
```
rwedge, line 110: >  For this paragraph, I think it could use a little more explanation on why we're using this example / what we gain from the example before diving into the specific of how we create the class.

This could sentence could be interpreted as if custom primitives are necessary in order to stack on features that have multiple outputs.  I think we should rephrase so that isn't implied.

rwedge (commented):
In the advanced custom guide, let's add an example of writing custom column names for a primitive with multiple output columns

docs/source/getting_started/primitives.ipynb
```diff
@@ -0,0 +1,382 @@
...
+    "        return absolute"
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    "Above, we created a new transform primitive that can be used with Deep Feature Synthesis by deriving a new primitive class using :class:`TransformPrimitive` and returning the implementation as a python function from :meth:`get_function`. Additionally, we set the input data types that the primitive applies to and the return data type.\n",
```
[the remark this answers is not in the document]
jeff-hernandez: I tried referring to the logic of the function as the implementation. What about just:
> returning a python function from `get_function`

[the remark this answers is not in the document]
jeff-hernandez: This is another alternative that comes to mind:
> returning a function that calculates the feature from `get_function`

rwedge (approved):
Looks good!

PR #1344, opened 2021-02-25
Add warning about dropping python 3.6 in the future

rwedge:
When featuretools is imported, checks to see if the environment is python 3.6.  If so, warns that Featuretools will be dropping python 3.6 support in an upcoming release.


gsheni (approved):
looks good

PR #1345, opened 2021-02-26
Update release branch naming instructions

rwedge:
Updating the release instructions so that the release branches do not use the names the release tags do

Updates the regex pattern for release branches in the release notes update check


gsheni (approved):
lgtm

PR #1351, opened 2021-03-08
Fixes for Koalas 1.7.0

thehomebrewnerd:
### Pull Request Description
This PR implements fixes for two issues for breaking changes that occurred with Koalas 1.7.0:
- Update `variable.py` to make sure `datetime64[ns]` is used properly in place of `datetime`
- Add a pytest.skip to a test that was failing due to Koalas throwing additional warnings. This is similar to the approach used in another test, but could warrant further investigation to see if the extra Koalas warning can be eliminated.

### DIFF
--- a/featuretools/tests/synthesis/test_dfs_method.py
+++ b/featuretools/tests/synthesis/test_dfs_method.py
@@ -321,6 +321,8 @@ def test_does_not_warn_with_stacking_feature(pd_es):
 
 
 def test_warns_with_unused_where_primitives(es):
+    if ks and any(isinstance(e.df, ks.DataFrame) for e in es.entities):
+        pytest.skip('Koalas throws extra warnings')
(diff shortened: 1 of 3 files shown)


featuretools/tests/synthesis/test_dfs_method.py
```diff
@@ -321,6 +321,8 @@ def test_does_not_warn_with_stacking_feature(pd_es):
 
 
 def test_warns_with_unused_where_primitives(es):
+    if ks and any(isinstance(e.df, ks.DataFrame) for e in es.entities):
```
rwedge, line 324: Can you add an issue for investigating the extra warning(s)

thehomebrewnerd: Created #1352 for this.

rwedge (approved):
Looks good!

PR #1354, opened 2021-03-08
Update performance check

rwedge:
Updates the performance check workflow to supply the commit hash as a parameter.

See https://github.com/alteryx/featuretools/actions/runs/633052796 for successful run


frances-h (approved):
looks good 👍

PR #1370, opened 2021-03-29
Refactor Relationship Creation

### LINKED ISSUE #1365: Refactor Relationship Creation
The current method for creating Relationships and adding them to an EntitySet will not work when Entities are replaced with Woodwork dataframes. The current approach is to specify a relationship by passing a parent variable and a child variable, and those variables have knowledge of the entity to which the belong and the entity has a reference to the Entityset. With Woodwork a data column does not have a reference to its parent dataframe and a dataframe does not have a reference to its EntitySet. Since relationship objects store information about the Entity and EntitySet they are associated with, we need a way to pass this information and we can't get it from a Woodwork column.

To resolve this issue we can refactor the relationship creation methods as follows:

`Relationship.__init__`
- Update the relationship initializer to accept five parameters: entityset (ft.EntitySet), parent_dataframe_name (str), parent_column_name (str), child_dataframe_name (str), child_column_name (str). From these inputs all the necessary Relationship attributes can be set

`EntitySet.add_relationship`
- Update this method to accept four parameters instead of a `Relationship` object: parent_dataframe_name (str), parent_column_name (str), child_dataframe_name (str), child_column_name (str). This method can then handle creating and storing the Relationship object and will prevent users from ever having to create a Relationship object directly.

`EntitySet.add_relationships`
- Update this method to accept a list of 4-tuples instead of a list of Relationship objects. Each tuple should be of the form `(parent_dataframe_name (str), parent_column_name (str), child_dataframe_name (str), child_column_name (str))` and can then call `EntitySet.add_relationship` for each tuple in the list, unpacking the values first.

These changes will allow creation of entityset relationships without ever creating a `Relationship` object, and will work properly with Woodwork dataframes and columns replacing Entities and Variables.

thehomebrewnerd:
### Refactor Relationship Creation

Closed #1365 

This PR refactors the process of creating relationships so that it no longer requires `Variable` objects. In addition `EntitySet.add_relationship` and `EntitySet.add_relationships` were also updated so relationships can be added to an EntitySet without the need for first creating a `Relationship` object.


thehomebrewnerd:
@rwedge I kept the current terminology of `parent_entity_id` and `parent_variable_id` inside the code. Not sure if we want to update that to something like `parent_dataframe_id` and `parent_column_id` instead? Wasn't sure if we had a preferred new set of terms to use as we replace Entity with a Woodwork Dataframe and Variable with Woodwork Column.

rwedge:
@thehomebrewnerd I think updating the terminology inside the code makes sense. 

Do we have a place where we are tracking the various API changes that will impact users?  This is one where the existing calls will be broken.

What do you think about `add_relationship` work with existing relationship objects:

```python
relationship = Relationship(es, "products", "product_id", "transactions", "product_id")
es.add_relationship(relationship)
```

thehomebrewnerd:
> Do we have a place where we are tracking the various API changes that will impact users? This is one where the existing calls will be broken.

No, not yet. There are a couple other breaking changes that are incorporated on the `woodwork-integration` branch as well based on methods being moved from Entity to EntitySet.

If you want to track all these do you have any ideas on where we should best track these? There are going to be a lot of them, of course, and we'll just need to be diligent in remembering to update our list as we make these changes.

thehomebrewnerd:
> What do you think about `add_relationship` work with existing relationship objects:
> 
> ```python
> relationship = Relationship(es, "products", "product_id", "transactions", "product_id")
> es.add_relationship(relationship)
> ```

Were you thinking of updating so that it would accept either a Relationship object or the column specifications, or only accept a Relationship object as it did previously?

The advantage of this new approach is that users never have to interact with a Relationship object. The disadvantage is that existing code will need to be updated to work with the new release. I'm fine with either approach, or a combination of the two.

rwedge:
We could make a list on the main PR's first comment so it's publicly visible, and move it to an issue if it grows unwieldy to keep it there

thehomebrewnerd:
> We could make a list on the main PR's first comment so it's publicly visible, and move it to an issue if it grows unwieldy to keep it there

@rwedge I guess I should also mention that we did document the other breaking changes that were made on the `woodwork-integration` branch in the release notes as breaking changes.

rwedge:
> Were you thinking of updating so that it would accept either a Relationship object or the column specifications, or only accept a Relationship object as it did previously?
> 
> The advantage of this new approach is that users never have to interact with a Relationship object. The disadvantage is that existing code will need to be updated to work with the new release. I'm fine with either approach, or a combination of the two.

I was thinking we allow either but in our documentation / examples we push the new approach.  It feels weird to be able to create a relationship object but not add it to an entityset.

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -32,8 +32,10 @@ def test_normalize_time_index_as_additional_variable(es):
 def test_cannot_re_add_relationships_that_already_exists(es):
     warn_text = "Not adding duplicate relationship: " + str(es.relationships[0])
     before_len = len(es.relationships)
+    rel = es.relationships[0]
     with pytest.warns(UserWarning, match=warn_text):
-        es.add_relationship(es.relationships[0])
+        es.add_relationship(rel._parent_dataframe_id, rel._parent_column_id,
```
rwedge, line 37:
```
es.add_relationship(relationship=es.relationships[0])
```

would be shorter

since we're testing the re-add validation specifically i don't think which input we use matters

or we could test both

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -91,25 +92,44 @@ def test_add_relationship_errors_child_v_index(es):
...
     es['log'].df = pd.DataFrame(columns=es['log'].df.columns)
     assert len(es['log'].df) == 0
     assert es['log'].df['session_id'].dtype == 'object'
 
     es.relationships.remove(relationship)
     assert(relationship not in es.relationships)
 
-    es.add_relationship(relationship)
+    es.add_relationship(relationship._parent_dataframe_id, relationship._parent_column_id,
+                        relationship._child_dataframe_id, relationship._child_column_id)
```
rwedge, line 110: same idea here, `relationship=` would be shorter

featuretools/tests/entityset_tests/test_es_metadata.py
```diff
@@ -3,13 +3,14 @@
...
-from featuretools import EntitySet, Relationship, variable_types
+from featuretools import EntitySet, variable_types
 from featuretools.tests.testing_utils import backward_path, forward_path
 
 
 def test_cannot_re_add_relationships_that_already_exists(es):
     before_len = len(es.relationships)
-    es.add_relationship(es.relationships[0])
+    rel = es.relationships[0]
+    es.add_relationship(rel._parent_dataframe_id, rel._parent_column_id, rel._child_dataframe_id, rel._child_column_id)
```
rwedge, line 13: same idea with using `relationships=`

featuretools/tests/entityset_tests/test_serialization.py
```diff
@@ -31,7 +26,7 @@
 BUCKET_NAME = "test-bucket"
 WRITE_KEY_NAME = "test-key"
 TEST_S3_URL = "s3://{}/{}".format(BUCKET_NAME, WRITE_KEY_NAME)
-TEST_FILE = "test_serialization_data_entityset_schema_5.1.0.tar"
+TEST_FILE = "test_serialization_data_entityset_schema_5.2.0.tar"
```
rwedge, line 29: if we're unable to read in the 5.1.0 test file successfully we need to move to 6.0.0

if so, this would also impact the feature serialization tests

featuretools/tests/entityset_tests/test_es.py
thehomebrewnerd: I think the way things are structured now the input doesn't matter, but if we ever changed the way we check for an existing relationship it might. With that in mind, I think it might be good to test both approaches here.

featuretools/tests/entityset_tests/test_serialization.py
thehomebrewnerd: Yeah, we can't read because of switching terminology from `entity` to `dataframe` and `variable` to `column`. I'll update to 6.0.0 instead.

docs/source/release_notes.rst
```diff
@@ -41,6 +42,11 @@ Release Notes
   which to add interesting values in the ``entity_id`` parameter.
 * ``Entity.set_secondary_time_index`` has been removed and replaced by ``EntitySet.set_secondary_time_index``
   with an added ``entity`` parameter to specify the entity on which to set the secondary time index.
+* ``Relationship`` initialization has been updated to accept four id values for the parent dataframe,
+  parent column, child dataframe and child column instead of accepting two ``Variable`` objects.
+* ``EntitySet.add_relationship`` has been updated to accept dataframe and column id values or a
+  ``Relationship`` object. Adding a relationship from a ``Relationship`` object now requires passing
+  the relationship as a keyword argument.
```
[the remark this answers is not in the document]
thehomebrewnerd: Here in the docstring or somewhere else?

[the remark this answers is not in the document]
rwedge: yep, here in the release notes

rwedge (approved):
Looks good!

PR #1382, opened 2021-04-01
Refactor and update test fixtures

### LINKED ISSUE #1364: Update test fixtures to avoid direct assignment of `Entity.df`
In some of the test fixtures, most notably for Dask and Koalas, a new dataframe is assigned to an Entity through an operation such as `Entity.df = new_dataframe`. With the new Woodwork accessor approach, this type of reassignment will not be possible. The test fixtures should be updated to rebuild the EntitySets with the proper dataframes using a different approach.

For example, here is the current code for the `dd_mock_customer` fixture:
```python
@pytest.fixture
def dd_mock_customer(pd_mock_customer):
    dd_mock_customer = copy.deepcopy(pd_mock_customer)
    for entity in dd_mock_customer.entities:
        entity.df = dd.from_pandas(entity.df.reset_index(drop=True), npartitions=4)
    return dd_mock_customer
```

This could be refactored to something like this instead to avoid direct assignment of the dataframe:
```python
@pytest.fixture
def dd_mock_customer(pd_mock_customer):
    entities = {}
    for entity in pd_mock_customer.entities:
        entities[entity.id] = (dd.from_pandas(entity.df.reset_index(drop=True), npartitions=4), entity.index, None, entity.variable_types)

    relationships = [(rel.parent_entity.id,
                      rel.parent_variable.name,
                      rel.child_entity.id,
                      rel.child_variable.name) for rel in pd_mock_customer.relationships]

    return ft.EntitySet(id=pd_mock_customer.id, entities=entities, relationships=relationships)
```

thehomebrewnerd:
### Refactor and update test fixtures

Closes #1364 

This PR refactors tests fixtures to avoid direct assignment of Entity.df, which will not be possible when a Woodwork dataframe is used to replace Entity.

During review of the test fixtures, two other fixtures were found to not be using Dask and/or Koalas entityset versions so these fixtures were also updated to include the Dask/Koalas versions.


thehomebrewnerd:
@rwedge Although the Woodwork integration work prompted the creation of this PR, nothing in this breaks the current implementation, and a couple of the fixture parameterization updates could be beneficial for the current code base. I am requesting a merge into `main` instead of the `woodwork-integrations` branch as a result.

featuretools/entityset/deserialize.py
```diff
@@ -161,7 +163,6 @@ def read_entity_data(description, path):
     else:
         error = 'must be one of the following formats: {}'
         raise ValueError(error.format(', '.join(FORMATS)))
-    dtypes = description['loading_info']['properties']['dtypes']
```
thehomebrewnerd, line 164: Changing the way the Dask entityset was created was causing a Dask error in `test_to_csv`. Dask was inferring the `log` entity `zipcode` column as `int64`, but it should be an `object` column based on the data it contains. The changes here were needed to tell Dask what the column types should be when reading the CSV, so it doesn't try to infer them. This is probably an overall safer way to read the file anyway since we know from serialization what the column dtypes should be.

I'm not exactly sure why changing the entityset construction highlighted this, but it did...

Also, I had to change the way the `dtypes` variable was created above because one of the tests for invalid input does not contain the `properties` entry in `loading_info` so using the old selection method resulted in a `KeyError`.

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -157,15 +158,59 @@ def simple_es():
...
+
+
+@pytest.fixture(params=['pd_simple_es', 'dd_simple_es', 'ks_simple_es'])
+def simple_es(request):
+    return request.getfixturevalue(request.param)
+
+
 def test_equal_categorical(simple_es):
+    # if ks and any(isinstance(e.df, ks.DataFrame) for e in simple_es.entities):
+    #     pytest.xfail("Koalas does not support categorical dtype")
```
rwedge, line 204: still needed?

thehomebrewnerd: No. Removed.

rwedge (approved):
Looks good

PR #1386, opened 2021-04-02
disable fail-fast behavior on CI tests

rwedge:
Disable the fail-fast strategy on our unit tests, which is cancelling jobs if a job in the same test matrix has failed (currently if the py38 optional unit tests fail, the py37 and py36 optionals will be cancelled).  I'd prefer to have the tests continue to so we'll know if there are issues with the other environments we're testing on.


thehomebrewnerd (approved):
Looks good!

PR #1394, opened 2021-04-06
Update boto3 and urllib3 requirements

rwedge:
Updating urllib3 to latest verstion to avoid security vulnerability.  Update boto3 requirement same time


thehomebrewnerd (approved):
LGTM!

PR #1395, opened 2021-04-08
Don't pass EntitySet to Relationship on init

thehomebrewnerd:
### Don't pass EntitySet to Relationship on init
This PR implements changes that would be needed if we did not pass an entityset to a relationship on init.


thehomebrewnerd:
Closing in favor of performing more detailed investigation on how to clean this up later.

PR #1400, opened 2021-04-13
Move time type check

### LINKED ISSUE #1369: Move Entity time type check off of Entity
The current code for `Entity` has a check to verify that if the Entity has a time index, that the time index is of the **same type as other Entities** in an EntitySet. This check happens during setting of the Entity time index in `Entity.set_time_index`. 

When Entity is replace with a Woodwork dataframe, this validation check will no longer be present and should be implemented in other places where necessary. One likely place where this check could be needed is in `calculate_feature_matrix` but there may be other places as well.

Current validation code included for reference:
```python
        if self.entityset.time_type is None:
            self.entityset.time_type = time_type
        elif self.entityset.time_type != time_type:
            raise TypeError("%s time index is %s type which differs from"
                            " other entityset time indexes" %
                            (self.id, time_type))
```

TLDR: **Time Index should be same type for all Entities in EntitySet**

jeff-hernandez:
Closes #1369 by moving the validation check for uniform time index to ``EntitySet``

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -156,6 +156,7 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
             raise ValueError(msg)
 
     target_entity = entityset[features[0].entity.id]
+
     cutoff_time = _validate_cutoff_time(cutoff_time, target_entity)
     entityset._check_time_indexes()
 
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -155,8 +155,8 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
             msg = "Using training_window is not supported with Dask Entities"
             raise ValueError(msg)
 
+    entityset._check_time_index()
     target_entity = entityset[features[0].entity.id]
-
     cutoff_time = _validate_cutoff_time(cutoff_time, target_entity)
 
     if isinstance(cutoff_time, pd.DataFrame):
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -155,7 +155,7 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
             msg = "Using training_window is not supported with Dask Entities"
             raise ValueError(msg)
 
-    entityset._check_time_index()
+    entityset._check_time_indexes()
     target_entity = entityset[features[0].entity.id]
     cutoff_time = _validate_cutoff_time(cutoff_time, target_entity)
 
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -155,9 +155,9 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
             msg = "Using training_window is not supported with Dask Entities"
             raise ValueError(msg)
 
-    entityset._check_time_indexes()
     target_entity = entityset[features[0].entity.id]
     cutoff_time = _validate_cutoff_time(cutoff_time, target_entity)
+    entityset._check_time_indexes()
 
     if isinstance(cutoff_time, pd.DataFrame):
         if instance_ids:
--- a/featuretools/entityset/entity.py
+++ b/featuretools/entityset/entity.py
@@ -5,14 +5,13 @@ import pandas as pd
 
 from featuretools import variable_types as vtypes
 from featuretools.utils.entity_utils import (
-    col_is_datetime,
     convert_all_variable_data,
     convert_variable_data,
     get_linked_vars,
     infer_variable_types
 )
 from featuretools.utils.gen_utils import import_or_none, is_instance
-from featuretools.utils.wrangle import _check_time_type, _dataframes_equal
+from featuretools.utils.wrangle import _dataframes_equal
 from featuretools.variable_types import Text, find_variable_types
 
 ks = import_or_none('databricks.koalas')
@@ -72,8 +71,10 @@ class Entity(object):
         self.set_index(index)
 
         self.time_index = None
+        entityset.entity_dict[self.id] = self
+
         if time_index:
-            self.set_time_index(time_index, already_sorted=already_sorted)
+            entityset.set_time_index(id, time_index, already_sorted=already_sorted)
 
         entityset.set_secondary_time_index(self, secondary_time_index)
 
@@ -282,42 +283,6 @@ class Entity(object):
         for v_id in variable_ids:
             v = self._get_variable(v_id)
             self.variables.remove(v)
-
-    def set_time_index(self, variable_id, already_sorted=False):
-        # check time type
-        if not isinstance(self.df, pd.DataFrame) or self.df.empty:
-            time_to_check = vtypes.DEFAULT_DTYPE_VALUES[self[variable_id]._default_pandas_dtype]
-        else:
-            time_to_check = self.df[variable_id].iloc[0]
-        time_type = _check_time_type(time_to_check)
-
-        if time_type is None:
-            raise TypeError("%s time index not recognized as numeric or"
-                            " datetime" % (self.id))
-
-        if self.entityset.time_type is None:
-            self.entityset.time_type = time_type
-        elif self.entityset.time_type != time_type:
-            raise TypeError("%s time index is %s type which differs from"
-                            " other entityset time indexes" %
-                            (self.id, time_type))
-
-        if is_instance(self.df, (dd, ks), 'DataFrame'):
-            t = time_type  # skip checking values
-            already_sorted = True  # skip sorting
-        else:
-            t = vtypes.NumericTimeIndex
-            if col_is_datetime(self.df[variable_id]):
-                t = vtypes.DatetimeTimeIndex
-
-        # use stable sort
-        if not already_sorted:
-            # sort by time variable, then by index
-            self.df = self.df.sort_values([variable_id, self.index])
-
-        self.convert_variable_type(variable_id, t, convert_data=False)
-
-        self.time_index = variable_id
 
     def set_index(self, variable_id, unique=True):
         """
--- a/featuretools/entityset/entity.py
+++ b/featuretools/entityset/entity.py
@@ -71,7 +71,10 @@ class Entity(object):
         self.df = df[[v.id for v in self.variables]]
         self.set_index(index)
 
-        self.set_time_index(time_index, already_sorted=already_sorted)
+        self.time_index = None
+        if time_index:
+            self.set_time_index(time_index, already_sorted=already_sorted)
+
         self.secondary_time_index = secondary_time_index
 
     def __repr__(self):
@@ -281,28 +284,31 @@ class Entity(object):
             self.variables.remove(v)
 
     def set_time_index(self, variable_id, already_sorted=False):
-        self.time_index = variable_id
-        self._already_sorted = already_sorted
-        self._check_time_index()
+        # check time type
+        if not isinstance(self.df, pd.DataFrame) or self.df.empty:
+            time_to_check = vtypes.DEFAULT_DTYPE_VALUES[self[variable_id]._default_pandas_dtype]
+        else:
+            time_to_check = self.df[variable_id].iloc[0]
+        time_type = _check_time_type(time_to_check)
 
-    def _check_time_index(self):
-        if self.time_index is None:
-            return
+        if time_type is None:
+            raise TypeError("%s time index not recognized as numeric or"
+                            " datetime" % (self.id))
 
-        time_type = self._get_time_type()
         if is_instance(self.df, (dd, ks), 'DataFrame'):
             t = time_type  # skip checking values
-            self._already_sorted = True  # skip sorting
+            already_sorted = True  # skip sorting
         else:
             t = vtypes.NumericTimeIndex
-            if col_is_datetime(self.df[self.time_index]):
+            if col_is_datetime(self.df[variable_id]):
                 t = vtypes.DatetimeTimeIndex
 
         # use stable sort
-        if not self._already_sorted:
+        if not already_sorted:
             # sort by time variable, then by index
-            self.df = self.df.sort_values([self.time_index, self.index])
-        self.convert_variable_type(self.time_index, t, convert_data=False)
+            self.df = self.df.sort_values([variable_id, self.index])
+        self.convert_variable_type(variable_id, t, convert_data=False)
+        self.time_index = variable_id
 
     def _get_time_type(self, variable_id=None):
         variable_id = variable_id or self.time_index
--- a/featuretools/entityset/entity.py
+++ b/featuretools/entityset/entity.py
@@ -5,6 +5,7 @@ import pandas as pd
 
 from featuretools import variable_types as vtypes
 from featuretools.utils.entity_utils import (
+    col_is_datetime,
     convert_all_variable_data,
     convert_variable_data,
     get_linked_vars,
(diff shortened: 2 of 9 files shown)


thehomebrewnerd:
@jeff-hernandez I see this is still draft, so maybe you are still thinking through the options, but when we replace `ft.Entity` with a Woodwork DataFrame, the `set_time_index` call will happen through Woodwork and not Featuretools. How will the validation work in that scenario since the Woodwork dataframe will not have a reference to the EntitySet to which it belongs?

jeff-hernandez:
@thehomebrewnerd is there a reason we don't want to set the time index from an entityset? It seems like, we need to reference the other dataframes to validate the time type. If the time type is valid, the `set_time_index` call through Woodwork can still happen inside `EntitySet.set_time_index`

thehomebrewnerd:
> @thehomebrewnerd is there a reason we don't want to set the time index from an entityset? It seems like, we need to reference the other dataframes to validate the time type. If the time type is valid, the `set_time_index` call through Woodwork can still happen inside `EntitySet.set_time_index`

I think the main reason is that users will be able to create an EntitySet from either a plan pandas dataframe or from a Woodwork dataframe. If users have already created a Woodwork dataframe and set the time index using Woodwork, I think it seems a little off to expect them to make two different `set_time_index` calls - one in Woodwork and one in FT.

You are correct in that we will need to be able to reference all the dataframes to validate the time type, so that will need to happen somewhere that uses the full EntitySet. @rwedge had originally suggested that we make the time type check to `calculate_feature_matrix`, when it becomes important that all the time types are the same.

jeff-hernandez:
> I think the main reason is that users will be able to create an EntitySet from either a plan pandas dataframe or from a Woodwork dataframe. If users have already created a Woodwork dataframe and set the time index using Woodwork, I think it seems a little off to expect them to make two different set_time_index calls - one in Woodwork and one in FT.

I wouldn't expect users to make two different `set_time_index` calls. If the time index is already set in the Woodwork DataFrame, I'd imagine the EntitySet can also use that time index when we create it from the Woodwork DataFrame.

It makes sense to validate when it becomes important in `calculate_feature_matrix`. I'll try using that approach.

thehomebrewnerd:
> I wouldn't expect users to make two different `set_time_index` calls. If the time index is already set in the Woodwork DataFrame, I'd imagine the EntitySet can also use that time index when we create it from the Woodwork DataFrame.

Yes, EntitySet could/would use any time index that is set previously within Woodwork and the time types could be validated when the dataframe is added to the EntitySet. However, there is also the problem that if a user calls `df.ww.set_time_index` on a Woodwork dataframe to change the time index after it has been added to an EntitySet, the EntitySet will not know that it needs to check for compatible time types upon that change. I think that is why moving the validation somewhere like cfm makes sense.

I am envisioning that there will no longer be a public `set_time_index` method anywhere in Featuretools once Woodwork is integrated, leaving the user only one way to set a time index - through Woodwork.

I might still be missing something here that would allow us to keep this check on EntitySet, but I wasn't thinking that was possible based on earlier code study.

jeff-hernandez:
> However, there is also the problem that if a user calls df.ww.set_time_index on a Woodwork dataframe to change the time index after it has been added to an EntitySet, the EntitySet will not know that it needs to check for compatible time types upon that change. I think that is why moving the validation somewhere like cfm makes sense.

I agree that scenario highlights why validating in cfm makes sense.

> I am envisioning that there will no longer be a public `set_time_index` method anywhere in Featuretools once Woodwork is integrated, leaving the user only one way to set a time index - through Woodwork.

Sounds good. This can help to make the API easier to use.

jeff-hernandez:
@thehomebrewnerd @rwedge this PR separates the time index type check into a private method so we can call whenever needed. There are few places where it seems necessary:
- `calculate_feature_matrix`
- `EntitySet.normalize_entity`
- `EntitySet.query_by_values`
- `EntitySet.update_dataframe`

featuretools/entityset/entityset.py
```diff
@@ -570,6 +554,8 @@ def entity_from_dataframe(self,
             make_index=make_index)
         self.entity_dict[entity.id] = entity
         self.reset_data_description()
+
```
rwedge, line 557: entity_from_dataframe should probably have `_check_time_index` in it.  It makes sense to validate as we add an entity.  This would probably remove some of the stray `_check_time_index` calls being added to tests

rwedge (commented):
I left some comments but in general I think this PR is doing a lot more than moving where we check the time type of entities (time type here referring to whether the time column is numeric or datetime, not all time index validation) I think we should either narrow the focus of the PR to just the time type, or work on a design for handling time validation that uses Woodwork TableAccessors directly

rwedge:
What if we revert the removal of `Entity.set_time_index`?  Woodwork has `TableAccessor.set_time_index` and takes `time_index` in the `__init__`.

Woodwork's `set_time_index` logic doesn't have the `time_type` setting logic, so to simulate we could remove it from `Entity.set_time_index` and handle this change

Woodwork doesn't have a `set_secondary_time_index` option so we could keep the validation there

featuretools/entityset/entityset.py
```diff
@@ -1203,14 +1192,67 @@ def update_dataframe(self, entity_id, df, already_sorted=False, recalculate_last
...
+            time_to_check = entity.df[variable_id].iloc[0]
+
+        time_type = _check_time_type(time_to_check)
+        if time_type is None:
+            info = "%s time index not recognized as numeric or datetime"
+            raise TypeError(info % entity.id)
+        return time_type
+
+    def _check_time_index(self):
+        for entity in self.entity_dict.values():
```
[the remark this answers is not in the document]
jeff-hernandez: Yes, we can add a method to check time indexes for a single entity. It can work as a private method on either the Entity or EntitySet. We will still need to separate the logic of checking whether the time index is the same across entities. This seems like something that can happen in `_check_time_indexes`

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -1392,15 +1415,12 @@ def test_normalize_with_numeric_time_index(int_es):
...
-                            new_entity_id="cancel_reason",
-                            index="cancel_reason",
-                            copy_variables=['upgrade_date'])
-    es['customers'].convert_variable_type('signup_date', variable_types.DatetimeTimeIndex)
+    es.normalize_entity(base_entity_id="customers",
+                        new_entity_id="cancel_reason",
+                        index="cancel_reason",
+                        copy_variables=['upgrade_date'])
+
+    assert isinstance(es['customers']['signup_date'], variable_types.DatetimeTimeIndex)
```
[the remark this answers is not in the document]
jeff-hernandez: There is a subtle difference. In the previous behavior, the variable type would get changed in `set_time_index`. If a user converts the variable type again after, it will be invalid when needed in `normalize_entity`. In this behavior, the variable type is not changed until it's needed in`normalize_entity`.

featuretools/entityset/entityset.py
jeff-hernandez: We can add in `entity_from_dataframe`. I wasn't sure if we were working towards limiting the time type check to only happen in CFM.

jeff-hernandez:
> What if we revert the removal of Entity.set_time_index? Woodwork has TableAccessor.set_time_index and takes time_index in the __init__.

We can keep `Entity.set_time_index` but we'll need to remove the logic that checks whether the time index is the same across entities.

jeff-hernandez:
@thehomebrewnerd @rwedge I made additional updates to hopefully clean things up a bit. Let me know what you think.

thehomebrewnerd:
> @thehomebrewnerd @rwedge I made additional updates to hopefully clean things up a bit. Let me know what you think.

@jeff-hernandez I maybe glossing over some details here, but there are still more changes here than I was expecting to see as part of this PR. My original thought was that we would just delete the lines linked below from `Entity` and leave everything else the same as most (hopefully all) of the other remaining validation checks will be handled by Woodwork in the future. EntitySet doesn't need to worry about setting time column variable types or sorting data - that will be handled by Woodwork.

I was thinking we would just remove these lines from Entity and replace with a comparable validation elsewhere, at a point that it becomes important for all entities to have the same time index column type (numeric or datetime).
https://github.com/alteryx/featuretools/blob/4439bef1eb78785a120f4471b98e71b1617e879d/featuretools/entityset/entity.py#L298-L303

Removing these lines means we would no longer validate that the time type is the same for all entity time indexes when an entity is added to an entityset, or when set_time_index is called. Consistency in the time index types doesn't really matter when creating the entityset. It really becomes important downstream, probably in `calculate_feature_matrix` and maybe `normalize_entity`? Not sure exactly everywhere it is needed as I haven't studied that in detail yet.

Anyway, this basically means that several of the tests we currently have in `test_es.py` that check the entityset time type will need to be updated, and new tests would be needed for making sure the error currently raised by the code linked above is still raised at the point where having the same type type between all entities is important.

Maybe I'm missing some details that make this impossible or impractical, but does this make sense or am I off base in my line of thinking here?

jeff-hernandez:
@thehomebrewnerd we can separted the validation check for consistency in the time index types. The question is where do we want to run this validation check to run. There are few places where they currently happen:
- `calculate_feature_matrix`
- `EntitySet.normalize_entity`
- `EntitySet.query_by_values`
- `EntitySet.entity_from_dataframe`
- `EntitySet.update_dataframe`
First, let's determine where we want this validation check to happen. Then, we can update the tests accordingly.

thehomebrewnerd:
@jeff-hernandez Maybe we can review with @rwedge as well, but I am thinking there is not much reason to validate that all entities have the same time type in `EntitySet.entity_from_dataframe` or `EntitySet.update_dataframe` because users would still be able to call `df.ww.set_time_index()` on the dataframe directly at a later point in time, invalidating the check that happened during those EntitySet method calls. Because of that, validating during those two calls is not sufficient to guarantee that all entity time types will remain the same.

I'll have to think about the other locations you call out a bit more though...

jeff-hernandez:
@thehomebrewnerd yeah sounds good. I'll put this on draft while we discuss with @rwedge and determine the validation points.

jeff-hernandez:
@thehomebrewnerd @rwedge based on the discussion, the PR has been updated and ready for review

featuretools/entityset/entity.py
```diff
@@ -284,40 +281,42 @@ def delete_variables(self, variable_ids):
...
-        else:
-            time_to_check = self.df[variable_id].iloc[0]
-        time_type = _check_time_type(time_to_check)
-
-        if time_type is None:
-            raise TypeError("%s time index not recognized as numeric or"
-                            " datetime" % (self.id))
+        self.time_index = variable_id
+        self._already_sorted = already_sorted
+        self._check_time_index()
```
rwedge, line 286: I think we should leave the `set_time_index` code largely unchanged, aside from moving the `time_type` logic. 

I don't think an `_already_sorted` attribute is necessary, both Entity and Woodwork Table sort by default.

featuretools/entityset/entity.py
```diff
@@ -284,40 +281,42 @@ def delete_variables(self, variable_ids):
...
         # use stable sort
-        if not already_sorted:
+        if not self._already_sorted:
             # sort by time variable, then by index
-            self.df = self.df.sort_values([variable_id, self.index])
+            self.df = self.df.sort_values([self.time_index, self.index])
+        self.convert_variable_type(self.time_index, t, convert_data=False)
 
-        self.convert_variable_type(variable_id, t, convert_data=False)
+    def _get_time_type(self, variable_id=None):
```
rwedge, line 307: We should move this outside of Entity, since we'll still want this after Entity has been replaced

featuretools/entityset/entityset.py
```diff
@@ -319,24 +319,8 @@ def add_relationship(self,
...
-                                " datetime" % (entity.id))
-            if self.time_type != time_type:
-                raise TypeError("%s time index is %s type which differs from"
-                                " other entityset time indexes" %
-                                (entity.id, time_type))
-            if time_index not in columns:
-                columns.append(time_index)
-
         entity.secondary_time_index = secondary_time_index
+        self._check_secondary_time_index(entity)
```
rwedge, line 323: Don't we want to validate before setting the secondary time index?  Even if `_check_secondary_time_index` reports an error, we'll still have set a bad secondary time index

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -156,8 +156,8 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
             raise ValueError(msg)
 
     target_entity = entityset[features[0].entity.id]
-
     cutoff_time = _validate_cutoff_time(cutoff_time, target_entity)
+    entityset._check_time_indexes()
```
thehomebrewnerd, line 160: If I'm thinking about this right, I think the only check that needs to happen here is that all of the entities (eventually woodwork dataframes) have the same time index type. Woodwork will prevent users from setting a time index column if that column does not contain valid numeric or datetime values, so that does not need to be checked again here.

featuretools/entityset/entity.py
jeff-hernandez: ✅  df5eb515

featuretools/entityset/entityset.py
jeff-hernandez: yeah, makes sense

✅  88fb837d

featuretools/entityset/entity.py
jeff-hernandez: ✅  d2b0c7c4

featuretools/computational_backends/calculate_feature_matrix.py
jeff-hernandez: ✅  cc461753

featuretools/entityset/entity.py
```diff
@@ -75,7 +75,7 @@ def __init__(self, id, df, entityset, variable_types=None,
         if time_index:
             self.set_time_index(time_index, already_sorted=already_sorted)
 
-        entityset.set_secondary_time_index(self, secondary_time_index)
+        self.secondary_time_index = secondary_time_index
```
thehomebrewnerd, line 78: Can this call say as it was? If the validation in `entityset.set_secondary_time_index` is successful it will then set the entity secondary time index to the specified value. Otherwise I think we could be setting it here before the validation check happens.

thehomebrewnerd:
@jeff-hernandez I think we are closing in on a solution here. I think it would be good to add a new test to verify that if an EntitySet contains entities with different time index types that the expected error is raised when calculate_feature_matrix is called, as that is the new wrinkle that we need to be sure to catch when we integrate Woodwork.

featuretools/entityset/entity.py
jeff-hernandez: The validation in `set_secondary_time_index` checks if the secondary time index is the same time type across entities. I don't think a Woodwork dataframe will have a reference to the entityset that contains the other dataframes, correct?

thehomebrewnerd: True, but the handling of the secondary time index is going to change significantly, and all the code here in `entity.py` will be deleted anyway. It's really only an issue in the short term. I think the current code should work fine until that deletion happens though, but not a big deal either way.

thehomebrewnerd: The only difference for now is that the new code sets the secondary time index on Entity without validation, while the previous code runs validation first and then sets it.

jeff-hernandez: I think it'd make sense to keep this validation check outside the entity since it gives an idea of where the validations are likely to happen once the Woodwork DataFrame gets integrated.

rwedge: I view this line as mimicking the user calling `EntitySet.set_secondary_time_index`.  Once Woodwork is added we expect users to call it that way.  In the meantime, `secondary_time_index` is still a valid argument to entity so we need to handle it in some way.

jeff-hernandez: Not sure quite sure if I understand the use case, but happy to revert this change

jeff-hernandez: ✅  7762b431

rwedge (approved):
Looks good!

PR #1402, opened 2021-04-15
Change to use GitHub Token rather than GitHub PAT

gsheni:
- Change from GitHub PAT to auto generated GitHub Token for dependency checker

### DIFF
--- a/.github/workflows/dependency_check.yml
+++ b/.github/workflows/dependency_check.yml
@@ -30,4 +30,3 @@ jobs:
         branch-suffix: short-commit-hash
         base: main
         reviewers: rwedge, thehomebrewnerd, frances-h
-        token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/dependency_check.yml
```diff
@@ -30,4 +30,4 @@ jobs:
         branch-suffix: short-commit-hash
         base: main
         reviewers: rwedge, thehomebrewnerd, frances-h
-        token: ${{ secrets.REPO_SCOPED_TOKEN }}
+        token: ${{ secrets.GITHUB_TOKEN }}
```
rwedge, line 33: `GITHUB_TOKEN` should be used by default, can we try dropping this line

gsheni: Done

rwedge (approved):
LGTM

PR #1403, opened 2021-04-19
Move default variable description logic to generate_description

rwedge:
Since `ColumnAccessor.description` defaults to `None`, updated `generate_description` to expect `None` as the default.  The `None` to default string conversion now happens inside `generate_description` instead of on `Variable`


frances-h (approved):
Looks good 👍

PR #1405, opened 2021-04-23
Replace Entity with Woodwork DataFrame

### LINKED ISSUE #1253: Replace Entity with DataFrame (using Woodwork Accessor)
- Featuretool's Entity class should be replace with Woodwork's DataFrame (using Woodwork typing via Accessor)
- secondary_time_index should go in metadata of the DataFrame - update example code in release notes
- the last_time_index should be a Series in the metadata of the DataTable (will be changed in #1254)
- For completion, this should allow a use create an EntitySet, add data via Woodwork DataFrame, and add relationships to the EntitySet. 
  - handles interesting_values

tamargrey:
Uses Woodwork DataFrames in place of Entities in EntitySet. 

ToDo

- [x] Update `add_last_time_indexes`
- [x] Update `add_interesting_values`
- [x] Determine implementation in uncertain areas
- [x] Determine behavior changes (either change FT or WW)
- [x] Fix tests
  - [x] update `make_ecommerce_entityset`
  - [x] update `load_mock_customer` (blocked by last time indexes)
  - [x] update `load_retail` (blocked by last time indexes)
  - [x] update test_es.py
  - [x] fix remaining tests in other files

To be done in other issues after this gets merged in
-  Fix fixtures that use deepcopy - need to be able to copy with ww retained.
- Update remaining methods
  - `plot`
  - serialization
  - `concat`
  - `query_by_values`

closes #1253


featuretools/demo/flight.py
```diff
@@ -94,30 +94,30 @@ def make_es(data):
...
-                                              'distance_group', 'carrier', 'flight_num'])
-
-    es.normalize_entity('flights', 'airlines', 'carrier',
-                        make_time_index=False)
-
-    es.normalize_entity('flights', 'airports', 'dest',
-                        additional_variables=['dest_city', 'dest_state'],
-                        make_time_index=False)
+    logical_types = {'flight_num': ww.logical_types.Categorical,
+                     'distance_group': ww.logical_types.Ordinal(order=[1, 2, 3, 8, 10]),
```
rwedge, line 98: Have we tried running this with `demo=False` to use the full data to see if there are other distance groups present?

featuretools/entityset/entityset.py
```diff
@@ -280,231 +295,240 @@ def add_relationship(self,
...
-        msg = u"Unable to add relationship because {} in {} is Pandas dtype {}"\
-            u" and {} in {} is Pandas dtype {}."
-        if not is_dtype_equal(parent_dtype, child_dtype):
-            raise ValueError(msg.format(parent_v, parent_e.id, parent_dtype,
-                                        child_v, child_e.id, child_dtype))
+        # --> Implementation: is this still relevant?
+        if isinstance(child_df, pd.DataFrame) and \
+                (child_df.empty and child_df[child_column].dtype == object and
+                 parent_df.ww.columns[parent_column].is_numeric):
+            child_df.ww[child_column] = pd.Series(name=child_column, dtype=np.int64)
```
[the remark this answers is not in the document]
rwedge: The comment mentions metadata so this will come up when we tackle serialization.  The lines are covered on codecov main so this scenario does get tested pre-ww.  Looks like it has to do with ensuring that if the child dataframe was empty for some reason (the metadata uses empty dataframes to keep memory footprint small), the foreign key column on the child has an integer pandas dtype instead of an object dtype, to ensure the dtypes match between empty parent df and empty child df to pass the dtype validation check just below

featuretools/demo/flight.py
tamargrey: Good call! We were missing `[4, 5, 6, 7, 9]`

featuretools/entityset/entityset.py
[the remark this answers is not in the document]
thehomebrewnerd: If I'm understanding this correctly - for Woodwork, as long as the parent column and the child column in the relationship have the same `LogicalType` we won't need to worry about manually doing this dtype conversion, as WW should handle it. If we create a new empty dataframe somewhere in the code, we will just need to make sure we set the correct logical types for the columns.

featuretools/entityset/entityset.py
```diff
@@ -280,231 +294,240 @@ def add_relationship(self,
...
+
+        parent_ltype = parent_df.ww.logical_types[parent_column]
+        child_ltype = child_df.ww.logical_types[child_column]
+        if parent_ltype != child_ltype:
+            warnings.warn(f'Logical type {child_ltype} for child column {child_column} does not match '
+                          f'parent column {parent_column} logical type {parent_ltype}. '
+                          'Changing child logical type to match parent.')
+            child_df.ww.set_types(logical_types={child_column: parent_ltype})
```
rwedge, line 326: Unsure if this change was already discussed, but it is changing from an error to a warn and try to continue anyway.   

If we do use this approach,  I think we should document this in release notes with the other noteworthy changes

featuretools/entityset/entityset.py
```diff
@@ -280,231 +294,240 @@ def add_relationship(self,
...
 
         self.relationships.append(relationship)
         self.reset_data_description()
         return self
 
-    def set_secondary_time_index(self, entity, secondary_time_index):
-        self._check_secondary_time_index(entity, secondary_time_index)
-        entity.secondary_time_index = secondary_time_index
+    def set_secondary_time_index(self, dataframe_id, secondary_time_index):
+        '''Sets the secondary time index for a dataframe in the EntitySet using its dataframe id'''
```
rwedge, line 333: Can we do a longer docstring that includes the Args format

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -44,72 +47,164 @@ def test_cannot_re_add_relationships_that_already_exists(es):
...
-def test_add_relationship_empty_child_convert_dtype(es):
-    relationship = ft.Relationship(es, "sessions", "id", "log", "session_id")
-    es['log'].df = pd.DataFrame(columns=es['log'].df.columns)
-    assert len(es['log'].df) == 0
-    assert es['log'].df['session_id'].dtype == 'object'
+# def test_add_relationship_empty_child_convert_dtype(es):
+#     relationship = ft.Relationship(es, "sessions", "id", "log", "session_id")
+#     empty_log_df = pd.DataFrame(columns=es['log'].columns)
+#     # --> schema isn't valid.... not sure what this is testing
+#     empty_log_df.ww.init(schema=es['log'].ww.schema)
```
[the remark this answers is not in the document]
thehomebrewnerd: I think this was testing that if you add a relationship between two columns, that the dtype of the child column gets updated to match the dtype of the parent columns, a condition which was enforced. This test seems to be related to  the entityset code block you highlighted above where the child column dtype was set to `int64` if the dataframe was empty.

I think the equivalent test with WW would be to create an entityset with one empty dataframe (the child), with the child column in the relationship having a different ltype/dtype than the parent column, and no relationships defined initially. Then add a relationship between those columns and verify that the logical types of the columns (and thus the dtypes) in the relationship are the same. This will probably also raise a warning now, if I'm following along correctly.

featuretools/entityset/entityset.py
```diff
@@ -280,231 +294,240 @@ def add_relationship(self,
...
+                      time_index=None,
+                      secondary_time_index=None,
+                      already_sorted=False):
         """
-        Load the data for a specified entity from a Pandas DataFrame.
+        Add a DataFrame to the EntitySet with Woodwork typing information.
 
         Args:
-            entity_id (str) : Unique id to associate with this entity.
+            dataframe_id (str) : Unique id to associate with this dataframe.
```
rwedge, line 502: since on the dataframe, this is going to be stored `df.ww.name` and `df.ww.id` doesn't exist, what if we changed the terminology to `dataframe_name`  (but still clarify the name should be unique in the docstring)?  I think it would help readability in other places when we have variables like `parent_datframe_id` - if we changed those to `parent_dataframe_name` i think it would be easier to intuit where that value came from

featuretools/entityset/entityset.py
tamargrey: I think it was discussed briefly in the larger context of how `Categorical` logical type was getting used for columns that were variable type `Id`, and it was causing issues with this line because sometimes the categories in the dtype weren't matching. Somehow we ended here with comparing logical types, and I think part of that was deciding that it was okay to try converting logical types. 

Happy to document in the release notes if you're okay with leaving the behavior here.

featuretools/entityset/entityset.py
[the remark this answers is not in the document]
tamargrey: Would the equivalent of an Entity with an empty dataframe just be a TableSchema, or are we we definitely going to need to have a Woodwork DataFrame where all columns have no data but they all have the correct dtypes for the schema?

featuretools/entityset/entityset.py
tamargrey: yeah, that's a good idea! Were you also thinking we'd update some of the column terminology like `parent_column_id` to `parent_column_name`?

featuretools/tests/entityset_tests/test_es.py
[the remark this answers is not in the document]
tamargrey: ahh got it. I feel like that's more of a test of the logical type change warning in general, which we have in a couple of places in `test_es`, and the Woodwork DataFrames with no data is maybe something we should be testing in Woodwork, but I think that maybe this test and the lines in entityset where we convert to `int64` are going to be covered by the logical type change, or, like you said the fact that if the logical types are the same, then the dtypes should be the same.

featuretools/entityset/entityset.py
rwedge: yes, similar situation where `Variable.id` is now `ColumnAccessor.name`

tamargrey: cool, I'll change all of those to use `name` instead of `id`

tamargrey: updated dataframe and column id variables to use name instead. Left `EntitySet.id` since that felt separate enough, but happy to change if you think it should match.

featuretools/entityset/entityset.py
tamargrey: updated!

featuretools/entityset/entityset.py
```diff
@@ -526,403 +556,459 @@ def entity_from_dataframe(self,
...
+        if make_time_index is None and base_dataframe.ww.time_index is not None:
             make_time_index = True
 
         if isinstance(make_time_index, str):
             # Set the new time index to make_time_index.
             base_time_index = make_time_index
-            new_entity_time_index = make_time_index
-            already_sorted = (new_entity_time_index == base_entity.time_index)
+            new_dataframe_time_index = make_time_index
+            already_sorted = (new_dataframe_time_index == base_dataframe.ww.time_index)
```
rwedge, line 725: `already_sorted` evaluating to `True` is not tested on main branch, we may want to make an issue to add a test

featuretools/entityset/entityset.py
```diff
@@ -526,403 +556,459 @@ def entity_from_dataframe(self,
...
+                new_dataframe_time_index = "first_%s_time" % (base_dataframe.ww.name)
 
             already_sorted = True
 
-            assert base_entity.time_index is not None, \
-                "Base entity doesn't have time_index defined"
+            assert base_dataframe.ww.time_index is not None, \
+                "Base dataframe doesn't have time_index defined"
+
+            if base_time_index not in [v for v in additional_columns]:
```
rwedge, line 737: I think this should be `copy_columns`, not `additional_columns`.  on lines 690-692 we already check for the base time index being in additional columns and error if it is

featuretools/entityset/entityset.py
```diff
@@ -246,265 +260,281 @@ def add_relationships(self, relationships):
...
+        for path in self.find_forward_paths(goal_dataframe_name, start_dataframe_name):
             # Reverse path
             yield path[::-1]
 
-    def _forward_entity_paths(self, start_entity_id, seen_entities=None):
+    def _forward_dataframe_paths(self, start_dataframe_name, seen_dataframes=None):
         """
-        Generator which yields the ids of all entities connected through forward
-        relationships, and the path taken to each. An entity will be yielded
+        Generator which yields the ids of all dataframes connected through forward
```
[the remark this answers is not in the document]
tamargrey: changing!

featuretools/entityset/entityset.py
```diff
@@ -246,265 +260,281 @@ def add_relationships(self, relationships):
...
-    def get_forward_entities(self, entity_id, deep=False):
+    def get_forward_dataframes(self, dataframe_name, deep=False):
         """
-        Get entities that are in a forward relationship with entity
+        Get dataframes that are in a forward relationship with dataframe
 
         Args:
-            entity_id (str): Id entity of entity to search from.
-            deep (bool): if True, recursively find forward entities.
+            dataframe_name (str): Name dataframe of dataframe to search from.
```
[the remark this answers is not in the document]
tamargrey: changing here and in `get_backward_dataframes`

featuretools/entityset/entityset.py
```diff
@@ -526,403 +556,459 @@ def entity_from_dataframe(self,
...
-                will be created across this variable.
+            index (str): Column in old dataframe
+                that will become index of new dataframe. Relationship
+                will be created across this column.
 
-            additional_variables (list[str]):
-                List of variable ids to remove from
-                base_entity and move to new entity.
+            additional_columns (list[str]):
+                List of column ids to remove from
```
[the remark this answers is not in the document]
tamargrey: changing!

featuretools/entityset/entityset.py
```diff
@@ -526,403 +556,459 @@ def entity_from_dataframe(self,
...
+                base_dataframe and move to new dataframe.
 
-            copy_variables (list[str]): List of
-                variable ids to copy from old entity
-                and move to new entity.
+            copy_columns (list[str]): List of
+                column ids to copy from old dataframe
```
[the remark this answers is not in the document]
tamargrey: changing!

featuretools/entityset/entityset.py
```diff
@@ -246,265 +260,281 @@ def add_relationships(self, relationships):
...
-            time_index (str, optional): Name of the variable containing
-                time data. Type must be in :class:`variables.DateTime` or be
-                able to be cast to datetime (e.g. str, float, or numeric.)
+            time_index (str, optional): Name of the column containing
+                time data. Type must be numeric or datetime in nature.
 
-            secondary_time_index (dict[str -> Variable]): Name of variable
-                containing time data to use a second time index for the entity.
+            secondary_time_index (dict[str -> list[str]]): Name of column
+                containing time data to use a second time index for the dataframe.
```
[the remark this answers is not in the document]
tamargrey: That seems like a good idea! 
> The dictionary key is the column in the dataframe to be used as the secondary time index, and the value is the list of columns associated with that secondary time index.

This has gotten me thinking: do we allow multiple secondary time indexes to be set? Because in `normalize_dataframe` we specify for `make_secondary_time_index` that "only one secondary time index is allowed" but I don't think we actually restrict this anywhere.

featuretools/entityset/entityset.py
```diff
@@ -526,403 +556,459 @@ def entity_from_dataframe(self,
...
+    def add_last_time_indexes(self, updated_dataframes=None):
         """
-        Calculates the last time index values for each entity (the last time
+        Calculates the last time index values for each dataframe (the last time
         an instance or children of that instance were observed).  Used when
         calculating features using training windows
         Args:
-            updated_entities (list[str]): List of entity ids to update last_time_index for
-                (will update all parents of those entities as well)
+            updated_dataframes (list[str]): List of dataframe ids to update last_time_index for
```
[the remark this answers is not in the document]
tamargrey: changing! also updated a few other places where `ids` is still being used.

featuretools/entityset/entityset.py
```diff
@@ -526,403 +556,459 @@ def entity_from_dataframe(self,
...
-            raise ValueError("'additional_variables' contains duplicate variables. All variables must be unique.")
+        if not isinstance(copy_columns, list):
+            raise TypeError("'copy_columns' must be a list, but received type {}"
+                            .format(type(copy_columns)))
 
-        if not isinstance(copy_variables, list):
-            raise TypeError("'copy_variables' must be a list, but received type {}"
-                            .format(type(copy_variables)))
+        if len(copy_columns) != len(set(copy_columns)):
+            raise ValueError("'copy_columns' contains duplicate columns. All columns must be unique.")
```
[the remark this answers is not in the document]
tamargrey: that makes sense to me! Consolidated the three checks for them being lists, not having duplicates, and not containing the index column.

featuretools/entityset/entityset.py
```diff
@@ -526,403 +556,459 @@ def entity_from_dataframe(self,
...
 
-        if len(copy_variables) != len(set(copy_variables)):
-            raise ValueError("'copy_variables' contains duplicate variables. All variables must be unique.")
-
-        for v in additional_variables + copy_variables:
+        for v in additional_columns + copy_columns:
             if v == index:
-                raise ValueError("Not copying {} as both index and variable".format(v))
+                raise ValueError("Not copying {} as both index and column in copy_columns".format(v))
```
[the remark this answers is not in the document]
tamargrey: This line is also currently not being tested on main, so I added a test that covers the copy_columns and additional_columns cases

featuretools/entityset/entityset.py
tamargrey: yep, that makes sense to me. 

I've been looking into the case where we do have the base dataframe's time index in `copy_columns`, and ran into an issue where we include the original time index's column name as an element in `transfer_types`, so it causes an error when initializing Woodwork, since we don't include that column name in the normalized dataframe. Added some tests to catch this case as well as a fix for it.

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -31,6 +31,46 @@ def test_normalize_time_index_as_additional_column(es):
...
+    assert "signup_date" in es["customers"].columns
+    es.normalize_dataframe(base_dataframe_name='customers',
+                           new_dataframe_name='cancellations',
+                           index='cancel_reason',
+                           make_time_index='signup_date',
+                           additional_columns=[],
+                           copy_columns=['signup_date'])
+
+    assert 'signup_date' in es['cancellations'].columns
+    assert es['cancellations'].ww.time_index == 'signup_date'
```
rwedge, line 44: is this sufficient to check that we didn't accidentally move the column instead of copy it? does ww reset the time index if the time index column is removed?

tamargrey: Ah, I see. If we add a check that `signup_date in es['customers']` and maybe a check that it's still the time index there as well, that will be a more clear check that it's being copied and not moved.

The answer to whether ww resets the time index if that column is removed is yes. The concept of what a "time index" is is reliant on a column that has a `time_index` tag, so if the column is not on the DataFrame, the DataFrame has no time index. But Featuretools probably shouldn't have to rely on that intimate a knowlege of ww.

tamargrey: added checks to all of these tests that confirm that `signup_date` is still in the original dataframe

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -0,0 +1,619 @@
+from datetime import datetime
+
+import dask.dataframe as dd
+import numpy as np
+import pandas as pd
+import pytest
+
+import woodwork as ww
```
rwedge, line 8: these lint fixes don't look right, what is woodwork being classified as that it gets its own block instead of being lumped in with the 3rd party libraries?

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -36,80 +78,173 @@ def test_cannot_re_add_relationships_that_already_exists(es):
...
+                     logical_types=log_logical_types,
+                     semantic_tags=log_semantic_tags,
+                     time_index='datetime')
+    assert 'log2' in es.dataframe_dict
+    assert es['log2'].ww.schema is not None
+    assert es['log2'].ww.logical_types['product_id'] == ltypes.Categorical()
+    assert es['products'].ww.logical_types['id'] == ltypes.Categorical
+
+    warning_text = 'Logical type Categorical for child column product_id does not match parent '\
+        'column id logical type Categorical. Changing child logical type to match parent.'
```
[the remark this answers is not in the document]
tamargrey: The difference here is that one logical type is instantiated and the other is not. There's a Woodwork change in progress that's actually going to have all logical types instantiated, so this test will probably stop raising an error once that's in main.

But your point still stands because if the difference was parameters, we'd get the same error that'd be confusing. I'll update the error message to include the logical type's parameters if there are any.

featuretools/tests/entityset_tests/test_last_time_index.py
```diff
@@ -83,194 +86,194 @@ class TestLastTimeIndex(object):
...
 
         # add extra value instance with no children
         row_values = {'value': 21.0,
                       'value_time': pd.Timestamp("2011-04-10 11:10:02"),
-                      'values_id': 11}
+                      'values_id': 21.0}
         # make sure index doesn't have same name as column to suppress pandas warning
-        row = pd.DataFrame(row_values, index=pd.Index([11]))
-        df = values.df.append(row, sort=True)
+        row = pd.DataFrame(row_values, index=pd.Index([21]))
```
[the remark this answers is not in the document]
tamargrey: Woodwork needs the underlying index to match the column that's set as the Woodwork index exactly. The row being added here is the 11th row, but because the value in the index column at that row is 21, the index value also needs to be 21. I can change the value in `row_values` to be 11 instead, I think, if that would make more sense.

featuretools/tests/entityset_tests/test_ww_es.py
tamargrey: I can revert these. I've been having some problems with linting as of yesterday, and was somewhat surprised when all of these showed up yesterday, so I'm not totally sure what's acting up.

Looking at the lint CI check, it's failing because `ERROR: Could not find a version that satisfies the requirement woodwork>=0.3.0 (from featuretools[koalas]) (from versions: 0.0.1, 0.0.2, 0.0.3, 0.0.4, 0.0.5, 0.0.6, 0.0.7, 0.0.8, 0.0.9, 0.0.10, 0.0.11, 0.1.0)`

Maybe I included Woodwork incorrectly in the requirements? At the time I started this PR, I just put the latest version of Woodwork that was out, but locally I've been using my local Woodwork repo, and Woodwork `0.3.0` is probably too low of a requirement at this point. I'm nut sure what's best here.

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -0,0 +1,619 @@
...
+
+def test_conflicting_dataframe_names(es):
+    new_es = EntitySet()
+
+    sessions_df = es['sessions'].ww.copy()
+
+    assert sessions_df.ww.name == 'sessions'
+
+    new_es.add_dataframe('different_name', sessions_df)
+    assert sessions_df.ww.name == 'different_name'
```
rwedge, line 298: thoughts on warning the user we're changing the name if they supply a dataframe initialized with a different woodwork name than the name arg they supplied?

thoughts on name being optional if supplying a ww name and we infer from the `ww.name` if so?

featuretools/tests/entityset_tests/test_ww_es.py
rwedge: You might try a fresh environment with a non-local woodwork install and see if you get the same linting suggestions.

featuretools/tests/entityset_tests/test_ww_es.py
tamargrey: I like the idea of raising a warning if we're changing the name.

For the optional name parameter, I think it could make sense to have the name parameter be optional for a woodwork dataframe that's passed in but still take precedence if it is specified and is different from the one set on the Woodwork dataframe.

rwedge: > For the optional name parameter, I think it could make sense to have the name parameter be optional for a woodwork dataframe that's passed in but still take precedence if it is specified and is different from the one set on the Woodwork dataframe.

for the other arguments that can be set on an initialized woodwork dataframe, we ignore the optional parameters in favor of the having the user configure them on the woodwork dataframe.  would `name` be the sole exception to this if allowed the user param to take precedent?

tamargrey: Currently it is the one exception. If we want to ignore it like the rest of the parameters, then we need to make the `name` attribute required like we do with `index`.

I think I'd prefer this over changing the name if it doesn't match and raising a warning. The main difference from the index behavior is that for normal dataframes, if no index is specified, we use the first column. In this case, we would probably just tell users that they have to specify a name when woodwork is not already initialized.

rwedge: sounds good to me 👍

featuretools/tests/entityset_tests/test_ww_es.py
tamargrey: Still don't totally understand what state I was in yesterday that caused the linting I was seeing, but I was able to recreate it in a fresh virtual environment that had everything but Woodwork installed. Installing Woodwork fixed it.

The version thing is just that Woodwork stopped supporting python 3.6 in `0.1.0`.

tamargrey: I see 3.6 support was already dropped in #1413. Does this mean we need to get all the changes in from main? Or are we okay waiting a bit on that (maybe just till this branch is merged in)

featuretools/tests/entityset_tests/test_dask_es.py
```diff
@@ -1,67 +1,69 @@
...
+              "Cannot add dataframe of type {} to an entityset with existing dataframes " \
+              "of type {}".format(type(df), type(dask_es.dataframes[0]))
 
     with pytest.raises(ValueError, match=err_msg):
-        dask_es.entity_from_dataframe(
-            entity_id="new_entity",
+        dask_es.add_dataframe(
+            dataframe_name="new_dataframe",
             dataframe=df,
             index="id")
```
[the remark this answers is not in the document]
tamargrey: good call! Updating here and in `test_koalas_es` where we have the same situation

featuretools/tests/entityset_tests/test_ww_es.py
rwedge: I think we can wait until after this PR

featuretools/tests/entityset_tests/test_last_time_index.py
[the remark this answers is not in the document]
rwedge: 21 is fine 👍

featuretools/tests/entityset_tests/test_ww_es.py
tamargrey: Implemented these changes. The one area that I thought might still need some consideration is how we want to handle users passing in a woodwork-initialized DataFrame at EntitySet.__init__. Here's the current behavior:

```python
    df.ww.init(index='id', name='table')
    warning = 'A Woodwork-initialized DataFrame was provided, so the following parameters were ignored: dataframe_name'
    with pytest.warns(UserWarning, match=warning):
        es = EntitySet('es', dataframes={'table': (df,)})
```

I'm not entirely happy with this, because it feels like we shouldn't get any sort of warning if the name in the dataframes dictionary matches the name of the DataFrame. 

My suggestion is this, I think:
1. Only raise the ignored warning in `add_Dataframe` if the name parameter doesn't match the dataframe's name
2. If there is a conflict, warn (from the __init__ method) and go with the dictionary value (the logic being that they chose that name the most recently, in theory. but idk)

featuretools/tests/entityset_tests/test_es.py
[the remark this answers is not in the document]
tamargrey: Added an element to the warning that will indicate if the parameters do not match.

featuretools/tests/entityset_tests/test_ww_es.py
tamargrey: Added the change where we only raise the ignored warning when the name does match, but error if the names do not match.

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -79,6 +77,13 @@ def test_init_es_with_woodwork_table(df):
     assert es['table'].ww.logical_types['category'] == Categorical
 
 
+def test_init_es_with_woodwork_table_diff_name_error(df):
+    df.ww.init(index='id', name='table')
+    error = "Cannot add dataframe with conflicting dictionary key, diff_name, from dataframe name, table"
```
rwedge, line 82: how about "Naming conflict in dataframes dictionary:  dictionary key 'diff_name' does not match dataframe name 'table'"?

tamargrey: updated

rwedge (approved):
Looks good!

Issue #1256, opened 2020-12-08
Replace Primitive input/return Variable Types with Logical Types (using ColumnSchema)


gsheni:
- The primitives should now use Woodwork Logical Types for their inputs, and return types.
- If a primitive has multiple inputs, the syntax is the same.
- If a primitive has different inputs (OR), the syntax is the same.

thehomebrewnerd:
Closed by #1411

PR #1411, opened 2021-04-27
Primitives use Woodwork ColumnSchema for input types and return type

rwedge:
Addresses #1256 

The standard input / return type values have been changed, still need to check other files for custom primitives / update the docs

Issues / concerns I've run into so far:
* When Woodwork validates the ColumnSchema, it expects Ordinal to be an instance with defined ordering.
* Previously, `list_primitives` would list all valid variables types for a primitive, but since variable_type longer exists, I'm returning stringified ColumnSchema objects instead.  Perhaps this could be improved


featuretools/tests/primitive_tests/test_agg_feats.py
```diff
@@ -114,9 +108,14 @@ def count_generate_name(self, base_feature_names, relationship_path_name,
                                    where_str,
                                    use_prev_str)
 
-    Count = make_agg_primitive(count_func, [[Index], [Variable]], Numeric,
-                               name="count", stack_on_self=False,
-                               cls_attributes={"generate_name": count_generate_name})
+    Count = make_agg_primitive(
+        count_func,
+        [[ColumnSchema(semantic_tags={'foreign_key', 'category'})], [ColumnSchema()]],
```
thehomebrewnerd, line 113: Same as prior comment regarding foreign key columns - I don't think `category` should be included...

featuretools/tests/primitive_tests/test_groupby_transform_primitives.py
```diff
@@ -189,12 +190,12 @@ def test_cum_max(pd_es):
 
 def test_cum_sum_group_on_nan(pd_es):
     log_value_feat = pd_es['log']['value']
-    pd_es['log'].df['product_id'] = (['coke zero'] * 3 + ['car'] * 2 +
+    pd_es['log']['product_id'] = (['coke zero'] * 3 + ['car'] * 2 +
```
thehomebrewnerd, line 193: Indentation needs fixed here

featuretools/tests/primitive_tests/test_groupby_transform_primitives.py
```diff
@@ -230,12 +231,12 @@ def cum_sum(values):
             return cum_sum
 
     log_value_feat = pd_es['log']['value']
-    pd_es['log'].df['product_id'] = (['coke zero'] * 3 + ['car'] * 2 +
+    pd_es['log']['product_id'] = (['coke zero'] * 3 + ['car'] * 2 +
```
thehomebrewnerd, line 234: Indentation here also

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -153,45 +154,47 @@ def pd_simple_es():
     })
 
     es = ft.EntitySet('equal_test')
-    es.entity_from_dataframe('values', df, index='id')
+    es.add_dataframe('values', df, index='id')
```
thehomebrewnerd, line 157: The kwarg order got changed on `add_dataframe`, so I think the dataframe needs to come first and then the name string.

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -508,20 +511,20 @@ def pd_boolean_mult_es():
                        "bool": [True, False, True],
                        "numeric": [2, 3, np.nan]})
 
-    es.entity_from_dataframe(entity_id="test",
-                             dataframe=df,
-                             index="index")
+    es.add_dataframe(dataframe_name=="test",
+                     dataframe=df,
```
thehomebrewnerd, line 515: `==` -> `=`

thehomebrewnerd (commented):
Generally I think this is looking fine. Added multiple comments about `Boolean` vs `BooleanNullable` logical types and a few other things.

featuretools/primitives/standard/cum_transform_feature.py
```diff
@@ -47,8 +47,9 @@ class CumCount(TransformPrimitive):
         [1, 2, 3, 4, 5, 6]
     """
     name = "cum_count"
-    input_types = [[Id], [Discrete]]
-    return_type = Numeric
+    input_types = [[ColumnSchema(semantic_tags={'foreign_key', 'category'})],
```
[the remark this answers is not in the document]
tamargrey: Yeah, I think the fact that `Id` variable type has `_default_pandas_dtype = int` but is a subclass of Categorical was one of those things we won't exactly be able to recreate between variables and Woodwork. Agree that we probably just want the `foreign_key` tag here, because it's not uncommon for the columns to have `Integer` logical types.

featuretools/primitives/standard/binary_transform.py
```diff
@@ -28,8 +23,10 @@ class GreaterThan(TransformPrimitive):
         [True, False, False]
     """
     name = "greater_than"
-    input_types = [[Numeric, Numeric], [Datetime, Datetime], [Ordinal, Ordinal]]
-    return_type = Boolean
+    input_types = [[ColumnSchema(semantic_tags={'numeric'}), ColumnSchema(semantic_tags={'numeric'})],
+                   [ColumnSchema(logical_type=Datetime), ColumnSchema(logical_type=Datetime)],
+                   [ColumnSchema(logical_type=Ordinal), ColumnSchema(logical_type=Ordinal)]]
+    return_type = ColumnSchema(logical_type=Boolean)
```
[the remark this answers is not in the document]
rwedge: Also did some limited testing with each input type including nan values and the return values were still strictly `True` or `False`

I think we can leave as Boolean for now

featuretools/primitives/standard/cum_transform_feature.py
```diff
@@ -47,8 +47,9 @@ class CumCount(TransformPrimitive):
...
+                   [ColumnSchema(semantic_tags={'category'})]]
+    return_type = ColumnSchema(semantic_tags={'numeric'})
```
rwedge, line 52: should this be the "Integer" logical type, since values are countable and `NaN` values are counted?

thehomebrewnerd: Yeah, that makes sense to me!

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -182,8 +179,8 @@ class Minute(TransformPrimitive):
         [0, 10, 45]
     """
     name = "minute"
-    input_types = [Datetime]
-    return_type = Numeric
+    input_types = [ColumnSchema(logical_type=Datetime)]
+    return_type = ColumnSchema(semantic_tags={'numeric'})
```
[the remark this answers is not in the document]
rwedge: update to use ordinal and added the ordinal ordering for Second, Minute, Hour, Week, and Month

thehomebrewnerd (approved):
Noticed one more small thing, but otherwise this looks good.

PR #1412, opened 2021-04-29
Lower max depth to 1 if single entity

rwedge:
Due to the constraints we've added that prevents transform features being stacked on other transform features directly, if an entityset has a single entity, features beyond depth 1 cannot be created (can't stack transforms on each other and no other entities to get direct or aggregation features from).  Currently in such a scenario, DFS will spend time creating stacked transform features and throwing them out due to the stacking rules.

This PR will update our build features logic to lower the max depth to 1 if there is only a single entity so that DFS skips looking at depth 2 features it won't include

With this change, ensuring seed features greater than max_depth are still included is a concern since max_depth is being changed, so I added a test with a complex seed feature to confirm it was still included in the returned features

### DIFF
--- a/featuretools/tests/synthesis/test_deep_feature_synthesis.py
+++ b/featuretools/tests/synthesis/test_deep_feature_synthesis.py
@@ -451,7 +451,7 @@ def test_bad_groupby_feature(es):
 
 
 def test_abides_by_max_depth_param(es):
-    for i in [1, 2, 3]:
+    for i in [0, 1, 2, 3]:
         dfs_obj = DeepFeatureSynthesis(target_entity_id='sessions',
                                        entityset=es,
                                        agg_primitives=[Sum],
@@ -461,8 +461,22 @@ def test_abides_by_max_depth_param(es):
         features = dfs_obj.build_features()
         for f in features:
             # last feature is identity feature which doesn't count
-            assert (f.get_depth() <= i + 1)
+            assert (f.get_depth() <= i)
 
+
+def test_max_depth_single_table(transform_es):
+    assert len(transform_es.entity_dict) == 1
+    for i in [0, 1, 2]:
+        dfs_obj = DeepFeatureSynthesis(target_entity_id='first',
+                                       entityset=transform_es,
+                                       trans_primitives=[AddNumeric],
+                                       max_depth=i)
+
+        features = dfs_obj.build_features()
+        # check transform features made if max_depth>=1
+        assert any([f.get_depth() == min(i, 1) for f in features])
+        # check all features depth 1 or less
+        assert all([f.get_depth() <= min(i, 1) for f in features])
 
 def test_drop_contains(es):
     dfs_obj = DeepFeatureSynthesis(target_entity_id='sessions',
--- a/featuretools/tests/synthesis/test_deep_feature_synthesis.py
+++ b/featuretools/tests/synthesis/test_deep_feature_synthesis.py
@@ -612,7 +612,7 @@ def test_dfs_includes_seed_features_greater_than_max_depth(es):
                                    max_depth=1,
                                    seed_features=[customer_agg])
     features = dfs_obj.build_features()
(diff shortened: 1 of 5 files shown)


rwedge:
I marked it for review but the docs are failing due to this new warning cropping up, I'll need to lint / go through the docs and add some max_depth values

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -461,7 +461,32 @@ def test_abides_by_max_depth_param(es):
         features = dfs_obj.build_features()
         for f in features:
             # last feature is identity feature which doesn't count
```
thehomebrewnerd, line 463: Is this comment still valid/needed since it looks like we are checking this for all the features?

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -461,7 +461,32 @@ def test_abides_by_max_depth_param(es):
...
+        if i == 2:
+            match = ("Only one entity in entityset, changing max_depth to 1 "
+                     "since deeper features cannot be created")
+            with pytest.warns(UserWarning, match=match):
+                dfs_obj = make_dfs_obj(i)
+        else:
+            dfs_obj = make_dfs_obj(i)
+
+        features = dfs_obj.build_features()
+        # check transform features made if max_depth>=1
```
thehomebrewnerd, line 486: It took me a second to reconcile this comment with the code. Just want to confirm I understand...are we just checking that there is at least one transform feature with a depth of 1 anytime `max_depth>0`?

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -461,7 +461,32 @@ def test_abides_by_max_depth_param(es):
...
+        assert any([f.get_depth() == min(i, 1) for f in features])
+        # check all features depth 1 or less
+        assert all([f.get_depth() <= min(i, 1) for f in features])
```
thehomebrewnerd, line 489: Is `min` needed here? Can we update to `assert all([f.get_depth() <= 1 for f in features])` instead?

thehomebrewnerd (approved):
Looks good. Just a couple minor things. I'll defer to you if you think they need to be updated or not.

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -460,8 +460,37 @@ def test_abides_by_max_depth_param(es):
...
+
+def test_max_depth_single_table(transform_es):
+    assert len(transform_es.entity_dict) == 1
+    for i in [-1, 0, 1, 2, None]:
+        def make_dfs_obj(max_depth=i):
+            dfs_obj = DeepFeatureSynthesis(target_entity_id='first',
+                                           entityset=transform_es,
+                                           trans_primitives=[AddNumeric],
+                                           max_depth=max_depth)
+            return dfs_obj
```
thehomebrewnerd, line 474: Doesn't really impact things, but should this be moved outside of the for loop and the default value for `max_depth` be removed since we are always calling it below with a value defined?

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -460,8 +460,37 @@ def test_abides_by_max_depth_param(es):
...
+        features = dfs_obj.build_features()
+        assert len(features) > 0
+        if i != 0:
+            # at least one depth 1 feature made
+            assert any([f.get_depth() == 1 for f in features])
+            # no depth 2 or higher even with max_depth=2
+            assert all([f.get_depth() <= 1 for f in features])
+        else:
+            # no depth 1 or higher features with max_depth=0
+            assert all([f.get_depth() == 0 for f in features])
```
thehomebrewnerd, line 493: I find this much easier to follow!

thehomebrewnerd (approved):
Looks good, assuming tests pass!

Issue #1330, opened 2021-02-01
Drop Python 3.6 Support


gsheni:
- Python 3.6 is no longer supported by numpy (1.20.0)
- We plan on supporting Python 3.6 for now and dropping support after in late Feb/March. 
- Please let us know if you have any problems with this plan.

gsheni:
Merged and release:
https://github.com/alteryx/featuretools/pull/1413

PR #1413, opened 2021-04-29
Drop Python 3.6 support

rwedge:
Addresses #1330


rwedge:
6 checks are python 3.6 actions that weren't triggered, when we're ready to merge I'll remove those checks rom the required checks list

gsheni (approved):

PR #1417, opened 2021-05-03
[TEST - DO NOT MERGE] Use list for interesting values

thehomebrewnerd:
Creating this as a test for using a list to store interesting values instead of a pandas Series.


thehomebrewnerd:
No longer needed.

PR #1421, opened 2021-05-04
Store interesting values on column metadata

### LINKED ISSUE #1375: Store variable/column interesting values on EntitySet
### Store variable/column interesting values on EntitySet

The current `Variable` class stores interesting values for variables, and also provides a setter method for assigning interesting values. When `Variable` is replaced with a Woodwork Series, the `interesting_values` property and corresponding setter will no longer be present.

As a result, setting and storing interesting values need to move somewhere else in the Featuretools code. One potential solution is to store these values on `EntitySet` instead, but other solutions can be investigated.

- Current API
```python
es["entity_name"]["variable_name"].interesting_values = ["Value 1", "Value 2"]
Entity.add_interesting_values()
```
- New API Proposal 1 --> Not ideal, too complex
```python
es["entity_name"].ww['variable_name'].ww.metadata['interesting_values'] = ["Value 1", "Value 2"]
```

- New API Proposal 2 --> More thought needed, seems good on first glance
```python
es.add_interesting_values(entity_id='entity_name', interesting_values ={'name': ["Value 1", "Value 2"]})
```

thehomebrewnerd:
### Store interesting values on column metadata

Closes #1375 

This PR updates `Entityset.add_interesting_values` to store the values as a list on the corresponding column metadata. With this approach the serialization/deserialization by Woodwork should be sufficient for handling these values. This PR does not update any serialization or deserialization within Featuretools. Also, DFS was not updated to pull the interesting values from the metadata instead of the Variable object, but this will be needed in the future.

We should wait to merge this until PR #1416 is merged as we still need to uncomment a test related to verbose output which requires changes from #1416.

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1,4 +1,4 @@
-# import copy
+import copy
 import logging
 import warnings
 from collections import defaultdict
@@ -301,7 +301,7 @@ class EntitySet(object):
             msg = "Unable to add relationship because child column '{}' in '{}' is also its index"
             raise ValueError(msg.format(child_column, child_df.ww.name))
         parent_df = relationship.parent_dataframe
-        parent_column = relationship.parent_column.name
+        parent_column = relationship._parent_column_id
         if 'foreign_key' not in child_df.ww.semantic_tags[child_column]:
             child_df.ww.add_semantic_tags({child_column: 'foreign_key'})
 
@@ -519,7 +519,7 @@ class EntitySet(object):
             time_index (str, optional): Name of the column containing
                 time data. Type must be numeric or datetime in nature.
 
-            secondary_time_index (dict[str -> Series]): Name of column
+            secondary_time_index (dict[str -> list[str]]): Name of column
                 containing time data to use a second time index for the dataframe.
 
             already_sorted (bool, optional) : If True, assumes that input dataframe
@@ -580,9 +580,9 @@ class EntitySet(object):
                               already_sorted=already_sorted)
             # If no index column is specified, set the first column
             if dataframe.ww.index is None:
-                dataframe.ww.set_index(dataframe.columns[0])
                 warnings.warn(("Using first column as index. "
                                "To change this, specify the index parameter"))
+                dataframe.ww.set_index(dataframe.columns[0])
 
         else:
             if dataframe.ww.index is None:
@@ -607,12 +607,12 @@ class EntitySet(object):
             # make sure name is set to match input dataframe_id
             dataframe.ww.name = dataframe_id
 
-        if secondary_time_index:
-            self._set_secondary_time_index(dataframe, secondary_time_index=secondary_time_index)
-
         if dataframe.ww.time_index is not None:
             self._check_uniform_time_index(dataframe)
             self._check_secondary_time_index(dataframe)
+
+        if secondary_time_index:
+            self._set_secondary_time_index(dataframe, secondary_time_index=secondary_time_index)
 
         self.dataframe_dict[dataframe_id] = dataframe
         self.reset_data_description()
@@ -667,14 +667,14 @@ class EntitySet(object):
                             .format(type(additional_columns)))
 
         if len(additional_columns) != len(set(additional_columns)):
-            raise ValueError("'additional_columns' contains duplicate variables. All variables must be unique.")
+            raise ValueError("'additional_columns' contains duplicate columns. All columns must be unique.")
 
         if not isinstance(copy_columns, list):
             raise TypeError("'copy_columns' must be a list, but received type {}"
                             .format(type(copy_columns)))
 
         if len(copy_columns) != len(set(copy_columns)):
-            raise ValueError("'copy_columns' contains duplicate variables. All variables must be unique.")
+            raise ValueError("'copy_columns' contains duplicate columns. All columns must be unique.")
 
         for v in additional_columns + copy_columns:
             if v == index:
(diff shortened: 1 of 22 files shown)


featuretools/entityset/entityset.py
```diff
@@ -1008,66 +1006,79 @@ def add_last_time_indexes(self, updated_dataframes=None):
     # #  Other ###############################################
     # ###########################################################################
 
-    # def add_interesting_values(self, max_values=5, verbose=False, entity_id=None):
-    #     """Find interesting values for categorical variables, to be used to generate "where" clauses
+    def add_interesting_values(self, max_values=5, verbose=False, dataframe_id=None, values=None):
+        """Find interesting values for categorical variables, to be used to generate "where" clauses
```
tamargrey, line 1010: Not sure if we want to say "categorical columns" here or something like "columns that are categorical in nature" or something more woodwork-specific.

featuretools/entityset/entityset.py
```diff
@@ -1008,66 +1006,79 @@ def add_last_time_indexes(self, updated_dataframes=None):
...
 
-    #     Args:
-    #         max_values (int) : Maximum number of values per variable to add.
-    #         verbose (bool) : If True, print summary of interesting values found.
-    #         entity_id (str) : The Entity for which to add interesting values. If not specified
-    #             interesting values will be added for all Entities.
+        Args:
+            max_values (int) : Maximum number of values per variable to add.
```
tamargrey, line 1013: variable usage - "per column to add"

featuretools/entityset/entityset.py
```diff
@@ -1008,66 +1006,79 @@ def add_last_time_indexes(self, updated_dataframes=None):
...
+        if dataframe_id:
+            dataframes = [self[dataframe_id]]
+        else:
+            dataframes = self.dataframes
+        for df in dataframes:
+            for column in df.columns:
+                # some heuristics to find basic 'where'-able variables
+                if df.ww.columns[column].is_categorical:
+                    # TODO - consider removing this constraints
+                    # don't add interesting values for entities in relationships
```
tamargrey, line 1043: Not really related to these changes, but if I'm understanding the logic below this for skipping the interesting value calculation, it's happening on the column level not the DataFrame level, right? 

We're currently iterating through every column in every dataframe, and at each column, we `continue` if the column is a child or parent column in a relationship, but because we're already in a nested for loop, we will just go onto the next column, not the next dataframe.

Is the goal here to just avoid parent and child columns or the dataframes that they come from? If it's the later, we will probably need to pull the `skip` variable to a wider scope so we can `continue` again in the outer for loop.

featuretools/entityset/entityset.py
```diff
@@ -1008,66 +1006,79 @@ def add_last_time_indexes(self, updated_dataframes=None):
...
+                    skip = False
+                    for r in self.relationships:
+                        if column in [r.child_column.name, r.parent_column.name]:
+                            skip = True
+                            break
+                    if skip:
+                        continue
+
+                    counts = df[column].value_counts()
```
tamargrey, line 1052: could we make use of woodwork's value_counts here? It already calculates value counts only for columns that are `is_categorical`.

I guess to use it, we'd end up calling it for the dataframe as a whole and then having to go through column by column and get the interesting values after the fact, so it'd be a structural change here, but thought I'd bring it up!

featuretools/entityset/entityset.py
thehomebrewnerd: All this code was carried over from the existing implementation, but it does seem like the comment doesn't quite match the code, so I'm not sure. Maybe this should read `don't add interesting values for columns in relationships` instead?

@rwedge ???

featuretools/entityset/entityset.py
thehomebrewnerd: Interesting idea! Might allow us to eliminate some looping here and simplify this code a bit. I'll create an issue to look into that further.

featuretools/entityset/entityset.py
gsheni: I think saying something like  "categorical columns" or "columns that are categorical in nature" is good. We want to avoid saying Woodwork-specific things in Featuretools to avoid putting a burden on the end-user to learn Woodwork.

thehomebrewnerd: Updated to `categorical columns`

featuretools/entityset/entityset.py
rwedge: I think the comment is wrong and the intent is to avoid computing intersting values for Id / foreign key columns

featuretools/entityset/entityset.py
thehomebrewnerd: https://github.com/alteryx/featuretools/issues/1426

featuretools/entityset/entityset.py
thehomebrewnerd: I'll update the comment - or should we just remove it completely?

rwedge: Updating seems fine

featuretools/entityset/entityset.py
```diff
@@ -1008,66 +1006,74 @@ def add_last_time_indexes(self, updated_dataframes=None):
...
-    #         max_values (int) : Maximum number of values per variable to add.
-    #         verbose (bool) : If True, print summary of interesting values found.
-    #         entity_id (str) : The Entity for which to add interesting values. If not specified
-    #             interesting values will be added for all Entities.
+        Args:
+            max_values (int) : Maximum number of values per column to add.
+            verbose (bool) : If True, print summary of interesting values found.
+            dataframe_id (str) : The dataframe in the EntitySet for which to add interesting values.
+                If not specified interesting values will be added for all dataframes.
+            values (dict): A dictionary mapping column names to the interesting values to set
```
tamargrey, line 1017: Is the purpose of adding a `values` parameter to give users the ability to avoid having to bring dask and koalas dataframes into memory? 

If so, maybe we should add a note in the docstring about it being useful for large datasets. 

The other things is that unless we're going to explicitly disallow finding the interesting values for dask and koalas dataframes, we should probably include them in the tests in some way, even if it's just xfailing and saying that we're recommending it not be used.

thehomebrewnerd: Previously there was a setter on the Variable that users could use to manually assign specific interesting values, but that no longer will exist after the Woodwork integration. The `values` parameter here is what is providing that same type of functionality.

We currently do not support adding interesting values automatically for Dask or Koalas as that would require pulling the data into memory, so Dask/Koalas users must pass via the `values` argument. This approach is tested in `test_add_interesting_values_specified_vals`. 

When we originally did the Dask and Koalas implementation for Featuretools we adopted the convention that if we ran a test only for pandas, that means we are not supporting that behavior for Dask and Koalas, and don't have plans to support, but if we included an xfail that was a feature that we don't currently support but would like to. The idea was that the xfails would point us to things to work on. I'm not sure we have followed that strictly though - might be worth a conversation to determine how we want to handle moving forward since we haven't used that approach in other libraries that I'm aware of.

tamargrey: Ah, I see. Thank you for the explanation! (I definitely have not been consciously following the xfail convention in woodwork, but it seems like a good one to have!)

I guess my question, then, becomes whether we should tell users in the docstring that they should supply `values` for dask and koalas dataframes. If I'm understanding correctly, though, it seems like featuretools assumes a user working with dask or koalas has read through the corresponding guides and knows about all the limitations through that. In which case, we should leave any additional notes about the limitations to the guides.

tamargrey (approved):
lgtm!

PR #1427, opened 2021-05-06
Clean up replace Entity Woodwork integration

tamargrey:
- Adds some tests in areas that were under-tested with respect to Woodwork's behavior in featuretools.
- cleans up Entity/Variable usage in docstrings, tests, and errors
- Brings up a few behavior questions

### DIFF
--- a/featuretools/tests/entityset_tests/test_es.py
+++ b/featuretools/tests/entityset_tests/test_es.py
@@ -1230,24 +1230,26 @@ def test_normalize_time_index_from_none(normalize_es):
 
 
 def test_raise_error_if_dupicate_additional_variables_passed(es):
-    error_text = "'additional_columns' contains duplicate variables. All variables must be unique."
+    error_text = "'additional_columns' contains duplicate columns. All columns must be unique."
     with pytest.raises(ValueError, match=error_text):
         es.normalize_dataframe('sessions', 'device_types', 'device_type',
                                additional_columns=['device_name', 'device_name'])
 
 
-def test_raise_error_if_dupicate_copy_variables_passed(es):
-    error_text = "'copy_columns' contains duplicate variables. All variables must be unique."
+def test_raise_error_if_dupicate_copy_columns_passed(es):
+    error_text = "'copy_columns' contains duplicate columns. All columns must be unique."
     with pytest.raises(ValueError, match=error_text):
         es.normalize_dataframe('sessions', 'device_types', 'device_type',
                                copy_columns=['device_name', 'device_name'])
 
 
-def test_normalize_dataframe_copies_variable_types(es):
+def test_normalize_dataframe_copies_logical_types(es):
     es['log'].ww.set_types(logical_types={'value': ltypes.Ordinal(order=[0.0, 1.0, 2.0, 3.0, 5.0, 7.0, 10.0, 14.0, 15.0, 20.0])})
 
     assert isinstance(es['log'].ww.logical_types['value'], ltypes.Ordinal)
+    assert len(es['log'].ww.logical_types['value'].order) == 10
     assert isinstance(es['log'].ww.logical_types['priority_level'], ltypes.Ordinal)
+    assert len(es['log'].ww.logical_types['priority_level'].order) == 3
     es.normalize_dataframe('log', 'values_2', 'value_2',
                            additional_columns=['priority_level'],
                            copy_columns=['value'],
@@ -1262,7 +1264,9 @@ def test_normalize_dataframe_copies_variable_types(es):
     assert 'value' in es['log'].columns
     assert 'value_2' in es['values_2'].columns
     assert isinstance(es['values_2'].ww.logical_types['priority_level'], ltypes.Ordinal)
+    assert len(es['values_2'].ww.logical_types['priority_level'].order) == 3
     assert isinstance(es['values_2'].ww.logical_types['value'], ltypes.Ordinal)
+    assert len(es['values_2'].ww.logical_types['value'].order) == 10
 
 
 # sorting not supported in Dask, Koalas
--- a/featuretools/tests/entityset_tests/test_es.py
+++ b/featuretools/tests/entityset_tests/test_es.py
@@ -20,7 +20,7 @@ from featuretools.utils.koalas_utils import pd_to_ks_clean
 ks = import_or_none('databricks.koalas')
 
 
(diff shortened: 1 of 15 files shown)


featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -146,23 +152,79 @@ def test_init_es_with_relationships(pd_df):
...
+
 
+def test_time_type_check_order(dates_df):
+    dates_df.ww.init(index='backwards_order', time_index='random_order')
+    es = EntitySet('es')
+
+    error = 'dates_table time index is numeric type which differs from other entityset time indexes'
+    with pytest.raises(TypeError, match=error):
+        # Because we set secondary time index before checking that time_index is valid, the time type will match the secondary time index
+        es.add_dataframe('dates_table', dates_df, secondary_time_index={'repeating_dates': ['random_order', 'special']})
```
tamargrey, line 187: In `EntitySet.add_dataframe`, we set the secondary time index before checking uniform time index for a set `time_index`. 

We still get an error if the time index type doesn't match the secondary time index type, but it seems a little odd to me that we set the time type based off of the secondary time index and not the time index. 

Is there any reason to not rearrange the order that we set the secondary time index?

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -146,23 +152,79 @@ def test_init_es_with_relationships(pd_df):
...
+    assert es._check_uniform_time_index(es['dates_table']) is None
+
+    # --> bug? IF a user sets the time index after initalization through woodwork, there's no way to check the time type
+    # --> should we add a entityset add time index method to help with this??
+    dates_df.ww.set_time_index('random_order')
+    assert dates_df.ww.time_index == 'random_order'
+
+    error = 'dates_table time index is numeric type which differs from other entityset time indexes'
+    with pytest.raises(TypeError, match=error):
+        es._check_uniform_time_index(es['dates_table'])
```
tamargrey, line 211: If a user wants to change the time index of their DataFrame in an EntitySet, they'll go through woodwork's `set_time_index`, which means that we don't get the time type check, and a user could set a time index that is the incorrect time type, and they won't catch it until later. 

Are we okay with this? If not, we might want to add a `EntitySet.set_dataframe_time_index` method that confirms the time type is valid for the EntitySet.

thehomebrewnerd: The idea behind PR #1400 was to try to gracefully catch time type issues for the scenario you describe. My thought is to try to keep things simple and not implement an additional method for setting the time index for now, but we could revisit if it proves to be a problem later.

tamargrey: ah, okay, I'm seeing that part of the conversation now.

The time type checking on EntitySet does a really good job of catching time type issues, so it sounds like we can leverage it later on in dfs to let users know if there's an issue with mismatched time types.

featuretools/tests/entityset_tests/test_ww_es.py
thehomebrewnerd: Seems like we should flip the order and set the time type from the time index and not the secondary time index. I can't think of a reason to keep the order as is.

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -146,23 +153,94 @@ def test_init_es_with_relationships(pd_df):
...
+
+    es.add_dataframe('dates_table', dates_df, secondary_time_index={'repeating_dates': ['random_order', 'special']})
+
+    assert dates_df.ww.metadata['secondary_time_index'] == {'repeating_dates': ['random_order', 'special', 'repeating_dates']}
+    assert es.time_type == Datetime
+
+    assert es._check_uniform_time_index(es['dates_table']) is None
+
+    # --> bug? IF a user sets the time index after initalization through woodwork, there's no way to check the time type
+    # --> should we add a entityset add time index method to help with this??
```
thehomebrewnerd, line 206: remove comments

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -497,3 +575,29 @@ def test_update_dataframe_last_time_index(es):
...
+                           additional_columns=['priority_level'],
+                           copy_columns=['value'],
+                           make_time_index=False)
+
+    # Metadata in the original dataframe is maintained, but it's lost in the new dataframe
+    assert 'interesting_values' in es['log'].ww.columns['value'].metadata
+    assert 'interesting_values' not in es['values_2'].ww.columns['value'].metadata
+    assert 'interesting_values' not in es['values_2'].ww.columns['priority_level'].metadata
+    assert es['log'].ww.columns['value'].description == 'a value column'
+    assert es['values_2'].ww.columns['value'].description is None
```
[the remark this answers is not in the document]
thehomebrewnerd: Seems to me like we should keep the metadata for the columns.

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -1036,11 +1021,19 @@ def test_sets_time_when_adding_entity(transactions_df):
...
 
 
-def test_checks_time_type_setting_time_index(es):
+def test_secondary_time_index_no_primary_time_index(es):
+    es['products'].ww.set_types(logical_types={'rating': 'Datetime'})
+    assert es['products'].ww.time_index is None
+    es.set_secondary_time_index('products', {'rating': ['url']})
+
+    assert 'secondary_time_index' in es['products'].ww.metadata
+    assert es['products'].ww.time_index is None
```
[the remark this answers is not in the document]
tamargrey: thanks for bringing this up! Meant to highlight it as well. 

Yeah, I think it doesn't really make sense to have a secondary time index if there's no primary one set.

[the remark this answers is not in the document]
tamargrey: Adding a check in `_check_secondary_time_index` to confirm that a primary time index is set on the dataframe where we're setting a secondary time index.

featuretools/tests/entityset_tests/test_ww_es.py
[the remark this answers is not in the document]
thehomebrewnerd: Ah, sorry. Both.

thehomebrewnerd (approved):
LGTM!

PR #1428, opened 2021-05-11
Add minimum dependency generator GitHub Action

gsheni:
- Add a CI workflow to generate the minimum dependencies
- Renamed `latest_dependencies.txt` to `latest_requirements.txt`
- Moved `latest_requirements.txt` file into 1 folder
- `fastparquet` had a release which broke our unit tests with latest dependencies, so I set a upper limit.


rwedge:
Should we make an issue to track the problem with the latest fastparquet version?

PR #1436, opened 2021-05-12
Update nbsphinx version to resolve docs build issue

thehomebrewnerd:
### Fix docs build issue
Update nbsphinx version to resolve docs build issue


rwedge:
Looks good, has the issue been raised with whatever dependency this is from?

thehomebrewnerd:
> Looks good, has the issue been raised with whatever dependency this is from?

Not sure - will have to check on this.

gsheni (approved):

PR #1445, opened 2021-05-18
Fix numpy installation for minimum unit tests

gsheni:
- Using a low numpy version causes the following issue with the unit tests, serialization, and fastparquet (which uses numpy header files):
```python
E   ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject
```
- This MR solves this, according to the following: https://stackoverflow.com/a/66743692/2512385

### DIFF
--- a/.github/workflows/unit_tests_with_minimum_deps.yml
+++ b/.github/workflows/unit_tests_with_minimum_deps.yml
@@ -36,14 +36,16 @@ jobs:
         name: Get numpy version for koalas
         run: |
           NUMPY_VERSION=$(cat featuretools/tests/requirement_files/minimum_koalas_requirements.txt | grep numpy)
+          python -m pip uninstall numpy -y
+          python -m pip install $NUMPY_VERSION --no-build-isolation
       - if: ${{ matrix.libraries == 'core' }}
         name: Get numpy version for core
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/unit_tests_with_minimum_deps.yml
```diff
@@ -32,6 +32,18 @@ jobs:
...
         run: |
           python -m pip install -e . --no-dependencies
+      - if: ${{ matrix.libraries == 'koalas' }}
+        name: Get numpy version for koalas
+        run: |
+          NUMPY_VERSION=$(cat featuretools/tests/requirement_files/minimum_koalas_requirements.txt | grep numpy)
+          python -m pip uninstall numpy -y
+          python -m pip install $NUMPY_VERSION --no-build-isolation
+      - if: ${{ matrix.libraries == 'core' }}
+        name: Get numpy version for core
```
rwedge, line 42: For these 2 names, perhaps "Install numpy version . . ." for each name field?

gsheni: fixed

rwedge (approved):
Looks good

PR #1451, opened 2021-05-26
Use woodwork 0.4.0

### LINKED ISSUE #1449: Update woodwork integration to use Woodwork version 0.4.0
The addition of `LogicalType.transform` and the fact that logical types in Woodwork objects are now instances means that the current Woodwork integration into featuretools, which is based off of Woodwork version 0.3.1, will not work for 0.4.0.

We should update the integration to use 0.4.0 and bump the min woodwork version as such.

tamargrey:
Updates woodwork integration to use woodwork version 0.4.0.

closes #1449

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -15,6 +15,7 @@ Future Release
         * Replaced ``Entity.update_data`` with ``EntitySet.update_dataframe`` (:pr:`1398`)
         * Move validation check for uniform time index to ``EntitySet`` (:pr:`1400`)
         * Replace ``Entity`` objects in ``EntitySet`` with Woodwork dataframes (:pr:`1405`)
+        * Update Woodwork integration to use Woodwork ``0.4.0``
     * Documentation Changes
         * Improve formatting of release notes (:pr:`1396`)
     * Testing Changes
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -15,7 +15,7 @@ Future Release
         * Replaced ``Entity.update_data`` with ``EntitySet.update_dataframe`` (:pr:`1398`)
         * Move validation check for uniform time index to ``EntitySet`` (:pr:`1400`)
         * Replace ``Entity`` objects in ``EntitySet`` with Woodwork dataframes (:pr:`1405`)
-        * Update Woodwork integration to use Woodwork ``0.4.0``
+        * Update Woodwork integration to use Woodwork ``0.4.0`` (:pr:`1451`)
     * Documentation Changes
         * Improve formatting of release notes (:pr:`1396`)
     * Testing Changes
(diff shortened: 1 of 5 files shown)


featuretools/tests/entityset_tests/test_es.py
```diff
@@ -131,16 +131,17 @@ def test_add_relationship_instantiated_logical_types(es):
...
-    assert es['products'].ww.logical_types['id'] == ltypes.Categorical
+    # --> see if we need to change this test to test something else now... not sure this will still fail :/
+    assert isinstance(es['log2'].ww.logical_types['product_id'], ltypes.Categorical)
+    assert isinstance(es['products'].ww.logical_types['id'], ltypes.Categorical)
 
     warning_text = 'Logical type Categorical for child column product_id does not match parent '\
         'column id logical type Categorical. There is a conflict between the parameters. '\
         'Changing child logical type to match parent.'
     with pytest.warns(UserWarning, match=warning_text):
         es.add_relationship(u'products', 'id', 'log2', 'product_id')
```
tamargrey, line 142: This warning no longer gets thrown, because initializing with `Categorical` or `Categorical()` now results in the same logical type. The way to trigger this warning now will be to have two logical types that are the same but have different parameters. For example `Datetime()` and `Datetime(datetime_format)`. However, in this case, the columns need to be parent and child columns in a relationship, and none of the index columns in the mock ecommerce entityset have parameters, so I'm not totally sure how we're going to cover this line. 

One option might be to change the logical types to be `Ordinal`, but the parameters would need to be different, and the `order` specified in Ordinal probably won't be different between parent and child ids. 

The other option is to create a new fixture for this test.

rwedge: New fixture / modifying the fixture when it gets passed to the test so that there are different parameters sounds good

tamargrey: figured out how to do it without having to create a new fixture! (I forgot that we can add elements to `order` that aren't in the series, so that way we can have different orders between child and parent columns!

tamargrey: figured out how to do it without having to create a new fixture! (I forgot that we can add elements to `order` that aren't in the series, so that way we can have different orders between child and parent columns!

rwedge (approved):
LGTM

docs/source/release_notes.rst
```diff
@@ -15,6 +15,7 @@ Future Release
         * Replaced ``Entity.update_data`` with ``EntitySet.update_dataframe`` (:pr:`1398`)
         * Move validation check for uniform time index to ``EntitySet`` (:pr:`1400`)
         * Replace ``Entity`` objects in ``EntitySet`` with Woodwork dataframes (:pr:`1405`)
+        * Update Woodwork integration to use Woodwork ``0.4.0``
```
jeff-hernandez, line 18: Missing ``(:pr:`1451`)`` at the end

jeff-hernandez (approved):

docs/source/release_notes.rst
tamargrey: whoopse 🤦‍♀️ thanks for pointing out!

rwedge: I think we can leave this out of the release notes

tamargrey: okay, I'll remove

Issue #1254, opened 2020-12-08
Put last_time_index in DataFrame metadata and move Series to DataFrame


gsheni:
- With #1253, the last_time_index is a Series in the metadata
- However, it should now be moved to be a Series on the actual DataFrame
- The column name should be kept in the metadata of the DataFrame

thehomebrewnerd:
When this is done, we should also review the implementation of `EntitySet.__sizeof__` to make sure the last_time_index is included correctly.

thehomebrewnerd:
As part of this issue, we should also make sure that `test_to_parquet_with_lti` in `test_serialization.py` is working correctly. This test is currently commented out because the lti series stored in the WW metadata is not JSON serializable and fails as a result.

PR #1452, opened 2021-05-26
WW/FT Serialization Updates

### LINKED ISSUE #1261: Update Entityset schema version with new DataFrame Accessor
- With the Woodwork integration, Entity will be replaced with a DataFrame (with Woodwork Accessor)
- This will change the EntitySet save to disk structure, and so it will need to be updated.

### LINKED ISSUE #1266: Update serialization to use DataFrame Accessor methods
- The serialization needs to be updated to use DataFrame's serialization (via Woodwork accessor)
- There needs be a discussion about what the new EntitySet structure looks like, and how breaking schema changes from Woodwork will be handled.

thehomebrewnerd:
### WW/FT Serialization Updates
Closes #1266 
Closes #1261 

Updates entityset serialization to work with Woodwork dataframes.

To Do:
 - [x] Verify to_csv tests work with koalas after [woodwork issue 950](https://github.com/alteryx/woodwork/issues/950) is resolved
 - [x] Confirm `test_to_parquet_with_lti` passes after last time index is moved out of WW metadata
 - [x] Confirm `test_operations_invalidate_metadata` passes for Koalas after last time index is moved out of WW metadata

### DIFF
--- a/featuretools/entityset/deserialize.py
+++ b/featuretools/entityset/deserialize.py
@@ -41,10 +41,6 @@ def description_to_entityset(description, **kwargs):
 
         entityset.add_dataframe(dataframe)
 
-        # description_to_dataframe(df, entityset, path=path)
-        # if entity['properties']['last_time_index']:
-        #     last_time_index.append(entity['id'])
-
     for relationship in description['relationships']:
         rel = Relationship.from_dictionary(relationship, entityset)
         entityset.add_relationship(relationship=rel)
--- a/featuretools/entityset/deserialize.py
+++ b/featuretools/entityset/deserialize.py
@@ -57,7 +57,7 @@ def empty_dataframe(description):
         df (DataFrame) : Empty dataframe with Woodwork initialized.
     '''
     # TODO: Can we update Woodwork to return an empty initialized dataframe from a description
-    # instead of using this function?
+    # instead of using this function? Or otherwise eliminate? Issue #1476
     logical_types = {}
     semantic_tags = {}
     column_descriptions = {}
--- a/featuretools/entityset/deserialize.py
+++ b/featuretools/entityset/deserialize.py
@@ -63,6 +63,7 @@ def empty_dataframe(description):
     column_descriptions = {}
     column_metadata = {}
     use_standard_tags = {}
+    category_dtypes = {}
     columns = []
     for col in description['column_typing_info']:
         col_name = col['name']
@@ -84,7 +85,14 @@ def empty_dataframe(description):
         column_metadata[col_name] = col['metadata']
         use_standard_tags[col_name] = col['use_standard_tags']
 
-    dataframe = pd.DataFrame(columns=columns)
+        if col['physical_type']['type'] == 'category':
+            # Make sure categories are recreated properly
+            cat_values = col['physical_type']['cat_values']
+            cat_dtype = col['physical_type']['cat_dtype']
+            cat_object = pd.CategoricalDtype(pd.Index(cat_values, dtype=cat_dtype))
+            category_dtypes[col_name] = cat_object
+    dataframe = pd.DataFrame(columns=columns).astype(category_dtypes)
+
     dataframe.ww.init(
         name=description.get('name'),
         index=description.get('index'),
--- a/featuretools/entityset/deserialize.py
+++ b/featuretools/entityset/deserialize.py
@@ -31,7 +31,6 @@ def description_to_entityset(description, **kwargs):
     path = description.get('path')
     entityset = EntitySet(description['id'])
 
-    last_time_index = []
     for df in description['dataframes'].values():
         if path is not None:
             data_path = os.path.join(path, 'data', df['name'])
@@ -44,9 +43,6 @@ def description_to_entityset(description, **kwargs):
     for relationship in description['relationships']:
         rel = Relationship.from_dictionary(relationship, entityset)
         entityset.add_relationship(relationship=rel)
-
-    if len(last_time_index):
-        entityset.add_last_time_indexes(updated_entities=last_time_index)
 
     return entityset
 
@@ -62,14 +58,11 @@ def empty_dataframe(description):
     '''
     # TODO: Can we update Woodwork to return an empty initialized dataframe from a description
     # instead of using this function?
-    loading_info = description['loading_info']
-    table_type = loading_info.get('table_type', 'pandas')
     logical_types = {}
     semantic_tags = {}
     column_descriptions = {}
     column_metadata = {}
     use_standard_tags = {}
-    category_dtypes = {}
     columns = []
     for col in description['column_typing_info']:
         col_name = col['name']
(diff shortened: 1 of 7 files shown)


rwedge:
I think it looks good so far.  Do we want to merge it and handle the TODOs after, or wait for the other issues to be resolved first?

thehomebrewnerd:
> I think it looks good so far. Do we want to merge it and handle the TODOs after, or wait for the other issues to be resolved first?

@rwedge I could go either way here. I was thinking it might be nice to confirm that everything works for Koalas once the WW serialization bug is fixed before we merge this in. I think the last time index related updates could be handled as a separate PR or as part of the PR that will be created for #1254 

Would it be beneficial to you to get this merged in sooner?

featuretools/entityset/deserialize.py
```diff
@@ -108,92 +52,66 @@ def description_to_entityset(description, **kwargs):
...
-        for column in latlongs:
-            if entity_type == 'dask':
-                meta = (column, tuple([float, float]))
-                dataframe[column] = dataframe[column].apply(parse_latlong_tuple,
-                                                            meta=meta)
-            elif entity_type == 'koalas':
-                dataframe[column] = dataframe[column].apply(parse_latlong_list)
-
+    # TODO: Can we update Woodwork to return an empty initialized dataframe from a description
+    # instead of using this function?
```
tamargrey, line 64: I feel like the empty dataframe might be a featuretools specific request (I'm still kind of wondering if we can find a way to just use a TableSchema here), but we could consider having the part of this function that's shared with woodwork's `_typing_information_to_woodwork_table` be a separate function that returns a TableSchema that can be used here to init with the empty dataframe. 

Not sure if initializing with a schema might pose problems, though

thehomebrewnerd: It's probably not a big deal either way, but this code is more or less duplicated from Woodwork and thought it might be nice to not have to duplicate it here. Maybe not something we need to worry too much about in the short term though.

featuretools/entityset/deserialize.py
```diff
@@ -90,12 +32,14 @@ def description_to_entityset(description, **kwargs):
...
-        # If path is None, an empty dataframe will be created for entity.
-        description_to_entity(entity, entityset, path=path)
-        if entity['properties']['last_time_index']:
-            last_time_index.append(entity['id'])
+    for df in description['dataframes'].values():
+        if path is not None:
+            data_path = os.path.join(path, 'data', df['name'])
+            dataframe = read_woodwork_table(data_path, validate=False, **kwargs)
+        else:
+            dataframe = empty_dataframe(df)
```
tamargrey, line 40: I don't think this matters because there's no data to actually load here, but I just wanted to point out that the woodwork table description as stored in `data_description.json` that gets used for the empty dataframe case has none of the loading info (location, type, encoding, etc) that is in the `woodwork_typing_info.json` as stored in the `data` directory.

featuretools/entityset/entityset.py
```diff
@@ -159,67 +158,65 @@ def __getitem__(self, dataframe_name):
...
-    #         self._data_description = deserialize.description_to_entityset(description)
+    @property
+    def metadata(self):
+        '''Returns the metadata for this EntitySet. The metadata will be recomputed if it does not exist.'''
+        if self._data_description is None:
+            description = serialize.entityset_to_description(self)
+            self._data_description = deserialize.description_to_entityset(description)
 
-    #     return self._data_description
+        return self._data_description
```
tamargrey, line 168: If metadata an entityset where the DataFrames are just a TableSchema on an empty DataFrame, I feel like it should be possible to create that without having to go through serialization and have this `empty_dataframe` function.

Here's one idea:
```python
    @property
    def metadata(self):
        '''Returns the metadata for this EntitySet. The metadata will be recomputed if it does not exist.'''
        if self._data_description is None:
            metadata_es = EntitySet(self.id)
            for df in self.dataframes:
                empty_df = df.loc[[], :]
                empty_df.ww.init(schema=df.ww.schema, validate=False)
                metadata_es.add_dataframe(empty_df)

            metadata_es.add_relationships(self.relationships)
            self._data_description = metadata_es

        return self._data_description
```

featuretools/tests/entityset_tests/test_serialization.py
```diff
@@ -7,207 +7,167 @@
...
-# Dask does not support to_pickle
+# Dask/Koalas don't support auto setting of interesting values with es.add_interesting_values()
+def test_to_csv_interesting_values(pd_es, tmpdir):
+    pd_es.add_interesting_values()
+    pd_es.to_csv(str(tmpdir))
+    new_es = deserialize.read_entityset(str(tmpdir))
+    assert pd_es.__eq__(new_es, deep=True)
+
+
+# TODO: Verify this works after WW Issue #950 is closed
```
tamargrey, line 105: able to remove this? Or waiting until Featuretools is using Woodwork 0.4.1?

tamargrey (commented):
There's a test in `test_es.py` that's commented out called `test_metadata_without_id` that can be uncommented now.

featuretools/tests/entityset_tests/test_serialization.py
thehomebrewnerd: I did test this works with the WW code on main, so I will go ahead and remove this comment. Will just need to make sure to update WW minimum version to 0.4.1 once it is released. I can make an issue so we don't forget about that.

featuretools/entityset/entityset.py
thehomebrewnerd: While I think this is a decent idea, it has a couple complications that might drive it beyond the scope of this PR:
- The current deserialization to empty dataframes always uses pandas dataframes. This approach would also create Dask and Koalas empty dataframes - based on some limited testing this was causing additional errors that would need to be fixed
- At least a couple tests rely on deserialization into an empty dataframe - so we would either need to keep that capability in deserialize or refactor those tests.

@tamargrey  I think it would be best if we set this aside for now and create an issue to revisit later if you are ok with that?

tamargrey: Ah yeah if it's going to create a bunch of additional things that need to be handled, definitely not worth going into now. Happy to set aside for now

thehomebrewnerd: Oh, one more thing - as I was looking over the `empty_dataframe` function, I did see some code there that I believe was not necessary as it wasn't being used, so I removed that.

As I was doing that I also realized that the metadata entityset won't have the categories stored in the column dtypes like the main entityset does. I'm not sure if this will create problems elsewhere or not?

tamargrey: ah... interesting. I guess I'm not entirely clear on what the metadata gets used for, but if we did need the categories, we could possibly store them elsewhere in the column schema or something.

thehomebrewnerd: Yeah, we have the information available in the json to recreated, but we would need to implement some additional code to to cast the columns properly in `empty_dataframe`. When deserializing with data, the dtypes are set during the read from disk calls, but since we aren't doing that in `empty_dataframe` the dtypes get set during Woodwork initialization.

As long as we don't need to access the category values in the metadata I think we should be fine, but I'm not entirely sure where all it gets used either...

thehomebrewnerd: Created issue #1476 for looking into removing `empty_dataframe`

thehomebrewnerd: After thinking about this a bit more, there was an easy way to recreate the category dtypes including categories for the empty dataframe, so I added this code.

rwedge (approved):

PR #1456, opened 2021-06-01
Store last time index as column on DataFrame

### LINKED ISSUE #1254: Put last_time_index in DataFrame metadata and move Series to DataFrame

tamargrey:
Stores `last_time_index` as a column in the Woodwork dataframe.

Identifying factors of the last time index column:
- Name of the column stored in `df.ww.metadata` under `'last_time_index'`
- The column will be names `last_time` (unless there's overlap between names - tbd what we'll do there)
- The semantic tag `last_time_index` will be added to the column in the woodwork table 

Some of these (the semantic tag) might be overkill, but I figured more things that indicate that the column is additional data the better.

To Do:

- [x] Determine why we started having to add `astype('object')` before converting to datetime (possibly related to the dtype being different somehow when the lti is coming off a woodwork dataframe??)
- [x] Unexplained bugs for `add_last_time_indexes` when called from update_dataframe
  - [x] In `test_update_dataframe_and_lti` find way to test koalas with timedelta and/or mismatched index
  - [x] `test_update_dataframe_last_time_index` koalas and dask failures
- [x] Fix koalas inability to reorder indices (check calculate feature matrix to see if there's something there)
- [x] Additional woodwork testing for last time index in `test_ww_es`
- [x] Update `__sizeof__` to no longer look at metadata for the last time index
- [x] confirm `test_to_parquet_with_lti` and `test_operations_invalidate_metadata` are working correctly (might have to wait till after) #1452 
- [x] confirm bug with not being able to deepcopy koalas dataframes is gone

 
closes #1254

### DIFF
--- a/featuretools/entityset/api.py
+++ b/featuretools/entityset/api.py
@@ -1,6 +1,6 @@
 # flake8: noqa
 from .deserialize import read_entityset
 from .entity import Entity
-from .entityset import EntitySet
+from .entityset import LTI_COLUMN_NAME, EntitySet
 from .relationship import Relationship
 from .timedelta import Timedelta
--- a/featuretools/entityset/api.py
+++ b/featuretools/entityset/api.py
@@ -1,6 +1,6 @@
 # flake8: noqa
 from .deserialize import read_entityset
 from .entity import Entity
-from .entityset import LTI_COLUMN_NAME, EntitySet
+from .entityset import EntitySet
 from .relationship import Relationship
 from .timedelta import Timedelta
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1056,8 +1056,8 @@ class EntitySet(object):
             lti = es_lti_dict[df.ww.name]
             if lti is not None:
                 if self.time_type == 'numeric':
-                    # --> don't totally understand why this is necessary when we can do the conversion normally?
                     if lti.dtype == 'datetime64[ns]':
+                        # Woodwork cannot convert from datetime to numeric
                         lti = lti.apply(lambda x: x.value)
                     lti = ww.init_series(lti, logical_type='Double')
                 else:
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -934,6 +934,7 @@ class EntitySet(object):
         explored = set()
 
         for df in queue:
+            # --> should we also be putting a column in the df yet? probs not
             df.ww.metadata['last_time_index'] = None
 
         # We will explore children of dataframes on the queue,
@@ -961,6 +962,10 @@ class EntitySet(object):
                         # Cannot have a category dtype with nans when calculating last time index
                         lti = lti.astype('object')
                         lti[:] = None
+
+                # --> this isn't the final lti, right? so maybe wait until the end to store on the dataframe?
+                # maybe a bit weird to change whats stored on metadata, but as long as we're ok with temp having series, then we can
+                # probably takes longer to add to the whole df
                 dataframe.ww.metadata['last_time_index'] = lti
 
             if dataframe.ww.name in children:
@@ -1034,14 +1039,20 @@ class EntitySet(object):
                             lti_df['last_time_old'] = lti_df['last_time_old'].astype('datetime64[ns]')
                             lti_df = lti_df.fillna(pd.to_datetime('1800-01-01 00:00')).max(axis=1)
                             lti_df = lti_df.replace(pd.to_datetime('1800-01-01 00:00'), pd.NaT)
-                    # lti_df = lti_df.apply(lambda x: x.dropna().max(), axis=1)
 
                     dataframe.ww.metadata['last_time_index'] = lti_df
                     dataframe.ww.metadata.get('last_time_index').name = 'last_time'
 
             explored.add(dataframe.ww.name)
-        self.reset_data_description()
 
+        # Cannot store last time index in Woodwork metadata - move to dataframe
+        for df in self.dataframes:
+            lti = df.ww.metadata.get('last_time_index')
+            if lti is not None:
+                df.ww['last_time'] = lti
+                df.ww.add_semantic_tags({'last_time': 'last_time_index'})
+                df.ww.metadata['last_time_index'] = 'last_time'
+        self.reset_data_description()
     # ###########################################################################
     # #  Other ###############################################
     # ###########################################################################
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1055,7 +1055,6 @@ class EntitySet(object):
         for df in self.dataframes:
             lti = es_lti_dict[df.ww.name]
             if lti is not None:
-                # --> maybe do this conversaion after creating the df
                 if self.time_type == 'numeric':
                     # --> don't totally understand why this is necessary when we can do the conversion normally?
                     # --> need to make sure this works for dask and koalas
@@ -1398,7 +1397,6 @@ class EntitySet(object):
             if updated_series is not series:
                 df[col_name] = updated_series
 
-        # --> WW bug: if metadata has a series in it, cannot deepcopy
         df.ww.init(schema=self[dataframe_name].ww._schema)
         # Make sure column ordering matches original ordering
         df = df.ww[old_column_names]
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1378,6 +1378,13 @@ class EntitySet(object):
         if not isinstance(df, type(self[dataframe_name])):
             raise TypeError('Incorrect DataFrame type used')
 
+        # If the original DataFrame has a last time index column and the new one doesnt
+        # remove the column and the reference to last time index from that dataframe
+        last_time_index_column = self[dataframe_name].ww.metadata.get('last_time_index')
+        if last_time_index_column is not None and last_time_index_column not in df.columns:
+            self[dataframe_name].ww.pop(last_time_index_column)
+            del self[dataframe_name].ww.metadata['last_time_index']
+
         old_column_names = list(self[dataframe_name].columns)
         if len(df.columns) != len(old_column_names):
             raise ValueError("Updated dataframe contains {} columns, expecting {}".format(len(df.columns),
@@ -1411,7 +1418,7 @@ class EntitySet(object):
 
         df_metadata = self[dataframe_name].ww.metadata
         self.set_secondary_time_index(dataframe_name, df_metadata.get('secondary_time_index'))
-        if recalculate_last_time_indexes and df_metadata.get('last_time_index') is not None:
+        if recalculate_last_time_indexes and last_time_index_column is not None:
             self.add_last_time_indexes(updated_dataframes=[dataframe_name])
         self.reset_data_description()
 
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -935,6 +935,7 @@ class EntitySet(object):
 
         for df in queue:
             # --> should we also be putting a column in the df yet? probs not
+            # --> maybbe need to drop a last_time column here to reset all ltis
             df.ww.metadata['last_time_index'] = None
 
         # We will explore children of dataframes on the queue,
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1,5 +1,6 @@
 import copy
 import logging
+from pdb import set_trace
 import warnings
 from collections import defaultdict
 
@@ -982,6 +983,8 @@ class EntitySet(object):
                     # so we didn't need this logic
                     for df in child_dataframes:
                         if df.ww.name not in explored and df.ww.name not in [q.ww.name for q in queue]:
+                            # must also reset last time index here
+                            es_lti_dict[df.ww.name] = None
                             queue.append(df)
                     queue.append(dataframe)
                     continue
@@ -1026,8 +1029,7 @@ class EntitySet(object):
                                                inplace=True)
 
                         lti_df.set_index(dataframe.ww.index, inplace=True)
-                        # --> shouldnt have to do astype object if we didn't before
-                        lti_df = lti_df.reindex(es_lti_dict[dataframe.ww.name].index).astype('object')
+                        lti_df = lti_df.reindex(es_lti_dict[dataframe.ww.name].index)
                         lti_df['last_time_old'] = es_lti_dict[dataframe.ww.name]
                     if not (lti_is_dask or lti_is_koalas) and lti_df.empty:
                         # Pandas errors out if it tries to do fillna and then max on an empty dataframe
@@ -1362,7 +1364,7 @@ class EntitySet(object):
         df_metadata = self[dataframe_name].ww.metadata
         self.set_secondary_time_index(dataframe_name, df_metadata.get('secondary_time_index'))
         if recalculate_last_time_indexes and df_metadata.get('last_time_index') is not None:
-            self.add_last_time_indexes(updated_dataframes=[self[dataframe_name].ww.name])
+            self.add_last_time_indexes(updated_dataframes=[dataframe_name])
         self.reset_data_description()
 
     def _check_time_indexes(self):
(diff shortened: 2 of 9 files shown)


featuretools/entityset/entityset.py
```diff
@@ -1034,14 +1044,20 @@ def add_last_time_indexes(self, updated_dataframes=None):
...
+                    es_lti_dict[dataframe.ww.name].name = 'last_time'
 
             explored.add(dataframe.ww.name)
-        self.reset_data_description()
 
+        # Store the last time index on the DataFrames
+        for df in self.dataframes:
+            lti = es_lti_dict[df.ww.name]
+            if lti is not None:
+                df.ww['last_time'] = lti
```
tamargrey, line 1057: This introduces nans any time the underlying index of the dataframe doesn't match the underlying index of `lti`. For pandas, this shouldn't be an issue because it handles differently ordered indexes well, and the underlying index will always match the woodwork index.

But because dask and koalas never set the underlying index of their dataframes, this means that unless the indexes happen to match, there will be nans. (this is most easily seen with `es.set_last_time_index(['products'])` since products has a categorical index. 

I tried mapping `lti`'s underlying index to the woodwork index, but that proved to be messy, and I'm not entirely sure possible with dask. I also tried using just the underlying index throughout last time index calculation, but because of the `groupby` on the woodwork index values, we'd still need a way of mapping to the actual underlying index.

tamargrey: Solved by using `merge` to combine the dataframe with the last time index. This creates a new DataFrame object for Dask and Koalas. Pandas still uses the setitem, so the object is the same.

featuretools/entityset/entityset.py
```diff
@@ -975,25 +982,28 @@ def add_last_time_indexes(self, updated_dataframes=None):
                     # so we didn't need this logic
                     for df in child_dataframes:
                         if df.ww.name not in explored and df.ww.name not in [q.ww.name for q in queue]:
+                            # must also reset last time index here
+                            es_lti_dict[df.ww.name] = None
```
tamargrey, line 986: This allows us to reset the children that don't get added to the queue until now (whereas the parents get added at the beginning, and we set the last time indexes to be None there)

featuretools/entityset/entityset.py
```diff
@@ -1034,14 +1044,74 @@ def add_last_time_indexes(self, updated_dataframes=None):
...
+        dfs_to_update = {}
+        for df in self.dataframes:
+            lti = es_lti_dict[df.ww.name]
+            if lti is not None:
+                if self.time_type == 'numeric':
+                    # --> don't totally understand why this is necessary when we can do the conversion normally?
+                    # --> need to make sure this works for dask and koalas
+                    if lti.dtype == 'datetime64[ns]':
+                        lti = lti.apply(lambda x: x.value)
+                    lti = ww.init_series(lti, logical_type='Double')
```
tamargrey, line 1063: I think we've always stored the last time index as Datetimes, but now that we're putting it into the DataFrame, I was thinking it should match the entityset's time type. Happy to remove these lines if it's not necessary, though.

featuretools/entityset/entityset.py
```diff
@@ -1034,14 +1044,74 @@ def add_last_time_indexes(self, updated_dataframes=None):
...
+                        time_index=df.ww.time_index,
+                        logical_types=df.ww.logical_types,
+                        semantic_tags={col_name: tags - {'index', 'time_index'} for col_name, tags in df.ww.semantic_tags.items()},
+                        table_metadata=df.ww.metadata,
+                        column_metadata={col_name: col_schema.metadata for col_name, col_schema in df.ww.columns.items()},
+                        use_standard_tags=df.ww.use_standard_tags
+                    )
+
+                    new_idx = new_df[new_df.ww.index]
+                    new_idx.name = None
```
tamargrey, line 1087: Added these lines to set the underlying index to match the Woodwork index for Dask because the underlying index was being completely changed by the merge operation (I think it has something to do with the partitions), but it seemed better to have the index match the Woodwork index rather than have a non-unique index.

featuretools/entityset/entityset.py
```diff
@@ -1034,14 +1044,74 @@ def add_last_time_indexes(self, updated_dataframes=None):
...
+
+                    new_df.ww.init(
+                        name=df.ww.name,
+                        index=df.ww.index,
+                        time_index=df.ww.time_index,
+                        logical_types=df.ww.logical_types,
+                        semantic_tags={col_name: tags - {'index', 'time_index'} for col_name, tags in df.ww.semantic_tags.items()},
+                        table_metadata=df.ww.metadata,
+                        column_metadata={col_name: col_schema.metadata for col_name, col_schema in df.ww.columns.items()},
+                        use_standard_tags=df.ww.use_standard_tags
```
tamargrey, line 1101: I'm not the biggest fan of this way of initializing Woodwork. It seems very likely to cause bugs/need updating every time Woodwork accessors get updated. 

But we can't initialize with a TableSchema here because the merge operations often change the dtypes, so the schema wouldn't be valid even after we went through the trouble of updating it to contain the last time index column. I played around with the idea of a function to allow initializing Woodwork with a partial schema (or maybe a partial columns dictionary?) while performing type inference on the columns without any schemas. I think this would also be useful for Woodwork's concat.

thehomebrewnerd (commented):
@tamargrey Just a couple quick comments on this and I'll do a deeper review later today.

I think there are two tests that still need to be fixed. The tests are `test_add_last_time_indexes` in `test_dask_es.py` and `test_koalas_es.py`.

Also, I created a test branch and merged in this code with the serialization update code from my PR and confirmed that `test_to_parquet_with_lti` and `test_operations_invalidate_metadata` both pass with the combined code. I went ahead and marked that item complete in the to-do list.

tamargrey:
@thehomebrewnerd thanks for checking the serialization stuff! Updated the koalas and dask tests and also realized that our `int_es` fixture was only testing pandas, so I updated it to also use dask and koalas in conftest. Confirmed that the other tests that use `int_es` and were running before were still running, but if if doesn't make sense to change that fixture because of the other tests that use it, happy to just have them exist for the woodwork tests.

thehomebrewnerd (commented):
Not looking too bad. I still want to think through the logic inside `add_last_time_indexes` a bit more, but in the meantime here is a bit more feedback.

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -624,23 +679,62 @@ def test_update_dataframe_different_dataframe_types():
...
+        new_time_index = ks.from_pandas(original_time_index.to_pandas() + pd.Timedelta(days=1))
+        expected_last_time_index = ks.from_pandas(original_last_time_index.to_pandas() + pd.Timedelta(days=1))
+    else:
+        new_time_index = original_time_index + pd.Timedelta(days=1)
+        expected_last_time_index = original_last_time_index + pd.Timedelta(days=1)
+
+    new_dataframe = es['log'].copy()
+    new_dataframe['datetime'] = new_time_index
+
+    es.update_dataframe('log', new_dataframe, recalculate_last_time_indexes=True)
```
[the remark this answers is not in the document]
thehomebrewnerd: This might also be an issue for the tests that follow this one.

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -286,11 +287,65 @@ def test_update_dataframe():
...
+def test_add_last_time_non_numeric_index(pd_es, ks_es, dask_es):
+    pd_es.add_last_time_indexes(['products'])
+    dask_es.add_last_time_indexes(['products'])
+    ks_es.add_last_time_indexes(['products'])
+
+    assert list(to_pandas(pd_es['products']['last_time']).sort_index()) == list(to_pandas(dask_es['products']['last_time']).sort_index())
+    assert list(to_pandas(pd_es['products']['last_time']).sort_index()) == list(to_pandas(ks_es['products']).sort_values('id')['last_time'])
+
+    assert pd_es['products'].ww.schema == dask_es['products'].ww.schema
+    assert pd_es['products'].ww.schema == ks_es['products'].ww.schema
```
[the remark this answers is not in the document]
tamargrey: Because `products` is the only dataframe in the mock entityset that doesn't have a numeric index, there were some bugs that were only showing up when the koalas and dask underlying indices (which will be numeric and therefore often happened to exactly match the woodwork index column) couldn't possibly match. 

So this test is confirming that the results of add_last_time index actually match in that case between pandas, dask, and koalas. It also confirms that the schemas match between the cases, because we're reinitializing Woodwork on a new DataFrame for the dask and koalas cases, and I wanted to make sure that was maintained as well.

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -286,11 +287,65 @@ def test_update_dataframe():
...
+    col.name = 'last_time'
+
+    es['customers'].ww['last_time'] = col
+
+    assert 'last_time' in es['customers'].columns
+    assert isinstance(es['customers'].ww.logical_types['last_time'], Boolean)
+
+    es.add_last_time_indexes(['customers'])
+
+    # Current behavior will attempt to overwrite this column.
```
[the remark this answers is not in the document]
tamargrey: That's a good point about the uniqueness of the name. My original thought processes around overwriting the column was to make sure that a user could call `add_last_time_indexes` twice without issues, but we probably want to minimize the chance that we overwrite an actual column. 

options to consider:
- `ft_last_time`
- `last_time_in_entityset`

Another option to avoid overwriting a user's column is to confirm that the column contains the `last_time_index` tag or that it's listed in the metadata as the last time index before proceeding. And raising an error telling users to rename otherwise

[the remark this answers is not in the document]
thehomebrewnerd: Yeah, I think it makes sense to allow overwriting a column that featuretools has set. I like the idea of using the semantic tag for this purpose - couple that with a unique column name and I think that should make it highly unlikely that we would overwrite user data.

Could also do something like add an underscore to the column name to make it less likely to be something already present up `_ft_last_time`, for example.

featuretools/tests/entityset_tests/test_ww_es.py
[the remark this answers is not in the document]
thehomebrewnerd: Got it. Thanks. Maybe a short code comment in the test would be helpful to help us remember that detail?

featuretools/tests/entityset_tests/test_ww_es.py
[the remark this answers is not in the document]
tamargrey: Oof, yeah, this definitely is something we need to handle. Good catch--added. 

In the case where we aren't recalculating last time index and the new dataframe has no last time index column, I opted to lose any reference to a last time index on the DataFrame, since the new dataframe won't have a last time index column added.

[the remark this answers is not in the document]
thehomebrewnerd: > In the case where we aren't recalculating last time index and the new dataframe has no last time index column, I opted to lose any reference to a last time index on the DataFrame, since the new dataframe won't have a last time index column added.

Seems reasonable

featuretools/entityset/entityset.py
thehomebrewnerd: > I played around with the idea of a function to allow initializing Woodwork with a partial schema

I like the idea of this type of behavior - use as much as we can and infer the rest. I think this might be useful in other situations beyond concat as well. Maybe we should add a couple issues - one to add this capability to Woodwork and a second to circle back and clean this up later?

tamargrey: Sounds good, I'll create the issues!

featuretools/tests/entityset_tests/test_ww_es.py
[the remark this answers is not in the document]
tamargrey: Added the column name change and semantic tag check

featuretools/tests/entityset_tests/test_ww_es.py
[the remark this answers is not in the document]
tamargrey: added a comment!

featuretools/entityset/entityset.py
```diff
@@ -898,11 +898,13 @@ def add_last_time_indexes(self, updated_dataframes=None):
...
         Calculates the last time index values for each dataframe (the last time
         an instance or children of that instance were observed).  Used when
-        calculating features using training windows
+        calculating features using training windows. Adds the last time index as
+        a series named _ft_last_time on the dataframe.
         Args:
             updated_dataframes (list[str]): List of dataframe names to update last_time_index for
                 (will update all parents of those dataframes as well)
         """
+        LTI_COLUMN_NAME = '_ft_last_time'
```
thehomebrewnerd, line 907: I like this change - easy to update in the future if we need to! What do you think about moving this out of the method and also using this same variable in our test assertions?

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -715,12 +735,32 @@ def test_update_dataframe_dont_recalculate_last_time_index(es):
...
+    new_dataframe = es['log'].copy()
+    new_dataframe['datetime'] = new_time_index
+    new_dataframe.pop('_ft_last_time')
+
+    es.update_dataframe('log', new_dataframe, recalculate_last_time_indexes=True)
+    pd.testing.assert_series_equal(to_pandas(es['log']['datetime']).sort_index(), to_pandas(new_time_index).sort_index(), check_names=False)
 
 
-def test_update_dataframe_recalculate_last_time_index(es):
+def test_update_dataframe_recalcuate_last_time_index_present(es):
```
thehomebrewnerd, line 763: `recalcuate` -> `recalculate`

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -715,12 +735,32 @@ def test_update_dataframe_dont_recalculate_last_time_index(es):
...
+        new_time_index = ks.from_pandas(original_time_index.to_pandas() + pd.Timedelta(days=10))
+    else:
+        new_time_index = original_time_index + pd.Timedelta(days=10)
+
+    new_dataframe = es['log'].copy()
+    new_dataframe['datetime'] = new_time_index
+    new_dataframe.pop('_ft_last_time')
+
+    es.update_dataframe('log', new_dataframe, recalculate_last_time_indexes=True)
+    pd.testing.assert_series_equal(to_pandas(es['log']['datetime']).sort_index(), to_pandas(new_time_index).sort_index(), check_names=False)
```
thehomebrewnerd, line 760: Do we also need to verify that the last time index column is now present after the update?

featuretools/tests/entityset_tests/test_ww_es.py
thehomebrewnerd: Also, if I'm following this test correctly, the original last time index for the log entity is the same as the recalculated last time index. I'm thinking we might want to do something here that causes the last time index after `update_dataframe` to change from what it was before - otherwise I'm not sure we can actually verify that it was recalculated as opposed to just carried over from before.

tamargrey: Yep, good call. Added a line where we confirm that the last time index column also matches the `new_time_index` here and in the test above this (which is the same except for whether or not the lti column is on the new dataframe).

featuretools/tests/entityset_tests/test_ww_es.py
tamargrey: yep--the check for whether the lti column matches what we expect accesses `es['log'][LTI_COLUMN_NAME]` so that should cover it!

featuretools/entityset/api.py
```diff
@@ -1,6 +1,6 @@
 # flake8: noqa
 from .deserialize import read_entityset
 from .entity import Entity
-from .entityset import EntitySet
+from .entityset import LTI_COLUMN_NAME, EntitySet
```
thehomebrewnerd, line 4: Does this need to be here?

thehomebrewnerd (approved):
Looking good. Only one minor comment this time around.

featuretools/entityset/api.py
tamargrey: probably not; I was trying to avoid two imports for `from featuretools.entityset import LTI_COLUMN_NAME, EntitySet` in `test_ww_es` but can just change it to `from featuretools.entityset.entityset import LTI_COLUMN_NAME, EntitySet`

rwedge (approved):

PR #1460, opened 2021-06-02
Update alteryx_open_src_update_checker to 2.0.0

gsheni:


frances-h (approved):

setup.py
```diff
@@ -8,7 +8,7 @@
 
 extras_require = {
     'tsfresh': ['featuretools-tsfresh-primitives >= 0.1.0'],
-    'update_checker': ['featuretools-update-checker >= 1.0.0'],
+    'update_checker': ['featuretools-update-checker >= 2.0.0'],
```
frances-h: should be `'update_checker': ['alteryx-open-src-update-checker >= 2.0.0']`

frances-h (changes requested):

setup.py
gsheni: Fixed

frances-h (dismissed):

featuretools/__init__.py
```diff
@@ -35,11 +35,11 @@
 logger = logging.getLogger('featuretools')
 
 # Call functions registered by other libraries when featuretools is imported
-for entry_point in pkg_resources.iter_entry_points('featuretools_initialize'):
+for entry_point in pkg_resources.iter_entry_points('alteryx_open_src_initialize'):
     try:
         method = entry_point.load()
         if callable(method):
-            method()
+            method('featuretools')
```
rwedge, line 42: won't this break any existing plugins that use the `featuretools_intialize` entry point?

gsheni: - There isn't any existing plugins (that we can search for) besides the update checker that uses that entry point:
  - https://github.com/search?q=featuretools_intialize&type=code

gsheni: I could make this change be in another for loop...

rwedge: add another loop for now and float an issue on whether to remove the old entry point?

gsheni: Sounds good. Fixed the MR and issue created: https://github.com/alteryx/featuretools/issues/1461

rwedge (approved):
Looks good

Issue #1453, opened 2021-05-27
Add function generate list of valid primitives, given an EntitySet


gsheni:
- As a user of Featuretools, I wish there was a utility function in Featuretools, which given an EntitySet, which provide me with a list of valid/applicable primitives. 
- For an EntitySet with a single DataFrame, this should be possible by looking at the typing information

```python
def get_valid_primitives(entityset, target_entity, max_depth=2, selected_primitives=None):
    """
        max_depth (int, optional) : maximum allowed depth of features.
    """

    # if just one entity, set max depth to 1 (transform stacking rule)
    if len(entityset.entity_dict) == 1 and (max_depth is None or max_depth > 1):
        warnings.warn("Only one entity in entityset, changing max_depth to "
                      "1 since deeper features cannot be created")
        max_depth = 1

    if selected_primitives is None:
        # use all primitives
    pass
```

rwedge:
Should this return valid primitives for a specific target entity or after considering every entity as the target entity?  Do we want a tunable max_depth parameter?

gsheni:
@rwedge good point. It should be for a specific target entity. I also added a max_depth parameter (defaults to 2 like DFS, but can be changed by the user).

thehomebrewnerd:
Closed by #1462

PR #1462, opened 2021-06-03
Add get_valid_primitives function

rwedge:
Addresses #1453 

* Still warns user about max_depth on single table
* Adds `get_entityset_type` method, could be made private instead


featuretools/tests/synthesis/test_get_valid_primitives.py
```diff
@@ -0,0 +1,98 @@
...
+    assert AddThree not in trans_prims
+
+    with pytest.raises(ValueError, match="'add_three' is not a recognized primitive name"):
+        agg_prims, trans_prims = get_valid_primitives(pd_es, "log", 2, [ThreeMostCommonCat, "add_three"])
+
+
+def test_get_valid_primitives_all_primitives(es):
+    agg_prims, trans_prims = get_valid_primitives(es, "customers")
+    assert Count in agg_prims
+    assert Hour in trans_prims
```
rwedge, line 89: I could test that all expected primitives are returned, but I'd want to make a toy entityset to handle this case so that the DFS call is not overly long due to the number of primitives being used

featuretools/synthesis/get_valid_primitives.py
```diff
@@ -0,0 +1,77 @@
...
+            elif prim in available_trans:
+                trans_primitives.append(available_trans[prim])
+            else:
+                raise ValueError(f"'{prim}' is not a recognized primitive name")
+    else:
+        entityset_type = get_entityset_type(entityset)
+        agg_primitives = [agg for agg in available_aggs.values()
+                          if entityset_type in agg.compatibility]
+        trans_primitives = [trans for trans in available_trans.values()
+                            if entityset_type in trans.compatibility]
```
tamargrey, line 57: does primitive compatibility also matter for primitives in `selected_primitives` if specified? Or can we assume that if a user is including a primitive in selected primitives, it's compatible?

featuretools/synthesis/get_valid_primitives.py
```diff
@@ -0,0 +1,77 @@
...
+
+    dfs_object = DeepFeatureSynthesis(target_entity, entityset,
+                                      agg_primitives=agg_primitives,
+                                      trans_primitives=trans_primitives,
+                                      max_depth=max_depth)
+
+    features = dfs_object.build_features()
+
+    trans, agg, groupby, where = _categorize_features(features)
```
tamargrey, line 66: groupby and where aren't getting used. Would it make sense to do `trans, agg, _, _ = _categorize_features(features)`?

featuretools/synthesis/get_valid_primitives.py
rwedge: If a user supplies incompatible primitives we could error, warn and ignore, or silently ignore.

Right now it errors but I could see going with one of the other approaches.  Warn seems like a nice mix of letting the user know why some primitives weren't used without stopping the process entirely

thehomebrewnerd (commented):
Generally I think this looks pretty good. I'm wondering how we should handle the situation where a primitive is only used when max_depth > 1 and a primitive gets used because of feature stacking. We could be returning a list of "valid" primitives, but if the user runs DFS with the depth 2 primitive and doesn't specify the corresponding depth 1 primitive, no features would be generated.  

So, how do we (or should we) communicate to the user that certain primitives would only generate features if used with other primitives? Maybe that is something we try to explain in the docstring?

featuretools/tests/synthesis/test_get_valid_primitives.py
```diff
@@ -0,0 +1,96 @@
...
+    msg = ("Selected primitive <class 'featuretools.variable_types.variable.Numeric'> "
+           "is not an AggregationPrimitive, TransformPrimitive, or str")
+    with pytest.raises(ValueError, match=msg):
+        get_valid_primitives(es, target_entity='log', selected_primitives=[Numeric])
+
+
+def test_primitive_compatibility(es):
+    _, trans_prims = get_valid_primitives(es, "customers", selected_primitives=[TimeSincePrevious])
+
+    if not all(isinstance(entity.df, pd.DataFrame) for entity in es.entities):
```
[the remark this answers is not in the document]
rwedge: we do this check in 38 places, sounds like a separate issue

featuretools/synthesis/utils.py
```diff
@@ -0,0 +1,72 @@
...
+
+def get_unused_primitives(specified, used):
+    """Get a list of unused primitives based on a list of specified primitives and a list of output features"""
+    if not specified:
+        return []
+    specified = {primitive if isinstance(primitive, str) else primitive.name for primitive in specified}
+    return sorted(list(specified.difference(used)))
+
+
+def get_entityset_type(entityset):
```
rwedge, line 64: should this be a util function or a propert/method on entityset?

thehomebrewnerd: Entityset property makes sense to me.

rwedge:
> So, how do we (or should we) communicate to the user that certain primitives would only generate features if used with other primitives? Maybe that is something we try to explain in the docstring?

Clarifying in the dosctring sounds good

featuretools/synthesis/utils.py
tamargrey: +1 on making it a property!

rwedge: I also agree with property making sense - how about making it a separate PR that handles the test case issues we pointed out as well as addressing other places we could use the property, like `entity_from_dataframe`

tamargrey (approved):
lgtm!

PR #1465, opened 2021-06-07
Implement deep copy on EntitySet to retain Woodwork typing info

### LINKED ISSUE #1362: EntitySet copy operations do not retain Woodwork typing information
Currently running `copy.deepcopy(entityset)` will cause any typing information contained in the Woodwork dataframes in the entityset to be lost. Add `__copy__` and `__deepcopy__` methods to `EntitySet` to make sure that Woodwork typing information is retained properly after copy operation.

Corresponding tests for these methods should also be added.

tamargrey:
Adds a `__deepcopy__` method to EntitySet that initializes Woodwork on the DataFrames in the `dataframe_dict` with the schemas from the original DataFrame. `copy.deepcopy(dataframe)` works for Dask and Pandas, so currently there is no implementation for Koalas unless or until we implement deepcopy for koalas.

Also discovered that `EntitySet.__eq__` seemed to be missing some attributes of entitysets that meant that equality checks could easily be incorrect. Added fixes as well as tests for them.

Note: This branch is currently branched off of the last time index PR's branch, so ideally, we'll merge that in first.

closes #1362

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -155,6 +155,17 @@ class EntitySet(object):
         name = self.id or "entity set"
         raise KeyError('DataFrame %s does not exist in %s' % (dataframe_name, name))
 
+    def __deepcopy__(self, memo):
+        # --> id, relationships, time type???
+        new_es = EntitySet()
+        for df_name, df in self.dataframe_dict.items():
+            new_df = copy.deepcopy(df)
+            schema = copy.deepcopy(self.dataframe_dict[df_name].ww.schema)
+            new_df.ww.init(schema=schema)
+            new_es.add_dataframe(new_df)
+
+        return new_es
+
     @property
     def dataframes(self):
         return list(self.dataframe_dict.values())
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1056,8 +1056,8 @@ class EntitySet(object):
             lti = es_lti_dict[df.ww.name]
             if lti is not None:
                 if self.time_type == 'numeric':
-                    # --> don't totally understand why this is necessary when we can do the conversion normally?
                     if lti.dtype == 'datetime64[ns]':
+                        # Woodwork cannot convert from datetime to numeric
                         lti = lti.apply(lambda x: x.value)
                     lti = ww.init_series(lti, logical_type='Double')
                 else:
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -934,6 +934,7 @@ class EntitySet(object):
         explored = set()
 
         for df in queue:
+            # --> should we also be putting a column in the df yet? probs not
             df.ww.metadata['last_time_index'] = None
 
         # We will explore children of dataframes on the queue,
@@ -961,6 +962,10 @@ class EntitySet(object):
                         # Cannot have a category dtype with nans when calculating last time index
                         lti = lti.astype('object')
                         lti[:] = None
+
+                # --> this isn't the final lti, right? so maybe wait until the end to store on the dataframe?
+                # maybe a bit weird to change whats stored on metadata, but as long as we're ok with temp having series, then we can
+                # probably takes longer to add to the whole df
                 dataframe.ww.metadata['last_time_index'] = lti
 
             if dataframe.ww.name in children:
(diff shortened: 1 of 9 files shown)


featuretools/entityset/entityset.py
```diff
@@ -124,6 +124,11 @@ def __sizeof__(self):
     #     return (EntitySet, serialize.entityset_to_description(self.metadata))
 
     def __eq__(self, other, deep=False):
+        # --> add data_description to equality check???
```
tamargrey, line 127: We were previously missing attributes of EntitySet that were not `dataframe_dict` or `relationships`, so I figured that `id` and `time_type` made sense to add (though, I do think that if the time type is mismatched, the equality check would fail anyway because the woodwork schemas wouldn't match, but this is a quicker check)

I wasn't sure about whether or not to include `_data_description`, though, because that seems to be something we recalculate at various points, and it mostly a recreation of the EntitySet's information (At least based off of my understanding).

featuretools/entityset/entityset.py
```diff
@@ -155,6 +162,21 @@ def __getitem__(self, dataframe_name):
         name = self.id or "entity set"
         raise KeyError('DataFrame %s does not exist in %s' % (dataframe_name, name))
 
+    def __deepcopy__(self, memo):
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        for k, v in self.__dict__.items():
+            copied_attr = copy.deepcopy(v, memo)
+            # --> figure out how to handle data description....
```
tamargrey, line 171: Similar to the equality check, I wasn't sure if we might want to handle data description in a special way for copying.

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -1,3 +1,4 @@
+
```
thehomebrewnerd: Extra blank line

featuretools/entityset/entityset.py
thehomebrewnerd: I'm thinking we don't need to worry about including `_data_description` (or metadata) in any equality checks. Since the data description is built from the other entityset information as you said, as long as all the other entityset data matches, the data description should as well (as long as it has been generated).

If we were going to check something here, I think it would maybe be best to check the entityset metadata equality as that should regenerate the description if it has been reset and we wouldn't need to worry about handling comparisons with the description possibly in different states between two entitysets. That would require merging in the serialization stuff though.

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -1747,6 +1747,98 @@ def test_entityset_equality(es):
...
+        assert new_df.ww.schema == original_df.ww.schema
+        assert new_df.ww._schema is not original_df.ww._schema
+
+        pd.testing.assert_frame_equal(to_pandas(new_df), to_pandas(original_df))
+        assert new_df is not original_df
+
+
+def test_deepcopy_entityset_woodwork_changes(es):
+    if ks and isinstance(es.dataframes[0], ks.DataFrame):
+        pytest.xfail('Cannot deepcopy Koalas DataFrames')
```
tamargrey, line 1819: Unless I'm misunderstanding something, the problem with being unable to deepcopy Koalas DataFrames comes from Koalas. The tests for `EntitySet.concat` xfail for koalas because `"Koalas deepcopy fails"`, so it seems like it's not the end of the world to wait for a potential deepcopy implementation for Koalas. 

But to build a workaround for that we have a few options, and if the change is simple enough, I'm happy to include it in this PR so that we can use concat for koalas EntitySets:

- bring the DataFrame into memory with `to_pandas` and bring back to koalas, which will create a copy of the data, and note that this can be very computationally intensive
- do a shallow copy of just the dataframe (which does work) for koalas and leave a note that the object is the only shallowly copied, so any changes should be made with that in mind. All the other attributes of the entityset would still be deepcopied

thehomebrewnerd: Since we are not currently supporting concat with Koalas entitysets, so I think it would be best to leave this as a separate issue/PR that we could tackle after the initial integration. Maybe capture your thoughts in a new issue though so we can revisit later?

featuretools/entityset/entityset.py
thehomebrewnerd: I'm thinking we don't need to handle this

thehomebrewnerd (commented):
Looking good to me so far...pending final decision on how to handle the data description.

featuretools/entityset/entityset.py
tamargrey: So we should just copy whatever is in the data description of the original entityset? Or not copy it and have `_data_description` on the new entityset be `None`?

thehomebrewnerd: Ah, good question. I guess I was thinking it didn't need any special treatment, so I guess that would mean whatever is in the original entityset data description would get copied over.

tamargrey: that makes sense to me!

featuretools/tests/entityset_tests/test_es.py
tamargrey: Created an issue: #1466

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -1747,6 +1747,98 @@ def test_entityset_equality(es):
...
     assert first_es != second_es
+    assert second_es != first_es
 
     second_es.add_relationship('customers', 'id', 'sessions', 'customer_id')
     assert first_es == second_es
+
+
+def test_entityset_id_equality(es):
+    first_es = EntitySet(id='first')
+    fisrt_es_copy = EntitySet(id='first')
```
thehomebrewnerd, line 1758: `fisrt` -> `first`

Also update below.

thehomebrewnerd (approved):
One minor spelling issue but otherwise looks good!

PR #1467, opened 2021-06-07
Update query_by_values for Woodwork Integration

### LINKED ISSUE #1420: Update EntitySet.query_by_values to use Woodwork
Allows querying instances that have columns containing certain values, returning a pandas DataFrame (with Woodwork initialized?). Also updates corresponding tests in `test_es.py` that are currently commented out.

thehomebrewnerd:
### Update query_by_values for Woodwork Integration

Closes #1420 

Updates `EntitySet.query_by_values` to work with Woodwork dataframes.

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1197,7 +1197,7 @@ class EntitySet(object):
                     else:
                         warnings.warn(
                             "Using training_window but last_time_index is "
-                            "not set for dataframe %s" % (dt.id)
+                            "not set for dataframe %s" % (dt.ww.name)
                         )
 
                     df = df[mask]
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1170,23 +1170,24 @@ class EntitySet(object):
         If the dataframe does not have a time index, return the original
         dataframe.
         """
-        dt = self[dataframe_name]
+
+        schema = self[dataframe_name].ww.schema
         if is_instance(df, ks, 'DataFrame') and isinstance(time_last, np.datetime64):
             time_last = pd.to_datetime(time_last)
-        if dt.ww.time_index:
+        if schema.time_index:
             df_empty = df.empty if isinstance(df, pd.DataFrame) else False
             if time_last is not None and not df_empty:
                 if include_cutoff_time:
-                    df = df[df[dt.ww.time_index] <= time_last]
+                    df = df[df[schema.time_index] <= time_last]
                 else:
-                    df = df[df[dt.ww.time_index] < time_last]
+                    df = df[df[schema.time_index] < time_last]
                 if training_window is not None:
                     training_window = _check_timedelta(training_window)
                     if include_cutoff_time:
-                        mask = df[dt.ww.time_index] > time_last - training_window
+                        mask = df[schema.time_index] > time_last - training_window
                     else:
-                        mask = df[dt.ww.time_index] >= time_last - training_window
-                    dt_lti = dt.ww.metadata.get('last_time_index')
+                        mask = df[schema.time_index] >= time_last - training_window
+                    dt_lti = schema.metadata.get('last_time_index')
                     if dt_lti is not None:
                         lti_slice = dt_lti.reindex(df.index)
                         if include_cutoff_time:
@@ -1197,12 +1198,12 @@ class EntitySet(object):
                     else:
                         warnings.warn(
                             "Using training_window but last_time_index is "
-                            "not set for dataframe %s" % (dt.ww.name)
+                            "not set for dataframe %s" % (dataframe_name)
                         )
 
                     df = df[mask]
 
-        dt_secondary_time_index = dt.ww.metadata.get('secondary_time_index') or {}
+        dt_secondary_time_index = schema.metadata.get('secondary_time_index') or {}
         for secondary_time_index, columns in dt_secondary_time_index.items():
             # should we use ignore time last here?
             df_empty = df.empty if isinstance(df, pd.DataFrame) else False
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1164,13 +1164,13 @@ class EntitySet(object):
     #         save_graph(graph, to_file, format_)
     #     return graph
 
-    def _handle_time(self, dataframe_id, df, time_last=None, training_window=None, include_cutoff_time=True):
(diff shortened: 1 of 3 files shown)


featuretools/entityset/entityset.py
```diff
@@ -1168,124 +1164,127 @@ def add_interesting_values(self, max_values=5, verbose=False, dataframe_name=Non
...
+                        lti_slice = dt_lti.reindex(df.index)
+                        if include_cutoff_time:
+                            lti_mask = lti_slice > time_last - training_window
+                        else:
+                            lti_mask = lti_slice >= time_last - training_window
+                        mask = mask | lti_mask
+                    else:
+                        warnings.warn(
+                            "Using training_window but last_time_index is "
+                            "not set on entity %s" % (dt.id)
```
tamargrey, line 1200: `entity` usage - maybe something like 

>  Using training_window but there is no last time index set for dataframe %s

featuretools/entityset/entityset.py
```diff
@@ -1168,124 +1164,127 @@ def add_interesting_values(self, max_values=5, verbose=False, dataframe_name=Non
...
-    # def query_by_values(self, entity_id, instance_vals, variable_id=None, columns=None,
-    #                     time_last=None, training_window=None, include_cutoff_time=True):
-    #     """Query instances that have variable with given value
+    def _handle_time(self, dataframe_id, df, time_last=None, training_window=None, include_cutoff_time=True):
+        """
+        Filter a dataframe for all instances before time_last.
+        If the dataframe does not have a time index, return the original
+        dataframe.
+        """
+        dt = self[dataframe_id]
```
tamargrey, line 1173: just wondering - why have we been using `dt` here? It doesn't seem like it's a reference to DataTables or anything, and I think it's getting used interchangeably with `df` here, so I'm not sure I understand its purpose

featuretools/entityset/entityset.py
```diff
@@ -1168,124 +1164,127 @@ def add_interesting_values(self, max_values=5, verbose=False, dataframe_name=Non
...
+    def query_by_values(self, dataframe_id, instance_vals, column_id=None, columns=None,
+                        time_last=None, training_window=None, include_cutoff_time=True):
+        """Query instances that have column with given value
 
-    #     training_window = _check_timedelta(training_window)
+        Args:
+            dataframe_id (str): The id of the dataframe to query
+            instance_vals (pd.Dataframe, pd.Series, list[str] or str) :
+                Instance(s) to match.
+            column_id (str) : Column to query on. If None, query on index.
```
tamargrey, line 1229: might be more consistent with the naming convention we've been going with to have the variables be `column_name` and `dataframe_name`

tamargrey (commented):
looking good so far--just a few comments on variable names and whatnot

featuretools/entityset/entityset.py
thehomebrewnerd: Yeah, that is unrelated to the Woodwork integration and that variable name is carried over from the code in `main`. The only reason that is used as far as I can tell is to access the various "entity" attributes like `index` and `time_index` which are needed for the filtering operations since those values aren't necessarily present on the dataframe that gets passed into the method.

featuretools/entityset/entityset.py
thehomebrewnerd: Good point - updated.

featuretools/entityset/entityset.py
thehomebrewnerd: Good catch - updated.

featuretools/entityset/entityset.py
tamargrey: Okay, I see the purpose; it's kind of like the way that we use typing information in some of the Woodwork statistics utils.

If we're only using it to access woodwork typing information, I think my suggestion would be to just pass the schema into the method. We don't use the `dataframe_name` for anything other than getting the `dt` variable, and passing around the schema is what we do in `_make_categorical_for_mutual_info` and `_replace_nans_for_mutual_info`, which are decent analogs for `_handle_time`.

thehomebrewnerd: Yep, we could do that. My only hesitation is related to whether that change obscures a little by what the function call is doing. Now by passing the entity name it is clear to me that we are handling time for the specified entity/dataframe, and passing a schema instead seems to make that a little less obvious to me, although I don't think that's a huge problem since this is a private method anyway.

One other option might be to continue to pass in the dataframe name but grab the schema and use that in place of `dt`. So instead of 
```
dt = self[dataframe_name]
```

we could use
```
schema = self[dataframe_name].ww.schema
```
and then update all the following references to `dt`, which I think would make it more clear what is happening inside this function.

I don't have a strong opinion here, but I think we can make some changes to clean this up a bit.

tamargrey: Using the dataframe name to get the schema seems like a good combination!

tamargrey (approved):
lgtm!

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -277,101 +277,115 @@ def test_add_relationship_error(es):
...
+        dataframe_name='log',
+        instance_vals=[0, 1, 2], column_name='session_id',
+        time_last=datetime(2011, 4, 9, 10, 50, 0),
+        training_window='15m')
+    df = to_pandas(df)
 
-#     assert list(df['id'].values) == [9]
-#     assert list(df['value'].values) == [0]
+    assert list(df['id'].values) == [9]
+    assert list(df['value'].values) == [0]
```
rwedge, line 378: for this test, should we pick an example where the last_time_index impacts the returned values? it would have to be a different dataframe

thehomebrewnerd: Probably a good idea. Do you think both of the training window tests should use the same dataframe so the difference between having lti set and not is clear?

rwedge: sure 👍

thehomebrewnerd: @rwedge The two tests that use training window have been updated to produce different results with a last time index set.

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -5,6 +5,7 @@
 
 import dask.dataframe as dd
 import numpy as np
+from numpy.core.fromnumeric import sort
```
gsheni, line 8: Is this actually used anywhere?

rwedge (approved):
LGTM

PR #1468, opened 2021-06-07
Update EntitySet.plot to use Woodwork

### LINKED ISSUE #1418: Update EntitySet.plot to use Woodwork
Converts the commented out code in EntitySet.plot and its corresponding tests to use Woodwork DataFrames.

Confirm that the plot matches the plot that shows up in main for `make_ecommerce_entityset` (last checked, the plot on the `woodwork-integration` branch had a bug where `_parent_column_id` was being used in place of `_parent_dataframe_id` at some point, and it had changed graph)

tamargrey:
Updates `EntitySet.plot` to use Woodwork logical types and semantic tags.

Here's a reminder of what the current plot looks like on Main:
image

Here's a peek at the updated plot (open to other suggestions for how to handle the combination of logical types and semantic tags without being too busy):
image

closes #1418

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1145,7 +1145,7 @@ class EntitySet(object):
         # Draw entities
         for df in self.dataframes:
             # --> need to finda nice way to show this
-            columns_string = '\l'.join([col_name + ': ' + str(col_schema)[14:-1]  # noqa: W605
+            columns_string = '\l'.join([col_name + ' : ' + str(col_schema.logical_type) + ', ' + str(list(col_schema.semantic_tags))  # noqa: W605
                                           for col_name, col_schema in df.ww.columns.items()])
             if isinstance(df, dd.DataFrame):  # entity is a dask entity
                 label = '{%s |%s\l}' % (df.ww.name, columns_string)  # noqa: W605
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1145,8 +1145,8 @@ class EntitySet(object):
         # Draw entities
         for df in self.dataframes:
             # --> need to finda nice way to show this
-            columns_string = '\l'.join([str(col_schema)  # noqa: W605
-                                          for col_schema in df.ww.columns])
+            columns_string = '\l'.join([col_name + ': ' + str(col_schema)[14:-1]  # noqa: W605
+                                          for col_name, col_schema in df.ww.columns.items()])
             if isinstance(df, dd.DataFrame):  # entity is a dask entity
                 label = '{%s |%s\l}' % (df.ww.name, columns_string)  # noqa: W605
             else:
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1131,8 +1131,13 @@ class EntitySet(object):
 
         Returns:
             graphviz.Digraph : Graph object that can directly be displayed in
-                Jupyter notebooks.
+                Jupyter notebooks. Nodes of the graph correspond to the DataFrames
+                in the EntitySet, showing the typing information for each column.
 
+        Note:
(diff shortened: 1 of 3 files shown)


featuretools/entityset/entityset.py
```diff
@@ -1121,52 +1121,61 @@ def add_interesting_values(self, max_values=5, verbose=False, dataframe_name=Non
...
+            for col_name, col_schema in df.ww.columns.items():
+                col_string = col_name + ' : ' + str(col_schema.logical_type)
+
+                tags = col_schema.semantic_tags - col_schema.logical_type.standard_tags
+                if tags:
+                    col_string += '; '
+                    col_string += ', '.join(tags)
+                column_typing_info.append(col_string)
+
+            columns_string = '\l'.join(column_typing_info)  # noqa: W605
```
tamargrey, line 1157: Wanted to provide another, more verbose option for what the plot could look like if we want to make sure we're clear about logical types vs semantic tags. This just makes use of the `ColumnSchema` repr

image

featuretools/entityset/entityset.py
```diff
@@ -1121,52 +1121,61 @@ def add_interesting_values(self, max_values=5, verbose=False, dataframe_name=Non
...
-    #         else:
-    #             label = '%s -> %s' % (rel._parent_column_name,
-    #                                   rel._child_column_name)
+        # Draw entities
+        for df in self.dataframes:
+            column_typing_info = []
+            for col_name, col_schema in df.ww.columns.items():
+                col_string = col_name + ' : ' + str(col_schema.logical_type)
+
+                tags = col_schema.semantic_tags - col_schema.logical_type.standard_tags
```
tamargrey, line 1151: In the context of a plot that will be easier to read if there's less to pick through for each column, the standard tags aren't usually adding much that the Logical Type doesn't already indicate (plus Variable types had no way to show this kind of hierarchy), so I'm thinking it makes sense to not include them.

featuretools/entityset/entityset.py
thehomebrewnerd: Thanks for providing this for comparison. I prefer the more simple view - I find it much easier to find the information when there is less to read through.

featuretools/entityset/entityset.py
thehomebrewnerd: I am a bit on the fence here since this means the information users see in other places (repr for example) will be inconsistent with the information they see here. I do agree that the standard tags don't provide much additional value, however.

If we decide to remove them from the plot, I think we should document this behavior in the docstring so that users can understand that the lack of standard tags in the plot is by design and not an error.

thehomebrewnerd (approved):
This looks good to me. I think we just need to decide how and what information we want to display in the plots, but otherwise I don't see anything that needs to be fixed or changed.

featuretools/entityset/entityset.py
tamargrey: Yes, and in addition to inconsistencies between here and other representations, I think another worry with losing the standard tags is that they generally are how we present input and return types in primitives.

If a user is trying to look through an entityset at this high of a level and figure out what primitives they might want to use, it's on them to understand that the `'numeric'` semantic tag will apply to `Integer` and `Double`. 

But I also think that the moment we remove the logical type and semantic tag labels from this information, we're leaving the meaning up to a user's interpretation, so I'm not sure the comparison between primitive input/return types and these lines in the plot will be ideal anyway.

thehomebrewnerd: My inclination is to leave off the standard tags as you have done, but include the tags that indicate index, time index and foreign key columns as I think that is important in the context of the graph. If we find that the absence of standard tags creates issues we can easily add them to the plots in a future PR.

Maybe just a quick mention in the docstring that standard tags won't be displayed in the plot?

rwedge (approved):

PR #1473, opened 2021-06-09
Add dataframe_type property to EntitySet

### LINKED ISSUE #1472: Add dataframe_type property to EntitySet
This property will return which dataframe library is being used by entities in the entityset.

rwedge:
Closes  #1472 

Adds dataframe_type property to entityset to make it easier to check which dataframe library is being used for that entityset

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,6 +7,7 @@ Future Release
 ==============
     * Enhancements
        * Add ``get_valid_primitives`` function (:pr:`1462`)
+       * Add ``EntitySet.dataframe_type`` attribute (:pr:`1473`)
     * Fixes
     * Changes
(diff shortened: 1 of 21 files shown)


thehomebrewnerd (commented):
This looks good - I think there might be a lot of other locations where we can use this property as well, acknowledging that this is to create a lot of merge conflicts with our woodwork-integration branch.

I just searched for `.entities` and looked through the results and found several statements like this (not a complete list, just some examples):
- `if ks and any(isinstance(e.df, ks.DataFrame) for e in es.entities):`
- `if any(isinstance(es.df, dd.DataFrame) for es in entityset.entities):`
- `if is_instance(self.entities[0].df, ks, 'DataFrame'):`
- `if isinstance(es.entities[0].df, dd.DataFrame):`

Not sure if we want to try to update some of these other calls now as well?

thehomebrewnerd (approved):
Looks good!

PR #1477, opened 2021-06-09
Replace list_variable_types with list_logical_types

### LINKED ISSUE #1258: Replace list_variable_types with list_logical_types
- To keep breaking changes to a minimum, the `list_variable_types` should return the DataFrame from `list_logical_types` in Woodwork.
- We should also add a `list_logical_types` (which should just return the same thing as Woodwork's `list_logical_types`
- Let's make sure to note this in the Breaking Changes.

jeff-hernandez:
Closes #1258 by replacing `list_variable_types` with `list_logical_types`

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -17,6 +17,7 @@ Future Release
         * Replace ``Entity`` objects in ``EntitySet`` with Woodwork dataframes (:pr:`1405`)
         * Refactor ``EntitySet.plot`` to work with Woodwork dataframes (:pr:`1468`)
         * Move ``last_time_index`` to be a column on the DataFrame (:pr:`1456`)
+        * Replace ``list_variable_types`` with ``list_logical_types`` (:pr:`1477`)
     * Documentation Changes
         * Improve formatting of release notes (:pr:`1396`)
     * Testing Changes
@@ -69,6 +70,8 @@ Breaking Changes
 * ``last_time_index``, ``secondary_time_index``, and ``interesting_values`` are no longer attributes
   of an entityset’s tables that can be accessed directly. Now they must be accessed through the metadata
   of the Woodwork DataFrame, which is a dictionary.
+* The helper function ``list_variable_types`` has been replaced with ``list_logical_types``.
+  To help minimize breaking changes, ``list_variable_types`` will return the same output as ``list_logical_types``.
 
 What's New in this Release
 ++++++++++++++++++++++++++
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -70,8 +70,8 @@ Breaking Changes
 * ``last_time_index``, ``secondary_time_index``, and ``interesting_values`` are no longer attributes
   of an entityset’s tables that can be accessed directly. Now they must be accessed through the metadata
   of the Woodwork DataFrame, which is a dictionary.
-* The helper function ``list_variable_types`` has been replaced with ``list_logical_types``.
-  To help minimize breaking changes, ``list_variable_types`` will return the same output as ``list_logical_types``.
+* The helper function ``list_variable_types`` will be removed in a future release and replaced by ``list_logical_types``.
+  In the meantime, ``list_variable_types`` will return the same output as ``list_logical_types``.
 
 What's New in this Release
 ++++++++++++++++++++++++++
--- a/featuretools/variable_types/api.py
+++ b/featuretools/variable_types/api.py
@@ -2,6 +2,7 @@
 from .utils import (
     find_variable_types,
     graph_variable_types,
+    list_logical_types,
     list_variable_types
 )
 from .variable import *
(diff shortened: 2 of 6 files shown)


featuretools/variable_types/api.py
```diff
@@ -2,6 +2,7 @@
 from .utils import (
     find_variable_types,
     graph_variable_types,
+    list_logical_types,
```
tamargrey, line 5: Are we eventually going to get rid of the `variable_types` directory? If so, we probably need somewhere else to put these. 

This also makes me wonder if we're going to need equivalents of `graph_variable_types` and `find_variable_types` (@gsheni )? We might be able to make use of the Woodwork type system's parent-child relationships, but idk

gsheni: We can completely remove `find_variable_types` and `graph_variable_types`. With WW we want to move away from the parent-child relationship. 

Yes, we should get rid of the variable types directory.

jeff-hernandez: Do we want users to import `from featuretools.logical_types import list_logical_types`?

gsheni: If we don't need to have a logical_types folder, let's not have it.`from featuretools import list_logical_types` seems good to me.

jeff-hernandez: Do we want to keep the reference in `featuretools.variable_types.utils` for now and make it possible to import from the top-level `from featuretools import list_logical_types`?

gsheni: My vote would be no. There is already going to be lots of breaking changes, and we can just be sure to note them in the release notes, and add it to the documentation page for V1 upgrade.

rwedge:
I think we should throw a warning if they use `list_variable_types` instead of `list_logical_types` letting them know they should use the new function

featuretools/variable_types/api.py
jeff-hernandez: Okay, so we want to import `from featuretools import list_logical_types` and wait to see where to move this function when we delete variable types in a future PR right?

jeff-hernandez: I can also just add another import statement to the reference in `featuretools.__init__`

gsheni: @jeff-hernandez its your call. If you want to do it in a followup MR, just be sure to create an issue. If you want to do it in this MR, that is also fine.

jeff-hernandez: Okay, i'll add another import statement at the top-level

tamargrey:
Can we add a test that checks the contents of `list_logical_types`? Maybe just confirming that it matches the dataframe produced by the direct woodwork call?

jeff-hernandez:
@tamargrey sure, where do we want to place this test?

tamargrey:
> @tamargrey sure, where do we want to place this test?

Hmmm I'd say wherever we're planning on putting `list_variable_types` once we remove the variables directory. Might be best to move it there now, and move its test to the corresponding testing directory so we can stick this new test there as well. 

Maybe `gen_utils.py` and `test_gen_utils.py`? If we end up having a series of woodwork-related utils we might end up creating a woodwork-utils file, but for now, gen utils seems the best fit to me.

jeff-hernandez:
@tamargrey I'll put the test in `test_gen_utils.py` for now and we can come back to `list_variable_types` whenever we remove that directory in a separate PR.

docs/source/release_notes.rst
```diff
@@ -69,6 +70,8 @@ Breaking Changes
 * ``last_time_index``, ``secondary_time_index``, and ``interesting_values`` are no longer attributes
   of an entityset’s tables that can be accessed directly. Now they must be accessed through the metadata
   of the Woodwork DataFrame, which is a dictionary.
+* The helper function ``list_variable_types`` has been replaced with ``list_logical_types``.
+  To help minimize breaking changes, ``list_variable_types`` will return the same output as ``list_logical_types``.
```
thehomebrewnerd, line 74: I think I'd tweak this wording a little bit. The first sentence says we replaced a function, but the second indicates the function still exists. Maybe we could say something like this:

The helper function `list_variable_types` will be removed in a future release and replaced by `list_logical_types`. In the meantime, `list_variable_types` will return the same output as `list_logical_types`.

thehomebrewnerd (approved):
Just one suggestion for the wording of the breaking change, but otherwise looks good to me.

docs/source/release_notes.rst
jeff-hernandez: fixed

docs/source/release_notes.rst
```diff
@@ -69,6 +70,8 @@ Breaking Changes
 * ``last_time_index``, ``secondary_time_index``, and ``interesting_values`` are no longer attributes
   of an entityset’s tables that can be accessed directly. Now they must be accessed through the metadata
   of the Woodwork DataFrame, which is a dictionary.
+* The helper function ``list_variable_types`` will be removed in a future release and replaced by ``list_logical_types``.
```
gsheni, line 73: We should add a section to `What's New in this Release` that describes the new list logical types and list semantic tags

Issue #1424, opened 2021-05-05
Include deep check of Woodwork dataframes in EntitySet equality check


tamargrey:
When `EntitySet.__eq__` was updated to use Woodwork DataFrames, there was no ability to pass a `deep` parameter to Woodwork's equality checks, so it was not included in the EntitySet equality check. With the current implementation, entityset equality checks will always do a deep check, looking at data equality (for Pandas dataframes only). We should allow a shallow check that does not check for data equality.

Now that Woodwork allows the `deep` parameter, we should add it in the equality check as `if not df.ww.__eq__(other[df_id].ww, deep=deep)`.

We should also update the corresponding entityset equality test to include some deep and shallow checks.

thehomebrewnerd:
The current `deep` equality check in Woodwork will not check that the underlying data is equal for Dask/Koalas dataframes. In the old Featuretools entity equality check a custom `_dataframes_equal` function was implemented for checking the equality of dataframes.

As part of this issue, we should also investigate whether we should also include use of the custom `_dataframes_equal` function for Dask/Koalas dataframes as well as investigate whether the current deep equality check implemented in Woodwork for pandas is significantly slower than the `_dataframes_equal` function to understand if the Woodwork implementation has introduced a significant performance degradation for Featuretools.

PR #1480, opened 2021-06-10
Allow deep equality check on EntitySet

tamargrey:
Passes the `deep` parameter for `EntitySet.__eq__` down to the Woodwork dataframe. 

#1424

### DIFF
--- a/featuretools/demo/flight.py
+++ b/featuretools/demo/flight.py
@@ -95,7 +95,7 @@ def make_es(data):
                         'taxi_in', 'taxi_out', 'air_time', 'dep_time']
 
     logical_types = {'flight_num': ww.logical_types.Categorical,
-                     'distance_group': ww.logical_types.Ordinal(order=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]),
+                     'distance_group': ww.logical_types.Ordinal(order=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]),
                      'canceled': ww.logical_types.Boolean,
                      'diverted': ww.logical_types.Boolean}
 
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -20,6 +20,8 @@ from featuretools.utils.plot_utils import (
 # import pandas.api.types as pdtypes
 
 
+# from featuretools.utils.wrangle import _check_timedelta
+
 ks = import_or_none('databricks.koalas')
 
 pd.options.mode.chained_assignment = None  # default='warn'
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -142,31 +142,6 @@ class EntitySet(object):
                 return False
         return True
 
-# --> remove maybe!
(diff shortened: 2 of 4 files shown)


featuretools/entityset/entityset.py
```diff
@@ -20,7 +20,7 @@
 # import pandas.api.types as pdtypes
 
 
-# from featuretools.utils.wrangle import _check_timedelta
+from featuretools.utils.wrangle import _dataframes_equal
```
thehomebrewnerd, line 23: Do we need this import now?

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -1777,6 +1777,60 @@ def test_entityset_time_type_equality(es):
...
+                           dataframe=es['sessions'].copy(),
+                           index='id',
+                           logical_types=es['sessions'].ww.logical_types,
+                           semantic_tags=get_df_tags(es['sessions']))
+    second_es.add_dataframe(dataframe_name='customers',
+                            dataframe=es['customers'].copy(),
+                            index='id',
+                            time_index='signup_date',
+                            logical_types=es['customers'].ww.logical_types,
+                            semantic_tags=get_df_tags(es['customers']))
```
thehomebrewnerd, line 1807: Super minor nitpick, but could we reorder these so we either add the dataframes to the entitysets in the same order or build the first entityset and then the second. I spent some time trying to figure out if there is a reason that we add customers to the first es and then sessions to the second, and then realized it doesn't matter. I found it a little harder to initially follow that we are building equal entitysets with this order.

featuretools/demo/flight.py
```diff
@@ -95,7 +95,7 @@ def make_es(data):
                         'taxi_in', 'taxi_out', 'air_time', 'dep_time']
 
     logical_types = {'flight_num': ww.logical_types.Categorical,
-                     'distance_group': ww.logical_types.Ordinal(order=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]),
+                     'distance_group': ww.logical_types.Ordinal(order=[1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11]),
```
thehomebrewnerd, line 98: Is this change needed for something in this PR?

thehomebrewnerd (commented):
Looks good - just a couple minor comments.

featuretools/demo/flight.py
tamargrey: nope, just was trying to look at this demo entityset and realized it was missing an element from the order

featuretools/entityset/entityset.py
tamargrey: nope, good call. removing

featuretools/tests/entityset_tests/test_es.py
tamargrey: yep, changing to building each entityset at a time.

tamargrey: oops, just pushed

thehomebrewnerd (approved):
Looks good!

PR #1486, opened 2021-06-11
Add function to list semantic tags

### LINKED ISSUE #1471: Add list_semantic_tags function to get all semantic tags
- Users should be able to import a `list_semantic_tags` from featuretools. 
- This will allow users to see all the standard/supported/suggested semantic tags in Featuretools
- This function will just return the Woodwork function:
  - https://woodwork.alteryx.com/en/latest/generated/woodwork.type_sys.utils.list_semantic_tags.html?highlight=list_semantic_tags

jeff-hernandez:
Closes #1471 by adding the function to list semantic tags


tamargrey:
Do this and `list_logical_types` need to go in the api reference? I'm not exactly sure where they would go (can you even put something in the api reference if it's just being pulled in from another package?) but I feel like we probably want them to be visible there.

jeff-hernandez:
@tamargrey yeah makes sense. I tried it out locally and it should render in the docs under the Featuretools namespace. Not exactly sure under what heading it should go under though.

gsheni:
@jeff-hernandez You can create a new header at the same level as the current Variable Utils Methods, called **Data Type Util Methods**

docs/source/api_reference.rst
```diff
@@ -376,6 +376,15 @@ Variable Utils Methods
     list_variable_types
```
gsheni, line 376: This line should be removed

jeff-hernandez: fixed

docs/source/release_notes.rst
```diff
@@ -22,6 +22,7 @@ Future Release
         * Refactor ``EntitySet.query_by_values`` to work with Woodwork dataframes (:pr:`1467`)
         * Replace ``list_variable_types`` with ``list_logical_types`` (:pr:`1477`)
         * Allow deep EntitySet equality check (:pr:`1480`)
+        * Add function to list semantic tags (:pr:`1486`)
```
gsheni, line 25: We should add a section to `What's New in this Release` that describes the new list logical types and list semantic tags

jeff-hernandez: Added section

gsheni (approved):

Issue #1096, opened 2020-07-31
Featuretools support for pandas 2.0.0


thehomebrewnerd:
Pandas has a running list of deprecations in versions `1.x` that will be removed in `2.0`. This list can be found [here](https://github.com/pandas-dev/pandas/issues/30228).

This issue will serve as a list of items that need to be updated in Featuretools to ensure compatibility with pandas 2.0.

- Update `Week` transform primitive to use new `Series.dt.isocalendar().week` approach and remove code that suppresses the `FutureWarning` related to this deprecation. This new approach is only available in pandas 1.1.0 and newer.
- Update pandas minimum version requirement in `requirements.txt` to the lowest possible version that works with all the other changes in this list

PR #1492, opened 2021-06-15
Test compatibility with upcoming pandas release 1.3.0

### LINKED ISSUE #1489: Test featuretools compatibility with Pandas 1.3 release candidate
We should make a branch that uses the release candidate and test if it's compatible with Featuretools currently

tamargrey:
FutureWarnings that show up:
- `test_transform_features.py::test_init_and_name[dask_es]` - FutureWarning: Series.dt.weekofyear and Series.dt.week have been deprecated.  Please use Series.dt.isocalendar().week instead
- `test_es.py::test_datetime64_conversion[dd_datetime3]` - FutureWarning: Using .astype to convert from timezone-naive dtype to timezone-aware dtype is deprecated and will raise in a future version.  Use ser.dt.tz_localize instead.
- `test_serialization.py::test_to_csv[dask_es]` - FutureWarning: The `inplace` parameter in pandas.Categorical.set_categories is deprecated and will be removed in a future version. Removing unused categories will always return a new Categorical object.
- `test_agg_feats.py::test_init_and_name[pd_es]` - FutureWarning: casting datetime64[ns] values to int64 with .astype(...) is deprecated and will raise in a future version. Use .view(...) instead.

locally, two tests failed:

- `test_some_instances_not_in_data` in `test_calculate_feature_matrix`

```python
        for x, y in zip(fm.columns, [ifeat_answer, prop_answer, dfeat_answer]):
>           np.testing.assert_array_equal(fm[x], y)
E           AssertionError: 
E           Arrays are not equal
E           
E           Mismatched elements: 5 / 10 (50%)
E           Max absolute difference: nan
E           Max relative difference: nan
E            x: array([False, 0.0, True, False, 0.0, nan, nan, nan, nan, nan],
E                 dtype=object)
E            y: array([ 0.,  0.,  1.,  0.,  0., nan, nan, nan, nan, nan])
```

- `test_transform` in `test_dask_primitives`
```python
>           dask_computed_fm = dask_fm.compute().set_index(entity.index).loc[fm.index][fm.columns]
...
>       naive = default.replace(**repl)
E       OverflowError: signed integer is greater than maximum
```
closes #1489

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -11,9 +11,10 @@ Future Release
     * Documentation Changes
         * Prevent logging on build (:pr:`1498`)
     * Testing Changes
+        * Test featuretools on pandas 1.3.0 release candidate and make fixes (:pr:`1492`)
 
     Thanks to the following people for contributing to this release:
-    :user:`frances-h`
+    :user:`frances-h`, :user:`tamargrey`
 
 v0.25.0 Jun 11, 2021
 ====================
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,6 +6,7 @@ Release Notes
 Future Release
 ==============
     * Enhancements
+        * Add ``replace_inf_values`` utility function for replacing ``inf`` values in a feature matrix (:pr:`1505`)
     * Fixes
     * Changes
     * Documentation Changes
@@ -14,7 +15,7 @@ Future Release
         * Test featuretools on pandas 1.3.0 release candidate and make fixes (:pr:`1492`)
 
     Thanks to the following people for contributing to this release:
-    :user:`frances-h`, :user:`tamargrey`
+    :user:`frances-h`, :user:`tamargrey`, :user:`thehomebrewnerd`
 
 v0.25.0 Jun 11, 2021
 ====================
--- a/featuretools/tests/computational_backend/test_calculate_feature_matrix.py
+++ b/featuretools/tests/computational_backend/test_calculate_feature_matrix.py
@@ -1611,7 +1611,7 @@ def test_some_instances_not_in_data(pd_es):
 
     assert all(fm.index.values == cutoff_time["instance_id"].values)
     for x, y in zip(fm.columns, [ifeat_answer, prop_answer, dfeat_answer]):
-        np.testing.assert_array_equal(fm[x], y)
+        np.testing.assert_array_equal(fm[x].astype('float64'), y)
 
 
 def test_missing_instances_with_categorical_index(pd_es):
--- a/featuretools/tests/testing_utils/mock_ds.py
+++ b/featuretools/tests/testing_utils/mock_ds.py
@@ -115,7 +115,7 @@ def make_ecommerce_dataframes(with_integer_time_index=False):
         'engagement_level': [1, 3, 2],
         'full_name': ['Mr. John Doe', 'Doe, Mrs. Jane', 'James Brown'],
         'email': ['john.smith@example.com', np.nan, 'team@featuretools.com'],
-        'phone_number': ['5555555555', '555-555-5555', '1-(555)-555-5555'],
+        'phone_number': ['555-555-5555', '555-555-5555', '1-(555)-555-5555'],
         'date_of_birth': customer_times['date_of_birth'],
     })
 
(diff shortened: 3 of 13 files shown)


tamargrey:
The behavior we're seeing in `test_some_instances_not_in_data` is because of a change in behavior in `pd.concat`. The following produces different results in the two different versions:
```python
import pandas as pd

df1 = pd.DataFrame(pd.Series([True, False, True, True], dtype='bool'))
df2 = pd.DataFrame(pd.Series([1,0,1], dtype='int64'))

pd.concat([df1,df2])
```

In pandas 1.2.4, concatenating a boolean column with a numeric column produced a numeric column with all the boolean values converted to `1`s or `0`s.
image


In the 1.3.0rc, the equivalent concat call produces an `object` dtype column with all the original values maintained.

image

tamargrey:
Opened [an issue](https://github.com/pandas-dev/pandas/issues/42092) in Pandas for the concat issue

tamargrey:
> Opened [an issue](https://github.com/pandas-dev/pandas/issues/42092) in Pandas for the concat issue

well it turns out this is the desired result!

rwedge:
We've been tracking deprecated Pandas functionality we still support in #1096, is this a new place we need to suppress that isocalendar warning?

featuretools/tests/testing_utils/mock_ds.py
```diff
@@ -115,7 +115,7 @@ def make_ecommerce_dataframes(with_integer_time_index=False):
         'engagement_level': [1, 3, 2],
         'full_name': ['Mr. John Doe', 'Doe, Mrs. Jane', 'James Brown'],
         'email': ['john.smith@example.com', np.nan, 'team@featuretools.com'],
-        'phone_number': ['5555555555', '555-555-5555', '1-(555)-555-5555'],
+        'phone_number': ['555-555-5555', '555-555-5555', '1-(555)-555-5555'],
```
gsheni, line 118: Why does this need to be changed?

tamargrey: This is because of the `OverflowError`. The string `'5555555555'` is now getting interpreted as numeric when it gets compared to datetimes with something like `df['phone_numbers'] == df['dates']` in Dask

tamargrey:
> We've been tracking deprecated Pandas functionality we still support in #1096, is this a new place we need to suppress that isocalendar warning?

@rwedge I think that the two warnings that have gotten through are now being raised inside of Dask. We can suppress the warnings in the test function, but the `filterwarnings` call in the Week primitive is doing its job still, I think

gsheni (dismissed):

docs/source/release_notes.rst
```diff
@@ -12,9 +12,10 @@ Future Release
     * Documentation Changes
         * Prevent logging on build (:pr:`1498`)
     * Testing Changes
+        * Test featuretools on pandas 1.3.0 release candidate and make fixes (:pr:`1492`)
```
thehomebrewnerd, line 15: Probably should update this to indicate we are now at the point of implementing changes for pandas 1.3.0 compatibility instead of testing changes on the release candidate.

gsheni (approved):

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -1611,7 +1611,7 @@ def test_some_instances_not_in_data(pd_es):
 
     assert all(fm.index.values == cutoff_time["instance_id"].values)
     for x, y in zip(fm.columns, [ifeat_answer, prop_answer, dfeat_answer]):
-        np.testing.assert_array_equal(fm[x], y)
+        np.testing.assert_array_equal(fm[x].astype('float64'), y)
```
thehomebrewnerd, line 1614: I'm on the fence about this change. We now get different results in our feature matrix depending on what version of pandas we use - this seems like a possible problem from a user perspective, no?

Here we are making the test pass, by type casting the result, but I think if we expect our feature matrix result to be different, we should change our expected answer `y`, not type cast the feature matrix.

Based on the notes in the related pandas issue, I'm not sure if this will be a change that will remain long term, and this is something we might want to revert in the future if new versions of pandas reimplement the old concat behavior.

tamargrey: @thehomebrewnerd The issue that caused this change will be fixed in pandas 1.3.1, I believe. https://github.com/pandas-dev/pandas/issues/42092

If we want to raise our min pandas version again, we can remove this dtype change

PR #1498, opened 2021-06-18
Fix docs to avoid logging demos

frances-h:
Set the testing header when building the docs to prevent them from logging the demos.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -9,9 +9,11 @@ Release Notes
     * Fixes
     * Changes
     * Documentation Changes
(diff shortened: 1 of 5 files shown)


docs/source/release_notes.rst
```diff
@@ -9,9 +9,11 @@ Release Notes
     * Fixes
     * Changes
     * Documentation Changes
+        * Prevent logging on build (:pr:`1498`)
```
gsheni, line 12: Un-comment above (remove dots and un-indent)
```
Future Release
==============
```

gsheni (approved):

Issue #1381, opened 2021-04-01
Replace variable_type in FeatureBase class


gsheni:
- With the upcoming integration, variable_type will no longer exist on FeatureBase. The primitive will be defined with a return type that includes Logical Type and Semantic Tag(s). 
- https://github.com/alteryx/featuretools/blob/1f437226cc0610de3af208baf9d788af376b7bb6/featuretools/feature_base/feature_base.py#L179
- This issues needs to update/modify the logic in `variable_type` property to account for this change. The name of the property should be changed as well (`woodwork_type`, `typing_info`, etc)

thehomebrewnerd:
Closed by #1501

PR #1501, opened 2021-06-22
Woodwork Integration - Features

rwedge:
Adresses #1381 

* Deletes Entity.py
* Deletes Variable.py.  Various places using variable types are either updated or left broken to be fixed later.
* Storing metadata about the entityset on each ColumnAccessor would allow us to keep the shortcut to define features as IdentityFeature(es['log']['value']), might be worth revisiting


featuretools/feature_base/feature_base.py
```diff
@@ -275,7 +269,8 @@ def __rdiv__(self, other):
     def __mul__(self, other):
         """Multiply by other"""
         if isinstance(other, FeatureBase):
-            if self.variable_type == Boolean and other.variable_type == Boolean:
+            if (self.column_schema.logical_type == ltypes.Boolean and
+                    other.column_schema.logical_type == ltypes.Boolean):
```
thehomebrewnerd, line 273: Not sure if it matters, but you could use `isinstance` here instead of checking equality:
`isinstance(self.column_schema.logical_type, ltypes.Boolean)`.

Adding the `is_boolean` property to WW would also help here...

featuretools/feature_base/feature_base.py
```diff
@@ -333,94 +328,95 @@ def __invert__(self):
...
+        self.return_type = column_accessor.ww.schema
+        super(IdentityFeature, self).__init__(entityset=entityset,
+                                              dataframe_name=dataframe_name,
                                               base_features=[],
                                               relationship_path=RelationshipPath([]),
                                               primitive=PrimitiveBase,
                                               name=name)
 
     @classmethod
     def from_dictionary(cls, arguments, entityset, dependencies, primitives_deserializer):
```
thehomebrewnerd, line 352: Do we need `dependencies` and `primitives_deserializer` as parameters here?

thehomebrewnerd (commented):
Some initial comments based on what I've reviewed so far.

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -109,7 +109,7 @@ class Day(TransformPrimitive):
     """
     name = "day"
     input_types = [ColumnSchema(logical_type=Datetime)]
-    return_type = ColumnSchema(logical_type=Ordinal)
+    return_type = ColumnSchema(logical_type=Ordinal, semantic_tags={'category'})
```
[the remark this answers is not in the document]
rwedge: perhaps this a bug / something we should look at, but the Woodwork equivalent to the "Discrete" variable type is the 'category' semantic tag, but I believe ColumnSchemas do not automatically add standard tags so `Mode` wasn't stacking on this primitive

featuretools/tests/computational_backend/test_feature_set_calculator.py
```diff
@@ -459,17 +458,17 @@ def test_make_3_stacked_agg_feats(df):
...
-        'e1': variable_types.Categorical,
-        'e2': variable_types.Categorical,
-        'e3': variable_types.Categorical,
-        'val': variable_types.Numeric
+    tags = {
+        'id': 'index',
+        'e1': 'category',
+        'e2': 'category',
+        'e3': 'category',
+        'val': 'numeric'
```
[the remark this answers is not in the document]
rwedge: Makes sense.

featuretools/tests/entityset_tests/test_entity.py
```diff
@@ -1,17 +1,16 @@
 from datetime import datetime
 
-import numpy as np
+# import numpy as np
```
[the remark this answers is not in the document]
rwedge: I'm all for deleting it once we've confirmed these test are no longer necessary / have been duplicated elsewhere

featuretools/primitives/standard/transform_primitive.py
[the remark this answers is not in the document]
thehomebrewnerd: @tamargrey and I talked about this behavior briefly. I think it is by design, and not a bug. Might be something to review eventually to see if we want standard tags added by default or not when creating a ColumnSchema directly.

featuretools/feature_base/feature_base.py
rwedge: I believe it's so the `from_dictionary` arguments are uniform across feature types so the feature deserializer doesn't have to think about what arguments to pass

featuretools/tests/primitive_tests/test_features_deserializer.py
```diff
@@ -11,7 +11,7 @@
 
 
 def test_single_feature(es):
-    feature = ft.IdentityFeature(es['log']['value'])
+    feature = ft.IdentityFeature(es, 'log', 'value')
```
thehomebrewnerd, line 14: In some files we are using `ft.IdentifyFeature` and in others we import `IdentifyFeature` and use it. Not sure if there is a good reason for that, but I just noticed it in case we want to clean up and be consistent in the tests.

featuretools/tests/primitive_tests/test_features_serializer.py
```diff
@@ -45,14 +45,14 @@ def test_base_features_in_list(es):
 
 
 def test_multi_output_features(es):
-    value = ft.IdentityFeature(es['log']['product_id'])
+    value = ft.IdentityFeature(es, 'log', 'product_id')
     threecommon = ft.primitives.NMostCommon()
-    tc = ft.Feature(es['log']['product_id'], parent_entity=es["sessions"], primitive=threecommon)
+    tc = ft.Feature(ft.IdentityFeature(es, 'log', 'product_id'), parent_dataframe_name="sessions", primitive=threecommon)
```
thehomebrewnerd, line 50: Use `value` in place of `ft.IdentityFeature(es, 'log', 'product_id')` here? And should value be named something else?

featuretools/tests/primitive_tests/test_direct_features.py
```diff
@@ -185,57 +185,57 @@ def pd_top3(x):
...
     error_text = 'No relationship from "customers" to "customers" found.'
     with pytest.raises(RuntimeError, match=error_text):
-        ft.DirectFeature(diamond_es['customers']['name'], diamond_es['customers'])
+        DirectFeature(IdentityFeature(diamond_es, 'customers', 'name'), 'customers')
 
 
 def test_serialization(es):
-    value = ft.IdentityFeature(es['products']['rating'])
-    direct = ft.DirectFeature(value, es['log'])
+    value = ft.IdentityFeature(es, 'products', 'rating')
```
thehomebrewnerd, line 234: `ft.IdentityFeature` -> `IdentifyFeature`?

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -838,9 +838,9 @@ def test_return_variable_types(es):
                                    agg_primitives=[Count, NMostCommon],
                                    trans_primitives=[Absolute, Hour, IsIn])
 
-    discrete = ft.variable_types.Discrete
-    numeric = ft.variable_types.Numeric
-    datetime = ft.variable_types.Datetime
+    discrete = "Discrete"
+    numeric = "Numeric"
+    datetime = "Datetime"
```
thehomebrewnerd, line 843: Just a temporary change?

thehomebrewnerd (commented):
Went through the files that I hadn't been through. Only noticed a handful of things, but overall don't notice any major issues.

featuretools/tests/primitive_tests/test_features_deserializer.py
rwedge: we're also very inconsistent on logical types, either using `ww.logical_types.Boolean`, `ltypes.Boolean`, or `Boolean`

thehomebrewnerd: This doesn't need to hold up anything here - maybe we should create a separate issue and clean that up later?

rwedge: Created #1510  to track the woodwork issue

For the featuretools imports, thoughts on the stance that we should avoid the `ft` import unless we are trying to test that we can access the class or method from the `ft` namespace?

thehomebrewnerd: I have no issues with that stance, and I think it will make the code cleaner without `ft.` sprinkled about everywhere...

featuretools/computational_backends/utils.py
```diff
@@ -261,16 +256,18 @@ def _check_cutoff_time_type(cutoff_time, es_time_type):
     if isinstance(cutoff_time, tuple):
         cutoff_time_value = cutoff_time[0]
         time_type = _check_time_type(cutoff_time_value)
-        is_numeric = time_type == NumericTimeIndex
-        is_datetime = time_type == DatetimeTimeIndex
+        is_numeric = time_type == 'numeric_time_index'
+        is_datetime = time_type == 'datetime_time_index'
     else:
+        raise NotImplementedError()
```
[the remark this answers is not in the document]
rwedge: I'll add a note to the TODO; don't want to implement just yet since we don't have the overall picture for CFM

featuretools/tests/primitive_tests/test_features_deserializer.py
rwedge: ceated #1511 to track the `ft.` use in tests issue

thehomebrewnerd (approved):
Changes all look good to me!

PR #1505, opened 2021-06-23
Add replace_inf_values util function

### LINKED ISSUE #1496: Handle divide by 0 in DivideNumeric primitives
- The Binary Transform Primitive in Featuretools which will divide 2 numeric columns
- However, this primitives does not currently handle divide by 0 (it outputs **inf**, when it should output **NaN**)
- The DivideNumeric primitives should handle the divide by 0 case and output NaN, not `inf` like it currently does.
  - Ideally this happens without significantly affecting primitive performance speed 

### Open Questions
- Does this apply to DivideNumericScalar and DivideByFeature too?

thehomebrewnerd:
### Add replace_inf_values util function
Closes #1496 

Adds a new utility function, `replace_inf_values`, that can be used to replace `np.inf` values in a feature matrix with `np.NaN` values (default) or with a value of the user's choosing (optional).

Also relocates Feature Selection section in API Reference, which was previously includes under the `EntitSet, Entity, Relationship, Variable Types` section instead of with the other feature related methods.

### DIFF
--- a/docs/source/api_reference.rst
+++ b/docs/source/api_reference.rst
@@ -251,6 +251,26 @@ Feature encoding
 
     encode_features
 
+Feature Selection
+~~~~~~~~~~~~~~~~~
+.. currentmodule:: featuretools.selection
+.. autosummary::
+    :toctree: generated/
+
+    remove_low_information_features
+    remove_highly_correlated_features
+    remove_highly_null_features
+    remove_single_value_features
+
+Feature utils
+~~~~~~~~~~~~~
+.. currentmodule:: featuretools.computational_backends
+.. autosummary::
+    :toctree: generated/
+
+    replace_inf_values
+    
+
 Saving and Loading Features
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 .. currentmodule:: featuretools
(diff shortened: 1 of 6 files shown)


featuretools/computational_backends/utils.py
```diff
@@ -274,3 +275,17 @@ def _check_cutoff_time_type(cutoff_time, es_time_type):
...
+
+        Args:
+            feature_matrix (DataFrame): DataFrame whose columns are feature names and rows are instances
+            replacement_value (int, float, str, optional): Value with which ``np.inf`` values will be replaced
+
+        Returns:
+            feature_matrix
+
+    """
+    return feature_matrix.replace([np.inf, -np.inf], replacement_value)
```
gsheni, line 291: Would it be using `inplace`?

thehomebrewnerd: We have generally avoided `inplace` operations as those are not supported by Dask.

gsheni: okay

docs/source/api_reference.rst
```diff
@@ -251,6 +251,26 @@ Feature encoding
...
+.. currentmodule:: featuretools.selection
+.. autosummary::
+    :toctree: generated/
+
+    remove_low_information_features
+    remove_highly_correlated_features
+    remove_highly_null_features
+    remove_single_value_features
+
+Feature utils
```
rwedge, line 265: how about Feature Matrix utils?

gsheni (approved):

rwedge (approved):

PR #1507, opened 2021-06-28
Add make_index functionality to Featuretools

### LINKED ISSUE #1491: Handle `make_index` logic directly in featuretools instead of passing to Woodwork
Woodwork will be removing the `make_index` logic in [this issue](https://github.com/alteryx/woodwork/issues/984), so Featuretools' `EntitySet.add_dataframe` should handle the logic for adding a column to the dataframe before initializing Woodwork on it.

That the DataFrame had an index created for it should be stored in the Woodwork metadata as `'make_index': True` to be used later on. The following changes should be made:
- handle index column creation directly in `EntitySet.add_dataframe`, adding `make_index` to the DataFrame's metadata.
- Remove any references to the Woodwork `make_index` attribute
- add test to confirm that the make index value is maintained across a deepcopy
- add test to confirm that the make index value is included in equality check (it will only be in the deep equality check, because shallow doesn't check the Woodwork metadata)
- Handle `update_dataframe` in a case where an index was made for the original dataframe, in cases where the new dataframe both does and does not have the made index column. (currently not handled at all)

thehomebrewnerd:
### Add make_index functionality to Featuretools
Closes #1491 

Adds `make_index` functionality back to Featuretools now that it has been removed from Woodwork. These changes also make it possible to support `make_index` with Koalas dataframes as in-place column additions are not needed.


featuretools/entityset/entityset.py
```diff
@@ -642,8 +638,6 @@ def add_dataframe(self,
             extra_params = []
             if index is not None:
                 extra_params.append('index')
-            if make_index:
-                extra_params.append('make_index')
```
rwedge, line 646: what if the user passes a ww dataframe but wants to use make index?

thehomebrewnerd: If I'm looking at things right, the user cannot use `make_index` with a ww dataframe (although they can try). If the user passes a ww dataframe, there is a check that the dataframe has an index column set, and if not a ValueError is raised: https://github.com/alteryx/featuretools/blob/bf3147c67667bd91ec81663bdcaa136f4b84c606/featuretools/entityset/entityset.py#L639-L640

It is possible that the user would pass a ww dataframe that already has an index set, and also pass make_index=True, in which case we would ignore the `make_index` parameter. So, I guess it does make sense to leave that check in place for the warning.

featuretools/entityset/entityset.py
```diff
@@ -620,19 +620,15 @@ def add_dataframe(self,
                 warnings.warn('Performing type inference on Dask or Koalas DataFrames may be computationally intensive. '
                               'Specify logical types for each column to speed up EntitySet initialization.')
 
+            created_index, index, dataframe = _create_index(index, make_index, dataframe)
```
tamargrey, line 623: can we change the name of `_create_index`? Or maybe split it up into separate `_validate_index` and `_create_index` functions? (I see that this came from the `Entity` class, but in this context the name makes the three variables returned from it kind of confusing)

featuretools/entityset/entityset.py
tamargrey: This means we're losing a behavior that previously existed where a user could init Woodwork on a dataframe before EntitySet init while also creating an index. 

Not that big of a deal to lose it, I think, but just wanted to point out.

tamargrey (commented):
Probably worth testing with `0.4.2` to confirm that everything works as expected with the `make_index` param actually removed from Woodwork

thehomebrewnerd:
> Probably worth testing with `0.4.2` to confirm that everything works as expected with the `make_index` param actually removed from Woodwork

That's actually what I did to create this PR! Everything here works with Woodwork 0.4.1 and 0.4.2 with no issues.

featuretools/entityset/entityset.py
```diff
@@ -620,19 +620,15 @@ def add_dataframe(self,
...
-                              make_index=make_index,  # TODO: handle make index logic outside of Woodwork
-                              already_sorted=already_sorted)
-            # If no index column is specified, set the first column
-            if dataframe.ww.index is None:
-                warnings.warn(("Using first column as index. "
-                               "To change this, specify the index parameter"))
-                dataframe.ww.set_index(dataframe.columns[0])
-
+                              already_sorted=already_sorted,
+                              table_metadata={'created_index': created_index})
```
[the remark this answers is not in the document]
thehomebrewnerd: That's an easy enough change to make, so I'll update to only add this key/value pair if `created_index` is not `None`.

featuretools/entityset/entityset.py
```diff
@@ -1540,3 +1532,37 @@ def _vals_to_series(instance_vals, variable_id):
...
+        warnings.warn(("Using first column as index. "
+                       "To change this, specify the index parameter"))
+        index = df.columns[0]
+    elif make_index and index in df.columns:
+        # Case 3: user wanted to make index but column already exists
+        raise RuntimeError(f"Cannot make index: column with name {index} already present")
+    elif index not in df.columns:
+        if not make_index:
+            # Case 4: user names index, it is not in df. does not specify
+            # make_index.  Make new index column and warn
```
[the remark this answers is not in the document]
thehomebrewnerd: I can go either way here. For now I was just replicating the previous behavior that existed in Featuretools. Generally I'm not a big fan of making assumptions about what the user wants unless we have to. With that in mind, I would be fine with removing this case.

@rwedge Any thoughts on this?

featuretools/entityset/entityset.py
thehomebrewnerd: Updating the function name to `_get_or_create_index` and changing `created_index` to a boolean `index_was_created` to better indicate whether the index column was created by Featuretools or not.

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -433,7 +430,7 @@ def test_make_index_any_location(df):
                      make_index=True,
                      logical_types=logical_types,
                      dataframe=df)
-    if dd and isinstance(df, dd.DataFrame):
+    if not isinstance(df, pd.DataFrame):
```
tamargrey, line 433: can use `es.dataframe_type` here if you'd like

thehomebrewnerd: Good call! Updated.

featuretools/entityset/entityset.py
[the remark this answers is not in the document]
rwedge: I think we should keep the existing behavior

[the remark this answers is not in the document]
thehomebrewnerd: Sounds good. I will keep this code path in to maintain the behavior that is currently in FT.

tamargrey (approved):
lgtm!

rwedge (approved):
:shipit:

PR #1509, opened 2021-06-29
Validate schema after dataframe update

jeff-hernandez:
Adds test case to check if new data conforms to original schema after the update.

### DIFF
--- a/featuretools/tests/entityset_tests/test_ww_es.py
+++ b/featuretools/tests/entityset_tests/test_ww_es.py
@@ -849,5 +849,5 @@ def test_update_dataframe_schema():
 
     assert not after.dtypes.equals(before.dtypes)
     es.update_dataframe(dataframe_name='data', df=after)
-    assert after.ww.schema == before.ww.schema
-    assert after.dtypes.equals(before.dtypes)
+    assert es['data'].ww.schema == before.ww.schema
+    assert es['data'].dtypes.equals(before.dtypes)
--- a/featuretools/tests/entityset_tests/test_ww_es.py
+++ b/featuretools/tests/entityset_tests/test_ww_es.py
@@ -847,7 +847,7 @@ def test_update_dataframe_schema():
         'latlong': [[5, 6], '7, 8', [np.nan, np.nan]],
     }).astype({'boolean': int, 'latlong': str})
 
-    assert not after.dtypes.equals(before.dtypes)
+    assert not after.dtypes.equals(es['data'].dtypes)
     es.update_dataframe(dataframe_name='data', df=after)
     assert es['data'].ww.schema == before.ww.schema
     assert es['data'].dtypes.equals(before.dtypes)
--- a/featuretools/tests/entityset_tests/test_ww_es.py
+++ b/featuretools/tests/entityset_tests/test_ww_es.py
@@ -824,3 +824,28 @@ def test_normalize_ww_init():
 
     assert es['new_df'].ww.name == 'new_df'
     assert es['new_df'].ww.schema.name == 'new_df'
+
+
+def test_update_dataframe_schema():
+    before = pd.DataFrame({
+        'index': [0, 1, 2],
+        'boolean': [True, False, True],
+        'latlong': [(1.0, 2.0), (3.0, 4.0), np.nan],
+    })
+
+    before.ww.init(index='index', logical_types={
+        'boolean': ww.logical_types.Boolean,
+        'latlong': ww.logical_types.LatLong,
+    })
+
+    es = EntitySet()
+    es.add_dataframe(dataframe_name='data', df=before)
+
+    after = pd.DataFrame({
+        'index': [0, 1, 2],
+        'boolean': [0, 1, 0],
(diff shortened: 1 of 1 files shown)


featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -824,3 +824,28 @@ def test_normalize_ww_init():
...
+        'latlong': ww.logical_types.LatLong,
+    })
+
+    es = EntitySet()
+    es.add_dataframe(dataframe_name='data', df=before)
+
+    after = pd.DataFrame({
+        'index': [0, 1, 2],
+        'boolean': [0, 1, 0],
+        'latlong': [[5, 6], '7, 8', [np.nan, np.nan]],
```
tamargrey, line 847: can we explicitly set the dtypes in this dataframe and add an assertion to confirm that they do not match `before`?

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -824,3 +824,28 @@ def test_normalize_ww_init():
...
+    })
+
+    es.update_dataframe(dataframe_name='data', df=after)
+    assert after.ww.schema == before.ww.schema
```
tamargrey, line 851: Lets also add a check that the dtypes of `before` and `after` match exactly as well

tamargrey: Also now that I think about it, I think there's a chance that the dataframe object in the entityset will not be `after`.  Should these be looking at `es['data']` instead of `after`?

featuretools/tests/entityset_tests/test_ww_es.py
jeff-hernandez: Yep fixed

featuretools/tests/entityset_tests/test_ww_es.py
jeff-hernandez: Sure, it might be more intuitive to reference from the entityset

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -824,3 +824,30 @@ def test_normalize_ww_init():
...
+        'latlong': [(1.0, 2.0), (3.0, 4.0), np.nan],
+    })
+
+    before.ww.init(index='index', logical_types={
+        'boolean': ww.logical_types.Boolean,
+        'latlong': ww.logical_types.LatLong,
+    })
+
+    es = EntitySet()
+    es.add_dataframe(dataframe_name='data', df=before)
```
thehomebrewnerd, line 842: I think this needs to be updated. If we are adding a Woodwork dataframe, it needs to have a `name` assigned. So we can either add a name in the `ww.init` call, or we can skip the init and specify the index and logical types during the add_dataframe call.

Also, the `df` parameter needs to be updated to `dataframe`

featuretools/tests/entityset_tests/test_ww_es.py
```diff
@@ -824,3 +824,30 @@ def test_normalize_ww_init():
...
+
+    after = pd.DataFrame({
+        'index': [0, 1, 2],
+        'boolean': [0, 1, 0],
+        'latlong': [[5, 6], '7, 8', [np.nan, np.nan]],
+    }).astype({'boolean': int, 'latlong': str})
```
thehomebrewnerd, line 848: I don't think this type casting is actually changing the dtype of the `latlong` column. Seems to be `object`. Maybe we should add another numeric column and use `float64` and `Int64` for before and after just to add another case to confirm it's working as expected? Maybe not strictly necessary but thinking it couldn't hurt.

thehomebrewnerd (commented):
@jeff-hernandez @tamargrey As I was wrapping up a review of this I was double-checking that we had things covered, and I took another look at `test_update_dataframe_different_dtypes` in `test_ww_es.py`. I think I might have missed that test before when I created the initial issue.

Do you think this new test is adding something of value that isn't covered in the existing test? I'm wondering now if we even need to add this...

tamargrey:
> Do you think this new test is adding something of value that isn't covered in the existing test? I'm wondering now if we even need to add this...

@thehomebrewnerd This test probably shows the behavior that we're trying to test more clearly, because it's less complicated, but I think between the test you mentioned and also `test_update_dataframe_data_transformation`, which tests latlong transformation, most of the same bases are being covered.

thehomebrewnerd:
> This test probably shows the behavior that we're trying to test more clearly, because it's less complicated

I do agree with you here on that point.

Having said that, the previous tests are also testing with Dask and Koalas, and if we leave the new test in, we should update to test with those libraries as well. I'm not sure it's worth it though.

@jeff-hernandez Do you have any concerns with closing this issue and PR without merging?

jeff-hernandez:
@thehomebrewnerd it does seem covered in another test already. I'm okay with closing the PR

thehomebrewnerd:
> @thehomebrewnerd it does seem covered in another test already. I'm okay with closing the PR

Sounds good. Sorry I missed that before!

PR #1518, opened 2021-07-01
Primitive options include entities overrides exclude entities

rwedge:
> Options for individual primitives or groups of primitives are set by the primitive_options parameter of DFS. This parameter maps any desired options to specific primitives. In the case of conflicting options, options set at this level will override options set at the entire DFS run level, and the include options will always take priority over their ignore counterparts.

Fixes bug where if an entity was in `include_entity` and `exclude_entity` for the same primitive, the primitive would be excluded from that entity instead of included


frances-h (dismissed):
👍 looks good!

frances-h (approved):

PR #1521, opened 2021-07-06
Remove entity tests

### LINKED ISSUE #1497: Remove Entity class definition
- With the Woodwork integration, the Entity class will be replaced with a Woodwork typed Dataframe
- The class definition and code should be completely removed, as it no longer be used

jeff-hernandez:
Closes #1497 by removing entity tests that are no longer needed


featuretools/utils/wrangle.py
```diff
@@ -100,20 +100,6 @@ def _check_time_against_column(time, time_column):
...
-    '''
-    Checks if `time` is an instance of common int, float, or datetime types.
-    Returns "numeric", "datetime", or "unknown" based on results
-    '''
-    time_type = None
-    if isinstance(time, (datetime, np.datetime64)):
-        time_type = "datetime_time_index"
-    elif isinstance(time, (int, float)) or np.issubdtype(time, np.integer) or np.issubdtype(time, np.floating):
-        time_type = "numeric_time_index"
-    return time_type
```
jeff-hernandez, line 114: Do we also want to remove `_check_cutoff_time_type` since it uses `_check_time_type`?
https://github.com/alteryx/featuretools/blob/dc5f69df4ad9a52aa0de581aa741a415bb32aaf1/featuretools/computational_backends/utils.py#L251-L273

jeff-hernandez:
It seems like the tests in `test_entity.py` are either already covered or no longer applicable

featuretools/utils/wrangle.py
thehomebrewnerd: @jeff-hernandez I think we will still need `_check_cutoff_time_type` to make sure that the cutoff time type is  consistent with the time index type specified in the Woodwork dataframes. Maybe the better approach here is to keep `_check_time_type` and `_check_cutoff_time_type` and refactor to use the updated values that we will have for `es_time_type` and to also work if the cutoff time is a WW dataframe (in the block that has a `NotImplementedError()` now).

What do you think?

tamargrey: +1 on keeping both functions and refactoring to use `'numeric'` and `Datetime`! (didn't realize that _check_time_type was used in this way, I think, when I was originally thinking we could remove it)

thehomebrewnerd (commented):
Just a couple small notes on the deleted tests.

featuretools/tests/entityset_tests/test_entity.py
```diff
@@ -1,236 +0,0 @@
...
-    to_pandas
-)
-from featuretools.utils.gen_utils import Library
-
-
-def test_is_index_column(es):
-    assert es['cohorts'].index == 'cohort'
-
-
-def test_reorders_index():
```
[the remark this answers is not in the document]
thehomebrewnerd: I made a note of this in the comment of the main integration PR so we can hopefully document later.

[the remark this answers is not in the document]
tamargrey: I do believe it's listed in the release notes' breaking changes `Featuretools will no longer re-order columns in DataFrames so that the index column is the first column of the DataFrame` but definitely good to include elsewhere as well

featuretools/utils/wrangle.py
jeff-hernandez: Sounds good, updated the values in `_check_time_type`

featuretools/tests/entityset_tests/test_entity.py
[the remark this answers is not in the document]
jeff-hernandez: Cool, sounds like this is resolved.

featuretools/utils/wrangle.py
```diff
@@ -100,17 +100,16 @@ def _check_time_against_column(time, time_column):
     return False
 
 
-# --> TODO possibly remove this or convert to Woodwork
 def _check_time_type(time):
     '''
     Checks if `time` is an instance of common int, float, or datetime types.
     Returns "numeric", "datetime", or "unknown" based on results
```
thehomebrewnerd, line 106: Not a big deal since this is a private function, but we probably should update the docstring here to match the code.

thehomebrewnerd (approved):
One docstring comment, but otherwise looks fine.

We will also need to update `_check_cutoff_time_type`, but that's getting a bit beyond the scope of this PR, so we can leave that for later when we update CFM.

PR #1522, opened 2021-07-07
Fix WW integration release notes

thehomebrewnerd:
### Fix WW integration release notes

Fixes several items related to the Woodwork integration that were no longer correct in the release notes.


tamargrey (approved):
lgtm! good call with the other changes as well

PR #1526, opened 2021-07-12
Standardize imports for Woodwork in codebase

### LINKED ISSUE #1510: Standardize how woodwork is imported in the codebase
Woodwork is imported in a handful of ways

`import woodwork as ww`
`from woodwork import logical_types as ltypes`
`from woodwork.logical_types import Numeric`
`from woodwork.column_schema import ColumnSchema`

We should look at the various scenarios where we import woodwork and decide on standard ways to import based on the scenario.

jeff-hernandez:
Closes #1510 by using direct imports of objects from Woodwork

### DIFF
--- a/featuretools/entityset/deserialize.py
+++ b/featuretools/entityset/deserialize.py
@@ -4,8 +4,8 @@ import tarfile
 import tempfile
 
 import pandas as pd
-import woodwork as ww
 from woodwork.deserialize import read_woodwork_table
+from woodwork.type_sys.type_system import type_system
 
 from featuretools.entityset.relationship import Relationship
 from featuretools.utils.gen_utils import check_schema_version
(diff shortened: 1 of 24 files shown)


featuretools/entityset/deserialize.py
```diff
@@ -4,8 +4,8 @@
 import tempfile
 
 import pandas as pd
-import woodwork as ww
 from woodwork.deserialize import read_woodwork_table
+from woodwork.type_sys.type_system import type_system
```
thehomebrewnerd, line 8: What do you think about importing this as `ww_type_system` to make it a little more clear when reading the code on line 73 that this is a woodwork method call?

thehomebrewnerd (approved):
Looks good. Just one minor suggestion, but it isn't important to update if you prefer to leave it like it is.

featuretools/entityset/deserialize.py
jeff-hernandez: Sure, I don't mind updating

PR #1550, opened 2021-07-22
Refactor add_interesting_values to leverage Woodwork

### LINKED ISSUE #1426: Refactor add_interesting_values

In the current `add_intersting_values` method, all of the columns in an entity are looped over, and value counts are computed for columns that are categorical. Woodwork implements a very similar functionality in the `value_counts` method. We should investigate updating the featuretools `add_interesting_values` method to make use of the Woodwork `value_counts` method. This might allow the featuretools code to be simplified and eliminate some looping or conditional checks that could be done by Woodwork instead.

thehomebrewnerd:
### Refactor add_interesting_values to leverage Woodwork

Closes #1426

Refactor `EntitySet.add_interesting_values` to leverage Woodwork value counts method for categorical columns instead of using a custom implementation.

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1204,7 +1204,7 @@ class EntitySet(object):
             total_count = len(df)
 
             for col, counts in value_counts.items():
-                if {'index', 'foreign_key'}.intersection(df.ww[col].ww.semantic_tags):
+                if {'index', 'foreign_key'}.intersection(df.ww.semantic_tags[col]):
                     continue
 
                 for i in range(min(max_values, len(counts))):
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1193,8 +1193,7 @@ class EntitySet(object):
 
         def add_value(df, col, val, verbose):
             if verbose:
-                msg = "Column {}: Marking {} as an "
-                msg += "interesting value"
+                msg = "Column {}: Marking {} as an interesting value"
                 logger.info(msg.format(col, val))
             interesting_vals = df.ww.columns[col].metadata.get('interesting_values', [])
             df.ww.columns[col].metadata['interesting_values'] = interesting_vals + [val]
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1190,43 +1190,37 @@ class EntitySet(object):
             dataframes = [self[dataframe_name]]
         else:
             dataframes = self.dataframes
+
+        def add_value(df, col, val, verbose):
+            if verbose:
+                msg = "Column {}: Marking {} as an "
+                msg += "interesting value"
+                logger.info(msg.format(col, val))
+            interesting_vals = df.ww.columns[col].metadata.get('interesting_values', [])
+            df.ww.columns[col].metadata['interesting_values'] = interesting_vals + [val]
+
         for df in dataframes:
-            for column in df.columns:
-                # some heuristics to find basic 'where'-able columns
-                # include categorical columns, exclude index or foreign key columns
-                col_schema = df.ww.columns[column]
-                col_is_valid = (col_schema.is_categorical and
-                                not {'index', 'foreign_key'}.intersection(col_schema.semantic_tags))
+            value_counts = df.ww.value_counts(top_n=max(25, max_values), dropna=True)
+            total_count = len(df)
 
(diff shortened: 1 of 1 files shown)


featuretools/entityset/entityset.py
```diff
@@ -1190,43 +1190,37 @@ def add_interesting_values(self, max_values=5, verbose=False, dataframe_name=Non
...
-                                msg = "Column {}: Marking {} as an "
-                                msg += "interesting value"
-                                logger.info(msg.format(column, idx))
-                            interesting_vals = df.ww.columns[column].metadata.get('interesting_values', [])
-                            df.ww.columns[column].metadata['interesting_values'] = interesting_vals + [idx]
+            value_counts = df.ww.value_counts(top_n=max(25, max_values), dropna=True)
+            total_count = len(df)
 
+            for col, counts in value_counts.items():
+                if {'index', 'foreign_key'}.intersection(df.ww[col].ww.semantic_tags):
```
tamargrey, line 1207: probably better to do `df.ww.semantic_tags[col]` so we don't have to go through the getitem and get the series

tamargrey (commented):
looking good, just a quick comment about woodwork syntax

featuretools/entityset/entityset.py
thehomebrewnerd: Yep, good call. Updated.

featuretools/entityset/entityset.py
```diff
@@ -1190,43 +1190,37 @@ def add_interesting_values(self, max_values=5, verbose=False, dataframe_name=Non
             dataframes = [self[dataframe_name]]
         else:
             dataframes = self.dataframes
+
+        def add_value(df, col, val, verbose):
+            if verbose:
+                msg = "Column {}: Marking {} as an "
+                msg += "interesting value"
```
rwedge, line 1197: should fit on one line now

featuretools/entityset/entityset.py
```diff
@@ -1190,43 +1190,37 @@ def add_interesting_values(self, max_values=5, verbose=False, dataframe_name=Non
...
+                logger.info(msg.format(col, val))
+            interesting_vals = df.ww.columns[col].metadata.get('interesting_values', [])
+            df.ww.columns[col].metadata['interesting_values'] = interesting_vals + [val]
```
rwedge, line 1200: we could append instead of concat

featuretools/entityset/entityset.py
```diff
@@ -1190,43 +1190,37 @@ def add_interesting_values(self, max_values=5, verbose=False, dataframe_name=Non
...
 
+            for col, counts in value_counts.items():
+                if {'index', 'foreign_key'}.intersection(df.ww.semantic_tags[col]):
+                    continue
+
+                for i in range(min(max_values, len(counts))):
+                    if len(counts) < 25:
+                        # Categorical columns will include counts of 0 for all values
+                        # in categories. Only include values that have nonzero counts.
+                        if counts[i]['count'] > 0:
```
rwedge, line 1214: can we move the check for count_value > 0 to before the if statement and break out of the loop once we encounter a count of 0?

featuretools/entityset/entityset.py
thehomebrewnerd: Indeed. Updated both instances of this. Also updated this in the interesting values refactor PR which touches these lines.

thehomebrewnerd: Updated

featuretools/entityset/entityset.py
thehomebrewnerd: Hmmm. Might be missing something in your suggestion, but we need to handle the situation where we get a new list back from the `get` call in 1199. That new list (with the new value added) then needs to be assigned back to the metadata, which was previously not set. I think using append here would require an extra line of code, something like this:

```python
interesting_vals = df.ww.columns[col].metadata.get('interesting_values', [])
intesting_vals.append(val)
df.ww.columns[col].metadata['interesting_values'] = interesting_vals
```

rwedge: It would require an extra line of code but it would save the creation of an additional list by skipping the concat

thehomebrewnerd: Sounds good. Happy to update, just wanted to make sure I understood your suggestion.

rwedge (approved):
LGTM

PR #1562, opened 2021-07-27
Fix additional reviewers to min and latest dependency checker

gsheni:
- Both workflows are failing on main due to a syntax error:
  - https://github.com/alteryx/featuretools/actions/runs/1068139279
  - https://github.com/alteryx/featuretools/actions/runs/1068139296
- This MR fixes it by referencing the correct team.

### DIFF
--- a/.github/workflows/latest_dependency_checker.yml
+++ b/.github/workflows/latest_dependency_checker.yml
@@ -31,4 +31,4 @@ jobs:
         branch: latest-dep-update
         branch-suffix: short-commit-hash
         base: main
-        team-reviewers: @alteryx/ml-tools
+        team-reviewers: alteryx/ml-tools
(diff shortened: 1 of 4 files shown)


.github/workflows/latest_dependency_checker.yml
```diff
@@ -31,4 +31,4 @@ jobs:
         branch: latest-dep-update
         branch-suffix: short-commit-hash
         base: main
-        team-reviewers: @alteryx/ml-tools
+        team-reviewers: alteryx/ml-tools
```
rwedge, line 34: based on how team-reviewers is used in [this example](https://github.com/peter-evans/create-pull-request#reference-example), do you think just `ml-tools` would work?

gsheni: Fixed. I also lowered the pandas version so we can test this.

rwedge:
to test, could we change the github action to happen on push to that branch?

gsheni:
No, we can't. The PR won't be generated on push because there won't be a diff with `main`

gsheni:
@rwedge No, we can't. The PR won't be generated on push because there won't be a diff with `main`

rwedge (approved):
LGTM

PR #1568, opened 2021-07-28
Fix remaining broken primitive tests

### LINKED ISSUE #1260: Update Feature schema version with new Logical Types/Semantic Tags
- With the Woodwork integration, primitives will now take in Logical Types/Semantic Tags.
- This will change the feature definitions, so the schema version will need to be updated.

### LINKED ISSUE #1547: Fix failing primitive tests
After PR #1533 is merged in there will still be approximately 17 tests in the `tests/primitive_tests` folder that are still failing. 

A fix should be implemented, allowing the following tests to pass. This may involve updates to the tests or the code in deep feature synthesis or calculate feature matrix.

thehomebrewnerd:
### Fix remaining broken primitive tests

Closes #1547 
Closes #1260 

This PR fixes the remaining broken tests in `tests/primitives` and updates the feature serialization schema version.


featuretools/tests/primitive_tests/test_direct_features.py
```diff
@@ -36,7 +36,11 @@ def test_direct_from_identity(es):
     df = calculator.run(np.array([0, 5]))
     df = to_pandas(df, index='id', sort_index=True)
     v = df[d.get_name()].tolist()
-    assert v == [0, 1]
+    if es.dataframe_type == Library.KOALAS.value:
+        expected = ['0', '1']
+    else:
+        expected = [0, 1]
+    assert v == expected
```
thehomebrewnerd, line 43: For Koalas all categorical columns are cast to a `string` dtype by Woodwork since Koalas does not support the `category` dtype. This causes the values to change from integers to strings here and in the test that follows.

featuretools/tests/primitive_tests/test_groupby_transform_primitives.py
```diff
@@ -271,7 +271,7 @@ def check(feature):
     for primitive in [CumSum, CumMean, CumMax, CumMin]:
         check(ft.Feature(pd_es, 'log', 'value', groupby=ft.IdentityFeature(pd_es, 'log', 'session_id'), primitive=primitive))
 
-    check(ft.Feature(ft.IdentityFeature(pd_es, 'log', 'session_id'), groupby=ft.IdentityFeature(pd_es, 'log', 'session_id'), primitive=CumCount))
+    check(ft.Feature(ft.Feature(pd_es, 'log', 'product_id'), groupby=ft.Feature(pd_es, 'log', 'product_id'), primitive=CumCount))
```
thehomebrewnerd, line 274: `session_id` is now an `Integer` logical type, so it is not recognized as a valid input for the `CumCount` primitive, so updated to use `product_id` instead, which is `Categorical`.

featuretools/tests/primitive_tests/test_agg_feats.py
```diff
@@ -220,6 +220,11 @@ def test_stack_on_self(es, test_primitive):
 def test_init_and_name(es):
     log = es['log']
 
+    # Add a BooleanNullable column so primitives with that input type get tested
+    boolean_nullable = log.ww['purchased']
+    boolean_nullable = boolean_nullable.ww.set_logical_type('BooleanNullable')
+    log.ww['boolean_nullable'] = boolean_nullable
```
tamargrey, line 226: I think `log.ww.set_types(logical_types={'purchased':'BooleanNullable'})` would work here as a single-liner if you want (unless this kind of inplace woodwork operation doesn't work as expected... in which case we should look into that)

thehomebrewnerd: We actually need to have two different columns in the dataframe here, one with `Boolean` logical type and one with `BooleanNullable` in order to test all the primitives with their valid input types. In any case, this test is still having some issues with Dask inputs, and may end up getting changed more, depending on the solution to that.

tamargrey: oh, just now seeing that the name of the second column is different!

thehomebrewnerd:
I'm having some trouble with the tests in `test_dask_primitives.py`, which appear to be a result of how pandas and/or Dask handle certain situations involving the new nullable dtypes. I have created issues in these repos as a result to try to get either a solution or more insight into what is happening and why.

Pandas concat issue with `bool` and `boolean` series:
https://github.com/pandas-dev/pandas/issues/42800

Dask result difference when running `series1.eq(series2)` with `category` and `string` dtypes:
https://github.com/dask/dask/issues/7957

Depending on the feedback on those issues, the tests in `test_dask_primitives.py` might need to be reworked to get around these issues.

featuretools/primitives/standard/binary_transform.py
```diff
@@ -280,6 +280,7 @@ def equal(x_vals, y_vals):
                 categories = set(x_vals.cat.categories).union(set(y_vals.cat.categories))
                 x_vals = x_vals.cat.add_categories(categories.difference(set(x_vals.cat.categories)))
                 y_vals = y_vals.cat.add_categories(categories.difference(set(y_vals.cat.categories)))
+                y_vals = y_vals.cat.reorder_categories(x_vals.cat.categories)
```
thehomebrewnerd, line 283: Certain comparisons were resulting in an error because the categories were not ordered the same between `x_vals` and `y_vals` so added this line here and below to fix.

tamargrey:
> Pandas concat issue with `bool` and `boolean` series:

The issue you created reminds me of the bug that came up when first testing `1.3.0`: https://github.com/pandas-dev/pandas/issues/42092

I wonder if they're related or if 1.3.0 behavior is different than 1.2.4 (which was the pandas version listed in the `show_versions` output)

thehomebrewnerd:
> > Pandas concat issue with `bool` and `boolean` series:
> 
> The issue you created reminds me of the bug that came up when first testing `1.3.0`: [pandas-dev/pandas#42092](https://github.com/pandas-dev/pandas/issues/42092)
> 
> I wonder if they're related or if 1.3.0 behavior is different than 1.2.4 (which was the pandas version listed in the `show_versions` output)

As far as I can tell the behavior has not changed in more recent versions.

featuretools/primitives/standard/binary_transform.py
rwedge: should we make an issue to revisit this to figure why it changed / how costly this reorder is?

thehomebrewnerd: Sure. Created #1584 for this.

rwedge (approved):
LGTM

PR #1581, opened 2021-07-29
Preserve EntitySet Woodwork schemas on pickling

### LINKED ISSUE #1541: Fix tests that use dask cluster
After PR #1533 is merged in, there are several tests that utilize a dask cluster for calculating a feature matrix. These tests are failing because Woodwork typing information is not being retained properly.

A fix should be implemented, allowing the following tests to pass.

In  `test_calculate_feature_matrix.py`:
- test_dask_kwargs
- test_dask_persisted_es
- test_warning_not_enough_chunks
- test_calls_progress_callback_cluster

In `test_dfs_method.py`:
- test_dask_kwargs
- test_calls_progress_callback_cluster

Note - these tests have been commented out in #1533

frances-h:
Add `__getstate__` and `__setstate__` methods to EntitySet to preserve and restore Woodwork schema when an entityset is pickled.

Closes #1541

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -7,6 +7,7 @@ import dask.dataframe as dd
 import numpy as np
 import pandas as pd
 from woodwork import init_series
+from woodwork.exceptions import WoodworkNotInitError
 from woodwork.logical_types import Datetime
 
 from featuretools.entityset import deserialize, serialize
@@ -1151,10 +1152,35 @@ class EntitySet(object):
             self.dataframe_dict[df.ww.name] = df
 
         self.reset_data_description()
+
+    # ###########################################################################
+    # #  Pickling ###############################################
+    # ###########################################################################
+    def __getstate__(self):
+        schemas = {}
+        for df_name, df in self.dataframe_dict.items():
+            try:
+                schema = df.ww.schema
+            except WoodworkNotInitError:
+                schema = None
+            schemas[df_name] = schema
+
+        return {
+            **self.__dict__,
+            'schemas': schemas
+        }
+
+    def __setstate__(self, state):
+        schemas = state.pop('schemas')
+        for df_name, df in state.get('dataframe_dict', {}).items():
+            if schemas[df_name] is not None:
+                df.ww.init(schema=schemas[df_name])
+        for k, v in state.items():
+            object.__setattr__(self, k, v)
+
     # ###########################################################################
     # #  Other ###############################################
     # ###########################################################################
-
     def add_interesting_values(self, max_values=5, verbose=False, dataframe_name=None, values=None):
         """Find or set interesting values for categorical columns, to be used to generate "where" clauses
 
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -25,6 +25,7 @@ pd.options.mode.chained_assignment = None  # default='warn'
 logger = logging.getLogger('featuretools.entityset')
 
 LTI_COLUMN_NAME = '_ft_last_time'
+WW_SCHEMA_KEY = '_ww__getstate__schemas'
 
 
 class EntitySet(object):
@@ -1158,16 +1159,15 @@ class EntitySet(object):
     def __getstate__(self):
         return {
             **self.__dict__,
-            'schemas': {df_name: df.ww.schema for df_name, df in self.dataframe_dict.items()}
+            WW_SCHEMA_KEY: {df_name: df.ww.schema for df_name, df in self.dataframe_dict.items()}
         }
 
     def __setstate__(self, state):
-        schemas = state.pop('schemas')
+        ww_schemas = state.pop(WW_SCHEMA_KEY)
         for df_name, df in state.get('dataframe_dict', {}).items():
-            if schemas[df_name] is not None:
-                df.ww.init(schema=schemas[df_name])
-        for k, v in state.items():
-            object.__setattr__(self, k, v)
+            if ww_schemas[df_name] is not None:
+                df.ww.init(schema=ww_schemas[df_name], validate=False)
+        self.__dict__ = state
 
     # ###########################################################################
     # #  Other ###############################################
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1167,7 +1167,7 @@ class EntitySet(object):
         for df_name, df in state.get('dataframe_dict', {}).items():
             if ww_schemas[df_name] is not None:
                 df.ww.init(schema=ww_schemas[df_name], validate=False)
-        self.__dict__ = state
+        self.__dict__.update(state)
 
     # ###########################################################################
     # #  Other ###############################################
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -7,7 +7,6 @@ import dask.dataframe as dd
 import numpy as np
 import pandas as pd
 from woodwork import init_series
-from woodwork.exceptions import WoodworkNotInitError
 from woodwork.logical_types import Datetime
 
 from featuretools.entityset import deserialize, serialize
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1157,17 +1157,9 @@ class EntitySet(object):
     # #  Pickling ###############################################
     # ###########################################################################
     def __getstate__(self):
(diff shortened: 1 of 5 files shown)


featuretools/entityset/entityset.py
```diff
@@ -1151,10 +1152,35 @@ def add_last_time_indexes(self, updated_dataframes=None):
...
     # ###########################################################################
-    # #  Other ###############################################
+    # #  Pickling ###############################################
     # ###########################################################################
+    def __getstate__(self):
+        schemas = {}
+        for df_name, df in self.dataframe_dict.items():
+            try:
+                schema = df.ww.schema
+            except WoodworkNotInitError:
```
tamargrey, line 1164: Does this `WoodworkNotInitError` get thrown from doing `df.ww.schema` on a DataFrame that doesn't have Woodwork initialized?

 I think that the `schema` property is one of the only properties that shouldn't throw that error, so I'd expect that `df.ww.schema` would just be `None` if woodwork is not initalized. So if it's getting thrown, there might be some other woodwork attribute that's being referenced.

Then again, I'd also expect that all the dataframes in `dataframe_dict` would have woodwork initialized, so I very well could be misunderstanding the state of things when this method gets called.

frances-h: Yeah, I just removed those lines since 1) `dataframe_dict` should initialize woodwork and 2) you're right that the `schema` would just be `None`.

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -1659,33 +1659,33 @@ def __call__(self, update, progress_percent, time_elapsed):
...
+    trans_per_customer = ft.Feature(ft.Feature(pd_mock_customer, "transactions", "transaction_id"), parent_dataframe_name="customers", primitive=Count)
+    features = [trans_per_session, ft.Feature(trans_per_customer, "sessions")]
 
-#     with cluster() as (scheduler, [a, b]):
-#         dkwargs = {'cluster': scheduler['address']}
-#         calculate_feature_matrix(features,
-#                                  entityset=pd_mock_customer,
-#                                  progress_callback=mock_progress_callback,
-#                                  dask_kwargs=dkwargs)
+    with cluster() as (scheduler, [a, b]):
```
davesque, line 1680: Same comment as above about cluster fixtures.

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -2016,3 +2017,19 @@ def test_dataframe_type_empty_es():
...
 def test_dataframe_type_pandas_es(pd_es):
     assert pd_es.dataframe_type == Library.PANDAS.value
+
+
+def test_pd_es_pickling(pd_es):
+    pkl = pickle.dumps(pd_es)
+    unpickled = pickle.loads(pkl)
+
+    assert pd_es.__eq__(unpickled, deep=True)
+    assert not hasattr(unpickled, 'schemas')
```
davesque, line 2027: This instance of the magic schemas key should probably also be using a module-level constant that is imported from the `entityset` module.

featuretools/tests/synthesis/test_dfs_method.py
```diff
@@ -467,49 +467,49 @@ def __call__(self, update, progress_percent, time_elapsed):
...
+            self.total_progress_percent = 0
+
+        def __call__(self, update, progress_percent, time_elapsed):
+            self.total_update += update
+            self.total_progress_percent = progress_percent
+            self.progress_history.append(progress_percent)
+
+    mock_progress_callback = MockProgressCallback()
+
+    with cluster() as (scheduler, [a, b]):
```
davesque, line 484: Same comment as above about cluster fixtures.

featuretools/tests/synthesis/test_dfs_method.py
```diff
@@ -467,49 +467,49 @@ def __call__(self, update, progress_percent, time_elapsed):
...
+
+def test_dask_kwargs(pd_dataframes, relationships):
+    cutoff_times_df = pd.DataFrame({"instance_id": [1, 2, 3],
+                                    "time": [10, 12, 15]})
+    feature_matrix, features = dfs(dataframes=pd_dataframes,
+                                   relationships=relationships,
+                                   target_dataframe_name="transactions",
+                                   cutoff_time=cutoff_times_df)
+
+    with cluster() as (scheduler, [a, b]):
```
davesque, line 504: Same comment as above about cluster fixtures.

davesque (commented):
Noticed a few things.  Otherwise, it looks like the general logic being used in the pickling methods is correct.

featuretools/entityset/entityset.py
```diff
@@ -1151,10 +1151,27 @@ def add_last_time_indexes(self, updated_dataframes=None):
...
+        return {
+            **self.__dict__,
+            'schemas': {df_name: df.ww.schema for df_name, df in self.dataframe_dict.items()}
+        }
+
+    def __setstate__(self, state):
+        schemas = state.pop('schemas')
+        for df_name, df in state.get('dataframe_dict', {}).items():
+            if schemas[df_name] is not None:
+                df.ww.init(schema=schemas[df_name])
```
thehomebrewnerd, line 1168: I *think* we can run init here with `validate=False` since we should know the schema is valid for the dataframe, and this could save a little time.
```
df.ww.init(schema=schemas[df_name], validate=False)
```

featuretools/entityset/entityset.py
```diff
@@ -1151,10 +1151,27 @@ def add_last_time_indexes(self, updated_dataframes=None):
...
+        for k, v in state.items():
+            object.__setattr__(self, k, v)
```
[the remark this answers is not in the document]
frances-h: I don't think it matters one way or another since `self.__dict__[k] = v` uses `__setattr__` anyways. It can be simplified to just `self.__dict__ = state` though so I've updated to use that.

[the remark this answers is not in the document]
davesque: Yeah I think that'll be simpler.  It might also make sense to do `self.__dict__.update(state)` just in case there's already something in `__dict__` for some reason.  There also appears to be a bit of precedent for using `update` in the python docs: https://docs.python.org/3/library/pickle.html#handling-stateful-objects.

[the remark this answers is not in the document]
davesque: > `self.__dict__[k] = v` uses `__setattr__` anyways

I didn't realize this so I just poked around in the console a bit to confirm.  However, it appears that setting a key on `__dict__` skips over the whole descriptor protocol and method resolution and doesn't actually hit `__setattr__`:
```python
In [1]: class Foo:
   ...:     def set_attr(self, k, v):
   ...:         self.__dict__[k] = v
   ...:     def __setattr__(self, k, v):
   ...:         print("__setattr__ called!")
   ...:         super().__setattr__(k, v)
   ...:

In [2]: f = Foo()

In [3]: f.set_attr("a", 1)

In [4]: vars(f)
Out[4]: {'a': 1}

In [5]: f.a = 2
__setattr__ called!

In [6]: vars(f)
Out[6]: {'a': 2}
```
I thought I'd just mention it here for posterity. :)

thehomebrewnerd (approved):
LGTM!

davesque (commented):
Nice, looks great!  Thanks for addressing all those comments!  Any ideas on what's causing the test failures?  In one case, there appeared to be a missing dependency.  In other cases, it looked more like there were maybe some failures relating to the changes you made.

thehomebrewnerd:
> Any ideas on what's causing the test failures?

@davesque Just curious what tests you are referring to here?

davesque:
I'm seeing that basically all tests except for the lint tests have failed:
Screen Shot 2021-08-02 at 9 52 51 AM

thehomebrewnerd:
> I'm seeing that basically all tests except for the lint tests have failed:
> Screen Shot 2021-08-02 at 9 52 51 AM

Yeah, we are in a weird state right now. That is to be expected. For this PR, I would mostly focus on the tests that were commented out, but are now included. The base branch still has a lot of failing tests that we are working on fixing for a variety of reasons, so as long as we aren't breaking tests that were previously passing before this PR, I don't think we need to be too worried at this point.

davesque:
@thehomebrewnerd @frances-h Ahh, okay.  Sorry I'm out of the loop on this.

davesque (approved):

tamargrey (approved):
lgtm!

PR #1582, opened 2021-07-30
Add reference to entity set in Woodwork columns

### LINKED ISSUE #1515: Add entityset information to columns in entities
Currently the woodwork integration deviates from Featuretools' old implementation for how features are written

old way: `ft.Feature(es['customers']['signup_date'])`

new way: `ft.Feature(es, 'customers', 'signup_date')`

not too different, but when making GroupByTransform features the new way can be a bit cumbersome:

old way: `ft.Feature(es['log']['value'], primitive=CumSum, groupby=es['log']['product_id'])`

new way: `ft.Feature(es, 'log', 'value', primitive=CumSum, groupby=ft.IdentityFeature(es, 'log', 'product_id'))`

The ability to make features the old way was tied to each `Variable` having references to the entityset, something the ColumnAccessors on each dataframe do not currently emulate.

Keeping the old functionality would be a nice to have but not crucial

jeff-hernandez:
Closes #1515


jeff-hernandez:
@rwedge would something like this resolve the issue?

rwedge:
Each column will also need to store the name of their dataframe since that's not stored by woodwork

Looking at how this is handled in current Featuretools, `Variable` has a direct reference to the entityset while features use the entityset metadata.

We want to be able to make a Feature by calling `ft.Feature(es['log']['value'])`

jeff-hernandez:
Similar to how we add the entity set to the column accessor, we could also add the dataframe name. Creating features would depend on a Woodwork column for those references, so I think it would be `ft.Feature(es['log'].ww['value'])`. Are we changing the API to `ft.Feature(es, 'log', 'value')`?

tamargrey:
(I can't remember if I've asked this before, so apologies if we've been through this) Do we need a reference to the whole entityset with dataframes and everything, or could we somehow replace the dataframes with just the `TableSchema` objects?

rwedge:
Currently the code in the integration branch has chang

> Creating features would depend on a Woodwork column for those references, so I think it would change to `ft.Feature(es['log'].ww['value'])`. Are we changing the API to `ft.Feature(es, 'log', 'value')`?

Currently the integration branch the API has changed to `ft.Feature(es, 'log', 'value')`, but one of the main goals of this issue/PR is to revert the API to `ft.Feature(es['log'].ww['value'])`, which is about as close as we can get to the old API

rwedge:
> (I can't remember if I've asked this before, so apologies if we've been through this) Do we need a reference to the whole entityset with dataframes and everything, or could we somehow replace the dataframes with just the TableSchema objects?

We'd need to take a closer look at how `Feature` uses `EntitySet` but I think a dictionary of `TableSchema` objects + the  relationships dictionary might be sufficient

jeff-hernandez:
It might be possible to store the reference in `_item_cache` of the pandas series.

```python
for column in dataframe.columns:
    dataframe[column]._item_cache['entityset'] = self
```

jeff-hernandez:
When indexing a Series from a DataFrame, reusing the same instance is a difference between Pandas and Dask behavior. Pandas does use the same instance whereas Dask does **not** and creates a new instance each time. The column accessor also follows this behavior, so storing references on the column accessor becomes problematic for Dask since the references do not exist on the new instance of the column accessor.

```python
import dask.dataframe as dd
import pandas as pd

pdf = pd.DataFrame({'a': range(3)})
assert id(pdf['a']) == id(pdf['a'])

ddf = dd.from_pandas(pdf, chunksize=3)
assert id(ddf['a']) != id(ddf['a'])
```

Instead of column accessors, using the column metadata and a dictionary works to store the entity set references. The column metadata cannot have a direct reference to the entity set, otherwise, this raises a recursion error during a deep copy. However, the column metadata can store the entity set id and the name of the DataFrame while a dictionary can store the entity set objects. Creating a feature from a Woodwork column would be done by getting the entity set id and DataFrame name from the column metadata. Then, using those keys to return the corresponding entity set and DataFrame from the dictionary. This PR implements the described approach and uses the following:
- `FeatureBase._entityset_ref` to store the entity set objects
- `Entityset._add_references_to_metadata` to add the entity set id and DataFrame name to the column metadata

featuretools/feature_base/feature_base.py
```diff
@@ -643,19 +646,18 @@ class TransformFeature(FeatureBase):
...
         # new_class_init method in make_trans_primitive
-        if hasattr(base_features, '__iter__'):
-            base_features = [_check_feature(bf) for bf in base_features]
+        base_features = _check_feature(base_features)
+        if isinstance(base_features, list):
             msg = "all base features must share the same dataframe"
             assert len(set([bf.dataframe_name for bf in base_features])) == 1, msg
         else:
-            base_features = [_check_feature(base_features)]
+            base_features = [base_features]
```
rwedge, line 654: Since this code block gets used in multiple classes and we already moved some logic into `_check_feature`, how about we rename it to `_validate_base_features` and move all this logic in there.

So on the classes it would look like

```python
base_features = _validate_base_features(base_features)
```
and life goes on

jeff-hernandez: Sounds good, will update

jeff-hernandez: ✅  39989675b01117d11a313abc10af8fbb1c6c2fd0

jeff-hernandez:
### Calculating feature matrix with dask parallel compute.

We want to check if the EntitySet dictionary affects parallel computing with dask, so we calculated a feature matrix based on different branches to compare results. The outcome indicates that the EntitySet dictionary works with parallel computing and doesn't significantly affect the runtime. The branches `main` and `woodwork-integration` have slightly larger runtimes than the `add_entityset_reference` branch, but runtimes vary across trials.

| Branch                  | Runtime  |
|-------------------------|----------|
| main                    | 176.93 s |
| woodwork-integration    | 168.50 s |
| add_entityset_reference | 165.61 s |

To reproduce this experiment, you can run the code snippet below. 

```python
import featuretools as ft
es = ft.demo.load_flight()
features = ft.dfs(target_dataframe_name='airports', entityset=es, features_only=True, max_depth=3)
fm = ft.calculate_feature_matrix(features=features, entityset=es, n_jobs=8, dask_kwargs={'diagnostics_port': 8786})
```

I also attached screen recordings of the dask runtime.

https://user-images.githubusercontent.com/12969559/130485474-848e344c-6ac4-4be1-8b2f-31df38a57868.mov

https://user-images.githubusercontent.com/12969559/130485551-2172ea92-f131-4ce7-bcdf-00cd375a8deb.mov

https://user-images.githubusercontent.com/12969559/130485388-b828eba3-4cc2-4fd6-acf3-89501293be2d.mov

jeff-hernandez:
### Calculating feature matrix with serialized features.

We also want to check if the EntitySet dictionary affects serialized features when calculating the feature matrix in a new python process. The outcome indicates serialized features (including pickled features) do work in new python processes.

#### Process 1

To rerun the experiment, create and save features in one python process.

```python
import featuretools as ft
es = ft.demo.load_flight()
features = ft.dfs(target_dataframe_name='airports', entityset=es, features_only=True, max_depth=2)
ft.save_features(features, 'features.json')
```

#### Process 2

Then, deserialize the features and calculate the feature matrix in a new python process.

```python
import featuretools as ft

es = ft.demo.load_flight()
features = ft.load_features('features.json')
fm = ft.calculate_feature_matrix(features=features, entityset=es)
```

featuretools/tests/computational_backend/test_feature_set_calculator.py
```diff
@@ -69,8 +69,8 @@ def test_make_dfeat(es):
     assert (v == 33)
 
 
-def test_make_agg_feat_of_identity_column(es):
-    agg_feat = ft.Feature(ft.Feature(es, 'log', 'value'), parent_dataframe_name='sessions', primitive=Sum)
+def test_make_agg_feat_of_identity_variable(es):
```
thehomebrewnerd, line 72: `variable` -> `column`?

featuretools/tests/computational_backend/test_feature_set_calculator.py
```diff
@@ -113,8 +113,8 @@ def test_full_dataframe_error_dask(dask_es):
         calculator.run(np.array([1]))
 
 
-def test_make_agg_feat_of_identity_index_column(es):
-    agg_feat = ft.Feature(ft.Feature(es, 'log', 'id'), parent_dataframe_name='sessions', primitive=Count)
+def test_make_agg_feat_of_identity_index_variable(es):
```
thehomebrewnerd, line 116: `variable` -> `column`

featuretools/tests/computational_backend/test_feature_set_calculator.py
```diff
@@ -69,8 +69,8 @@ def test_make_dfeat(es):
     assert (v == 33)
 
 
-def test_make_agg_feat_of_identity_column(es):
-    agg_feat = ft.Feature(ft.Feature(es, 'log', 'value'), parent_dataframe_name='sessions', primitive=Sum)
+def test_make_agg_feat_of_identity_variable(es):
+    agg_feat = ft.Feature(ft.Feature(es['log'].ww['value']), parent_dataframe_name='sessions', primitive=Sum)
```
thehomebrewnerd, line 73: I see a lot of calls like this, but I'm not sure we need the inner `ft.Feature()` call now? We needed them before this update, but I'm thinking we don't now? This could become:

```
agg_feat = ft.Feature(es['log'].ww['value'], parent_dataframe_name='sessions', primitive=Sum)
```

If these nested `ft.Feature` calls aren't needed any more, I think we should remove them before merging this.

featuretools/tests/primitive_tests/test_feature_base.py
```diff
@@ -72,46 +72,46 @@ def test_get_depth(es):
...
 def test_return_type_inference_numeric_time_index(int_es):
-    last = ft.Feature(int_es, "log", "datetime", parent_dataframe_name="customers", primitive=Last)
+    last = ft.Feature(int_es["log"].ww["datetime"], parent_dataframe_name="customers", primitive=Last)
     assert "numeric" in last.column_schema.semantic_tags
 
 
 def test_return_type_inference_id(es):
-    # direct features should keep foreign key tag
-    direct_id_feature = ft.Feature(es, "sessions", "customer_id", "log")
+    # direct features should keep Id variable type
```
thehomebrewnerd, line 109: Should leave this as foreign key tag

featuretools/tests/primitive_tests/test_feature_base.py
```diff
@@ -72,46 +72,46 @@ def test_get_depth(es):
...
+    direct_id_feature = ft.Feature(es["sessions"].ww["customer_id"], "log")
     assert "foreign_key" in direct_id_feature.column_schema.semantic_tags
 
-    # aggregations of foreign key types should get converted
-    last_feat = ft.Feature(es, "log", "session_id", parent_dataframe_name="customers", primitive=Last)
+    # aggregations of Id variable types should get converted
```
thehomebrewnerd, line 113: Should leave this as foreign key tag

featuretools/tests/primitive_tests/test_feature_descriptions.py
```diff
@@ -181,40 +181,40 @@ class CustomTrans(TransformPrimitive):
...
     assert describe_feature(custom_agg) == custom_agg_description
 
-    custom_trans = TransformFeature(IdentityFeature(es, 'log', 'zipcode'), CustomTrans)
+    custom_trans = TransformFeature(IdentityFeature(es['log'].ww['zipcode']), CustomTrans)
     custom_trans_description = 'The result of applying CUSTOM_TRANSFORM to the "zipcode".'
     assert describe_feature(custom_trans) == custom_trans_description
 
 
-def test_column_description(es):
+def test_variable_description(es):
```
thehomebrewnerd, line 197: variable -> column

PR #1593, opened 2021-08-03
Update Specifying Primitive Options guide for Woodwork

### LINKED ISSUE #1570: Update Specifying Primitive Options Page with Woodwork Dataframe
- https://featuretools.alteryx.com/en/latest/guides/specifying_primitive_options.html
- We should also at the same time **use Jupyter notebooks, rather than RST files.**

thehomebrewnerd:
### Update Specifying Primitive Options guide for Woodwork

Closes #1570


frances-h (commented):
Looks good overall, just need to clear the notebook outputs

frances-h (approved):
lgtm!

thehomebrewnerd:
> can we switch the `dfs` calls to use `features_only` -- we don't use the feature matrix and it would reduce the indent on these calls

Yep - updated!

Issue #1596, opened 2021-08-05
Fix remaining Woodwork integration tests that have been skipped


thehomebrewnerd:
The following tests need to be fixed as part of the Woodwork integration. The reason these tests are failing are documented along with the test. All of these tests have been skipped with pytest.skip for the time being. Once the issues are resolved, these xfails should be removed

**featuretools/tests/entityset_tests/test_es.py::test_concat_with_lti[dask_es]**
- This test fails for Dask because Woodwork creates an error that shows up on .compute() caused by this line `updated_series = self[dataframe_name].ww.logical_types[col_name].transform(series)` in `EntitySet.update_dataframe`. This error happens when trying to run `transform` for the two `LatLong` columns in the `log` dataframe
- [Woodwork Issue 1099](https://github.com/alteryx/woodwork/issues/1099) has been created for this problem

**featuretools/tests/primitive_tests/test_dask_primitives.py::test_transform**
- This test is failing because of a Dask issue with series.eq returning different values than the pandas equivalent. This is documented in [Dask Issue 7957](https://github.com/dask/dask/issues/7957).

**featuretools/tests/selection/test_selection.py::test_remove_highly_correlated_features**
- This code executed as part of this test attempts to use Woodwork `select` on the feature matrix. However, at this stage Woodwork is not initialized on the returned feature matrix, so this results in an error. Once Featuretools returns a Woodwork initialized feature matrix this test can be fixed.

**featuretools/tests/synthesis/test_koalas_dfs.py::test_single_table_ks_entityset_cutoff_time_df**
- This failure should be revisited after the feature matrix has WW initialized. In the current release the column `IS_WEEKEND(dates` is returned as a float64 column (instead of a boolean) because of two missing values. Without Woodwork initialized on the feature matrix, this column is now being returned as a type `object` for pandas with values of `False, NaN, Nan, 0.0`. Koalas returns a column with `False, None, None, False`. Once this column is set with the proper dtype, these values should match. The values in other columns match but there are some dtype differences.

gsheni:
@thehomebrewnerd moved this into our current sprint.

gsheni:
As this is not required to finish the integration, moving this issue to the follow up to Woodwork integration Epic

thehomebrewnerd:
Support for Dask dataframes is being dropped.

PR #1597, opened 2021-08-05
Updates for WW 0.6.0 and fix other failing tests

### LINKED ISSUE #1583: Upgrading from Woodwork 0.5.0 to 0.5.1 (and upcoming 0.6.0) introduces multiple test failures
### Upgrading from Woodwork 0.5.0 to 0.5.1 introduces multiple test failures

Upgrading from Woodwork 0.5.0 to 0.5.1 introduces 37 new Featuretools test failures. We need to investigate what is causing these failures and determine if they are Woodwork bugs, or updates that are needed inside of Featuretools.

Initial investigation indicates many of the failures are related to two items:
- Woodwork added the `origin` attribute which breaks serialization tests
- There are issues storing the inferred datetime format when calling `init_series` with Dask or Koalas input

thehomebrewnerd:
### Updates for WW 0.6.0 and fix other failing tests

Closes #1583 

This PR implements changes to upgrade to WW 0.6.0. Several other failing tests were also fixed and the remaining four failing tests were skipped pending further resolution in Issue #1596


featuretools/synthesis/utils.py
```diff
@@ -59,7 +59,7 @@ def get_unused_primitives(specified, used):
 def _schemas_equal(feature_schema, primitive_schema):
     """Checks if a feature schema matches a primitive schema"""
     if primitive_schema.logical_type and \
-            primitive_schema.logical_type != feature_schema.logical_type:
+            not isinstance(feature_schema.logical_type, type(primitive_schema.logical_type)):
```
rwedge, line 62: could this potentially cause issues comparing two ordinals with different orders?

thehomebrewnerd: Ah - the issue that caused me to change this was involved comparing two datetimes that had different `datetime_format` values. A `Datetime` column with a format was not found to be a match for a primitive that had a `Datetime` input because of the format value difference.

I think the idea here is that if we define a primitive input schema as `ColumnSchema(logical_type=Ordinal)` that we want it to match a column that might have a logical type instance of `Ordinal` with order values defined. The previous check of `primitive_schema.logical_type != feature_schema.logical_type` would fail to find a match if one had order values defined and the other did not.

So having said that I, think we want to match on logical type alone and not specific attributes of the instantiated logical types?

thehomebrewnerd: Maybe the name `_schemas_equal` isn't quite right given that behavior?

featuretools/tests/synthesis/test_koalas_dfs.py
```diff
@@ -202,6 +202,7 @@ def test_single_table_ks_entityset_single_cutoff_time():
 
 @pytest.mark.skipif('not ks')
 def test_single_table_ks_entityset_cutoff_time_df():
+    pytest.skip("TODO: Investigate failure root cause and resolve")
```
jeff-hernandez, line 205: Do we have an issue open for this? This one would also be helpful to reference.

featuretools/tests/primitive_tests/test_dask_primitives.py
```diff
@@ -12,6 +13,7 @@
 
 
 def test_transform(pd_es, dask_es):
+    pytest.skip("TODO: Dask issue with `series.eq`. Fix once Dask Issue #7957 is closed.")
```
jeff-hernandez, line 16: Nice, this is helpful for reference.

jeff-hernandez (approved):
Looks good, just some minor comments

featuretools/tests/synthesis/test_koalas_dfs.py
thehomebrewnerd: Nothing specific to this failure - haven't fully determined the root cause for this one yet, but the issue is documented broadly in #1596

thehomebrewnerd: If the root cause is established and leads to another issue before this PR is closed, I will update this text to include more detail and the issue number.

featuretools/synthesis/utils.py
rwedge: yeah.  the logic of this method seems practically identitacal to `is_valid_input`, should we just have one method?

thehomebrewnerd: Seems like we should be able to have a single method. I'll update.

thehomebrewnerd: Replaced `_schemas_equal` with `is_valid_input` and updated associated tests.

rwedge (approved):

Issue #1578, opened 2021-07-28
Update FAQ with Woodwork Dataframe


gsheni:
- The page below should be updated with the new function calls, parameters, and the accessor approach.
- https://featuretools.alteryx.com/en/latest/resources/frequently_asked_questions.html
- We should also at the same time **use Jupyter notebooks, rather than RST files.**

tamargrey:
We should add a question under the EntitySets section on `How do I update a column's description or metadata` that explains that you'll need to go through `df.ww.columns[col_name]` to update attributes like description and metadata and that doing `df.ww[col_name]` will give you a new Series object, so if you use that syntax, any changes won't get made on the EntitySet itself.

jeff-hernandez:
Closed by #1649

PR #1603, opened 2021-08-06
Update feature description guide to use Woodwork

tamargrey:
Removes the feature descriptions rst file and creates a `feature_descriptions.ipynb` notebook containing updated code and language to match the Woodwork integration.


docs/source/guides/feature_descriptions.ipynb
```diff
@@ -0,0 +1,329 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "69e42e65",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "feature_descriptions = {\n",
+    "    'sessions: SUM(transactions.amount)': 'the total transaction amount for a session'}\n",
+    "\n",
+    "display(feature_defs[14])\n",
```
rwedge, line 150: squashed output into one cell again

rwedge (commented):
I think we should a note about the gotcha of trying to set the ColumnSchema.description by doing `table.ww[column].schema.description = "foo"`

docs/source/guides/feature_descriptions.ipynb
```diff
@@ -0,0 +1,329 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "07d48d9b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "print(feature_defs[9])\n",
+    "ft.describe_feature(feature_defs[9])"
+   ]
+  },
```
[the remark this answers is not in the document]
tamargrey: ah, got it. I can put each line in its own cell here and anywhere else I do this

docs/source/guides/feature_descriptions.ipynb
```diff
@@ -0,0 +1,329 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "655399b9",
+   "metadata": {},
+   "source": [
+    "## Improving Descriptions\n",
+    "\n",
+    "While the default descriptions can be helpful, they can also be further improved by providing custom definitions of `ColumnSchema` objects and features, and by providing alternative templates for primitive descriptions. \n",
```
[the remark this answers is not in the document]
tamargrey: That seems like a good rule of thumb!

docs/source/guides/feature_descriptions.ipynb
```diff
@@ -127,12 +145,24 @@
...
+  {
+   "cell_type": "raw",
+   "id": "27013806",
+   "metadata": {},
+   "source": [
+    ".. note::\n",
+    "\n",
+    "    When setting a description on a column in a DataFrame as described above, you will have to go through the ``columns`` attribute of the Woodwork DataFrame, ``df.ww.columns[col_name].description``, in order to set the column's description in a way that will be propogated to the feature description. \n",
+    "    \n",
+    "    The other option for selecting the series from the DataFrame with ``df.ww[col_name].ww.description`` will create an entirely new Woodwork Series that is not related to the EntitySet from which feature descriptions are built."
```
tamargrey, line 157: @rwedge Is this along the lines of how you were thinking we could warn users off of going through the `getitem` in woodwork?

rwedge: @tamargrey yes, although I think this note could focus solely on why not to do `df.ww[col_name]` since we've already shown in the example that `df.ww.columns` works.  We could probably skip re-iterating the correct way to do it

tamargrey: Pushed an update that puts the bad syntax first. I did keep a quick reference to the correct syntax after the fact, since I worry that the difference between `es['customers'].ww.columns['join_date'].description` and `es['customers'].ww['join_date'].ww.description` may not be immediately obvious.

docs/source/guides/feature_descriptions.ipynb
```diff
@@ -152,9 +152,7 @@
    "source": [
     ".. note::\n",
     "\n",
-    "    When setting a description on a column in a DataFrame as described above, you will have to go through the ``columns`` attribute of the Woodwork DataFrame, ``df.ww.columns[col_name].description``, in order to set the column's description in a way that will be propogated to the feature description. \n",
-    "    \n",
-    "    The other option for selecting the series from the DataFrame with ``df.ww[col_name].ww.description`` will create an entirely new Woodwork Series that is not related to the EntitySet from which feature descriptions are built."
+    "    When setting a description on a column in a DataFrame as described above, be careful to avoid setting the description via ``df.ww[col_name].ww.description``. The use of ``df.ww[col_name]`` creates an entirely new Series object that is not related to the EntitySet from which feature descriptions are built. Therefore, setting the description in any way other than going through the ``columns`` attribute will not set the column's description in a way that will be propogated to the feature description. "
```
rwedge, line 155: I like this a lot.

We could also add on at the end "This also applies when updating column metadata."  Not necessary for the note but might be useful to share

tamargrey: I'm thinking this gotcha might be an important thing to reference in the Woodwork typing guide. Since, like you said, it'll be relevant to updating information stored in he metadata as well.

There isn't really a section in the guide on accessing descriptions and metadata, but it might be worth it to add a section on "Updating a Columns Typing Information" at the bottom

tamargrey: I'm thinking that adding the situation here to the FAQs in #1578 will mean that we shouldn't need to reference metadata here. I think it's still useful to have a note specific to the situation at hand in these guides when this situation comes up but that we don't need to reference the other situations.

rwedge (approved):
LGTM

frances-h (approved):
looks good!

PR #1604, opened 2021-08-09
Update Koalas Guide to use Woodwork

### LINKED ISSUE #1573: Update Using Koalas EntitySets with Woodwork Dataframe
- https://featuretools.alteryx.com/en/latest/guides/using_koalas_entitysets.html
- We should also at the same time **use Jupyter notebooks, rather than RST files.**

tamargrey:
Creates a `using_koalas_entitysets` jupyter notebook in place of the rst file, and updates the language to match the new Featuretools api with Woodwork.

Also makes a couple of small changes to the Dask guide to fix typos or add clarification.

Closes #1573


docs/source/guides/using_dask_entitysets.ipynb
```diff
@@ -194,17 +194,17 @@
...
     "\n",
     "### EntitySet Limitations\n",
     "\n",
     "When creating a Featuretools `EntitySet` that will be made of Dask DataFrames, all of the DataFrames used to create the `EntitySet` must be of the same type, either all Dask DataFrames or all pandas DataFrames. Featuretools does not support creating an `EntitySet` containing a mix of Dask and pandas DataFrames.\n",
     "\n",
-    "Additionally, the `EntitySet.add_interesting_values()` method is not supported when using a Dask `EntitySet`. If needed, users can manually set `interesting_values` for DataFrames by assigning them directly with syntax similar to this: `es[\"dataframe_name\"].ww[\"variable_name\"].ww.metadata[\"interesting_values\"] = [\"Value 1\", \"Value 2\"]`.\n",
+    "Additionally, the `EntitySet.add_interesting_values()` method is not supported when using a Dask `EntitySet`. If needed, users can manually set `interesting_values` for DataFrames by assigning them directly with syntax similar to this: `es[\"dataframe_name\"].ww.columns[\"column_name\"].metadata[\"interesting_values\"] = [\"Value 1\", \"Value 2\"]`.\n",
```
tamargrey, line 207: I believe the original suggestion here `es["dataframe_name"].ww["column_name"].ww.metadata["interesting_values"] = ["Value 1", "Value 2"]` will fall into the trap of setting the metadata on the separate Series object created by `es["dataframe_name"].ww["column_name"]`. 

Switched to going through the `columns` attr to avoid that. @rwedge had brought this gotcha up in the feature descriptions guide as an important thing to warn users away from doing. I'm thinking it might make sense to add that warning here and in the koalas guide as well.

docs/source/guides/using_koalas_entitysets.ipynb
```diff
@@ -0,0 +1,236 @@
...
+   "cell_type": "raw",
+   "id": "ac77ea82",
+   "metadata": {
+    "raw_mimetype": "text/restructuredtext"
+   },
+   "source": [
+    ".. note::\n",
+    "    Support for Koalas EntitySets is still in Beta. While the key functionality has been implemented, development is ongoing to add the remaining functionality.\n",
+    "    \n",
+    "        All planned improvements to the Featuretools/Koalas integration are `documented on Github <https://github.com/alteryx/featuretools/issues?q=is%3Aopen+is%3Aissue+label%3AKoalas>`_. If you see an open issue that is important for your application, please let us know by upvoting or commenting on the issue. If you encounter any errors using Koalas dataframes in EnitySets, or find missing functionality that does not yet have an open issue, please create a `new issue on Github <https://github.com/alteryx/featuretools/issues>`_."
```
thehomebrewnerd, line 21: `EnitySets` -> `EntitySets`

docs/source/guides/using_koalas_entitysets.ipynb
```diff
@@ -0,0 +1,236 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "bed6a8c8",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "primitives_df = ft.list_primitives()\n",
+    "koalas_compatible_df = primitives_df[primitives_df[\"koalas_compatible\"] == True]\n",
+    "koalas_compatible_df.head()\n",
+    "koalas_compatible_df.tail()"
```
thehomebrewnerd, line 166: Split into multiple cells so output from both `head` and `tail` are displayed.

docs/source/guides/using_dask_entitysets.ipynb
thehomebrewnerd: Maybe the better approach is to demonstrate using the `add_interesting_values` function directly, similar to what we do in the test below?
```python
def test_add_interesting_values_specified_vals(es):
    product_vals = ['coke zero', 'taco clock']
    country_vals = ['AL', 'US']
    interesting_values = {
        'product_id': product_vals,
        'countrycode': country_vals,
    }
    es.add_interesting_values(dataframe_name='log', values=interesting_values)
```

I guess I had forgotten that we reworked this to allow for this type of usage. This approach should work fine with Dask/Koalas entitysets.

docs/source/guides/using_koalas_entitysets.ipynb
```diff
@@ -0,0 +1,236 @@
...
+    "\n",
+    "When using a pandas DataFrames, the ordering of the underlying DataFrame rows is maintained by Featuretools. For a Koalas DataFrame, the ordering of the DataFrame rows is not guaranteed, and Featuretools does not attempt to maintain row order in a Koalas DataFrame. If ordering is important, close attention must be paid to any output to avoid issues.\n",
+    "\n",
+    "### EntitySet Limitations\n",
+    "\n",
+    "When creating a Featuretools ``EntitySet`` that will be made of Koalas dataframes, all of the dataframes used to create the ``EntitySet`` must be of the same type, either all Koalas dataframe, all Dask dataframes, or all pandas dataframes. Featuretools does not support creating an ``EntitySet`` containing a mix of Koalas, Dask, and pandas dataframes.\n",
+    "\n",
+    "Additionally, the ``EntitySet.add_interesting_values()`` method is not supported when using a Koalas ``EntitySet``. If needed, users can manually set ``interesting_values`` for dataframes by assigning them directly with syntax similar to this:  `es[\"dataframe_name\"].ww.columns[\"column_name\"].metadata[\"interesting_values\"] = [\"Value 1\", \"Value 2\"]`.\n",
```
thehomebrewnerd, line 190: Update suggested approach here to be consistent with Dask.

docs/source/guides/using_dask_entitysets.ipynb
tamargrey: oh yeah, forgot about that as well. Definitely a nicer way to set these values; I'll update both guides to show an example

tamargrey: Updated here and in the koalas guide to show an example with the call to `add_interesting_values`

docs/source/guides/using_dask_entitysets.ipynb
```diff
@@ -194,18 +194,37 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "a3c6b9b8",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "values_dict = {'values': [12, 103]}\n",
+    "es.add_interesting_values(dataframe_name='dask_input_df', values=values_dict )\n",
```
thehomebrewnerd, line 218: Super minor nitpick, but there is an extra space between `values_dict` and the closing `)`.

docs/source/guides/using_koalas_entitysets.ipynb
```diff
@@ -187,7 +196,27 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "6e13e5ef",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "values_dict = {'values': [12, 103]}\n",
+    "es.add_interesting_values(dataframe_name='koalas_input_df', values=values_dict )\n",
```
thehomebrewnerd, line 210: Extra space before `)`

thehomebrewnerd (approved):
Just a couple really minor nitpicks, but otherwise looks good!

PR #1608, opened 2021-08-10
Update Glossary with Woodwork terms

### LINKED ISSUE #1579: Update Glossary with Woodwork DataFrame terms
- https://featuretools.alteryx.com/en/latest/resources/usage_tips/glossary.html
- We remove `entity` and `entities` from the page.
- We should remove `variable` from the page
- We should add logical type, semantic tag to the page.

frances-h:
Updates the glossary to remove references to "entity" and "variables" to match the updated woodwork integration. Also adds "logical type", "semantic tags", and "ColumnSchema" to the glossary.

Closes #1579

### DIFF
--- a/docs/source/resources/usage_tips/glossary.rst
+++ b/docs/source/resources/usage_tips/glossary.rst
@@ -41,4 +41,4 @@ Glossary
         Optional additional information on the column about the meaning or potential uses of data. Used to determine which primitives can be applied to a column to generate features.
 
     ColumnSchema
-        All of Woodwork's column type information including the logical type and any semantic tags.
+        All of a Woodwork column's type information including the logical type and any semantic tags.
--- a/docs/source/resources/usage_tips/glossary.rst
+++ b/docs/source/resources/usage_tips/glossary.rst
@@ -8,7 +8,7 @@ Glossary
     :sorted:
 
     feature
-        A transformation of data used for machine learning.  featuretools has a custom language for defining features as described :ref:`here <primitives>`. All features are represented by subclasses of :class:`FeatureBase`.
+        A transformation of data used for machine learning.  Featuretools has a custom language for defining features as described :ref:`here <primitives>`. All features are represented by subclasses of :class:`FeatureBase`.
 
     feature engineering
         The process of transforming data into representations that are better for machine learning.
@@ -32,13 +32,16 @@ Glossary
         A dataframe that references another dataframe via relationship. The "many" in a one-to-many relationship.
 
     relationship
-        A mapping between a parent dataframe and a child dataframe. The child dataframe must contain a column referencing the ID column on the parent dataframe. Represented by the :class:`.Relationship` class.
+        A mapping between a parent dataframe and a child dataframe. The child dataframe must contain a column referencing the index column on the parent dataframe. Represented by the :class:`.Relationship` class.
 
     logical type
-        Additional information about how data should be interpreted or parsed beyond how the data is stored on disk or in memory.
+        Additional information about how a column should be interpreted or parsed beyond how the data is stored on disk or in memory. Used to determine which primitives can be applied to a column to generate features.
 
     semantic tag
-        Optional additional information on the column about the meaning or potential uses of data.
+        Optional additional information on the column about the meaning or potential uses of data. Used to determine which primitives can be applied to a column to generate features.
+
+    ColumnSchema
+        All of Woodwork's column type information including the logical type and any semantic tags.
 
 .. todo
 .. label maker,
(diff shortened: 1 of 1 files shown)


docs/source/resources/usage_tips/glossary.rst
```diff
@@ -13,33 +13,32 @@ Glossary
...
+        A dataframe that is referenced by another dataframe via relationship. The "one" in a one-to-many relationship.
 
-    child entity
-        An entity that references another entity via relationship. The "many" in a one-to-many relationship.
+    child dataframe
+        A dataframe that references another dataframe via relationship. The "many" in a one-to-many relationship.
 
     relationship
-        A mapping between a parent entity and a child entity. The child entity must contain a variable referencing the ID variable on the parent entity. Represented by the :class:`.Relationship` class.
+        A mapping between a parent dataframe and a child dataframe. The child dataframe must contain a column referencing the ID column on the parent dataframe. Represented by the :class:`.Relationship` class.
```
thehomebrewnerd, line 35: I think we should change `ID` to `index` here.

docs/source/resources/usage_tips/glossary.rst
```diff
@@ -13,33 +13,32 @@ Glossary
...
+
+    logical type
+        Additional information about how data should be interpreted or parsed beyond how the data is stored on disk or in memory.
```
thehomebrewnerd, line 38: Maybe this should say `Additional information about how a column should be interpreted...` instead?

docs/source/resources/usage_tips/glossary.rst
```diff
@@ -13,33 +13,32 @@ Glossary
...
+
+    semantic tag
+        Optional additional information on the column about the meaning or potential uses of data.
```
thehomebrewnerd, line 41: Maybe we should mention in both of these that they are used to determine what primitives can be applied to a column to generate features?

thehomebrewnerd (commented):
In addition to the other comments a couple suggestions:
- Capitalize Featuretools in second sentence on line 11
- Maybe we should add a definition here for `column schema` as that is an important concept to understand to write primitives and understand how they get applied

docs/source/resources/usage_tips/glossary.rst
```diff
@@ -8,38 +8,40 @@ Glossary
...
-        A mapping between a parent entity and a child entity. The child entity must contain a variable referencing the ID variable on the parent entity. Represented by the :class:`.Relationship` class.
+        A mapping between a parent dataframe and a child dataframe. The child dataframe must contain a column referencing the index column on the parent dataframe. Represented by the :class:`.Relationship` class.
+
+    logical type
+        Additional information about how a column should be interpreted or parsed beyond how the data is stored on disk or in memory. Used to determine which primitives can be applied to a column to generate features.
+
+    semantic tag
+        Optional additional information on the column about the meaning or potential uses of data. Used to determine which primitives can be applied to a column to generate features.
+
+    ColumnSchema
```
tamargrey, line 43: Should this be `column schema` to match the format of the other entries? We do usually reference `ColumnSchema`, but in theory the CamelCase version should be saved for when we're talking about the class/object.

frances-h: In my opinion CamelCase is fine here since we never refer to it as "column schema" and we're basically explaining what the class is/represents. For what it's worth, EntitySet was CamelCase in the original file too, so there is some precedence to having entries that break the pattern. 

We could also capitalize each entry which might make the CamelCasing stick out less.

docs/source/resources/usage_tips/glossary.rst
```diff
@@ -8,38 +8,40 @@ Glossary
...
+        All of Woodwork's column type information including the logical type and any semantic tags.
 
 .. todo
 .. label maker,
```
rwedge, line 47: these todo terms are relics that have been "todo" for 4 years, let's remove them

docs/source/resources/usage_tips/glossary.rst
tamargrey: okay, happy to keep as is!

docs/source/resources/usage_tips/glossary.rst
```diff
@@ -32,15 +32,13 @@ Glossary
...
     semantic tag
-        Optional additional information on the column about the meaning or potential uses of data.
+        Optional additional information on the column about the meaning or potential uses of data. Used to determine which primitives can be applied to a column to generate features.
 
-.. todo
-.. label maker,
-.. lag
-..     The amount of time before the prediction time that data is used to make a prediction. This is the time between the beginning and end of a learning window.
+    ColumnSchema
+        All of Woodwork's column type information including the logical type and any semantic tags.
```
thehomebrewnerd, line 44: `Woodwork's column` -> `Woodwork column's`

PR #1618, opened 2021-08-16
Feature selection

simha104:
### Pull Request Description
Updated function calls to accommodate for changed parameter names. Attempting to change the doc so that the feature matrix is woodwork initialized before passing it into the `remove_highly_correlated_features` function

-----


gsheni:
@simha104 can you sign the CLA agreement? https://cla-assistant.io/alteryx/featuretools?pullRequest=1618

docs/source/release_notes.rst
```diff
@@ -24,6 +24,8 @@ Future Release
         * Allow deep EntitySet equality check (:pr:`1480`)
         * Update ``EntitySet.concat`` to work with Woodwork DataFrames (:pr:`1490`)
         * Add function to list semantic tags (:pr:`1486`)
+        * Remove categorical-encoding as an add-on library (:pr:`1632`)
```
thehomebrewnerd, line 27: Unrelated?

docs/source/release_notes.rst
```diff
@@ -24,6 +24,8 @@ Future Release
...
+        * initialize Woodwork on feature matrix in ``remove_highly_correlated_features`` if necessary (:pr:`1618`)
```
thehomebrewnerd, line 28: Not a big deal as I suspect this is all going to be reworked before release anyway, but maybe capitalize `initialize` to be consistent with other entries.

featuretools/selection/selection.py
```diff
@@ -109,11 +109,14 @@ def remove_highly_correlated_features(feature_matrix, features=None, pct_corr_th
                 the feature list will not be returned. For consistent results,
                 do not change the order of features outputted by dfs.
     """
+    if feature_matrix.ww.schema is None:
+        feature_matrix.ww.init()
```
thehomebrewnerd, line 113: Is the intent to remove this after the feature matrix is returned with WW initialized, or do we want this to work regardless?

If we leave it, is it possible that we would end up inferring column types to be different than they would normally be based on the primitive return type? If so, there would be no mechanism for a user to fix, and maybe it doesn't matter...

If we intend to remove, we might want to create an issue to fix this later so we don't forget.

featuretools/selection/selection.py
```diff
@@ -109,11 +109,14 @@ def remove_highly_correlated_features(feature_matrix, features=None, pct_corr_th
...
+
     if pct_corr_threshold < 0 or pct_corr_threshold > 1:
         raise ValueError("pct_corr_threshold must be a float between 0 and 1, inclusive.")
 
     if features_to_check is None:
-        features_to_check = feature_matrix.columns
+        features_to_check = list(feature_matrix.columns)
```
tamargrey, line 119: This change was to accommodate https://github.com/alteryx/woodwork/issues/1107

thehomebrewnerd (approved):
A couple small comments, but looks good.

featuretools/selection/selection.py
tamargrey: I think the idea is that a user might lose the woodwork typing info on their feature matrix at some point (maybe they copied it), and we want to allow them to run feature selection even in that case. @rwedge came up with the idea, I think.

Plus, the Woodwork types that get selected are `['numeric', Boolean, BooleanNullable]`, which aren't likely to have been set by us in cfm and then get reinferred incorrectly (even `Age` columns, which wouldn't get reinferred as `Age` would still get inferred as a numeric column).

It could be a problem, though, if a user has done something to their feature matrix that changes the dtypes, or if, in returning a feature matrix with Woodwork initialized slightly differently than on the original feature matrix, we'll confuse users. Maybe if woodwork wasn't initialized on the fm passed in, we un-init woodwork before returning (not sure that this is possible without a copy operation). 

I do think it makes sense to add a note to the docstring that woodwork will be initialized on the fm passed in iff need be, so if the types differ slightly, that would be why.

docs/source/release_notes.rst
tamargrey: oops copied accidentally in a merge conflict. Thanks!

featuretools/selection/selection.py
tamargrey: @gsheni @jeff-hernandez does this behavior make sense to you?

jeff-hernandez: I think the features matrix will commonly result in numeric, categorical, or encoded categorical types. If a feature is an encoded categorical type and Woodwork sets as Categorical in the feature matrix. Would inferring without typing information change encoded categorical types to numeric?

gsheni: We should make our feature select work with OR without Woodwork typing. If the user hasn't typed their dataframe according to Woodwork (`init`), then our feature selection should do it for them. 

So we should keep the `init` in selection.py

jeff-hernandez:
Not related to the changes in this PR, but it would be nice to change this cell format in the notebook to `reST`.

https://github.com/alteryx/featuretools/blob/fa0337dcf8b6058f13fd91e3cdc753e92d087fe8/docs/source/guides/feature_selection.ipynb#L121-L129

PR #1630, opened 2021-08-19
Use make index to re-create index on new DataFrame in EntitySet.replace_dataframe

### LINKED ISSUE #1484: EntitySet `update_dataframe` does not handle `make_index`
As a Featuretools user, I would like to be able to call `update_dataframe` when `make_index=True` using a DataFrame that does not have the created index.

If a user adds a DataFrame to an EntitySet via `add_dataframe` with `make_index=True`, featuretools will tell Woodwork to create an index column on the DataFrame. 

Woodwork also stores the `make_index` value as a property on the DataFrame accessor in order to make use of it in situations where a user might want to use the same schema for another DataFrame that wouldn't have the created index column on it. This is pretty much the exact situation that `EntitySet.update_dataframe` is built for, and currently, we do not handle the `make_index` case at all.

In fact, if a DataFrame has `make_index=True`, that value gets lost on the resulting DataFrame, because we call `df.ww.init(schem=schema)`, which will always set the `make_index` value to False. 

`EntitySet.update_dataframe` should check the `make_index` value of the original DataFrame, and, if the made index column is not present on the new dataframe, create it in the same way. This will allow users to pass DataFrames without this created index to `update_dataframe`, which is currently not possible.

tamargrey:
This PR does two things:
1. renames `EntitySet.update_dataframe` to `EntitySet.replace_dataframe`
2. Accesses the `created_index` key in the DataFrame's Woodwork metadata to recreate the index column on the new DataFrame (if it's not present) at `replace_dataframe`

closes #1484

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -45,7 +45,7 @@ Breaking Changes
 * ``EntitySet.add_relationship`` has been updated to accept dataframe and column name values or a
   ``Relationship`` object. Adding a relationship from a ``Relationship`` object now requires passing
   the relationship as a keyword argument.
-* ``Entity.update_data`` has been removed. To update the dataframe, call ``EntitySet.update_dataframe`` and use the ``dataframe_name`` paramter.
+* ``Entity.update_data`` has been removed. To update the dataframe, call ``EntitySet.replace_dataframe`` and use the ``dataframe_name`` paramter.
 * The data in an ``EntitySet`` is no longer stored in ``Entity`` objects. Instead, dataframes
   with Woodwork typing information are used. Accordingly, most language referring to “entities”
   will now refer to “dataframes”, references to “variables” will now refer to “columns”, and
@@ -134,11 +134,11 @@ this approach the ``relationship`` parameter name must be included.
 
 **Update DataFrame**
 
-To update a dataframe in an EntitySet, call ``EntitySet.update_dataframe`` and pass in the name of the dataframe to update along with the new data.
+To update a dataframe in an EntitySet, call ``EntitySet.replace_dataframe`` and pass in the name of the dataframe to update along with the new data.
 
 .. code-block:: python
 
-    >>> es.update_dataframe(dataframe_name='log', df=df)
+    >>> es.replace_dataframe(dataframe_name='log', df=df)
 
 **List Logical Types and Semantic Tags**
 
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,7 +7,7 @@ Future Release
 ==============
     * Enhancements
         * Add support for creating EntitySets from Woodwork DataTables (:pr:`1277`)
-        * Add ``EntitySet.__deepcopy__` that retains Woodwork typing information (:pr:`1465`)
+        * Add ``EntitySet.__deepcopy__`` that retains Woodwork typing information (:pr:`1465`)
     * Fixes
     * Changes
         * Remove ``add_interesting_values`` from ``Entity`` (:pr:`1269`)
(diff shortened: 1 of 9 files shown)


featuretools/entityset/entityset.py
```diff
@@ -1433,6 +1434,14 @@ def update_dataframe(self, dataframe_name, df, already_sorted=False, recalculate
             self[dataframe_name].ww.pop(last_time_index_column)
             del self[dataframe_name].ww.metadata['last_time_index']
 
+        # If the original DataFrame had an index created via make_index, we may need to remake the index
+        created_index = self[dataframe_name].ww.metadata.get('created_index')
+        if created_index is not None:
+            if created_index not in df.columns:
+                df = _create_index(df, created_index)
+            else:
+                warnings.warn(f'New DataFrame, {dataframe_name}, already has the created index column, {created_index}. ')
```
tamargrey, line 1443: Decided to allow users to pass in a DataFrame that has the index column present. Figure this could be important for instances where someone is making changes to the DataFrame that's currently in the EntitySet, since they wouldn't necessarily want their data re-indexed. I'm not sure that this warning is actually necessary, as it's warning on the lack of action being taken. 

Happy to change to an error, if this behavior seems confusing or unexpected, since we do raise an error at `add_dataframe` if `make_index` is true but the `index` column is already in the DataFrame.

tamargrey: If we do make this error, we'll have to update `EntitySet.concat` which calls `replace_dataframe` on the concated dataframes, and should probably keep whatever index was created from the concat call and not reindex there.

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -459,6 +459,49 @@ def test_make_index_any_location(df):
...
+
+    assert 'id' in has_idx_df.columns
+
+    original_idx_col = [100, 1, 2]
+
+    warning = "New DataFrame, test_df, already has the created index column, id."
+    with pytest.warns(UserWarning, match=warning):
+        es.replace_dataframe('test_df', has_idx_df)
+    assert es['test_df'].ww.index == 'id'
+    assert all(original_idx_col == to_pandas(es['test_df']['id']))
```
thehomebrewnerd, line 502: This approach looks fine, but I think we should split this into a separate test since it is testing different behavior. Will keep the individual tests smaller and easier to understand.

thehomebrewnerd (commented):
Just a couple small things, but generally looks fine.

featuretools/entityset/entityset.py
tamargrey: Removing the warning makes sense to me, as I did feel weird about raising a warning on our lack of action.

docs/source/release_notes.rst
```diff
@@ -44,7 +44,7 @@ Breaking Changes
 * ``EntitySet.add_relationship`` has been updated to accept dataframe and column name values or a
   ``Relationship`` object. Adding a relationship from a ``Relationship`` object now requires passing
   the relationship as a keyword argument.
-* ``Entity.update_data`` has been removed. To update the dataframe, call ``EntitySet.update_dataframe`` and use the ``dataframe_name`` paramter.
+* ``Entity.update_data`` has been removed. To update the dataframe, call ``EntitySet.replace_dataframe`` and use the ``dataframe_name`` paramter.
```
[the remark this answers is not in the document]
tamargrey: changed!

featuretools/tests/entityset_tests/test_es.py
tamargrey: splitting up makes sense to me!

thehomebrewnerd (approved):
LGTM!

PR #1640, opened 2021-08-24
Revert changes to Equal and NotEqual primitives

thehomebrewnerd:
### Revert changes to Equal and NotEqual primitives

Earlier updates were made to the `Equal` and `NotEqual` primitives to accommodate comparisons of categorical columns with different orderings. After reverting these changes for troubleshooting, the original error that prompted the update cannot be reproduced. This PR reverts those primitive changes until the error is reproducible.


rwedge (approved):
LGTM

PR #1644, opened 2021-08-25
Remove add on libraries from install docs, setup.py and CI tests

gsheni:
In anticipation of Featuretools 1.0.0rc1 being released, removing the add-on libraries as they will be supported right away.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -11,6 +11,4 @@ nbconvert==6.0.6
 nbsphinx==0.8.5
 Sphinx==3.2.1
 pydata_sphinx_theme==0.4.0
-nlp_primitives>=1.0.0
-autonormalize >= 1.0.1
 -r koalas-requirements.txt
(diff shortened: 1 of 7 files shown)


dev-requirements.txt
```diff
@@ -11,6 +11,4 @@ nbconvert==6.0.6
 nbsphinx==0.8.5
 Sphinx==3.2.1
 pydata_sphinx_theme==0.4.0
-nlp_primitives>=1.0.0
```
thehomebrewnerd, line 14: I think we can remove the `featuretools-sklearn-transformer` requirement at the top of this file as well.

thehomebrewnerd (approved):
One minor thing I think we can clean up

docs/source/api_reference.rst
```diff
@@ -32,14 +32,6 @@ Wrappers
 ~~~~~~~~
 .. currentmodule:: featuretools
 
-Scikit-learn (BETA)
```
thehomebrewnerd, line 35: We can also remove the four lines above this to get rid of the `Wrappers` section

dev-requirements.txt
gsheni: Fixed

docs/source/api_reference.rst
gsheni: Fixed

PR #1662, opened 2021-08-30
Update primitives loading to be more robust

davesque:
* Emit warning when primitives entry point throws error
* Prevent overwriting of names in primitive namespace
* Avoid loading non-class objects and the `AggregationPrimitive` and `TransformPrimitive` base classes

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -8,13 +8,21 @@ Future Release
     * Fixes
     * Changes
         * Remove autonormalize, tsfresh, nlp_primitives, sklearn_transformer, caegorical_encoding as an add-on libraries (will be added back later) (:pr:`1644`)
+        * Emit a warning message when a ``featuretools_primitives`` entrypoint
+          throws an exception (:pr:`1662`)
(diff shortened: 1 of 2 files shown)


davesque:
There didn't appear to be any tests already for the `featuretools_primitives` entrypoint namespace so I didn't bother adding any.  If we think we need that, we can add it.  This was just a sort of incidental work item that came up during the other stuff I've been involved in.  Didn't want to rabbit hole too hard on it.

gsheni:
@davesque I think we avoided testing the entry_point because it would have been too much work to create a fake package, and tests that the entry point would work. We can create an issue to address this in the future, but it hasn't been a problem so far.

davesque:
@gsheni Yeah, I'm having a hard time even imagining an elegant way to test this without doing a lot of crazy monkey patching.

thehomebrewnerd (approved):
I'm not super well-versed in the way this entry point stuff works, so take this with a grain of salt, but nothing here jumps out to me as a concern.

gsheni (approved):

jeff-hernandez (approved):

Issue #1559, opened 2021-07-26
Use feature origin attribute from ColumnSchema when generating feature matrix


gsheni:
- Woodwork v0.5.0 added a feature origin attribute
  - https://github.com/alteryx/woodwork/pull/1056
- We should use when after we generate the feature matrix to specify which features were **engineered** and which features were **source**
- You can use the feature definitions to specify the **origin**. 

- We should also update the documentation (Woodwork Typing in Featuretools guide) to specify where this information, what it means and how users can access it.

tamargrey:
Should seed features get their own origin value?

rwedge:
I think they can be seen as user-specified engineered features, I don't think they need a separate value

PR #1664, opened 2021-08-31
Calculate feature matrix returns woodwork dataframe

### LINKED ISSUE #1393: Calculate Feature Matrix return a Woodwork typed DataFrame
- Our DFS/calculate_feature_matrix call can return a Woodwork typed DataFrame.
- This Woodwork typed DataFrame would not need to be validated 
  - (`validation=False, already_sorted=True` during `df.ww.init`).
- The order of the DataFrame returned should be same as the cutoff_time dataframe (if provided) (this is the current behavior).
- We should use the ColumnSchema on primitives to init the **feature_matrix**.

rwedge:
Fixes #1393, #1559


featuretools/computational_backends/utils.py
```diff
@@ -291,3 +292,43 @@ def replace_inf_values(feature_matrix, replacement_value=np.nan, columns=None):
     else:
         feature_matrix[columns] = feature_matrix[columns].replace([np.inf, -np.inf], replacement_value)
     return feature_matrix
+
+
+def get_ww_types_from_features(features, entityset, pass_columns, cutoff_time):
+    """ TODO """
```
thehomebrewnerd, line 298: Should add a docstring here...or delete this. Ideally we would have a brief docstring though.

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -215,7 +215,7 @@ class Week(TransformPrimitive):
         """
     name = "week"
     input_types = [ColumnSchema(logical_type=Datetime)]
-    return_type = ColumnSchema(logical_type=Ordinal(order=range(1, 53)), semantic_tags={'category'})
+    return_type = ColumnSchema(logical_type=Ordinal(order=list(range(1, 53))), semantic_tags={'category'})
```
thehomebrewnerd, line 218: Should this be `range(1, 54)` here?  I think some years can technically have 53 weeks depending on how the calendar falls.

featuretools/computational_backends/utils.py
```diff
@@ -291,3 +292,52 @@ def replace_inf_values(feature_matrix, replacement_value=np.nan, columns=None):
...
+            if logical_types[name] is None and "numeric" in semantic_tags[name]:
+                logical_types[name] = Double
+        if all([f.primitive is None for f in feature.get_dependencies(deep=True)]):
+            origins[name] = "base"
+        else:
+            origins[name] = "engineered"
+
+    for column in pass_columns:
+        logical_types[column] = cutoff_time.ww[column].ww.logical_type
+        semantic_tags[column] = cutoff_time.ww[column].ww.semantic_tags.copy()
```
tamargrey, line 326: a slightly better way to get the logical type and semantic tags here will be to go through the logical type and semantic tag attributes:

```python
logical_types[column] = cutoff_time.ww.logical_types[column]
semantic_tags[column] = cutoff_time.ww.semantic_tags[column].copy()
```

This will avoid creating a series and just access the info directly

featuretools/computational_backends/utils.py
```diff
@@ -291,3 +292,52 @@ def replace_inf_values(feature_matrix, replacement_value=np.nan, columns=None):
...
+        origins[column] = "base"
+
+    if entityset.dataframe_type in (Library.DASK.value, Library.KOALAS.value):
+        target_dataframe_name = features[0].dataframe_name
+        table_accessor = entityset[target_dataframe_name].ww
```
tamargrey, line 331: I don't think there's anything wrong with using the accessor like this, so this is just a suggestion for slightly different syntax, but you could store the table schema as a variable, and that might be more in line with how woodwork expects users to access typing info

```python
table_schema = entityset[target_dataframe_name].ww.schema
index_col = table_schema.index
logical_types[index_col] = table_schema.logical_types[index_col]
semantic_tags[index_col] = table_schema.semantic_tags[index_col] # no need to copy, because the .schema already copied

```

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -304,15 +308,15 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
...
                     pd.MultiIndex.from_frame(cutoff_time[["instance_id", "time"]],
                                              names=feature_matrix.index.names))
             else:
                 # Maintain index dtype
                 index_dtype = feature_matrix.index.get_level_values(0).dtype
-                feature_matrix = feature_matrix.reindex(cutoff_time[1].astype(index_dtype), level=0)
+                feature_matrix = feature_matrix.ww.reindex(cutoff_time[1].astype(index_dtype), level=0)
             if not cutoff_time_in_index:
-                feature_matrix.reset_index(level='time', drop=True, inplace=True)
+                feature_matrix.ww.reset_index(level='time', drop=True, inplace=True)
```
tamargrey, line 319: is it ever possible that any of these reindexing could invalidate the schema by making the underlying index not match the woodwork index column?

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -751,3 +751,20 @@ def update_progress_callback_parameters(progress_bar, previous_progress):
...
+        fm.ww.init(**ww_init_kwargs)
+
+    fm = feature_matrix[0]
+
+    for other in feature_matrix[1:]:
+        fm = fm.ww.append(other)
+        if not fm.ww.schema:
+            # Koalas bug sometimes causes schema to be invalidated
+            # after `append` so need to reinitialize WW
+            fm.ww.init(**ww_init_kwargs)
```
tamargrey, line 767: another option that might handle this case is to do the `append` call outside of Woodwork and then reinit with the schema. Then you wouldn't get the `TypingInfoMismatchWarning`.

```python
schema = fm.ww.schema
fm = fm.append(other) #or can use pd.concat now that you don't need to go through woodwork accessor
fm.ww.init(schema=schema)
```
This will make it so that the init call can update any dtypes or underlying index (since it'll use the partial schema init unlike the `fm.ww.append` call which goes through the full schema init, which requires the schema be valid). The downside would be if you wanted to be able to catch that `not fm.ww.schema` case so that you can re-init in the exact way you are now, you'd lose that control.

tamargrey: Though, this is also a point in the column of adding `ww.concat_rows` or updating `ww.concat_cols` to be a more general concat.

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -320,7 +320,7 @@ class Weekday(TransformPrimitive):
     """
     name = "weekday"
     input_types = [ColumnSchema(logical_type=Datetime)]
-    return_type = ColumnSchema(logical_type=Ordinal(order=range(7)), semantic_tags={'category'})
+    return_type = ColumnSchema(logical_type=Ordinal(order=list(range(7))), semantic_tags={'category'})
```
tamargrey, line 323: Is it a good rule of thumb that if a primitive's return type is `Ordinal` you need to include the `order` parameter? I notice that Ordinals in input types don't have orders

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -186,7 +186,7 @@ class Minute(TransformPrimitive):
     """
     name = "minute"
     input_types = [ColumnSchema(logical_type=Datetime)]
-    return_type = ColumnSchema(logical_type=Ordinal(range(59)), semantic_tags={'category'})
+    return_type = ColumnSchema(logical_type=Ordinal(list(range(59))), semantic_tags={'category'})
```
tamargrey, line 189: for consistency's sake, this and line 165 (the Second primitive) should probably specify the `order=` for the parameter passed in

tamargrey (commented):
A couple of comments about woodwork syntax!

featuretools/primitives/standard/transform_primitive.py
rwedge: If the return type `Ordinal` doesn't have the `order` parameter, Woodwork will error when trying to initialize on the feature matrix

I added an order to the `Year` primitive but it's very arbitrary and is just there to avoid an error

featuretools/computational_backends/calculate_feature_matrix.py
rwedge: @thehomebrewnerd didn't you come acrosss something like this earlier today?

thehomebrewnerd: I did, when I was attempting to update `_validate_cutoff_time` to retain typing information if the user already initialized WW. There is a reset index call in that function as well. It does cause the WW schema to be invalidated in some cases, which was causing at least one test failure.

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -751,3 +751,18 @@ def update_progress_callback_parameters(progress_bar, previous_progress):
     progress_percent = (progress_bar.n / progress_bar.total) * 100
     time_elapsed = progress_bar.format_dict["elapsed"]
     return (update, progress_percent, time_elapsed)
+
+
+def init_ww_and_concat_fm(feature_matrix, ww_init_kwargs):
+    for fm in feature_matrix:
+        fm.ww.init(**ww_init_kwargs)
```
jeff-hernandez, line 758: We also initialize the feature matrix after concat. Is this init still necessary?

rwedge: Those inits are to ensure each smaller matrix has compatible types for when we do the larger concat

as an example, there could be what's supposed to be a `BooleanNullable` column and the smaller matrices have these values:

series 1: [True, False], bool dtype
series 2: [1.0, np.nan], float dtype
seires 3: [np.nan, np.nan], float dtype

and after the concat we get

series: [True, False, 1.0, np.nan, np.nan, np.nan], object dtype

and when try to cast to  `BooleanNullable` pandas is unable to convert the type

featuretools/computational_backends/utils.py
```diff
@@ -291,3 +292,54 @@ def replace_inf_values(feature_matrix, replacement_value=np.nan, columns=None):
...
+
+def get_ww_types_from_features(features, entityset, pass_columns=None, cutoff_time=None):
+    '''Given a list of features and entityset (and optionally a list of pass
+    through columns and the cutoff time dataframe), returns the logical types,
+    semantic tags,and origin of each column in the feature matrix.  Both
+    pass_columns and cutoff_time will need to be supplied in order to get the
+    type information for the pass through columns
+    '''
+    if pass_columns is None:
+        pass_columns = []
```
jeff-hernandez, line 305: This line doesn't seem to be covered by tests. Do we want to test this function with pass_columns as None?

rwedge: It should have been covered but I forgot to update `encode_features.py` to use the default values

rwedge:
There was an issue building the docs due to an example where a "labels" column in the entityset was inferred as Boolean in the training dataset and Unknown in the testing dataset, so when doing the feature matrix typing it couldn't convert the NaN values to Boolean.

I didn't like that the example would allow DFS to stack features on top of the label column so I updated the example so DFS was configured to include the label column but not allow stacking on it, and then removing the labels feature completely when calculating the features using the test dataset.

Perhaps a simpler solution would be to set the logical type for the label column to BooleanNullable and calculate the feature on both dataset, but I think generally you wouldn't need the label column for the test features.

featuretools/computational_backends/calculate_feature_matrix.py
rwedge: The feature matrix does not set a woodwork index so this shouldn't be an issue I believe

tamargrey (approved):
lgtm!

rwedge:
> Looks good. I think there are a couple more tests that we could fix now if we want to, or we can do this later as part of #1596.
> 
> Based on my local testing it seems we can now remove the `pytest.skip` from `test_concat_with_lti`.
> 
> I think we also might be able to fix `test_single_table_ks_entityset_cutoff_time_df` in `test_koalas_dfs.py` if we set the logical type of `values` to be `IntegerNullable` instead of `Integer` and then fix the categoricals before comparison like we do in `test_ks_entityset_secondary_time_index`.

Testing the `test_single_table_ks` change locally now, will probably include both

docs/source/resources/frequently_asked_questions.ipynb
```diff
@@ -958,8 +961,8 @@
     "                           \"gender\": [\"f\", \"m\", \"m\", \"m\", \"m\"],\n",
     "                           \"signup_date\": pd.date_range('2014-01-01 01:41:50', periods=5, freq='25min')})\n",
     "\n",
-    "# lets add NaN label column to the test Dataframe\n",
-    "test_train['labels'] = np.nan\n",
+    "# lets remove the labels feature\n",
+    "test_features = [feature for feature in features_enc if feature.get_name() != \"labels\"]\n",
```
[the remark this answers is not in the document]
rwedge: Something like this?

> Now we can use the feature definitions to calculate our feature matrix for the test data, and avoid label leakage.  Since the test data does not include labels, we will remove the labels feature from the list of encoded features before calculating on the test data.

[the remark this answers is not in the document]
jeff-hernandez: That works, but maybe we say something like:

> Since we do not want the test data to include the labels, let's remove the labels feature from the list of encoded features before calculating on the test data.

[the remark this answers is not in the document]
jeff-hernandez: Thinking about this a bit more, how do we want users to get the labels associated with the test data?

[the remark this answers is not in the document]
jeff-hernandez: Should users include the labels feature, then remove it after for scoring?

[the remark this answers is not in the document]
rwedge: Yeah we could that, but in that case we should use actual boolean values for the column instead of NaNs.

Normally we would recommend adding the labels to the cutoff time dataframe and they'd be passed through that way, but this is a scenario with no cutoff times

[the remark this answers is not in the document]
jeff-hernandez: Cool, I think it'd be helpful to reflect how users would go about separating the associated labels from the test data.

jeff-hernandez (approved):
Looks good!

PR #1672, opened 2021-09-03
Fix typos in transition guide

tamargrey:


thehomebrewnerd (approved):
Looks good, after all tests pass - still needs a release notes update.

tamargrey:
My local build doesn't have any spaces in `set_types` or `dfs(target_dataframe_name=‘dataframe_name’, …)` (though I didn't change anything)
image

The build the docs version here still has the spaces at set_types and the dfs call:
image

@thehomebrewnerd @gsheni is this worth pursuing further to try and fix the spaces?

thehomebrewnerd:
@tamargrey I think that is the same issue we were seeing before with the table formatting getting messed up...seems to be happening only on readthedocs, which maybe means the build environment there is different than our local environments somehow? I never saw these issues with local builds either.

Anyway, I think you found some stack overflow discussion about something similar. If there is some easy fix we can find, it might be worth resolving, but I think it's probably not that critical to clean up relative to some other issues.

tamargrey:
> Anyway, I think you found some stack overflow discussion about something similar. If there is some easy fix we can find, it might be worth resolving, but I think it's probably not that critical to clean up relative to some other issues.

Yeah, I'll look into that for a bit, but if I can't find something by end of day, probably not worth drawing it out longer

PR #1675, opened 2021-09-07
Fix foreign_key tag bug

### LINKED ISSUE #1674: Bug in EntitySet.add_relationship can cause `foreign_key` tag to be lost
### Bug in EntitySet.add_relationship can cause `foreign_key` tag to be lost

In `EntitySet.add_relationship`, the `foreign_key` semantic tag is added to the child column in a relationship. There is also a check to make sure that the logical type of the child column matches the logical type of the parent column and if not, the logical type for the child column is updated to match the parent.

The `foreign_key` tag is added before the logical type check. As a result, when the logical type is changed, the previously set foreign key tag is lost.

The code should be updated to add the `foreign_key` tag at the end of the method, after the logical type has been changed.

thehomebrewnerd:
### Fix foreign_key tag bug

Closes #1674 

Updates `EntitySet.add_relationship` to make sure `foreign_key` tags are retained when changing child column logical type.


rwedge (approved):
LGTM

PR #1678, opened 2021-09-07
Remove old categorical code - woodwork-integration branch update

thehomebrewnerd:
### Remove old categorical code - woodwork-integration branch update

Removes old code related to categorical columns. The code was implemented as a workaround for a pandas issue which has since been closed.


rwedge:
I think we should go about including this as a merge from main instead

PR #1679, opened 2021-09-07
Merge code from main removing old code

thehomebrewnerd:
### Merge code from main removing old code

Merges in changes from main which removed workaround code that is no longer necessary.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,6 +6,18 @@ Release Notes
 Future Release
 ==============
     * Enhancements
+    * Fixes
+    * Changes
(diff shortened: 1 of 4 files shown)


thehomebrewnerd:
> To make the merge less messy, what if we title woodwork integrations future release as "1.0.0.dev" or something and put the featuretools main Future Release section below it. Otherwise we're gonna have to pull out parts of the future release section if there's another release of main

Sure, I can make that update. I am also not sure what is up with the lint tests, so I'll have to investigate. It's passing locally for me.

rwedge:
I think Future Release should go below 1.0.0.dev because this is to account for the edge case where Future Release happens before 1.0.0

PR #1689, opened 2021-09-10
Update WW version to 0.8.0

thehomebrewnerd:
### Update WW version to 0.8.0

Update WW requirement to `>=0.8.0` to take advantage of performance improvement in latest version.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -54,6 +54,7 @@ Release Notes
         * Remove remaining references to ``Entity`` and ``Variable`` in code (:pr:`1612`)
         * Update Woodwork to version 0.7.1 with changed initialization (:pr:`1648`)
         * Remove unused ``_dataframes_equal`` and ``camel_to_snake`` functions (:pr:`1683`)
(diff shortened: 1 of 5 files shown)


tamargrey:
Latest dependency tests are passing, but minimum dependency tests arent--does the woodwork version need to be updated in those files as well?

thehomebrewnerd:
> Latest dependency tests are passing, but minimum dependency tests arent--does the woodwork version need to be updated in those files (minimum requirements text files) as well?

Maybe so - I thought those other requirements files were supposed to be auto-generated but I haven't really looked closely so might have misunderstood how that works.

tamargrey:
> Maybe so - I thought those other requirements files were supposed to be auto-generated but I haven't really looked closely so might have misunderstood how that works.

Maybe they just get updated automatically on the MachineFL PRs? Seems like there's logic in the minimum dependency checker for that. So if the automated dependency updates had already happened, then the minimum requirements would have been updated?

tamargrey (approved):
lgtm once the last tests pass!

PR #1699, opened 2021-09-16
Encode Features - prefer runtime over space if not inplace

rwedge:
* Remove tqdm progress bar for step that was removed previously
* Speed up adding encoded columns to the encoded matrix by adding them all at once if `inplace=False` 
* Skip dropping the original columns if `inplace=False`, omit them from the concat instead

### DIFF
--- a/featuretools/synthesis/encode_features.py
+++ b/featuretools/synthesis/encode_features.py
@@ -103,7 +103,7 @@ def encode_features(feature_matrix, features, top_n=DEFAULT_TOP_N, include_unkno
 
     new_feature_list = []
     new_columns = []
-    encoded_columns = set()
+    encoded_columns = []
 
     for f in iterator:
         # TODO: features with multiple columns are not encoded by this method,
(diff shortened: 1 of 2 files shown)


featuretools/synthesis/encode_features.py
```diff
@@ -147,30 +147,38 @@ def encode_features(feature_matrix, features, top_n=DEFAULT_TOP_N, include_unkno
...
     ww_init_kwargs = get_ww_types_from_features(new_feature_list, entityset)
-    new_X.ww.init(**ww_init_kwargs)
-    return new_X, new_feature_list
+
+    # Grab ww metadata from feature matrix since it may be more exact
+    for column in kept_columns:
+        column_info = feature_matrix.ww.columns[column]
+        ww_init_kwargs["logical_types"][column] = column_info.logical_type
+        ww_init_kwargs["semantic_tags"][column] = column_info.semantic_tags
+        ww_init_kwargs["column_origins"][column] = column_info.origin
```
rwedge, line 181: @tamargrey any issues with this syntax?

featuretools/synthesis/encode_features.py
```diff
@@ -147,30 +147,38 @@ def encode_features(feature_matrix, features, top_n=DEFAULT_TOP_N, include_unkno
...
+        X = pd.concat([X[kept_columns]] + encoded_columns, axis=1)
 
     entityset = new_feature_list[0].entityset
     ww_init_kwargs = get_ww_types_from_features(new_feature_list, entityset)
-    new_X.ww.init(**ww_init_kwargs)
-    return new_X, new_feature_list
+
+    # Grab ww metadata from feature matrix since it may be more exact
+    for column in kept_columns:
+        column_info = feature_matrix.ww.columns[column]
```
thehomebrewnerd, line 178: I don't think there is any weird looping inside Woodwork happening when we call `feature_matrix.ww.columns` that would cause major performance problems, but to be safe you could get the column dict outside the loop instead hitting the WW accessor each time inside the loop. The accessor does forward this call to the schema object each time, so there might be a slight bit of overhead there in getting the columns dict.

thehomebrewnerd (approved):
Looks good - just one minor suggestion

featuretools/synthesis/encode_features.py
tamargrey: looks good! I like nate's idea of creating the `feature_matrix.ww.columns` object once outside of the loop!

tamargrey (approved):
lgtm!

jeff-hernandez (approved):

PR #1719, opened 2021-09-23
Add notebook cleaner

bchen1116:
Remove executed output of notebooks

git status

Example changes:
Pasted Graphic 10
diff --git adocssourceindex ipynb bdocssourceindex ipynb

Lint:
make lint

Lint-fix:
make lint-fix

Remaining diff:
index 3fd8d6d 0dla8ab 100644

**NOTE**: We see that the python version changes to my local version, and the `id` values for the cells are removed. The removal of IDs don't seem to be detrimental for anything, and there is no set python version, so these changes should be fine.

Added support of removing last cell if it is empty, which can be seen [here](https://github.com/alteryx/woodwork/pull/1153#issuecomment-926084592)

### DIFF
--- a/docs/notebook_cleaner.py
+++ b/docs/notebook_cleaner.py
@@ -1,6 +1,7 @@
-import click
 import json
 import os
+
+import click
 
 DOCS_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)), "source")
 
--- a/docs/notebook_cleaner.py
+++ b/docs/notebook_cleaner.py
@@ -20,15 +20,16 @@ def _get_ipython_notebooks(docs_source):
 def _check_delete_empty_cell(notebook, delete=True):
     with open(notebook, "r") as f:
         source = json.load(f)
-        cell = source["cells"][-1]
-        if cell["cell_type"] == "code" and cell["source"] == []:
-            # this is an empty cell, which we should delete
-            if delete:
-                source["cells"] = source["cells"][:-1]
-            else:
-                return False
+    cell = source["cells"][-1]
+    if cell["cell_type"] == "code" and cell["source"] == []:
+        # this is an empty cell, which we should delete
+        if delete:
+            source["cells"] = source["cells"][:-1]
+        else:
+            return False
     if delete:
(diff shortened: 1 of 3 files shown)


docs/notebook_cleaner.py
```diff
@@ -0,0 +1,86 @@
...
+def _fix_execution_and_output(notebook):
+    with open(notebook, "r") as f:
+        source = json.load(f)
+        for cells in source["cells"]:
+            if cells["cell_type"] == "code" and cells["execution_count"] != None:
+                cells["execution_count"] = None
+                cells["outputs"] = []
+        source["metadata"]["kernelspec"]["display_name"] = "Python 3"
+        source["metadata"]["kernelspec"]["name"] = "python3"
+    json.dump(source, open(notebook, "w"), ensure_ascii=False, indent=1)
```
rwedge, line 54: can we use a `with open` context for this dump statement?

docs/notebook_cleaner.py
```diff
@@ -0,0 +1,131 @@
...
+    else:
+        return True
+
+
+def _check_execution_and_output(notebook):
+    with open(notebook, "r") as f:
+        source = json.load(f)
+        for cells in source["cells"]:
+            if cells["cell_type"] == "code" and (cells["execution_count"] is not None or cells['outputs']!= []):
+                return False
```
rwedge, line 41: this code block doesn't need to happen within the file loading block

docs/notebook_cleaner.py
```diff
@@ -0,0 +1,131 @@
...
+
+def _fix_execution_and_output(notebook):
+    with open(notebook, "r") as f:
+        source = json.load(f)
+        for cells in source["cells"]:
+            if cells["cell_type"] == "code" and cells["execution_count"] is not None:
+                cells["execution_count"] = None
+                cells["outputs"] = []
+        source["metadata"]["kernelspec"]["display_name"] = "Python 3"
+        source["metadata"]["kernelspec"]["name"] = "python3"
```
rwedge, line 53: same as above, can move this outside the context manager

rwedge (approved):
LGTM

PR #1726, opened 2021-09-28
Add primitive init tests

### LINKED ISSUE #1717: Add test to verify all primitives can be initialized without passing parameter values
## Add test verify all primitives can be initialized without passing parameter values

We should add a test to make sure that all standard primitives can be initialized without passing any parameter values. If initialization is not possible, running  `get_valid_primitives` will result in an error. Adding a test will make sure any new standard primitives are created with default values specified for all parameters.

thehomebrewnerd:
### Add primitive init tests

Closes #1717 

Adds explicit tests to verify all primitives can be initialized without specifying any parameter values.


rwedge (approved):
LGTM

PR #1733, opened 2021-10-11
Fix get_unused_primitives only recognizes lowercase primitive strings

### LINKED ISSUE #1729: get_unused_primitives only recognizes lowercase primitive strings
When running dfs with passed in lists of primitives that are strings, any snake_case string can be used ("Month" vs "month" vs "MONTH" for example), however the only primitive that will be recognized in `get_unused_primitives` as having been used in dfs is the all lowercase string ("month"). This means that any primitive string  that is not all lowercase will show up in the `UnusedPrimitiveWarning` even if the primitive was actually used in dfs and shows up in the feature matrix. 

We should update `get_unsused_primitives` to cast all the strings to lowercase before the `sorted(list(specified.difference(used)))` call that is currently the source of this bug.

```python
from featuretools.tests.testing_utils import (
    make_ecommerce_entityset,
)
import featuretools as ft

fm, f = ft.dfs(entityset=es, target_dataframe_name='customers', agg_primitives=['Count'])
fm
```

The above dfs run will raise the following warning: `UnusedPrimitiveWarning: Some specified primitives were not used during DFS: agg_primitives: ['Count']`

The below dfs run will not:

```python
fm, f = ft.dfs(entityset=es, target_dataframe_name='customers', agg_primitives=['count'])
fm
```

HenryRocha:
Fixes #1729

Cast specified primitives to lowercase strings in `get_unused_primitives`.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -8,7 +8,7 @@ Future Release
     * Enhancements
     * Fixes
         * Pin upper bound for pyspark (:pr:`1748`)
-        * Fix `get_unused_primitives` only recognizes lowercase primitive strings (:pr:`1733`)
+        * Fix ``get_unused_primitives`` only recognizes lowercase primitive strings (:pr:`1733`)
     * Changes
         * Add new version of nlp_primitives as an add-on library (:pr:`1743`)
     * Documentation Changes


tamargrey:
Thanks for the contribution @HenryRocha! This fix makes sense to me. Would you be able to add a test for the behavior being fixed here in `test_dfs_method.py`? There's a couple of tests there that test the unused primitives warnings like `test_warns_with_unused_primitives` that will be good templates for this test.

thehomebrewnerd:
@davesque You raise some interesting things to think about. Here are a couple of my thoughts:

- I could be wrong, but I think DFS expects snake case strings as input to the primitives lists, so if you supplied something like `"CumCount"` you would get an unknown primitive error since that string doesn't match any primitive names. The input doesn't seem to be strictly snake case though because `"Cum_Count"` does get recognized as a valid primitive.
- An alternate solution could be to require that the primitive names supplied as strings match the primitive name exactly and we don't do any lowercasing of the DFS inputs like we seem to be doing now before we match them to primitive names.
- I do think this problem stems from the flexibility we allow in the inputs. However, as a user I definitely like having the ability to specify primitives by string rather than by class objects or instances. If we only allowed class inputs we would have to add extra imports for those classes, which is inconvenient from a usability point of view in my opinion.

rwedge:
The "variable types" typing system used before we started using Woodwork originally had a "name" attribute that was defined separately from the class name, but we changed it to be an automatically generated snake_case version of the class name 

Switching to snake case for the string name probably added some unnecessary potential for user error

rwedge:
> @rwedge Cool. Useful context. Does having the separate snake case name seem like something that might still be required for some reason?

@davesque - I don't think it'd be required but it could break existing user workflows if we don't recognize the snake case names that were used previously

Issue #1739, opened 2021-10-14
Difference between attribute of stack_on and base_of in class AggregationPrimitive


c787297017:
What's the difference between attribute of stack_on(stack_on_exclude) and base_of(base_of_exclude) in class AggregationPrimitive? 

I try to put a custom primitive that is for transformation into the list of stack_on_exclude or base_of_exclude in a primitive class inherited from primitive class Sum(standard primitive in featuretools), because I don't want to see the result of my custom primitive is aggregated by primitive sum, but it doesn't work.

Looking forward for your reply and many thanks for your work.

jeff-hernandez:
@c787297017  thanks for the question! `stack_on_exclude` will exclude the given primitives as inputs, whereas `base_of_exclude` will exclude itself as an input to the given primitives. For example, if you run this code you'll get the listed feature definitions.

```python
from woodwork.column_schema import ColumnSchema
import featuretools as ft
import woodwork as ww


class Abs(ft.primitives.TransformPrimitive):
    input_types = [ColumnSchema(semantic_tags={'numeric'})]
    return_type = ColumnSchema(semantic_tags={'numeric'})
    name = 'abs'

    def get_function(self):
        return abs


class Sum(ft.primitives.AggregationPrimitive):
    input_types = [ColumnSchema(semantic_tags={'numeric'})]
    return_type = ColumnSchema(semantic_tags={'numeric'})
    name = 'sum'

    def get_function(self):
        return sum


if __name__ == '__main__':
    es = ft.demo.load_mock_customer(return_entityset=True)

    fd = ft.dfs(
        entityset=es,
        features_only=True,
        agg_primitives=[Sum],
        trans_primitives=[Abs],
        target_dataframe_name='sessions',
    )
```
```
[<Feature: customer_id>,
 <Feature: device>,
 <Feature: SUM(transactions.amount)>,
 <Feature: customers.zip_code>,
 <Feature: SUM(transactions.ABS(amount))>,
 <Feature: ABS(SUM(transactions.amount))>,
 <Feature: customers.SUM(transactions.amount)>]
```

You can avoid generating `<Feature: SUM(transactions.ABS(amount))>` by using `base_of_exclude` to exclude `Abs` from being an input to `Sum`.

```python
class Abs(ft.primitives.TransformPrimitive):
    ...
    base_of_exclude = [Sum]
```

Alternatively, you can use `stack_on_exclude` to exclude `Abs` as an input primitive.

```python

[...]

PR #1749, opened 2021-10-18
Check `base_of_exclude` attribute on Primitive

jeff-hernandez:
Currently, we check the `base_of_exclude` attribute on  Feature classes instead of Primitive classes which raises an error.

#### Example

```python
from woodwork.column_schema import ColumnSchema
import featuretools as ft
import woodwork as ww


class Abs(ft.primitives.TransformPrimitive):
    input_types = [ColumnSchema(semantic_tags={'numeric'})]
    name = 'abs'

    def get_function(self):
        return abs

class Sum(ft.primitives.AggregationPrimitive):
    input_types = [ColumnSchema(semantic_tags={'numeric'})]
    base_of_exclude = [Abs]
    name = 'sum'

    def get_function(self):
        return sum

fd = ft.dfs(
    entityset=es,
    features_only=True,
    agg_primitives=[Sum],
    target_dataframe_name='customers',
)
```
```
AttributeError: 'AggregationFeature' object has no attribute 'base_of_exclude'
```

Related to #1739


featuretools/tests/primitive_tests/test_agg_feats.py
```diff
@@ -200,6 +200,11 @@ def test_base_of_and_stack_on_heuristic(es, test_primitive):
     child.primitive.base_of = [test_primitive]
     assert check_stacking(test_primitive(), [child])
 
+    test_primitive.stack_on = []
+    child.primitive.base_of = []
+    child.primitive.base_of_exclude = [test_primitive]
+    assert not check_stacking(test_primitive(), [child])
```
rwedge, line 206: if `stack_on` and `base_of` are both set to `[]`. `check_stacking` will return False even if `test_primitive` is not in `base_of_exclude`

jeff-hernandez: Oh okay, interesting. I changed `stack_on` and `base_of` to None so they can return True. It kind of seems like whether `None` or `[]` they should return the same value.

rwedge: I think `None` means no whitelist while `[]` means an empty whitelist, so nothing allowed

rwedge (approved):
LGTM

PR #1765, opened 2021-10-29
Fix pass through columns for parallel pandas computation

### LINKED ISSUE #1763: Bug with parallel chunk calculation and cutoff time df
#### Bug Description
In my case I used dfs with cutoff time as df and n_jobs == -1
In the previous featuretools version everything worked fine. But after v.1.0.0 release I have found out that dfs breaks when running it with cutoff time which contains extra columns (pass_columns) and n_jobs > 1. The error is the following
`AttributeError: 'NoneType' object has no attribute 'logical_types'`

It happens in the following function `get_ww_types_from_features`
In the following lines
```
    if pass_columns:
        cutoff_schema = cutoff_time.ww.schema
        for column in pass_columns:
            logical_types[column] = cutoff_schema.logical_types[column]
            semantic_tags[column] = cutoff_schema.semantic_tags[column]
            origins[column] = "base"
```

However, when n_jobs == 1 everything works fine. I have investigated the source code and have found that it is probably caused by the fact that when cutoff_times is cut into chunks in the `parallel_calculate_chunks` function, the resulting chunks loose woodwork extension. 

So to solve this issue, I think, it is needed to make .ww.init() for every chunk after division

<details>

Featuretools version: 1.0.0

SYSTEM INFO
-----------
python: 3.7.5.final.0
python-bits: 64
OS: Darwin
OS-release: 19.6.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: ru_RU.UTF-8
LOCALE: ru_RU.UTF-8

INSTALLED VERSIONS
------------------
numpy: 1.21.1
pandas: 1.3.2
tqdm: 4.62.2
PyYAML: 5.4.1
cloudpickle: 1.6.0
dask: 2021.9.0
distributed: 2021.9.0
psutil: 5.8.0
pip: 19.2.3
setuptools: 41.2.0

</details>

rwedge:
Fixes #1763 

WIP for now, ideally we wouldn't need to pass schema separately.  Initialize woodwork on the chunked cutoff times before sending them to the dask workers had an issue where when the worker got them they wouldn't be initialized

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -336,7 +336,7 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
 
 def calculate_chunk(cutoff_time, chunk_size, feature_set, entityset, approximate, training_window,
                     save_progress, no_unapproximated_aggs, cutoff_df_time_col, target_time,
-                    pass_columns, progress_bar=None, progress_callback=None, include_cutoff_time=True):
+                    pass_columns, progress_bar=None, progress_callback=None, include_cutoff_time=True, schema=None):
 
     if not isinstance(feature_set, FeatureSet):
         feature_set = cloudpickle.loads(feature_set)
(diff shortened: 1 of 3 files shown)


featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -374,6 +374,8 @@ def update_progress_callback(done):
         feature_matrix.append(_feature_matrix)
 
     else:
+        if schema:
+            cutoff_time.ww.init_with_full_schema(schema=schema)
```
gsheni, line 378: @rwedge it doesn't seem like this line is getting hit with the unit test you added.

gsheni (approved):

Issue #1758, opened 2021-10-27
Add rolling gap primitives to the Handling Time guide in documentation


tamargrey:
Once #1756 , #1757 , #1735 , and #1730 have been completed, we'll have added a significant amount of new functionalities for time series feature engineering. It could be helpful to add a section on the Primitives in Featuretools that have been built specifically with time series feature engineering in mind.

We can talk about the rolling gap primitives, delayed primitive(s), and any additional time series primitives that have been added from the issues above, how they're used, and why they're specifically useful for time series problems.

Issue #1757, opened 2021-10-27
Allow date offsets as parameters to rolling Transform primitives


tamargrey:
Once we allow rolling gap transform primitives in #1756 that work with integers as the parameters that specify the window of aggregation, we should add the ability to use [pandas date offset](https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#dateoffset-objects) strings as `window_length` and `gap` parameters to the rolling primitives as well. 

This should still maintain the ability to specify `window_length` and `gap` with integers that indicate number of rows, but this will add the ability to get consistent windows for data that does not have a uniform sampling frequency.

tamargrey:
Assuming the shift + rolling implementation works, we should be able to do something like this:

shifted = s.shift(freq='1d')
rolling_max = shifted.rolling('3d').max()
pd.concat([s, shifted, rolling_max], axis=1)

We'll just need to make sure a datetime is set as the series' index and have some logic to handle combinations of string/integer parameters (do we want to allow combos? I think it'd be possible)

tamargrey:
In looking into using `s.shift` with the `freq` parameter, realized that it would cause problems if `gap` was set to be an offset that was not an even multiple of the data's own frequency (so if gap was `4d` and the data's freq was `2d` that'd okay, but if the gap was `3d` in that situation it wouldn't work.

Currently planning on just setting any value prior to the first date + gap to be null. This assumes that the data is sorted.

tamargrey:
In looking into using s.shift with the freq parameter, realized that it would cause problems if gap was set to be an offset that was not an even multiple of the data's own frequency (so if gap was 4d and the data's freq was 2d that'd okay, but if the gap was 3d in that situation it wouldn't work.

Currently planning on shifting by the number of rows that are less than the datetime that is (first date + gap). This assumes that the data is sorted.

PR #1770, opened 2021-11-04
Add Rolling Transform primitives

### LINKED ISSUE #1756: Add rolling gap primitives to Transform Primitives with integer parameters
As a user, I wish I could use Featuretools to engineer transform features that are generated over a rolling window within a single column. The window over which aggregations are calculated should be defined with two parameters: `window_length` and `gap`. These sorts of rolling aggregations can be useful for solving time series machine learning problems, which benefit from looking at previous observations. Therefore, adding these primitives will increase Featuretools' capabilities for time series feature engineering.

We should add primitives that can calculate rolling aggregations for the following operations: min, max, mean, std, count (?), and sum (?).

Window length and gap should be specified with integers representing the number of rows over which the gap and window should span. Note: For the best usage in time series problems, this assumes that the data has a uniform observation frequency (daily, hourly ,etc). This implementation will not enforce this uniform frequency.

Input types will require a Datetime time index to be passed in along with the column over which rolling aggregations will be calculated.

tamargrey:
Adds five rolling transform primitives that all takeinteger  `window_length`, `gap` and `min_periods` parameters.
closes #1756

### DIFF
--- a/featuretools/primitives/standard/rolling_transform_primitive.py
+++ b/featuretools/primitives/standard/rolling_transform_primitive.py
@@ -63,7 +63,6 @@ class RollingMax(TransformPrimitive):
         self.window_length = window_length
         self.gap = gap
         self.min_periods = min_periods
-        # -->determine if we need to add the uses_full_dataframe = True right now
 
     def get_function(self):
         def rolling_max(datetime, numeric):
--- a/featuretools/primitives/standard/rolling_transform_primitive.py
+++ b/featuretools/primitives/standard/rolling_transform_primitive.py
@@ -249,7 +249,7 @@ class RollingSTD(TransformPrimitive):
     name = "rolling_std"
     input_types = [ColumnSchema(logical_type=Datetime, semantic_tags={'time_index'}), ColumnSchema(semantic_tags={'numeric'})]
     return_type = ColumnSchema(logical_type=Double, semantic_tags={'numeric'})
-    compatibility = [Library.PANDAS, Library.DASK, Library.KOALAS]
+    compatibility = [Library.PANDAS, Library.KOALAS]
 
     def __init__(self, window_length=3, gap=0, min_periods=1):
         self.window_length = window_length
--- a/featuretools/primitives/standard/rolling_transform_primitive.py
+++ b/featuretools/primitives/standard/rolling_transform_primitive.py
@@ -44,7 +44,7 @@ class RollingMax(TransformPrimitive):
         >>> rolling_max = RollingMax(window_length=3, gap=1)
         >>> times = pd.date_range(start='2019-01-01', freq='1min', periods=5)
         >>> rolling_max(times, [4, 3, 2, 1, 0]).tolist()
-        [nan, 4.0, 4.0, 3.0, 2.0]
+        [nan, 4.0, 4.0, 4.0, 3.0]
 
         We can also control the minimum number of periods required for the rolling calculation.
 
@@ -57,7 +57,7 @@ class RollingMax(TransformPrimitive):
     name = "rolling_max"
     input_types = [ColumnSchema(logical_type=Datetime, semantic_tags={'time_index'}), ColumnSchema(semantic_tags={'numeric'})]
     return_type = ColumnSchema(logical_type=Double, semantic_tags={'numeric'})
-    compatibility = [Library.PANDAS, Library.DASK, Library.KOALAS]
+    compatibility = [Library.PANDAS, Library.KOALAS]
 
     def __init__(self, window_length=3, gap=0, min_periods=1):
         self.window_length = window_length
@@ -107,7 +107,7 @@ class RollingMin(TransformPrimitive):
         >>> rolling_min = RollingMin(window_length=3, gap=1)
         >>> times = pd.date_range(start='2019-01-01', freq='1min', periods=5)
         >>> rolling_min(times, [4, 3, 2, 1, 0]).tolist()
-        [nan, 3.0, 2.0, 1.0, 0.0]
+        [nan, 4.0, 3.0, 2.0, 1.0]
 
         We can also control the minimum number of periods required for the rolling calculation.
 
@@ -120,7 +120,7 @@ class RollingMin(TransformPrimitive):
     name = "rolling_min"
     input_types = [ColumnSchema(logical_type=Datetime, semantic_tags={'time_index'}), ColumnSchema(semantic_tags={'numeric'})]
     return_type = ColumnSchema(logical_type=Double, semantic_tags={'numeric'})
-    compatibility = [Library.PANDAS, Library.DASK, Library.KOALAS]
+    compatibility = [Library.PANDAS, Library.KOALAS]
 
     def __init__(self, window_length=3, gap=0, min_periods=1):
         self.window_length = window_length
@@ -185,7 +185,7 @@ class RollingMean(TransformPrimitive):
     name = "rolling_mean"
     input_types = [ColumnSchema(logical_type=Datetime, semantic_tags={'time_index'}), ColumnSchema(semantic_tags={'numeric'})]
     return_type = ColumnSchema(logical_type=Double, semantic_tags={'numeric'})
-    compatibility = [Library.PANDAS, Library.DASK, Library.KOALAS]
+    compatibility = [Library.PANDAS, Library.KOALAS]
 
     def __init__(self, window_length=3, gap=0, min_periods=0):
         self.window_length = window_length
@@ -291,7 +291,7 @@ class RollingCount(TransformPrimitive):
         >>> import pandas as pd
         >>> rolling_count = RollingCount(window_length=3)
         >>> times = pd.date_range(start='2019-01-01', freq='1min', periods=5)
-        >>> rolling_count(times, [4, 3, 2, 1, 0]).tolist()
+        >>> rolling_count(times).tolist()
         [1.0, 2.0, 3.0, 3.0, 3.0]
 
         We can also control the gap before the rolling calculation.
@@ -299,7 +299,7 @@ class RollingCount(TransformPrimitive):
         >>> import pandas as pd
         >>> rolling_count = RollingCount(window_length=3, gap=1)
         >>> times = pd.date_range(start='2019-01-01', freq='1min', periods=5)
-        >>> rolling_count(times, [4, 3, 2, 1, 0]).tolist()
+        >>> rolling_count(times).tolist()
         [nan, 1.0, 2.0, 3.0, 3.0]
 
         We can also control the minimum number of periods required for the rolling calculation.
@@ -307,7 +307,7 @@ class RollingCount(TransformPrimitive):
         >>> import pandas as pd
         >>> rolling_count = RollingCount(window_length=3, min_periods=3)
         >>> times = pd.date_range(start='2019-01-01', freq='1min', periods=5)
-        >>> rolling_count(times, [4, 3, 2, 1, 0]).tolist()
+        >>> rolling_count(times).tolist()
         [nan, nan, 3.0, 3.0, 3.0]
     """
     name = "rolling_count"
--- a/featuretools/primitives/standard/rolling_transform_primitive.py
+++ b/featuretools/primitives/standard/rolling_transform_primitive.py
@@ -6,7 +6,7 @@ from woodwork.logical_types import (
     Double,
     Datetime
 )
-from featuretools.primitives.utils import roll_series_with_gap
+from featuretools.primitives.utils import _roll_series_with_gap
 from featuretools.primitives.base.transform_primitive_base import (
     TransformPrimitive
 )
@@ -20,7 +20,7 @@ class RollingMax(TransformPrimitive):
         Given a list of numbers and a corresponding list of
         datetimes, return a rolling maximum of the numeric values,
         starting at the row `gap` rows away from the current row and looking backward
-        over the specified window (by `window_length` and `gap`). 
+        over the specified window (by `window_length` and `gap`).
 
         Input datetimes should be monotonic.
 
@@ -72,10 +72,10 @@ class RollingMax(TransformPrimitive):
     def get_function(self):
         def rolling_max(datetime, numeric):
             x = pd.Series(numeric.values, index=datetime.values)
-            rolled_series = roll_series_with_gap(x,
-                                                 self.window_length,
-                                                 gap=self.gap,
-                                                 min_periods=self.min_periods)
+            rolled_series = _roll_series_with_gap(x,
+                                                  self.window_length,
+                                                  gap=self.gap,
+                                                  min_periods=self.min_periods)
             return rolled_series.max().values
         return rolling_max
 
@@ -135,10 +135,10 @@ class RollingMin(TransformPrimitive):
     def get_function(self):
         def rolling_min(datetime, numeric):
             x = pd.Series(numeric.values, index=datetime.values)
-            rolled_series = roll_series_with_gap(x,
-                                                 self.window_length,
-                                                 gap=self.gap,
-                                                 min_periods=self.min_periods)
+            rolled_series = _roll_series_with_gap(x,
+                                                  self.window_length,
+                                                  gap=self.gap,
+                                                  min_periods=self.min_periods)
             return rolled_series.min().values
         return rolling_min
 
@@ -200,10 +200,10 @@ class RollingMean(TransformPrimitive):
     def get_function(self):
         def rolling_mean(datetime, numeric):
             x = pd.Series(numeric.values, index=datetime.values)
-            rolled_series = roll_series_with_gap(x,
-                                                 self.window_length,
-                                                 gap=self.gap,
-                                                 min_periods=self.min_periods)
+            rolled_series = _roll_series_with_gap(x,
+                                                  self.window_length,
+                                                  gap=self.gap,
+                                                  min_periods=self.min_periods)
             return rolled_series.mean().values
         return rolling_mean
 
@@ -215,7 +215,7 @@ class RollingSTD(TransformPrimitive):
         Given a list of numbers and a corresponding list of
         datetimes, return a rolling standard deviation of
         the numeric values, starting at the row `gap` rows away from the current row and
-        looking backward over the specified time window 
+        looking backward over the specified time window
         (by `window_length` and `gap`). Input datetimes should be monotonic.
 
     Args:
@@ -264,10 +264,10 @@ class RollingSTD(TransformPrimitive):
     def get_function(self):
         def rolling_std(datetime, numeric):
             x = pd.Series(numeric.values, index=datetime.values)
-            rolled_series = roll_series_with_gap(x,
-                                                 self.window_length,
-                                                 gap=self.gap,
-                                                 min_periods=self.min_periods)
+            rolled_series = _roll_series_with_gap(x,
+                                                  self.window_length,
+                                                  gap=self.gap,
+                                                  min_periods=self.min_periods)
             return rolled_series.std().values
         return rolling_std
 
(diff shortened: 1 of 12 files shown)


featuretools/primitives/standard/rolling_transform_primitive.py
```diff
@@ -0,0 +1,343 @@
...
+        >>> times = pd.date_range(start='2019-01-01', freq='1min', periods=5)
+        >>> rolling_max(times, [4, 3, 2, 1, 0]).tolist()
+        [NaN, NaN, 4.0, 3.0, 2.0]
+    """
+    name = "rolling_max"
+    input_types = [ColumnSchema(logical_type=Datetime, semantic_tags={'time_index'}), ColumnSchema(semantic_tags={'numeric'})]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={'numeric'})
+    compatibility = [Library.PANDAS, Library.DASK, Library.KOALAS]
+
+    def __init__(self, window_length=3, gap=0, min_periods=1):
```
tamargrey, line 62: Previously the default window-length was `1d`, which makes sense and will be a valid window with multiple instances present if the data's sampling frequency is less than 1 day. At this point in the implementation, though, we do not have access to date offsets, and setting the window length to `1` doesn't make much sense, as it would never end up with more than one instance present in the window. 

I think I might prefer using `window_length=7` or something that has some time-related meaning. But chose 3 for the moment just to show the arbitrary nature of having to set a default for this parameter. 

Also, when we implement #1757, we can go back to setting the default to be `1d`, but for now, I don't think there's a way to ensure this won't be a breaking change when that's implemented.

featuretools/primitives/utils.py
```diff
@@ -227,3 +227,46 @@ def _find_class_in_descendants(self, search_key):
...
+        Most rolling calculations act this way. The implication of that here is that in order to
+        achieve the gap, we insert NaNs at the beinning of the series, which would cause `count` to calculate
+        on windows that technically should not have the correct number of periods. In the RollingCount primitive,
+        we handle this case manually, replacing those values with NaNs. Any primitive that uses this function
+        should determine whether this kind of handling is also necessary.
+
+    """
+    # Workaround for pandas' fixed but unreleased bug: https://github.com/pandas-dev/pandas/issues/43016
+    # Can remove when upgraded to pandas 1.4.0
+    if str(series.dtype) == 'Int64':
```
tamargrey, line 265: Created an issue to remove this eventually https://github.com/alteryx/featuretools/issues/1771

freddyaboulton (commented):
@tamargrey The behavior of these primitives makes sense to me! The only thing that stood out to me was the difference in parameter definitions between EvalML and Featuretools that we chatted about. When we get to datestrings we will need your help in transforming the EvalML definitions to featuretools, but we agreed the "add 1" operation is possible. 

Don't think EvalML will ever set min_periods to be different than window_length but that may be useful for other kinds of problems. 

If users use `RollingSTD` with min_periods=1, there will be one extra nan compared to the other primitives because the standard deviation of a single observation is not defined. That makes sense to me.

featuretools/primitives/utils.py
```diff
@@ -227,3 +227,46 @@ def _find_class_in_descendants(self, search_key):
...
+    """Provide rolling window calculations with window specified with both a window length
+        and a gap that indicates the amount of time between each instance and its window.
+
+    Args:
+        series (Series): The series over which the windows will be created. Must be numeric in nature
+            and have a datetime64[ns] index.
+        window_length (int): The number of rows to be included in each window. For data
+            with a uniform sampling frequency, for example of one day, the window_length will
+            correspond to a period of time, in this case, 7 days for a window_length of 7.
+        gap (int, optional): The number of rows prior to each instance to be skipped before
```
freddyaboulton, line 242: nit-pick: The use of "skipped" confused me at first. If gap=1, and we're saying that 1 row prior to each instance will be skipped, that makes me think it's doing what gap=2 actually does.

thehomebrewnerd (commented):
Overall this is looking decent, but just have a couple things:
- For the tests in `test_rolling_primitive.py` and `test_deep_feature_synthesis.py` we are only using the pandas version of the test fixture series. Should/could we also test with Dask/Koalas?
- Do we have any concerns with changing the init arguments for these primitives from their previous definitions? I think this means any existing applications where these were used as premium primitives will likely break, but I'm not sure that is an issue. Just wanted to mention in case it is important.

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -36,12 +36,16 @@
...
+    RollingMax,
     Sum,
     TimeSincePrevious,
     TransformPrimitive,
     Trend,
     Year
 )
+from featuretools.primitives.standard.rolling_transform_primitive import (
+    RollingCount
+)
```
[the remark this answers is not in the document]
tamargrey: yep, changed!

tamargrey:
@thehomebrewnerd 
> For the tests in `test_rolling_primitive.py` and `test_deep_feature_synthesis.py` we are only using the pandas version of the test fixture series. Should/could we also test with Dask/Koalas?

Using all three was originally my plan, which was why I created all three fixtures, but when I started running into various issues (koalas' implementation of rolling sometimes has different behavior than pandas, and dask was having other unexplained failutes), I ended up decided to just allow Pandas, which was how they previously were. I still left the dask and koalas fixtures in place, since `_roll_series_with_gap` is still tested with them, so in theory we could add the dask and koalas compatibility if it's requested a bit easier. 

Happy to remove the extra fixtures if it doesn't make sense to keep them in if the primitives aren't yet compatible. Or, if we're okay with this taking a little longer to get in, I could investigate the dask and koalas issues  further to try and get the primitives compatible. @gsheni may have thoughts on whether it'd be worth it.

> Do we have any concerns with changing the init arguments for these primitives from their previous definitions? I think this means any existing applications where these were used as premium primitives will likely break, but I'm not sure that is an issue. Just wanted to mention in case it is important.

We'd need to get the date offset implementation in at the same time to have the primitives not break if they are currently in use anywhere. That would be a bigger ask for this PR, that I think we didn't want to have to wait for.  Also ran this by @gsheni a while back to confirm this was okay. Can revisit if it's going to be problematic to break them.

gsheni:
@tamargrey is correct. We are punting on the Koalas/Dask implementation of these primitives. Additionally, we are okay with this being a breaking change for the RollingX primitives (compared to premium primitives)

thehomebrewnerd:
I think my vote would be to remove the Dask and Koalas fixtures if we are not supporting those types of inputs now, and we can add them back in when they are needed. Otherwise the could end up hanging around for a long time without really being needed.

featuretools/primitives/standard/rolling_transform_primitive.py
```diff
@@ -0,0 +1,335 @@
...
+        starting at the row `gap` rows away from the current row and looking backward
+        over the specified window (by `window_length` and `gap`).
+
+        Input datetimes should be monotonic.
+
+    Args:
+        window_length (int): The number of rows to be included in each frame. For data
+            with a uniform sampling frequency, for example of one day, the window_length will
+            correspond to a period of time, in this case, 7 days for a window_length of 7.
+        gap (int, optional): The number of rows prior to each instance to be skipped before
```
rwedge, line 27: How's this sound

> The number of rows backward from the target instance to be skipped before the window of usable data begins.  Defaults to 0, which will include the target instance in the window

If `include_cutoff_time` is False does gap=0 function the same as gap=1 would normally?

tamargrey: I like your suggestion. The one thing I might change is just completely getting rid of the word "skipped" based off of @freddyaboulton's comment above. How about

> The number of rows backward from the target instance before the window of usable data begins. Defaults to 0, which will include the target instance in the window

I think just talking about "n rows back" better encapsulates the idea that one row back is immediately above and zero rows back is just that instance. Instead of also including the idea of "n rows skipped" which would lead me to think there would be `n` entire rows that are ignored before starting the window.

tamargrey: To your second point: unless I'm missing something, setting a cutoff time and having `include_cutoff_time=False` shouldn't have an impact on the values included in any window. 

In the example below where we have primitives set with gap=0 and gap=1 and it just impacts the final row.
image

If I set cutoff_times for each row to be its own time index value (and now we have to have include_cutoff_time=True or else we'll have no data even though windows technically would be before the cutoff time--may need to be a separate issue if we wanted to handle that case specifically), it has no impact at all. (If we also add `uses_full_dataframe = True` to the primitive because of that separate bug)

image

rwedge: Thanks for the explanation, I think I was confusing using the cutoff time to create a window with using the target instance for a window

tamargrey: Ahh fair. Anything I can add to the docstrings to help clarify what happens in these primitives?

rwedge: I think it's pretty clear

gsheni: We could put this information in our documentation. Add it as a section or a new page on Time Series. Use it as a chance to explain rolling primitives

tamargrey: Have an issue to do that (#1758 - could also just be adding an entirely new Time series page). Was imagining it could happen once we have a lot of the time series stuff in, but we could always start it now and update it as we go along

gsheni: @tamargrey your call. happy with either.

featuretools/tests/primitive_tests/test_primitive_utils.py
```diff
@@ -135,3 +140,96 @@ def test_errors_no_primitive_in_file(bad_primitives_files_dir):
...
+    # Make min periods the size of the window
+    no_partial_values = to_pandas(_roll_series_with_gap(rolling_series,
+                                                        window_length,
+                                                        gap=gap,
+                                                        min_periods=window_length).count())
+    num_null_aggregates = len(no_partial_values.loc[pd.isna(no_partial_values)])
+    num_partial_aggregates = len(no_partial_values.loc[no_partial_values < window_length])
+
+    # Koalas doesn't handle min_periods for count
+    if ks and not isinstance(rolling_series, ks.Series):
```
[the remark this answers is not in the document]
tamargrey: removed the koalas fixture

featuretools/primitives/standard/rolling_transform_primitive.py
tamargrey: Gonna leave the documentation to later. Should be a bit easier to create a more cohesive narrative when everything is implemented.

rwedge:
@tamargrey can you pull in the latest changes?

featuretools/tests/conftest.py
```diff
@@ -565,3 +565,8 @@ def divide_by_zero_es_koalas(divide_by_zero_es_pd):
                          index=df.ww.index,
                          logical_types=df.ww.logical_types)
     return es
+
+
+@pytest.fixture
+def rolling_series():
```
rwedge, line 571: what if we kept the "rolling_series_pd" name to avoid having to check the definition to see if its a pandas only fixture or a parameterized one?

tamargrey: sure, happy to keep as `rolling_series_pd`

rwedge (dismissed):
LGTM assuming test pass

gsheni (dismissed):

featuretools/primitives/utils.py
```diff
@@ -227,3 +227,47 @@ def _find_class_in_descendants(self, search_key):
...
+            over the window. Can only be as large as window_length. Defaults to 1.
+
+    Returns:
+        pandas.core.window.rolling.Rolling: The Rolling object for the series passed in.
+
+    Note:
+        Certain operations, like `pandas.core.window.rolling.Rolling.count` that can be performed
+        on the Rolling object returned here may treat NaNs as periods to include in window calculations.
+        So a window [NaN, 1, 3]  when `min_periods=3` will proceed with count, saying there are three periods
+        but only two values wnd would return count=2. The calculation `max` on the other hand,
```
thehomebrewnerd, line 255: Spelling error: `wnd`

featuretools/tests/primitive_tests/test_primitive_utils.py
```diff
@@ -135,3 +138,94 @@ def test_errors_no_primitive_in_file(bad_primitives_files_dir):
...
+@pytest.mark.parametrize(
+    "window_length, gap",
+    [
+        (6, 2),
+        (6, 0)  # No gap - changes early values
+    ]
+)
+def test_roll_series_with_gap_early_values(window_length, gap, rolling_series_pd):
+    # Default min periods is 1 - will include all
+    default_partial_values = to_pandas(_roll_series_with_gap(rolling_series_pd,
```
thehomebrewnerd, line 195: Leftover `to_pandas` call?

featuretools/tests/primitive_tests/test_primitive_utils.py
```diff
@@ -135,3 +138,94 @@ def test_errors_no_primitive_in_file(bad_primitives_files_dir):
...
+                                                             gap=gap).count())
+    num_empty_aggregates = len(default_partial_values.loc[default_partial_values == 0])
+    num_partial_aggregates = len((default_partial_values
+                                  .loc[default_partial_values != 0])
+                                 .loc[default_partial_values < window_length])
+    assert num_empty_aggregates == gap
+    assert num_partial_aggregates == window_length - 1
+
+    # Make min periods the size of the window
+    no_partial_values = to_pandas(_roll_series_with_gap(rolling_series_pd,
```
thehomebrewnerd, line 206: Remove `to_pandas` call?

featuretools/tests/primitive_tests/test_primitive_utils.py
```diff
@@ -135,3 +138,94 @@ def test_errors_no_primitive_in_file(bad_primitives_files_dir):
...
+    window_length = 3
+    gap = 2
+    # Because we're inserting nans, confirm that nullability of the dtype doesn't have an impact on the results
+    nullable_series = rolling_series_pd.astype('Int64')
+    non_nullable_series = rolling_series_pd.astype('int64')
+
+    nullable_rolling_max = _roll_series_with_gap(nullable_series, window_length, gap=gap).max()
+    non_nullable_rolling_max = _roll_series_with_gap(non_nullable_series, window_length, gap=gap).max()
+
+    pd.testing.assert_series_equal(to_pandas(nullable_rolling_max), to_pandas(non_nullable_rolling_max))
```
thehomebrewnerd, line 231: Remove `to_pandas` calls?

featuretools/tests/primitive_tests/test_primitive_utils.py
```diff
@@ -135,3 +138,94 @@ def test_errors_no_primitive_in_file(bad_primitives_files_dir):
...
+
+    # because we shift, gap is included as nan values in the series.
+    # Count treats nans in a window as values that don't get counted,
+    # so the gap rows get included in the count for whether a window has "min periods".
+    # This is different than max, for example, which does not count nans in a window as values towards "min periods"
+    assert num_null_aggregates == window_length - 1
+    assert num_partial_aggregates == gap
+
+
+def test_roll_series_with_gap_nullable_types(rolling_series_pd):
```
thehomebrewnerd, line 221: Any need to also test these with an input series that contains null values to confirm they are handled as expected?

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -412,6 +414,28 @@ def test_bad_groupby_feature(es):
                              groupby_trans_primitives=['max'])
 
 
+def test_make_rolling_features(pd_es):
```
thehomebrewnerd, line 417: Might be overkill, but you could create a loop here and test all the new primitives, except maybe count which has a different naming scheme. That probably already gets tested elsewhere though.

featuretools/tests/primitive_tests/test_rolling_primitive.py
```diff
@@ -0,0 +1,144 @@
...
+    RollingSTD
+)
+from featuretools.primitives.utils import _roll_series_with_gap
+
+
+def test_rolling_max(rolling_series_pd):
+    window_length = 5
+    gap = 2
+
+    expected_vals = _roll_series_with_gap(rolling_series_pd,
```
thehomebrewnerd, line 18: Wondering just a little bit about the robustness of these tests. If I'm looking at this correctly we are more or less duplicating the primitive function here and confirming that the actual primitive function returns the same value as the version we have in the test. If we did something to change `_roll_series_with_gap` that causes incorrect results, that would also get propagated here and the test would pass, but hopefully be caught in the tests specific to that function.

Not sure if it is better to hard code the expected values or not? Not a blocker for me one way or the other, just something to consider.

tamargrey: Yeah, my goal in the tests for this feature would be to catch any wrong or unexpected rolling behavior in the tests specific to `_roll_series_with_gap`. 

My goal with the primitive tests was to confirm that they behave exactly like `_roll_series_with_gap` so we could catch any differing behavior that could get introduced within the primitive classes themselves. Or, in the case of `RollingCount` where we're specifically making changes to the output of `_roll_series_with_gap`, test just that differing behavior.

With regard to whether we should hardcode the values, my hesitation with doing that here would just be in the fact that it can get confusing to update hardcoded values if/when things do change. I've had situations when I've broken hardcoded tests in a repo I wasn't super familiar with, and it was a pretty big struggle to figure out whether they broke because the hardcoded value should now just be updated, or if it was because my changes had unintended side effects. I have less of a problem with hardcoded values in the `_roll_series_with_gap` tests, though, so if I should be doing a better job of that, happy to make changes.

featuretools/primitives/utils.py
tamargrey: thanks for catching!

featuretools/tests/primitive_tests/test_primitive_utils.py
tamargrey: yep, will remove all the to_pandas calls in this file!

featuretools/tests/primitive_tests/test_primitive_utils.py
tamargrey: yep, good idea--will add

featuretools/tests/synthesis/test_deep_feature_synthesis.py
tamargrey: yeah, the test below `test_make_rolling_count_off_datetime_feature` will cover count. Happy to iterate over all the rolling primitives

tamargrey: ended up parameterizing the various primitives instead of iterating over them

thehomebrewnerd: Even better!

featuretools/tests/primitive_tests/test_rolling_primitive.py
thehomebrewnerd: Sounds good to me. Know need to make any changes.

thehomebrewnerd (approved):
LGTM. Seems like the test coverage CI failure is due to some flaky Dask behavior and nothing in this MR.

featuretools/utils/plot_utils.py
```diff
@@ -34,7 +34,11 @@ def get_graphviz_format(graphviz, to_file):
                              " so that the format can be inferred")
 
         format_ = split_path[-1]
-        valid_formats = graphviz.backend.FORMATS
+        try:
+            valid_formats = graphviz.backend.FORMATS
+        except AttributeError:
+            from graphviz.parameters import FORMATS
+            valid_formats = FORMATS
```
rwedge, line 41: try using graphviz.FORMATS

gsheni (approved):

PR #1784, opened 2021-11-16
Clean up install page with correct python versions

gsheni:


docs/source/install.rst
```diff
@@ -1,7 +1,7 @@
 Install
 *******
 
-Featuretools is available for Python 3.7 and 3.8. The recommended way to install Featuretools is using ``pip`` or ``conda``::
+Featuretools available for Python 3.7, 3.8, and 3.9. It can be installed from PyPI, conda, or from source.::
```
rwedge, line 4: I think we want to keep the "is" in the first sentence.

The period is unnecessary and when built reads like this:

> It can be installed from PyPI, conda, or from source.:

Should it be Anaconda instead of conda if we're saying PyPI instead of pip?

I feel like users will be a bit more familiar with pip and conda compared to PyPI and Anaconda

gsheni: Fixed to say pip

docs/source/install.rst
```diff
@@ -1,7 +1,7 @@
 Install
 *******
 
-Featuretools is available for Python 3.7 and 3.8. The recommended way to install Featuretools is using ``pip`` or ``conda``::
+Featuretools available for Python 3.7, 3.8, and 3.9. It can be installed from pip, conda, or from source.::
```
rwedge, line 4:
```suggestion
Featuretools is available for Python 3.7, 3.8, and 3.9. It can be installed from pip, conda, or from source::
```

rwedge (approved):

PR #1786, opened 2021-11-17
Fix formatting issue in Improving Computational Performance guide

### LINKED ISSUE #1778: Bug with parallel feature matrix calculation within sklearn cross-validation
### Bug with parallel feature matrix calculation within sklearn cross-validation

-----
#### Bug Description

Hello, guys! Thank you for the quick release of featuretools 1.1.0 ! 

During my research I have faced the following bug:
I have an estimator which is actually an `imblearn Pipeline`. The estimator consists of several steps including my custom transformer which calculates feature matrix with `featuretools`. And I want to check the quality of the model with `sklearn cross_validate` function . If I set `n_jobs` > 1 both in `featuretools.calculate_feature_matrix` and in `sklearn.cross_validate`, then I get an unexpected error `ValueError: cannot find context for 'loky'`. When either one of `n_jobs` is set to 1, then everything works fine.

I googled for some time and I understood that such error might happen when parallelization is used without `if __name__ == '__main__'` - but it's the best information I've got - nothing more valuable. So for me it looks like there is some conflict in parallelization usage in `sklearn` and `featuretools`. And as far both of the libraries are essential as well as parallelization working with big data, i really hope you will be able to find a way to fix it :)

P.S this problem was actual before 1.0.0 release - previously I used 0.24.0 and still faced it

<details>

Featuretools version: 1.1.0

SYSTEM INFO
-----------
python: 3.7.5.final.0
python-bits: 64
OS: Darwin
OS-release: 19.6.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: ru_RU.UTF-8
LOCALE: ru_RU.UTF-8

INSTALLED VERSIONS
------------------
numpy: 1.21.1
pandas: 1.3.2
tqdm: 4.62.2
PyYAML: 5.4.1
cloudpickle: 1.6.0
dask: 2021.10.0
distributed: 2021.10.0
psutil: 5.8.0
pip: 19.2.3
setuptools: 41.2.0

</details>

thehomebrewnerd:
### Fix formatting issue in Improving Computational Performance guide

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -686,11 +686,11 @@ def parallel_calculate_chunks(cutoff_time, chunk_size, feature_set, approximate,
     except Exception:
         raise
     finally:
-        if client is not None:
-            client.close()
-
-        if 'cluster' not in dask_kwargs and cluster is not None:
-            cluster.close()
+        if 'cluster' not in dask_kwargs:
+            if cluster is not None:
+                cluster.close()
+            if client is not None:
+                client.close()
 
     ww_init_kwargs = get_ww_types_from_features(feature_set.target_features, entityset, pass_columns, cutoff_time)
     feature_matrix = init_ww_and_concat_fm(feature_matrix, ww_init_kwargs)
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -687,10 +687,10 @@ def parallel_calculate_chunks(cutoff_time, chunk_size, feature_set, approximate,
         raise
     finally:
         if 'cluster' not in dask_kwargs:
-            if cluster is not None:
-                cluster.close()
             if client is not None:
                 client.close()
+            if cluster is not None:
+                cluster.close()
 
     ww_init_kwargs = get_ww_types_from_features(feature_set.target_features, entityset, pass_columns, cutoff_time)
     feature_matrix = init_ww_and_concat_fm(feature_matrix, ww_init_kwargs)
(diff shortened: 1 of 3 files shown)


rwedge:
Doesn't Featuretools always create a client itself?  The user can pass in a cluster but not a client

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -686,11 +686,11 @@ def parallel_calculate_chunks(cutoff_time, chunk_size, feature_set, approximate,
...
-        if client is not None:
-            client.close()
-
-        if 'cluster' not in dask_kwargs and cluster is not None:
-            cluster.close()
+        if 'cluster' not in dask_kwargs:
+            if cluster is not None:
+                cluster.close()
+            if client is not None:
+                client.close()
```
rwedge, line 693: in the past we swapped the order to close the client before the cluster so that there wouldn't be these benign warnings from the client about losing contact with the cluster

thehomebrewnerd:
> Doesn't Featuretools always create a client itself? The user can pass in a cluster but not a client

Maybe @davesque can help with some of the details here, because I'm a little fuzzy on the details of exactly what happens when setting up a cluster/client.

What I observed during my testing with the latest code I posed in #1778 is this:
- When I ran the example on Featuretools `main`, the process would hang after the feature matrix computation tasks were completed (this was with `client.close()` being called inside `parallel_calculate_chunks`)
- When I updated FT to move the `client.close()` call to happen only when we create the cluster inside FT, the code did not hang and completed successfully.

I might be misunderstanding something, but that seemed to me like we were killing communication with the scheduler somehow with the `client.close()` call which caused things to stop working.

featuretools/computational_backends/calculate_feature_matrix.py
thehomebrewnerd: Deleted this code change for now - will revisit later if needed.

rwedge (approved):
LGTM

Issue #1730, opened 2021-10-08
Spike: Refactor Rolling primitives


tamargrey:
Feature engineering for time series problems often makes use of "delayed" (or lagged) copies of the original data that shifts relevant columns in the original dataset in order to provide each row with information about previous observations. This is useful for time series problems where the data has an inherent, temporal ordering and knowing information about earlier observations can inform predictions about future observations. 

The sort of features that featuretools should be able to create for a value at time `t=0` are:
- previous observation - the observation immediately prior to `t` at `t - 1`
- the observation a specified amount of time prior to `t` - the observation at `t - 7` to get the observation a week ago for data with a daily frequency.
  - or any "meaningful" multiples of the data frequency. So data with a daily frequency is meaningful at [7, 14, 28, 365]. Data with a weekly frequency could be meaningful at [2, 4, 52].
- any number of these observations
  - the last `k` observations - the observations between `range(t - k, t)` or, if we only want to include one observation each week over the last `k` weeks of daily data: `range(t - 7k, t, 7)`

EvalML's `DelayedFeatureTransformer` makes use of this kind of feature engineering via the `series.shift` pandas operation [here](https://github.com/alteryx/evalml/blob/27198716f31a0e1661b0d4733c960be0d94c1acf/evalml/pipelines/components/transformers/preprocessing/delayed_feature_transformer.py#L127
). There, EvalML uses the `start_delay` and `max_delay` attributes to determine the length of the window of time each value will have access to, creating new columns for the feature matrix. 

Featuretools may be able to perform this kind of feature engineering by viewing the columns created as multi-output features created from a transform primitive that sets `number_output_features` to be equal to a `window_length` parameter. The primitive may also need `uses_full_dataframe = True`, as it will access earlier values in the dataframe. This primitive should also take in a `gap` parameter that determines whether the prior observations are taken from immediately before a value or whether there is a gap of time. `gap` and `window_length` have to be functional with EvalML's `start_delay` and `max_delay`.

As the usefulness of shifted data depends on the existence of a time index, the one of the inputs should be a time index column. Additionally, users may need to be aware that, for this primitive to function, the data passed in will have to have a constant interval of time between observations (frequency).

To determine
1. Is there a way for DFS to restrict use of this primitive to columns with a known frequency? Or will correct usage require the creation of seed features? Of is a frequency not actually necessary?
[...]

PR #1788, opened 2021-11-18
[Draft] Add Lag primitive

tamargrey:
Adds a `Lag` primitive that shifts an array by a defined number of periods.

To Do:

- [ ] Understand woodwork init problem with non nullable columns
- [ ] Add tests
  - [ ] Test primitive behavior
  - [ ] Test with DFS - woodwork init!
- [ ] Add to API ref

### DIFF
--- a/featuretools/primitives/standard/transform_primitive.py
+++ b/featuretools/primitives/standard/transform_primitive.py
@@ -863,3 +863,45 @@ class EmailAddressToDomain(TransformPrimitive):
                 emails_df['domain'] = emails_df['email'].str.strip().str.split('@', expand=True)[1]
             return emails_df.domain.values
         return email_address_to_domain
+
+
+class Lag(TransformPrimitive):
+    """Shifts an array of values by a specified number of periods.
+
+    Args:
+        periods (int): The number of periods by which to shift the input.
+            Default is 1. Periods correspond to rows.
+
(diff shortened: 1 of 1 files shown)


tamargrey:
The failure in the unit tests is because of Lag's `None` return type trying to initialize lagged feature columns that contain nulls as the logical type from the EntitySet. If the logical type is not nullable, Woodwork init will fail. There are several routes we could take to deal with this situation:
- Handle the nans in feature matrix problem within featuretools - possibly in `get_ww_types_from_features` maybe by not specifying a logical type for woodwork init if the logical type would be non-nullable or by always specifying a nullable version instead (kind of like how we do Double when the numeric tag is present). 
- Handle the nans in non-nullable logical types falling back to nullable logical types in Woodwork (https://github.com/alteryx/woodwork/issues/1119)
- Force users to utilize `fill_value` in the primitive - the fill value will need to work for any input type, so it may need to be a dictionary so we can handle numeric fills vs boolean fills, vs string fills, etc. 
- Require that users only use this primitive to create seed features or specify `ignore_columns` as the non-nullable columns if any non-nullable logical types are present

tamargrey:
Closing for now, since we're waiting to complete #1730 until we've handled the nullability situation described above in  the Woodwork nullability epic: https://github.com/alteryx/woodwork/issues/1195

PR #1797, opened 2021-11-30
Add NumericLag transform primitve

### LINKED ISSUE #1790: Add NumericLag primitive
Delayed features are useful for time series problems, and although we can't make a general `Lag` primitive that can take in any column type because of the null values introduced, we can create a `NumericLag` primitive that just lags on numeric columns. This will let us start using Featuretools more for time series problems until we have a solution to the nullability problem.

tamargrey:
closes #1790 

The numeric lag primitive requires that the DataFrame also have a time index, though whether the time index is numeric or datetime in nature shouldn't matter.

### DIFF
--- a/featuretools/primitives/standard/transform_primitive.py
+++ b/featuretools/primitives/standard/transform_primitive.py
@@ -10,6 +10,7 @@ from woodwork.logical_types import (
     BooleanNullable,
     Categorical,
     Datetime,
+    Double,
     EmailAddress,
     LatLong,
     NaturalLanguage,
@@ -893,14 +894,15 @@ class NumericLag(TransformPrimitive):
         [100, 1, 2, 3]
     """
     name = "lag"
-    input_types = [ColumnSchema(semantic_tags={'numeric'})]
-    return_type = ColumnSchema(semantic_tags={'numeric'})
+    input_types = [ColumnSchema(logical_type=Datetime), ColumnSchema(semantic_tags={'numeric'})]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={'numeric'})
 
     def __init__(self, periods=1, fill_value=None):
         self.periods = periods
         self.fill_value = fill_value
 
     def get_function(self):
-        def lag(x):
-            return x.shift(periods=self.periods, fill_value=self.fill_value)
+        def lag(datetime, numeric):
+            x = pd.Series(numeric.values, index=datetime.values)
+            return x.shift(periods=self.periods, fill_value=self.fill_value).values
         return lag
--- a/featuretools/primitives/standard/transform_primitive.py
+++ b/featuretools/primitives/standard/transform_primitive.py
@@ -865,15 +865,14 @@ class EmailAddressToDomain(TransformPrimitive):
         return email_address_to_domain
 
 
-class Lag(TransformPrimitive):
+class NumericLag(TransformPrimitive):
     """Shifts an array of values by a specified number of periods.
 
     Args:
         periods (int): The number of periods by which to shift the input.
             Default is 1. Periods correspond to rows.
 
-        # --> cant use same fill value for different types - string wont work for an otherwise numeric column
-        fill_value (int, float, string, bool): The value to use to fill in
+        fill_value (int, float, optional): The value to use to fill in
             the gaps left after shifting the input. Default is None.
 
     Examples:
@@ -894,8 +893,8 @@ class Lag(TransformPrimitive):
         [100, 1, 2, 3]
     """
     name = "lag"
-    input_types = [ColumnSchema()]
-    return_type = None
+    input_types = [ColumnSchema(semantic_tags={'numeric'})]
+    return_type = ColumnSchema(semantic_tags={'numeric'})
 
     def __init__(self, periods=1, fill_value=None):
         self.periods = periods
--- a/featuretools/primitives/standard/transform_primitive.py
+++ b/featuretools/primitives/standard/transform_primitive.py
@@ -895,7 +895,7 @@ class NumericLag(TransformPrimitive):
     """
     name = "numeric_lag"
     input_types = [ColumnSchema(semantic_tags={'time_index'}), ColumnSchema(semantic_tags={'numeric'})]
-    return_type = ColumnSchema(logical_type=Double, semantic_tags={'numeric'})
+    return_type = ColumnSchema(semantic_tags={'numeric'})
     uses_full_dataframe = True
 
     def __init__(self, periods=1, fill_value=None):
--- a/featuretools/primitives/standard/transform_primitive.py
+++ b/featuretools/primitives/standard/transform_primitive.py
@@ -896,6 +896,7 @@ class NumericLag(TransformPrimitive):
     name = "numeric_lag"
     input_types = [ColumnSchema(semantic_tags={'time_index'}), ColumnSchema(semantic_tags={'numeric'})]
     return_type = ColumnSchema(logical_type=Double, semantic_tags={'numeric'})
+    uses_full_dataframe = True
 
     def __init__(self, periods=1, fill_value=None):
         self.periods = periods
--- a/featuretools/primitives/standard/transform_primitive.py
+++ b/featuretools/primitives/standard/transform_primitive.py
@@ -10,7 +10,6 @@ from woodwork.logical_types import (
     BooleanNullable,
     Categorical,
     Datetime,
-    Double,
     EmailAddress,
     LatLong,
     NaturalLanguage,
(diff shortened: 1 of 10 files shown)


featuretools/primitives/standard/transform_primitive.py
```diff
@@ -863,3 +864,45 @@ def email_address_to_domain(emails):
...
+        [nan, nan, nan, 1.0, 2.0]
+
+        You can specify the fill value to use
+
+        >>> lag_fill_value = NumericLag(fill_value=100)
+        >>> lag_fill_value(pd.Series(pd.date_range(start="2020-01-01", periods=4, freq='D')), [1, 2, 3, 4]).tolist()
+        [100, 1, 2, 3]
+    """
+    name = "numeric_lag"
+    input_types = [ColumnSchema(semantic_tags={'time_index'}), ColumnSchema(semantic_tags={'numeric'})]
```
thehomebrewnerd, line 897: Is there a reason for requiring a time index as input to this primitive?

tamargrey: Having a lagging feature only really makes sense if the data has a temporal ordering to it.

Otherwise a `NumericLag` feature with `periods=k` is essentially just choosing to include other row values from the dataset without any rhyme or reason other than the fact that they just happened to be the row that is `k` periods away.

thehomebrewnerd: Agreed, but we aren't actually doing anything with the ordering inside the primitive. So, I guess my question maybe is this - should we require the user to have a time index set and have FT pass it to the primitive unused, or should we document the behavior of this primitive and expect the user to use it properly only when they have correctly ordered data?

I'm not really advocating for one or the other at this point, just bringing it up for discussion.

rwedge:
This primitive should have the `uses_full_dataframe` flag,  otherwise with unique cutoff times the other rows won't be present to shift correctly

featuretools/primitives/standard/transform_primitive.py
tamargrey: We do the same thing with the rolling primitives, though in the future we will use pandas' offset aliases, which will require those datetime values. So that might have been a good enough reason to require the time index. 

We could do the same thing here, actually, and at some point allow the lag primitive to be used with offset aliases.

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -863,3 +864,46 @@ def email_address_to_domain(emails):
...
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={'numeric'})
```
thehomebrewnerd, line 898: Not sure how important it is, but should we just leave this as `return_type = ColumnSchema(semantic_tags={'numeric'})` instead? I suspect the logical type is specified to try and avoid type inference, but this will force all columns to get converted to Double, which theoretically could cause a loss of information, especially if used with very large number inputs.

I think we should try to be specific with our return logical types when possible, but if we truly don't know that the output will always be of a certain type, maybe we should let inference happen so don't end up with an unnecessary type conversion that alters our data?

tamargrey: I definitely see your point. Happy to change to just the tag.

Though if I'm reading the way woodwork types are determined in the feature matrix, we may be automatically using `Double` when the `numeric` tag is present and no logical type has been specified ( I think that this is what keeps the nullability problem from showing up here): https://github.com/alteryx/featuretools/blob/main/featuretools/computational_backends/utils.py#L318

featuretools/tests/synthesis/test_deep_feature_synthesis.py
```diff
@@ -448,6 +449,35 @@ def test_make_rolling_count_off_datetime_feature(pd_es):
     assert feature_with_name(features, rolling_transform_name)
 
 
+def test_numeric_lag_works_with_non_nullable(pd_es):
```
thehomebrewnerd, line 452: Just curious about the intended purpose of this test. Were you intending to make sure we can calculate feature values correctly and that we don't get errors if we have a non-nullable input that gets lagged and introduces null values?

If so, I think you actually need to go to the step of computing the feature matrix instead of just the features. Without computing the features, I don't think that type of error will show up.

Feel free to disregard if the intention was something else here.

tamargrey: yep, you're right. I can add the calculate feature matrix step here. Though maybe that means this should be a more targeted test in `test_calculate_feature_matrix` or `test_feature_set_calculator`

featuretools/primitives/standard/transform_primitive.py
thehomebrewnerd: Yeah, think you are right, and we will end up with Double regardless. I forgot about that code in `get_ww_types_from_features`. 

I'm still thinking it would be better to leave the logical type off here, so if we ever change the logic in `get_ww_types_from_features` we won't be forcing an unnecessary conversion based on the primitive return type.

featuretools/tests/synthesis/test_deep_feature_synthesis.py
thehomebrewnerd: Yeah, probably would be best to relocate the test in that case.

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -1868,3 +1870,34 @@ def test_cfm_with_invalid_time_index(es):
...
+    assert isinstance(pd_es['new_log'].ww.logical_types['value'], Integer)
+
+    lag_primitive = NumericLag(periods=5)
+    features = ft.dfs(target_dataframe_name='new_log',
+                      entityset=pd_es,
+                      agg_primitives=[],
+                      trans_primitives=[lag_primitive],
+                      features_only=True)
+
+    fm = calculate_feature_matrix(features=features, entityset=pd_es)
```
thehomebrewnerd, line 1897: Could remove `feautres_only=True) and combine these into a single call:
```
    fm, features = ft.dfs(target_dataframe_name='new_log',
                          entityset=pd_es,
                          agg_primitives=[],
                          trans_primitives=[lag_primitive],)
```

Also, should we do any checks on the features values or logical types? Not sure it's critical, but you could check that we end up with null values as expected. If we don't check anything in the feature matrix you could change the assignment of `fm, features` to `_, features` since we are just throwing away the feature matrix data in that case.

tamargrey: good call. Added checks for the null values in the feature matrix, and it means now we're not using the `features` list, so changed that to `_`.

Also moved the test to `test_transform_features`, because it seems that that's the better location for how a specific primitive's feature ends up looking in the feature matrix

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -1113,3 +1114,35 @@ def test_time_since_primitive_matches_all_datetime_types(es):
...
+                    index='id', time_index='datetime',
+                    name='new_log')
+    pd_es.add_dataframe(new_log)
+    rels = [('sessions', 'id', 'new_log', 'session_id'),
+            ('products', 'id', 'new_log', 'product_id')]
+    pd_es = pd_es.add_relationships(rels)
+
+    assert isinstance(pd_es['new_log'].ww.logical_types['value'], Integer)
+
+    lag_primitive = NumericLag(periods=5)
```
thehomebrewnerd, line 1134: Not a big deal, but you could use a variable for periods here and below instead of hardcoding `5` everywhere. That way it might be more obvious that the assertions are checking that the number of nulls introduced is equal to the periods used for lagging.

thehomebrewnerd (dismissed):
Just one more non-critical, non-blocking suggestion for the cfm test.

featuretools/tests/primitive_tests/test_transform_features.py
tamargrey: done!

thehomebrewnerd (approved):
LGTM!

rwedge:
For the CFM test can we use unique cutoff times as well to test that scenario?

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -1133,10 +1133,12 @@ def test_cfm_with_lag_and_non_nullable_column(pd_es):
 
     periods = 5
     lag_primitive = NumericLag(periods=periods)
+    cutoff_times = pd_es['new_log'][['id', 'datetime']]
```
tamargrey, line 1136: @rwedge Was this what you had in mind for adding individual cutoff times for each row?

rwedge: yep!

rwedge (approved):
LGTM

PR #1809, opened 2021-12-14
Add ability to use rolling primitives with pandas' offset aliases

### LINKED ISSUE #1757: Allow date offsets as parameters to rolling Transform primitives

tamargrey:
Allows users to pass strings like `1D` or `1H` as `gap` and `window_length` parameters to the rolling transform primitives. This allows data that does not have a uniform sampling frequency to be used accurately with the rolling transform primitives.

closes #1757 

To Do:

- [x] Handle input exceptions
  - [x] Non offset strings?
  - [x] Combination of parameter types?
  - [x] No datetime present on the input series to _roll_series_with_gap?
  - [x] Min periods fine for offset string but too big for integer input?
- [x] Update docstrings
  - [x] Parameters
  - [x] Note limitations
    - [x] Min periods can only be as big as window length for integers but no limitation for offset strings so be careful
    - [x] Window length can only be a fixed frequency (day or hour; not month or week) offset due to pandas' limitation; gap can only be fixed frequency due to introduced ft limitaion (we add gap to window length so they need to be fixed)
  - [x] Examples in primitive docstrings
- [x] Add edge case tests
  - [x] large min periods because data's frequency is much higher than the observation window (hourly data with a window of 7 days)
  - [x] Empty series?
  - [x] Mismatched parameters
- [x] Clean up
  - [x] remove unnecessary commnts
  - [x] Code quality/clarity
  - [x] Test clarity/repetition


featuretools/primitives/standard/rolling_transform_primitive.py
```diff
@@ -210,14 +323,38 @@ class RollingSTD(TransformPrimitive):
...
+            and it will indicate a length of time between a target instance and the beginning of its window.
+            Defaults to 0, which will include the target instance in the window.
+        min_periods (int, optional): Minimum number of observations required for performing calculations
+            over the window. Can only be as large as window_length when window_length is an integer.
+            When window_length is an offset alias string, this limitation does not exist, but care should be taken
+            to not choose a min_periods that will always be larger than the number of observations in a window.
+            Defaults to 1.
+
+    Note:
+        Only offset aliases with fixed frequencies can be used when defining gap and window_lengt.
```
rwedge, line 347: missing the 'h' in `window_length`, happens in several other places as well

featuretools/tests/primitive_tests/utils.py
```diff
@@ -0,0 +1,5 @@
+def get_number_of_days(offset):
```
thehomebrewnerd, line 1: Probably not a big deal since we are only using this in tests, but I think this would return the wrong value if we ever used this in a test with a number above 9.

featuretools/primitives/standard/rolling_transform_primitive.py
```diff
@@ -62,13 +97,19 @@ def __init__(self, window_length=3, gap=0, min_periods=1):
         self.gap = gap
         self.min_periods = min_periods
 
+    def _offset_max(self, series):
+        return _apply_roll_with_offset_gap(series, self.gap, max, self.min_periods)
+
```
thehomebrewnerd, line 102: I'm not sure it matters too much, but do we need these `_offset*` functions as methods on the primitive? I haven't tested this to confirm, but I think you can just call `_apply_roll_with_offset_gap` directly in the `rolled_series.apply` call below, passing in the additional required arguments. Something like this:

```
if isinstance(self.gap, str):
    additional_args = (self.gap, max, self.min_periods)
    return rolled_series.apply(_apply_roll_with_offset_gap, args=additional_args)
```

thehomebrewnerd (commented):
There is a lot to digest here, but here are a couple things I've noticed so far.

featuretools/tests/primitive_tests/utils.py
tamargrey: Yeah, I can either clarify the limitations in a docstring/what the intended use is or update it to return any numeric characters before a string. I think my preference for now would be to keep it simple and add a docstring.

featuretools/primitives/standard/rolling_transform_primitive.py
tamargrey: Good call! Updated to do this

featuretools/primitives/standard/rolling_transform_primitive.py
```diff
@@ -251,13 +386,20 @@ def __init__(self, window_length=3, gap=0, min_periods=1):
         self.gap = gap
         self.min_periods = min_periods
 
+    def _pandas_std(self, series):
+        return series.std()
```
tamargrey, line 390: Tried using `np.std`, but it seems like there's a slightly different handling of nans that produces different results from the pandas implementation

thehomebrewnerd: If you want, you can define this as a function inside the scope of `rolling_std` since that is the only place it's used, rather than adding it as a class method.

featuretools/tests/primitive_tests/test_rolling_primitive.py
```diff
@@ -109,14 +192,13 @@ def test_rolling_count(rolling_series_pd):
     [
         (0, 2),  # 0 and 1 get treated the same
         (1, 2),
+        (1, 2),
```
thehomebrewnerd, line 195: Accidental duplication?

featuretools/tests/primitive_tests/utils.py
thehomebrewnerd: Still need to add the docstring here?

thehomebrewnerd (commented):
A couple more minor things - I'd still like to go through this again to solidify my understanding of what is happening exactly, but nothing major is jumping out to me so far.

featuretools/primitives/standard/rolling_transform_primitive.py
tamargrey: that makes sense! changing

featuretools/tests/primitive_tests/utils.py
tamargrey: oops, yeah forgot to push

featuretools/tests/primitive_tests/test_rolling_primitive.py
tamargrey: yep, removed!

thehomebrewnerd (approved):

davesque (commented):
I looked over the recent commit that includes the value checking stuff I suggested.  That part looks good!  The rest is quite a lot so I'll have to spend some more time on it.

featuretools/primitives/utils.py
```diff
@@ -261,17 +273,129 @@ def _roll_series_with_gap(series, window_size, gap=0, min_periods=1):
...
 
-    gap_applied = series
-    if gap > 0:
-        gap_applied = series.shift(gap)
+    functional_window_length = window_size
+    if isinstance(gap, str):
+        # Add the window_size and gap so that the rolling operation correctly takes gap into account.
+        # That way, we can later remove the gap rows in order to apply the primitive function
+        # to the correct window
+        functional_window_length = to_offset(window_size) + to_offset(gap)
```
freddyaboulton, line 303: This implementation makes sense to me

featuretools/tests/primitive_tests/test_rolling_primitive.py
```diff
@@ -133,12 +214,113 @@ def test_rolling_count_primitive_min_periods_nans(min_periods, expected_num_nams
...
+def test_rolling_count_with_no_gap(window_length, gap, min_periods, expected_num_nams, rolling_series_pd):
     primitive_instance = RollingCount(window_length=window_length, gap=gap, min_periods=min_periods)
     primitive_func = primitive_instance.get_function()
     vals = pd.Series(primitive_func(rolling_series_pd.index))
 
     assert vals.isna().sum() == expected_num_nams
+
+
+@pytest.mark.parametrize("primitive", [RollingCount, RollingMax, RollingMin, RollingMean])
+def test_rolling_primitives_non_uniform_data(primitive):
```
freddyaboulton, line 227: Might be worth it to actually pre-compute the expected series and check with `assert_series_equal`? My reasoning is that in two weeks when we ask ourselves how this works, it'll be easier/faster to just look at the expected answer in the unit tests. 

That being said, I stepped through this manually and the resulting series matches what I'd expect. Right now, the window size in this test only includes one observation so all the primitives return the same result. I think it'd be good if we test on some situations where there is more than one observation per window.

Maybe something like this could work:

```python
import pandas as pd
from featuretools.primitives import RollingMean, RollingCount

# When the data isn't uniform, this impacts the number of values in each rolling window
datetimes = (list(pd.date_range(start='2017-01-01', freq='1d', periods=3)) +
             list(pd.date_range(start='2017-01-10', freq='2d', periods=4)) +
             list(pd.date_range(start='2017-01-22', freq='1d', periods=7)))

no_freq_series = pd.Series(range(len(datetimes)), index=datetimes)
```

freddyaboulton (commented):
@tamargrey Thank you for the work on this! I agree with the implementation you chose where we compute window_end - gap to account for the gap. I left a comment on adding coverage but what you have here (and some local tests in a jupyter notebook) matches what I'd expect.

featuretools/tests/primitive_tests/test_rolling_primitive.py
tamargrey: @freddyaboulton both good ideas! I added two tests in 4f90dd2bd94b094d8cce556e454990761fe9b9de that test the primitives on a series that has windows with multiple observations present and use `assert_series_equal`. If the tests from this commit were along the lines of what you were thinking, I'll go ahead and update the other tests.

tamargrey: Also updated a test in `test_primitive_utils` to use the same approach for non uniform series: b0e5e25

test-requirements.txt
```diff
@@ -11,3 +11,4 @@ boto3>=1.17.46
 composeml>=0.3.0
 urllib3>=1.26.5
 pyarrow>=3.0.0
+mock==4.0.3
```
rwedge, line 14: Can we use the builtin version of mock instead?

https://docs.python.org/dev/library/unittest.mock.html#unittest.mock.patch

tamargrey: yep!

featuretools/tests/primitive_tests/test_primitive_utils.py
```diff
@@ -174,10 +218,15 @@ def test_roll_series_with_gap(window_length, gap, rolling_series_pd):
...
-    window_length = 3
-    gap = 0
-    actual_rolling = _roll_series_with_gap(rolling_series_pd, window_length, gap=gap).mean()
+@pytest.mark.parametrize(
+    "window_length",
+    [
+        3,
+        "3d"
+    ]
+)
```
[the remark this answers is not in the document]
tamargrey: collapsed to one line!

rwedge (approved):
LGTM!

PR #1810, opened 2021-12-16
Update cfm to use nullable types when needed

### LINKED ISSUE #1692: CFM with instance ids that are entirely after cutoff time raises TypeConversionError for non nullable logical types
When calculate feature matrix is called with a cutoff time that is a single datetime and instance ids, the former implementation would include all instance ids as rows in the resulting feature matrix, and if any instance ids were entirely after the cutoff time, that row in the feature matrix would be fully null.

In Woodwork, this causes issues in columns whose logical types are not nullable, for example, an `Integer` column.

If we look at the three transaction ids `[1, 2, 3]` in the transactions dataframe, we can see that `1` and `3` have a `transaction_time` that is after the cutoff time `"2014-1-1 04:00"`.

image

So if we use those transaction ids as instance ids with the `"2014-1-1 04:00"` cutoff time, the rows in the resulting DataFrame should be filled with nans.

```python

es = ft.demo.load_mock_customer(return_entityset=True, random_seed=0)

                      target_dataframe_name='transactions',
                      cutoff_time=pd.Timestamp("2014-1-1 04:00"),
                      instance_ids=[1,2,3] #1 and 3 are entirely after the cutoff time
                     )
```
However the following error is thrown when calculate feature matrix is executed in the above code: `TypeConversionError: Error converting datatype for session_id from type float64 to type int64. Please confirm the underlying data is consistent with logical type Integer.`

This is because the logical type of `session_id` is `Integer`, which cannot handle null values. 

We should either update the logical type in this case to `IntegerNullable` or consider dropping the rows altogether (This is what happens when no instance ids are supplied).

thehomebrewnerd:
### Update cfm to use nullable types when needed

Closes #1692 

Updates the CFM process to use Woodwork nullable logical types when needed, instead of the non nullable types. This is important for situations where the cutoff time settings cause null values to be introduced into the feature matrix for columns that previously did not have null values present.

### DIFF
--- a/.github/workflows/auto_approve_dependency_PRs.yml
+++ b/.github/workflows/auto_approve_dependency_PRs.yml
@@ -1,7 +1,7 @@
 name: Auto Approve Dependency PRs
 on:
   schedule:
-    - cron: '*/30 * * * *'
+    - cron: '*/5 * * * *'
   workflow_dispatch:
 jobs:
   build:
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -10,6 +10,7 @@ import cloudpickle
 import dask.dataframe as dd
 import numpy as np
 import pandas as pd
+from woodwork.logical_types import Age, AgeNullable, Boolean, BooleanNullable, Integer, IntegerNullable
 
 from featuretools.computational_backends.feature_set import FeatureSet
 from featuretools.computational_backends.feature_set_calculator import (
@@ -759,7 +760,27 @@ def update_progress_callback_parameters(progress_bar, previous_progress):
 
 
 def init_ww_and_concat_fm(feature_matrix, ww_init_kwargs):
+    cols_to_check = {col for col, ltype in
+                     ww_init_kwargs["logical_types"].items()
+                     if isinstance(ltype, (Age, Boolean, Integer))}
+    replacement_type = {
+        "age": AgeNullable(),
+        "boolean": BooleanNullable(),
+        "integer": IntegerNullable(),
+    }
     for fm in feature_matrix:
+        updated_cols = set()
+        for col in cols_to_check:
+            # Only convert types for pandas if null values are present
+            # Always convert for Dask/Koalas to avoid pulling data into memory for null check
+            is_pandas_df_with_null = isinstance(fm, pd.DataFrame) and fm[col].isnull().any()
+            is_dask_df = isinstance(fm, dd.DataFrame)
+            is_koalas_df = is_instance(fm, ks, 'DataFrame')
+            if is_pandas_df_with_null or is_dask_df or is_koalas_df:
+                current_type = ww_init_kwargs["logical_types"][col].type_string
+                ww_init_kwargs["logical_types"][col] = replacement_type[current_type]
+                updated_cols.add(col)
+        cols_to_check = cols_to_check - updated_cols
         fm.ww.init(**ww_init_kwargs)
 
     if any(isinstance(fm, dd.DataFrame) for fm in feature_matrix):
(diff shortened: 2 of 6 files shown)


featuretools/tests/synthesis/test_dask_dfs.py
```diff
@@ -55,7 +55,7 @@ def test_single_table_dask_entityset():
 
     # Use the same columns and make sure both indexes are sorted the same
     dask_computed_fm = dask_fm.compute().set_index('id').loc[fm.index][fm.columns]
-    pd.testing.assert_frame_equal(fm, dask_computed_fm)
+    pd.testing.assert_frame_equal(fm, dask_computed_fm, check_dtype=False)
```
thehomebrewnerd, line 58: I updated these test checks to ignore the dtypes, because this update means we can no longer expect the dask and koalas feature matrices to have the same dtypes for all the columns. With pandas, the logical types would only change when they need to, but with Dask and Koalas they would always end up using the nullable types in the feature matrix.

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -759,7 +767,27 @@ def update_progress_callback_parameters(progress_bar, previous_progress):
 
 
 def init_ww_and_concat_fm(feature_matrix, ww_init_kwargs):
+    cols_to_check = {col for col, ltype in
+                     ww_init_kwargs["logical_types"].items()
+                     if isinstance(ltype, (Age, Boolean, Integer))}
+    replacement_type = {
+        "age": AgeNullable(),
+        "boolean": BooleanNullable(),
+        "integer": IntegerNullable(),
```
tamargrey, line 776: Could we use the logical type strings for the values as well here?

rwedge:
Have we done a performance test run on the branch?

thehomebrewnerd:
> 

@rwedge Just now launched a run to compare commit `d01e7c7` against the latest release.

featuretools/computational_backends/calculate_feature_matrix.py
thehomebrewnerd: @tamargrey  Yeah, I think we could use the strings instead without issue, but I did it this way to be consistent with the other values in the logical types dictionary which are all `LogicalType` instances. I'm not sure that's important though. Do you have a preference here, or is there an advantage to using the strings other than avoiding an import?

thehomebrewnerd:
> > 
> 
> @rwedge Just now launched a run to compare commit `d01e7c7` against the latest release.

Pandas performance didn't change much with this update. 
[d01e7c7c3_v1.3.0__2022-01-06T14.16.37.733426.html.zip](https://github.com/alteryx/featuretools/files/7823007/d01e7c7c3_v1.3.0__2022-01-06T14.16.37.733426.html.zip)

featuretools/computational_backends/calculate_feature_matrix.py
tamargrey: was just thinking to avoid the import, but I think keeping the types consistent with the rest of the dictionary is a fine reason to use the objects here. It'd make our lives easier if we ever try using this dictionary for anything further

tamargrey (dismissed):
LGTM! Also I confirmed these changes allow us to use a `Lag` primitive where nans are introduced on columns that have non nullable logical types!

.github/workflows/auto_approve_dependency_PRs.yml
```diff
@@ -1,7 +1,7 @@
 name: Auto Approve Dependency PRs
 on:
   schedule:
-    - cron: '*/30 * * * *'
+    - cron: '*/5 * * * *'
```
gsheni, line 4: I asked Nate for this change (rather than putting up a 1 line PR). I want this workflow to run more often.

thehomebrewnerd:
> LGTM! Also I confirmed these changes allow us to use a `Lag` primitive where nans are introduced on columns that have non nullable logical types!

@tamargrey Nice - thanks for checking that out!

PR #1815, opened 2021-12-27
Add Datetime transform primitives - IsRushHour, IsNoonHour, IsNightHour

gsheni:


gsheni:
Will hold off on these for now, as it's not clear they would make for a better model.

PR #1816, opened 2021-12-27
[DO NOT MERGE] Separate offset gap logic from numeric gap object

tamargrey:


featuretools/primitives/utils.py
```diff
@@ -379,3 +412,23 @@ def _check_rolling_inputs(window_size, gap):
...
         except ValueError:
             raise ValueError(f"Cannot roll series. Gap, {gap}, is not a valid offset alias.")
+
+
+def _check_window_size(window_size):
+    if isinstance(window_size, str):
+        try:
+            to_offset(window_size)
+        except ValueError:
+            raise ValueError(f"Cannot roll series. Window length, {window_size}, is not a valid offset alias.")
```
davesque, line 422: I see in the body of this function, we're wrapping the entire function body in this if statement.  Would it make more sense to begin the function with a guard like this:
```suggestion
    if not isinstance(window_size, str):
        raise TypeError("...")
    try:
        to_offset(window_size)
    except ValueError:
        raise ValueError(f"Cannot roll series. Window length, {window_size}, is not a valid offset alias.")
```
The main difference in behavior would be that the function would fail loudly if a value of the wrong type was passed in.  Or was it intentional that we're essentially ignoring values of the wrong type?

tamargrey: It was intentional! `window_size` can be either an integer or a string; it's only when `gap` is a string that `window_size` is restricted to also being a string. This check is more about confirming that strings are offset strings, so maybe the function should be named something more appropriate like `_check_rolling_offset_strings`.

That being said, you're right that there's no overall checking that the inputs are the type that we expect, so we definitely could add additional checks that confirm that, if they're not strings, they're integers, otherwise raise that `TypeError`.

davesque: Thanks for clarifying!  Yeah, I'd be in favor of adding the more thorough type check just so things are more clear to future readers of the code.

davesque: And I'm not sure but there may also be other places in this MR where the same changes could be made.

tamargrey: Sounds good! I'll go ahead and add some explicit type checks into this PR, and if we decide to not go with this implementation, I'll make sure they get into the original PR (#1809) as well

tamargrey: On second thought, gI'm going to wait until we've decided whether or not to adopt the changes in this PR to add the type checking, so putting this here for now:

- Confirm gap is a string or integer >= 0
- Confirm window length is a string or integer >= 0 
- confirm that min_periods is an integer >= 0

Also, might be worth checking these on the primitives themselves instead of just the utils, to catch potential places where we might have some operation in the primitive itself that would fail before these checks get called in the utils.

davesque: > Also, might be worth checking these on the primitives themselves instead of just the utils, to catch potential places where we might have some operation in the primitive itself that would fail before these checks get called in the utils.

Yep, makes sense.  Although I still think it might be good to include value checks as close as possible to the code for which it makes a difference.  So if one of the utility functions only works for certain types of values, the checks should be in that function.  And if there are some constraints on values for logic that exists inside of methods on the actual primitive, the checks should be in those methods.  It's all for the sake of readability and to avoid the situation where some check doesn't make sense unless you look at code in some other module or function.

davesque: That is of course unless it's really clear that those functions are only used in a single part of the code.  Then I guess you can rely on checks that were performed further up the call stack.

davesque: Anyway, I'll leave it to you :).  I think you probably get what I mean.

tamargrey: @davesque Ended up making these changes over on the original PR: ca4eb469d6e94be6fabf0d5795d0bc9e8e394f74

PR #1821, opened 2021-12-28
Fix setting of environment variables for auto approve dependency workflow

gsheni:


gsheni:
Proof the first half the workflow works:
- https://github.com/alteryx/featuretools/runs/4653371031?check_suite_focus=true

tamargrey (approved):
lgtm assuming tests pass

PR #1825, opened 2022-01-03
Fix s3 credentials test issue

rwedge:
Some unit tests were failing to download public files if existing AWS credentials were detected

This PR shifts testing S3 deserializing with credentials to only in tests with mocked S3 fixtures, leaving the real S3 tests that do not use credentials


davesque (approved):
Not super familiar with that test suite but LGTM!

PR #1833, opened 2022-01-07
Increase minimum numpy to 1.21.0 and remove fastparquet

gsheni:
- Removed `fastparquet`, as it cannot handle nullable pandas dtypes.
- Removed `pyyaml`, as it will be handled by dask
- Bump minimum numpy to 1.21.0
- Bump minimum koalas to 1.8.1 (due to more recent numpy req)
- Bump minimum scipy to 1.3.3 (due to more recent numpy req)


featuretools/tests/requirement_files/latest_requirements.txt
```diff
@@ -2,9 +2,10 @@ click==8.0.3
 cloudpickle==2.0.0
 dask==2021.12.0
 distributed==2021.12.0
-numpy==1.20.3
+numpy==1.21.5
 pandas==1.3.5
 psutil==5.9.0
+PyYAML==6.0
```
thehomebrewnerd, line 8: Remove?

rwedge:
Last time we updated PyYAML was due to a security vulnerability and dask does not set a min requirement,  should we keep the requirement for now / suggest dask set a min requirement?

featuretools/tests/requirement_files/latest_requirements.txt
gsheni: Fixed.

gsheni:
@rwedge If we are not using a dependency in our library, then its the other library's (dask) job to manage that dependency.

rwedge:
Do we want to remove PyYAML from the `show_info` command?

https://github.com/alteryx/featuretools/blob/9814869c7ec50d0313dd590b079c37c0fd24765a/featuretools/utils/cli_utils.py#L12-L14

https://github.com/alteryx/featuretools/blob/9814869c7ec50d0313dd590b079c37c0fd24765a/featuretools/tests/utils_tests/test_cli_utils.py#L32-L35

gsheni:
@rwedge good catch. Removed.

rwedge (approved):
LGTM

PR #1837, opened 2022-01-07
Update auto approve dependency PR to run on workflow completion

gsheni:
- This will make it so that when a workflow (unit tests) completes on a branch, the auto approve workflow will run.

### DIFF
--- a/.github/workflows/auto_approve_dependency_PRs.yml
+++ b/.github/workflows/auto_approve_dependency_PRs.yml
@@ -1,7 +1,5 @@
 name: Auto Approve Dependency PRs
 on:
-  schedule:
-    - cron: '*/5 * * * *'
   workflow_dispatch:
   workflow_run:
     workflows: ["Unit Tests - Latest Dependencies"]
--- a/.github/workflows/auto_approve_dependency_PRs.yml
+++ b/.github/workflows/auto_approve_dependency_PRs.yml
@@ -3,6 +3,13 @@ on:
   schedule:
     - cron: '*/5 * * * *'
   workflow_dispatch:
+  workflow_run:
+    workflows: ["Unit Tests - Latest Dependencies"]
+    branches:    
+      - 'latest-dep-update-[a-f0-9]+'
+      - 'min-dep-update-[a-f0-9]+'
+    types:
+      - completed
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/auto_approve_dependency_PRs.yml
```diff
@@ -3,6 +3,13 @@ on:
   schedule:
     - cron: '*/5 * * * *'
   workflow_dispatch:
+  workflow_run:
```
gsheni, line 6: https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#workflow_run

.github/workflows/auto_approve_dependency_PRs.yml
```diff
@@ -3,6 +3,13 @@ on:
...
+    workflows: ["Unit Tests - Latest Dependencies"]
+    branches:    
+      - 'latest-dep-update-[a-f0-9]+'
```
gsheni, line 9: Need a double check on this syntax
- https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet
```
dep_update = '^latest-dep-update-[a-f0-9]{7}$'
min_dep_update = '^min-dep-update-[a-f0-9]{7}$'
```

.github/workflows/auto_approve_dependency_PRs.yml
rwedge: Can we remove the "schedule" event and rely on workflow_run only?

.github/workflows/auto_approve_dependency_PRs.yml
rwedge: Not sure if the github syntax supports `[a-f0-9]{7}` but we could put `[a-f0-9]` 7 times in a row . . .

gsheni: That's why I put `+`, which means `+: Matches one or more of the preceding character.`.

.github/workflows/auto_approve_dependency_PRs.yml
gsheni: Removed

rwedge (approved):
LGTM

Issue #1761, opened 2021-10-28
Allow users to pass camelCase to DFS for primitives


gsheni:
- We currently allow primitives name (as strings) to be passed to DFS. `add_numeric`
- We should modify this behavior to allow camelCase (which should match `primitive.__name__`) (`addNumeric`)
  - This is different than TitleCase (`AddNumeric`)
  - If possible, we should make it work with both. 
- Users can still pass primitive names.
- Eventually we will phase out primitive names (https://github.com/alteryx/featuretools/issues/1750)

thehomebrewnerd:
@gsheni Was looking at this more closely. I think `primitive.__name__` returns the TitleCase version of the primitive - the string equivalent of the class name:

```>>> import featuretools as ft
>>> ft.primitives.standard.IsNull.__name__
'IsNull'
>>> ft.primitives.standard.AddNumeric.__name__
'AddNumeric'
```

gsheni:
@thehomebrewnerd I see. Is your point that we don't store the camelCase anywhere?

gsheni:
My opinion would be yes we do. It makes it less likely that a Featuretools user will error when using DFS.

gsheni:
Completed in #1854

PR #1841, opened 2022-01-11
make primitive string names case and underscore insensitive

tuethan1999:
### Pull Request Description
#1761 

### Feature Request
Provide convenience to a user by allowing them to pass a primitive name in either snake_case, TitleCase or camelCase

### Implementation
make the string to primitive matching case and underscore insensitive. Therefore `num_true`, `NumTrue`, `numTrue` and even `nU_mT_ru_e` all boil down to `numtrue`.


davesque (commented):
Just want to plug my concerns again that we're enabling too broad a range of behaviors with this kind of functionality (previous discussion for reference: #1733).  But thanks all the same @tuethan1999 for getting it done. :)

Just had the one comment about a possible change.

May as well provide a link to the previous discussion for reference.

featuretools/tests/synthesis/test_dfs_method.py
```diff
@@ -339,6 +339,29 @@ def test_no_warns_with_different_case_string_primitives(es):
         assert not record
 
 
+def test_no_warns_with_and_without_snakecase_string_primitives(es):
+    # Should not raise a UnusedPrimitiveWarning warnings
+    for trans_primitive in ["is_null", "isNull", "IsNull", "_iSnu__Ll_", "i_s_n_u_L_l"]:
```
thehomebrewnerd, line 344: Am I understanding this correctly that we will accept `i_s_n_u_l_l` as a valid primitive name with this change? 

If so, is that what we intend here? That seems a bit too flexible to me.

thehomebrewnerd: If we don't want to break the current behavior, I think we need to allow `is_null`, but beyond that, it seems like the only other things we should match on is the exact value returned by `primitive.__name__` (`IsNull`). I'm not even sure we should match on camel case, but that just my two cents here.

tuethan1999:
I think we shouldn't try to support both `primitive.name` and `primitive.__name__` for the case where the customizable `primitive.name` conflicts with another primitive's `primitive.__name__`. Consider the case:

```python
class NumUnique():
  name = "foobar"

class UniqueCount():
  name = "num_unique"
```

PR #1848, opened 2022-01-18
Add DateToHoliday transform primitive

### LINKED ISSUE #1846: Add DateToHoliday primitive
- Transforms time of an instance into the holiday name, if there is one.

dvreed77:
- Fixes #1846


dev-requirements.txt
```diff
@@ -14,3 +14,4 @@ pydata_sphinx_theme==0.7.1
 nlp_primitives>=2.0.0
 autonormalize >= 1.0.2
 -r koalas-requirements.txt
+holidays>=0.12
```
thehomebrewnerd, line 17: You are going to want this to be in `requirements.txt` as most users wouldn't install the requirements in `dev-requirements.txt` - those are primarily for developers contributing to the codebase.

docs/source/release_notes.rst
```diff
@@ -9,6 +9,7 @@ Future Release
     * Fixes
     * Changes
         * Add autonormalize as an add-on library (:pr:`1840`)
+        * Add Date to Holiday Transform Primitive (:pr: `1846`)
```
thehomebrewnerd, line 12: Minor detail here, but I would stylize this as DateToHoliday - use the class name without spaces.  See line 24 below for an example.  Also be sure to add your username to the list of contributors to the release on line 19.

dev-requirements.txt
dvreed77: yes, of course. It should be in both place tho, correct?

thehomebrewnerd: @dvreed77 This requirement only needs to be in one place. The typical path for installing requirements for development would be to run `make installdeps`. After upgrading pip, that make command then installs featuretools in editable mode, and that process will grab everything from `requirements.txt`. Finally, the requirements from `dev-requirements.txt` are installed by the make command, which also includes references to the `test-requirements.txt` and `koalas-requirements.txt` files, so essentially `make installdeps` installs everything found in all of our requirements files.

dvreed77: @thehomebrewnerd should I update `contributing.md` to run `make installdeps`? Guide currently has this:

```bash
  virtualenv venv
  source venv/bin/activate
  python -m pip install -e .
  python -m pip install -r dev-requirements.txt
  ```

File: docs/source/release_notes.rst (hunk 2 above)
Comment (dvreed77) [reply docs/source/release_notes.rst]: Fixed

File: dev-requirements.txt (hunk 1 above)
Comment (thehomebrewnerd) [reply dev-requirements.txt:17]: Yeah, I think we should replace the `python*` commands with `make installdeps`. The only real difference is that `make installdeps` will also make sure pip is upgraded first, which should be a good thing to have first.

File: docs/source/release_notes.rst [hunk 3]
```diff
@@ -9,6 +9,7 @@ Future Release
     * Fixes
     * Changes
         * Add autonormalize as an add-on library (:pr:`1840`)
+        * Add DateToHoliday Transform Primitive (:pr: `1846`)
```
tamargrey, line 12: Seems like the release notes ci check is failing because the PR number is slightly wrong--should be 1848!

dvreed77: ahh thanks! I was wondering what was wrong here

dev-requirements.txt
dvreed77: @thehomebrewnerd I added `holiday` package back to dev-requirements because it looks like github actions uses this for install before running unit tests

featuretools/tests/requirement_files/minimum_test_requirements.txt
```diff
@@ -11,3 +11,4 @@ boto3==1.17.46
 composeml==0.3.0
 urllib3==1.26.5
 pyarrow==3.0.0
+holidays>=0.12
```
thehomebrewnerd, line 14: Hmm...I don't think you should need this here. These should just be packages that are needed for running the tests, but not already included in the other requirements files .

dvreed77: CleanShot 2022-01-19 at 10 45 15

This test was failing

dvreed77: last line does a pip instal from `minimum_test_requirements.txt`

thehomebrewnerd: I think the holidays install for these tests should get handled a little lower in this file. It will try to install requirements from `minimum_core_requirements.txt` or `minimum_koalas_requirements.txt` depending on the version of the test being run, so as long as you have `holidays` included in those two files I think you should be good.

image

thehomebrewnerd: One other point of clarification here, in the `minimum_*` files, we will want to pin the version to a specific number to make sure we always get that lowest version installed, instead of using `>=version` syntax which would allow newer versions to be installed.

docs/source/release_notes.rst
```diff
@@ -9,6 +9,7 @@ Future Release
     * Fixes
     * Changes
         * Add autonormalize as an add-on library (:pr:`1840`)
+        * Add DateToHoliday Transform Primitive (:pr:`1848`)
```
thehomebrewnerd, line 12: You should add your username to the list of contributors on line 19.

dev-requirements.txt
thehomebrewnerd: I still think this should be removed from dev requirements and should be covered by the other requirements installs.

thehomebrewnerd (commented):
Just a couple small things yet to clean up, but overall looking good.

featuretools/tests/primitive_tests/test_datetoholiday_primitives.py
```diff
@@ -0,0 +1,30 @@
...
+        datetime(2016, 2, 27),
+        datetime(2017, 5, 29, 10, 30, 5),
+        datetime(2018, 7, 4)])
+
+    holiday_series = date_to_holiday(dates).tolist()
+
+    assert holiday_series[0] == "New Year's Day"
+    assert np.isnan(holiday_series[1])
+    assert holiday_series[2] == "Memorial Day"
+    assert holiday_series[3] == "Independence Day"
```
gsheni, line 24: Do we want to add some additional tests?

```
    def test_nat(self):
        primitive_instance = self.primitive()
        case = pd.Series([
            '2019-10-14',
            'NaT',
            '2016-02-15',
            'NaT',
        ]).astype('datetime64')
        answer = ['Columbus Day', np.nan, "Washington's Birthday", np.nan]
        given_answer = primitive_func(case).astype('str')
        np.testing.assert_array_equal(given_answer, answer)

    def test_valid_country(self):
        primitive_instance = self.primitive(country='Canada')
        case = pd.Series([
            '2016-07-01',
            '2016-11-11',
            '2017-12-26',
            '2018-09-03',
        ]).astype('datetime64')
        answer = ['Canada Day', np.nan, 'Boxing Day', 'Labour Day']
        given_answer = primitive_func(case).astype('str')
        np.testing.assert_array_equal(given_answer, answer)

    def test_multiple_countries(self):
        primitive_mexico = self.primitive(country='Mexico')
        primitive_func = primitive_mexico.get_function()
        case = pd.Series([datetime(2000, 9, 16),
                          datetime(2005, 1, 1)])
        assert len(primitive_func(case)) > 1
        primitive_india = self.primitive(country='IND')
        primitive_func = primitive_mexico.get_function()
        case = pd.Series([datetime(2048, 1, 1),
                          datetime(2048, 10, 2)])
        primitive_func = primitive_india.get_function()
        assert len(primitive_func(case)) > 1
        primitive_uk = self.primitive(country='UK')
        primitive_func = primitive_uk.get_function()
        case = pd.Series([datetime(2048, 3, 17),
                          datetime(2048, 4, 6)])
        assert len(primitive_func(case)) > 1
        countries = ['Argentina', 'AU', 'Austria', 'BY', 'Belgium',
                     'Brazil', 'Canada', 'Colombia', 'Croatia',
                     'England', 'Finland', 'FRA', 'Germany',
                     'Germany', 'Italy', 'NewZealand', 'PortugalExt',
                     'PTE', 'Spain', 'ES', 'Switzerland', 'UnitedStates',
                     'US', 'UK', 'UA', 'CH', 'SE', 'ZA']
        for x in countries:
            self.primitive(country=x)
```

dvreed77: yes, thank you

gsheni (approved):

thehomebrewnerd (approved):
LGTM!

PR #1853, opened 2022-01-20
Add DistanceToHoliday transform primitive

### LINKED ISSUE #1847: Add DistanceToHoliday primitive
- Computes the number of days before or after a given holiday

dvreed77:
- fixes #1847


featuretools/primitives/standard/transform_primitive.py
```diff
@@ -1078,3 +1080,85 @@ def date_to_holiday(x):
             df = df.merge(holidays_df, how='left')
             return df.names.values
         return date_to_holiday
+
+
+class DistanceToHoliday(TransformPrimitive):
```
gsheni, line 1085: Should we try to move some of this into a utility? Since it uses similar functionality to our other primitive (DateToHoliday)?

We did this with CountString & WhiteSpaceCount
- https://github.com/alteryx/nlp_primitives/blob/main/nlp_primitives/whitespace_count.py#L6
- https://github.com/alteryx/nlp_primitives/blob/main/nlp_primitives/count_string.py#L9

We have a primitive util file too:
- https://github.com/alteryx/featuretools/blob/main/featuretools/primitives/utils.py

dvreed77: @gsheni to me, there seems to be enough difference between them that additional abstraction isn't really needed.

gsheni: Gotcha, sounds good.

dvreed77: @gsheni upon second look I think you had a good point. Let me know what you think

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -1059,22 +1057,89 @@ class DateToHoliday(TransformPrimitive):
...
+    input_types = [ColumnSchema(logical_type=Datetime)]
+    return_type = ColumnSchema(semantic_tags={'numeric'})
+    default_value = 0
+
+    def __init__(self, holiday="New Year's Day", country="US"):
+        self.country = country
+        self.holiday = holiday
+        self.holidayUtil = HolidayUtil(country)
+
+        available_holidays = list(set(self.holidayUtil.federal_holidays.values()))
```
gsheni, line 1123: Do we need to do `self` here?
```suggestion
        self.available_holidays = list(set(self.holidayUtil.federal_holidays.values()))
```

featuretools/primitives/utils.py
```diff
@@ -293,3 +294,21 @@ def _haversine_calculate(lat_1s, lon_1s, lat_2s, lon_2s, unit):
...
+            available_countries = 'https://github.com/dr-prodigy/python-holidays#available-countries'
+            error = 'must be one of the available countries:\n%s' % available_countries
+            raise ValueError(error)
+
+        self.federal_holidays = getattr(holidays, country)(years=range(1950, 2100))
+
+    def to_df(self):
+        holidays_df = pd.DataFrame(sorted(self.federal_holidays.items()),
+                                   columns=['holiday_date', 'names'])
+        holidays_df.holiday_date = holidays_df.holiday_date.astype('datetime64')
```
gsheni, line 313: @thehomebrewnerd @rwedge Do we need to do this `astype` or will Featuretools automatically do this?

featuretools/primitives/standard/transform_primitive.py
gsheni: yep, looks good.

featuretools/primitives/standard/transform_primitive.py
dvreed77: its not used outside of __init__ so I don't think so

featuretools/tests/primitive_tests/test_distancetoholiday_primitives.py
```diff
@@ -0,0 +1,62 @@
...
+        'NaT',
+        '2012-05-31',
+        'NaT',
+    ]).astype('datetime64')
+    answer = [0, np.nan, -151, np.nan]
+    given_answer = date_to_holiday(case).astype('float')
+    np.testing.assert_array_equal(given_answer, answer)
+
+
+def test_valid_country():
```
thehomebrewnerd, line 52: I think the original version of this test was perhaps testing something slightly different since it was named `test_holiday_out_of_range`, but I'm not exactly clear now what the intent of that test was. Do you have any ideas? I see the original test introduced `np.nan` values in the result, but this one does not.

featuretools/tests/primitive_tests/test_distancetoholiday_primitives.py
```diff
@@ -0,0 +1,62 @@
...
+def test_distanceholiday():
+    distance_to_holiday = DistanceToHoliday("New Year's Day")
+    dates = [datetime(2010, 1, 1),
+             datetime(2012, 5, 31),
+             datetime(2017, 7, 31),
+             datetime(2020, 12, 31)]
+
+    distance_list = distance_to_holiday(dates).tolist()
+
+    assert distance_list[0] == 0
```
thehomebrewnerd, line 21: Minor suggestion, but can we create a list of expected values and assert the same way we do in `test_nat` and `test_valid_country` instead of checking each value individually?

featuretools/tests/primitive_tests/test_distancetoholiday_primitives.py
dvreed77: @thehomebrewnerd I dug into this a bit. The test above isn't replacing that test, I actually didn't see the original tests when moving things over. 

This out of range test brought to my attention some unexpected behavior when using the holidays library. It appears the observed holidays are only present in the dataset. For example, the test date of 2010-1-1 results in a NaN because there is no holiday named "Boxing Day" in the data set for 2009 and 2010 since they fall on Saturday and Sunday respectively. Instead, the dataset contains "Boxing Day (Observed)".

I think this is confusing behavior in my opinion

thehomebrewnerd (approved):

thehomebrewnerd (approved):

PR #1854, opened 2022-01-20
Allow TitleCase and camelCase

tuethan1999:
### Pull Request Description
issue: #1761 
implementation that allowed all variations of case and underscore: #1841

### Feature Request
Provide convenience to a user by allowing them to pass a primitive name in either snake_case, TitleCase or camelCase

### Implementation
Reduce TitleCase and camelCase into snake_case before matching them internally. This means we can't accept primitive names like "cOuNt" because it will be converted to "c_ou_nt". I think this is the desired behavior though

Assumptions:
- primitive name is camel case


featuretools/primitives/utils.py
```diff
@@ -40,6 +41,11 @@ def get_transform_primitives():
     return {prim.name.lower(): prim for prim in transform_primitives}
 
 
+def camel_and_title_to_snake(name):
+    name = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', name)
+    return re.sub('([a-z0-9])([A-Z])', r'\1_\2', name).lower()
```
rwedge, line 46: I would expect "Top3Words" to be converted to "top_3_words" but currently this will return "top3_words"

rwedge:
A test just for the `camel_and_title_to_snake` method with a few inputs could be good

tamargrey:
Is there somewhere in the documentation that we can put the new constraints on what primitive strings will be recognized? I guess we never specifically mentioned that they had to me the same as `primitve.name` in the past, but now that there's more than one valid option, it might be nice to have the information written down somewhere.

tamargrey:
@dvreed77 Not sure if we need to handle this case, but wanted to note a potentially problematic string for `camel_and_title_to_snake`: `"NaN"`. Something like `ft.utils.gen_utils.camel_and_title_to_snake('hasNaNs')` results in `has_na_ns`.

I think we can avoid dealing with this specific example if we just use the word "Null" an time we might use "nan" in a primitive name, but it's something to keep in mind bc we can't stop users from trying to pass in that string

featuretools/tests/utils_tests/test_gen_utils.py
```diff
@@ -68,3 +69,16 @@ def test_list_semantic_tags():
     ft_semantic_tags = ft.list_semantic_tags()
     ww_semantic_tags = list_semantic_tags()
     assert ft_semantic_tags.equals(ww_semantic_tags)
+
+
+def test_camel_and_title_to_snake():
```
rwedge, line 74: can we include a snake case test case to confirm snake case passed in is unaltered?

dvreed77: done

rwedge (dismissed):
LGTM

rwedge (approved):

PR #1860, opened 2022-01-22
Split transform primitives into separate files

jacobboney:
### Pull Request Description

I starting by only splitting off longitude and latitude so I could get my bearings in the code base. I took those classes and moved them into their own file "lat_long_primitive.py" and also importing their dependencies. I those classed from "transform_primitive.py" as well as added the new file into the api. 

As far as I can tell everything seems to be pointing to each other so I don't believe there is any disconnect. However, when running "make test" I am getting a sorted/formatting error. I am uncertain if this is due to the file I created or there is some issue with the "get_transform_primitives()" function in "utils.py".

I apologize in advance if this is a simple oversight on my part. I'm still in the process of learning and any advice or direction would be appreciated. Thank you!

*Additionally, I realize I don't need to commit as much as I did, still getting the hang of version control.

-----


jacobboney:
*UPDATE*
I was able to find and fix my issue. It was a conflict with the imports. I solved the problem with isort python utility. I was able to run the tests to the finish. I'll now work on splitting more primitives from "transform_primitive.py". 

If there is a preference for separation feel free to let me know, otherwise, I'll use my best judgement and we can review the changes when finished.

Issue #1740, opened 2021-10-14
Adding unnamed DataFrames to EntitySets (and have Featuretools generate a name for df)


chukarsten:
### Adding unnamed DataFrames to EntitySets

-----

#### Bug/Feature Request Description

Currently, in EvalML's [DFSTransformer](https://github.com/alteryx/evalml/blob/main/evalml/pipelines/components/transformers/preprocessing/featuretools.py), we run into issues with un-named dataframes being passed to the DFSTransfomer's .fit() method.  

Calling the DFSTransformer.fit() results in the dataframe being added to an EntitySet() via the _make_entity_set() method which uses FT 1.0.0's new add_dataframe() method.  add_dataframe() now requires either:
1.) a dataframe that is not woodwork initialized, whereupon it initializes it and names it according to a parameter passed in via add_dataframe()
2.) a dataframe that is woodwork initialized and named.

EvalML currently supports woodwork initialized dataframes that are unnamed.

#### Expected Output

We'd suggest the following changes:

```
        if dataframe.ww.schema is None:
            ...

        else:
            if dataframe.ww.name is None:
                if dataframe_name is None:
                    raise ValueError('Cannot add a Woodwork DataFrame to EntitySet without a name or a proposed name')
                else:
                    dataframe.ww.name = dataframe_name
            if dataframe.ww.index is None:
                if index is None or not make_index:
                    raise ValueError('Cannot add Woodwork DataFrame to EntitySet without index')
                else:
                    # do index stuff
```
Hopefully we could do something similar to accommodate woodwork initialized dataframes and just give them a name, per the add_dataframe() function.

#### Output of ``featuretools.show_info()``

Featuretools version: 1.0.0

gsheni:
@chukarsten Is something we need to prioritize? Is it negatively affecting EvalML?

PR #1862, opened 2022-01-24
Add __setitem__ method on EntitySet to overload add_dataframe method

### LINKED ISSUE #1685: Add dataframe with entityset setter
- With Featuretools 1.0.0 we add a dataframe to an EntitySet with the following:
```
es = ft.EntitySet('new_es')

es.add_dataframe(dataframe=orders_df,
                 dataframe_name='orders',
                 index='order_id',
                 time_index='order_date')
```
# Improvement
- However, you could also change the EntitySet setter to add it with this approach:
```
es = ft.EntitySet('new_es')
es['orders'] = orders_df
```
- In this case, Featuretools would assume the 1st column in the dataframe is the index (current behavior in Feature < 1.0.0)
- The time index would be None
- already_sorted would be False

dvreed77:
### Pull Request Description
- Added a `__setitem__` method on EntitySet to overload `add_dataframe` method
- fixes #1685


thehomebrewnerd (commented):
I'm thinking we will want to update [this guide](https://featuretools.alteryx.com/en/stable/getting_started/using_entitysets.html#Adding-dataframes) in the documentation to discuss this approach to adding a dataframe to an EntitySet. 

As part of this, we should mention the limitations of this approach, namely that if the dataframe does not have Woodwork initialized, the first column will be assumed to be the index column and all column types will be inferred by Woodwork. If users want to set the time index or have control over logical types, they should first initialize Woodwork on the dataframe with the proper parameters before assigning the dataframe to the EntitySet with this method.

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -8,6 +8,7 @@
 import numpy as np
 import pandas as pd
 import pytest
+from mock import patch
```
rwedge, line 11: mock is part of the standard library now

`from unittest.mock import patch`

docs/source/release_notes.rst
```diff
@@ -35,6 +35,7 @@ v1.4.0 Jan 10, 2022
         * Skip code coverage for specific dask usage lines (:pr:`1829`)
         * Increase minimum required numpy version to 1.21.0, scipy to 1.3.3, koalas to 1.8.1 (:pr:`1833`)
         * Remove pyyaml as a requirement (:pr:`1833`)
+        * Adds `__setitem__` method to overload `add_dataframe` method on EntitySet (:pr:`1862`)
```
thehomebrewnerd, line 38: This needs to move to the Future Release section above. Also, in order to get code formatting to work correctly, you will need to use double back-tics around the methods.

docs/source/release_notes.rst
```diff
@@ -45,7 +46,7 @@ v1.4.0 Jan 10, 2022
         * Remove fastparquet as a test requirement (:pr:`1833`)
 
     Thanks to the following people for contributing to this release:
-    :user:`davesque`, :user:`gsheni`, :user:`rwedge`, :user:`thehomebrewnerd`
+    :user:`davesque`, :user:`gsheni`, :user:`rwedge`, :user:`thehomebrewnerd`, :user:`dvreed77`
```
thehomebrewnerd, line 49: Move to Future Release instead

docs/source/getting_started/using_entitysets.ipynb
```diff
@@ -71,7 +71,19 @@
     "\n",
     "* The ``index`` parameter specifies the column that uniquely identifies rows in the dataframe.\n",
     "* The ``time_index`` parameter tells Featuretools when the data was created.\n",
-    "* The ``logical_types`` parameter indicates that \"product_id\" should be interpreted as a Categorical column, even though it is just an integer in the underlying data."
+    "* The ``logical_types`` parameter indicates that \"product_id\" should be interpreted as a Categorical column, even though it is just an integer in the underlying data.\n",
+    "\n",
+    "You can also use a setter on the ``EntitySet`` object to add dataframes\n",
```
thehomebrewnerd, line 76: I think it might flow better if we move this a little lower in the section, after we demonstrate how to use `add_dataframe`.

Also, what do you think about putting this in a Note callout, like we do below with the description of `plot`?

docs/source/getting_started/using_entitysets.ipynb
```diff
@@ -71,7 +71,19 @@
...
+    "\n",
+    "```python\n",
+    "es[\"transactions\"] = transactions_df\n",
+    "```\n",
+    "\n",
+    "Note, that this will use the default implementation of `add_dataframe`, notably the following:\n",
+    "\n",
+    "* if the DataFrame does not have [Woodwork](https://woodwork.alteryx.com/) initialized, the first column will be the index column\n",
+    "* if the DataFrame does not have Woodwork initialized, all columns will be inferred by Woodwork.\n",
+    "* control over time index and logical types will need be initialized through Woodwork before adding."
```
thehomebrewnerd, line 86: Suggested rewording:
`if control over the time index column and logical types is needed, Woodwork should be initialized before adding the dataframe`

thehomebrewnerd (commented):
Nothing major, just a couple things to clean up in release notes and a couple suggestions for the documentation.

docs/source/getting_started/using_entitysets.ipynb
dvreed77: I tried to put this into a note callout, but they don't support code blocks or lists. I guess I could reword it

thehomebrewnerd: Ah, ok, didn't realize that. No big deal on the note formatting, but if we can try somehow separate it from the flow of the example, that might be good.

dvreed77: Actually lists work, but can't get codeblocks to work. I reformatted it

docs/source/getting_started/using_entitysets.ipynb
dvreed77: fixed

thehomebrewnerd (approved):
Looks good to me, but might be good to get someone else to review and approve before merging, just to make sure I'm not forgetting something that we wanted to do with this update.

docs/source/getting_started/using_entitysets.ipynb
```diff
@@ -104,9 +111,22 @@
    "source": [
     ".. currentmodule:: featuretools\n",
     "\n",
+    "\n",
+    ".. note ::\n",
+    "\n",
+    "    You can also use a setter on the ``EntitySet`` object to add dataframes\n",
```
tamargrey, line 117: Repeat of the sentence above?

docs/source/getting_started/using_entitysets.ipynb
```diff
@@ -104,9 +111,22 @@
...
+    "\n",
+    "    ``es[\"transactions\"] = transactions_df``\n",
+    "\n",
+    "    that this will use the default implementation of `add_dataframe`, notably the following:\n",
+    "\n",
+    "    * if the DataFrame does not have `Woodwork <https://woodwork.alteryx.com/>`_ initialized, the first column will be the index column\n",
+    "    * if the DataFrame does not have Woodwork initialized, all columns will be inferred by Woodwork.\n",
```
tamargrey, line 124: wording nitpick - can this say 

> all columns' logical types will be inferred by Woodwork

featuretools/entityset/entityset.py
```diff
@@ -671,6 +671,9 @@ def add_dataframe(self,
         self._add_references_to_metadata(dataframe)
         return self
 
+    def __setitem__(self, key, value):
+        self.add_dataframe(dataframe=value, dataframe_name=key)
```
tamargrey, line 675: Would it be worth handling the situation where a user passes in a woodwork-initialized dataframe that doesn't have a `df.ww.name` set (or if the `df.ww.name` is different from `key`) differently than `add_dataframe` does? 

Currently the first situation would end up with an error `Cannot add a Woodwork DataFrame to EntitySet without a name` and the second raises a warning that the `dataframe_name` parameter is going to be ignored in favor of the name at `df.ww.name`. 

I know there was some discussion with @thehomebrewnerd  and @rwedge when we decided on that behavior for `add_dataframe` in the first place, but somehow `setitem` feels different to me. I know in woodwork's setitem, we just use the key value for a series' name even if we have to overwrite an already-existing name, and I think it might be work updating the woodwork name on the dataframe before calling `add_dataframe`.

dvreed77: @tamargrey: When looking at the behavior of `add_dataframe` when using a Woodwork initialized dataframe without a name, I do find it strange to get an error. I would think `add_dataframe`, when given a name, would add that name to the Woodwork dataframe. Maybe that behavior should change? I feel pretty strongly that `__setitem__` should be a very thin layer over `add_dataframe` and call it with its default values, and let `add_dataframe` handle errors, warnings, etc.

dvreed77: Actually, a partially initialized Woodwork Dataframe causes all the issues that are described in the Note text, so maybe that is enough in this case

rwedge: I think the first case should be handled by having `add_dataframe` add a name to woodwork-initialized dataframes that lack a name.  There's a larger issue #1740 discussing it

dvreed77: Maybe @gsheni can chime in here, but I believe the underlying sprite of this ticket is to improve DX with Featuretools. So if I have an EntitySet and I just want to quickly add a dataframe (that hasn't been WW-initialized), I can just do `es['dataframe'] = df`. If the user wants more granular control, they can do that by using `add_dataframe` and additionally initializing the df using WW before adding it.

rwedge: It'd be nice for this to also work with already woodworked dataframes but I'm fine with those edge cases being handled in other tickets

gsheni: Yes, let's go for a good initial pass and then we can improve on it later on.

PR #1891, opened 2022-02-09
Fix primitive entrypoint

jeff-hernandez:
Using `featuretools_primtives` as an entry point to load external primitive results in an error because we attempt to iterate over the module.

```
  File "/Users/jeff.hernandez/Code/featuretools/featuretools/primitives/__init__.py", line 33, in <module>
    for primitive in loaded:
TypeError: 'module' object is not iterable
```

This fix iterates over the keys in the module to extract primitive objects.


featuretools/primitives/__init__.py
```diff
@@ -5,41 +5,47 @@
...
+
+            if (
+                inspect.isclass(primitive) and
+                issubclass(primitive, base_primitives) and
+                primitive not in base_primitives
+            ):
+                scope = globals()
+                if primitive.__name__ in scope:
+                    error = f"primitive with name \"{primitive.__name__}\" already exists"
+                    raise RuntimeError(error)
```
jeff-hernandez, line 46: If we want to be more specific, we can include which modules contain the conflicting primitives.

rwedge:
Can we add a test case similar to what we do with the plugin_tests?

I was thinking we'd update the test folder structure like this:
* featuretools/
  * tests/
    * entry_point_tests/
      * plugin_tests/
      * primitive_tests/

featuretools/primitives/__init__.py
```diff
@@ -5,41 +5,47 @@
...
+
+def _load_primitives():
+    """Load in a list of primitives registered by other libraries into Featuretools.
+
+        Example entry_points definition for a library using this entry point:
+        entry_points={
+            "featuretools_primitives": [
+                other_library = other_library:LIST_OF_PRIMITIVES
+            ]
+        }
```
gsheni, line 17:
```suggestion
        If using setup.py:
        
        entry_points={
            "featuretools_primitives": [
                other_library = other_library.file_with_primitives_list:LIST_OF_PRIMITIVES
            ]
        }
        
        If using setup.cfg:
        
            [options.entry_points]
            featuretools_primitives =
                other_library = other_library.file_with_primitives_list:LIST_OF_PRIMITIVES
```

gsheni: @jeff-hernandez Should the other_library have the primitive list in the `__init__.py` or a separate file? I feel like having it in the `__init__.py` might lead to a circular import

jeff-hernandez: Pointing to the `__init__.py` isn't necessary, just the module that contains the primitive objects. Ideally, the other primitive libraries will have access to all the primitives from the top-level module.

gsheni: But shouldn't be put an example here + note that says not to use `__int__.py` in your other_library? To avoid a circular import? Because if you try to import you other_library, it might have to import `featuretools` which has to import the other library.

jeff-hernandez: I think `other_library = other_library` and `other_library = other_library.__init__.py` are equivalent, so if a circular import raises in one then it should raise in the other. I haven't tested whether or not an `import featuretools` statement inside that module would raise a circular error, but it's worth testing.

gsheni: With this entrypoint, you just have to point to the file where all primitives are imported (which could be the __init__.py` of the library).

jeff-hernandez:
@rwedge ready for review

featuretools/primitives/__init__.py
```diff
@@ -2,44 +2,65 @@
...
+                issubclass(primitive, base_primitives) and
+                primitive not in base_primitives
+            ):
+                name = primitive.__name__
+                scope = globals()
+
+                if name in scope:
+                    this_module, that_module = primitive.__module__, scope[name].__module__
+                    message = f'Ignoring primitive "{name}" from "{this_module}" '
+                    message += f'because it already exists in "{that_module}"'
```
rwedge, line 60: I think we'd want to add some context to this warning to make it clearer this is happening when featuretools is loading primitives from the primitives entry point

rwedge, line 60: "because it already exists in {that_module}"

I think 

"because a primitive with that name already exists in {that_module}"

would be more precise

jeff-hernandez: Thoughts on 'Ignoring {entry_point.name} primitive "{name}" from "{this_module}..."?

featuretools/tests/entry_point_tests/test_primitives.py
```diff
@@ -0,0 +1,20 @@
+from featuretools.tests.entry_point_tests.utils import (
+    import_featuretools,
+    install_featuretools_primitives,
+    uninstall_featuretools_primitives
+)
+
+
+def test_entry_point():
```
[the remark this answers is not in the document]
jeff-hernandez: sure we can test that

featuretools/primitives/__init__.py
rwedge: I was thinking something like

"While loading primitives via "featuretools_primitives" entry point, ignored primitive {name} from {this_module} . . ."

Basically I want to be clearer that this is happening during entry point loading

Feels like it'll just pop up out of nowhere during `import featuretools` and I'd like to add some context

jeff-hernandez: sounds good, updated

featuretools/tests/entry_point_tests/utils.py
```diff
@@ -3,26 +3,39 @@
...
-    if level:
-        c += 'import os;'
-        c += 'os.environ["FEATURETOOLS_LOG_LEVEL"] = "%s";' % level
-
-    c += 'import featuretools;'
-    return python('-c', c)
-
-
-def install_featuretools_plugin():
+def get_relative_path(*args):
```
rwedge, line 6: this method name is a little vague

`join_to_entry_point_tests_module_path` would be more descriptive but perhaps not the best

jeff-hernandez: thoughts on `_get_path_to_add_on_package`?

rwedge: `_get_path_to_add_ons` and have it return `pwd + 'add-ons'` by default?

jeff-hernandez: yeah sounds good

rwedge (approved):
LGTM

Issue #1680, opened 2021-09-07
Quirk if dataframes have NaN in index and Woodwork integration


rwedge:
Featuretools allows creating an entity/dataframe with NaN as one of the index values, but when calculating aggregations for that value it does retrieve the value succesfully.  In this example, user 4 is a user with a null favorite color.  Calculating the feature "last user with a favor color of null" you'd expect the result to be 4 since they're the only user with that "color" chosen, but instead featuretools returns NaN for that feature.

```python
import featuretools as ft
import pandas as pd

df = pd.DataFrame({'user': [0, 1, 2, 3, 4],
                   'join_date': pd.date_range(start="2001-1-1", periods=5),
                   'favorite_number': ["blue", "red", "blue", "green", None]})

es = ft.EntitySet("test")

es.entity_from_dataframe(entity_id="users",
                         dataframe=df,
                         index="user",
                         time_index="join_date")

es.normalize_entity(base_entity_id="users",
                    new_entity_id="colors",
                    index="favorite_color")

feature_matrix, feature_list = ft.dfs(entityset=es, target_entity="users", agg_primitives=["last"])

feature_matrix['colors.LAST(users.user)']
```

gsheni:
Closing in favor of https://github.com/alteryx/woodwork/issues/1191, which will prevent this

PR #1897, opened 2022-02-10
Remove null index values on normalized dataframe

### LINKED ISSUE #1874: normalize_dataframe can introduce a null value into the new DataFrame's index, causing Woodwork error
Woodwork 0.12.0 no longer allowing null values in index columns has caused some of our normalize dataframe tests to fail, because the intended index column contains nans, which introduces a nan into the normalized index.

We should determine what the expected behavior is when there are nans in the intended index column. We can
- Ignore the nan, and only include non-null values in the normalized dataframe
- Not allow this situation since nans aren't allowed in indices and raise an error

rwedge:
Fixes #1874

This PR doesn't address #1680 as with the NA index value removed from the dataframe at normalization, there's no NA-indexed data to group on, even if we enabled the NA GROUP


tamargrey:
We can probably close #1873 now @dvreed77

featuretools/entityset/entityset.py
```diff
@@ -803,6 +803,7 @@ def normalize_dataframe(self, base_dataframe_name, new_dataframe_name, index,
 
         new_dataframe2 = new_dataframe. \
             drop_duplicates(index, keep='first')[selected_columns]
+        new_dataframe2 = new_dataframe2.dropna(subset=[index])
```
rwedge, line 806: doing the drop here might cause a scenario where `make_secondary_time_index` logic reintroduces null index values but I'm not positive

rwedge:
@alteryx/ml-tools, any ideas on why upgrading to WW 0.12.0 would impact this test?
https://github.com/alteryx/featuretools/blob/79eea5fd3db16d4059141af77cd69f5864f20f14/featuretools/tests/entityset_tests/test_es.py#L2175-L2180

tamargrey (approved):
lgtm assuming tests pass!

PR #1909, opened 2022-02-16
Update urls to test deserialization

rwedge:
The newest woodwork version handles LatLong nans slightly differently and requires a new file in order to deserialize to the same entityset as the tests expect.  This new file will work with older version of woodwork as well.


thehomebrewnerd (approved):
LGTM!

PR #1919, opened 2022-02-18
Better error when DFS returns an empty list of features

### LINKED ISSUE #1309: Raise better error when DFS returns an empty list of features
With some combinations of data and primitives, `DeepFeatureSynthesis.build_features()` can return an empty list of features. When this empty list is subsequently passed to `calculate_feature_matrix` an error is raised: `AssertionError: features must be a non-empty list of features`. This error can be confusing for users as they have not specified the features manually in this scenario.

We could catch this situation in the `dfs` function prior to calling `calculate_feature_matrix` and raise a more meaningful error so that users can better understand what path to take to resolve the issue.

#### Expected Output
A more meaningful error should be raise, perhaps something along the lines of "No features can be generated from the specified primitives. Please make sure the primitives you are using are compatible with the variable types in your data."

#### Code to reproduce current AssertionError:
```python
import featuretools as ft
import pandas as pd

data = ft.demo.load_mock_customer(return_entityset=False, n_customers=5)
# Drop zip_code to prevent `zip_code` IdentityFeature from being created
data['customers'] = data['customers'].drop(columns='zip_code')

es = ft.EntitySet()
es.entity_from_dataframe(entity_id='customers', dataframe=data['customers'], index='customer_id', time_index='join_date')
es.entity_from_dataframe(entity_id='sessions', dataframe=data['sessions'], index='session_id', time_index='session_start')
es.add_relationship(ft.Relationship(es['customers']['customer_id'], es['sessions']['customer_id']))

ft.dfs(entityset=es, target_entity='customers', trans_primitives=[], agg_primitives=['max'])
```

DavidQi:
- Fixes #1309

### DIFF
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -147,7 +147,7 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
     """
     assert (isinstance(features, list) and features != [] and
             all([isinstance(feature, FeatureBase) for feature in features])), \
-        "features must be a non-empty list of features"
+        "No features can be generated from the specified primitives. Please make sure the primitives you are using are compatible with the variable types in your data."
 
     # handle loading entityset
     from featuretools.entityset.entityset import EntitySet
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -147,7 +147,7 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
     """
     assert (isinstance(features, list) and features != [] and
             all([isinstance(feature, FeatureBase) for feature in features])), \
-        "No features can be generated from the specified primitives. Please make sure the primitives you are using are compatible with the variable types in your data."
+        "features must be a non-empty list of features"
 
     # handle loading entityset
     from featuretools.entityset.entityset import EntitySet
--- a/featuretools/synthesis/dfs.py
+++ b/featuretools/synthesis/dfs.py
@@ -3,7 +3,6 @@ import warnings
 from featuretools.computational_backends import calculate_feature_matrix
 from featuretools.entityset import EntitySet
 from featuretools.exceptions import UnusedPrimitiveWarning
-from featuretools.feature_base import FeatureBase
 from featuretools.synthesis.deep_feature_synthesis import DeepFeatureSynthesis
 from featuretools.synthesis.utils import (
     _categorize_features,
--- a/featuretools/synthesis/dfs.py
+++ b/featuretools/synthesis/dfs.py
@@ -9,6 +9,7 @@ from featuretools.synthesis.utils import (
     get_unused_primitives
 )
 from featuretools.utils import entry_point
+from featuretools.feature_base import FeatureBase
 
 
 @entry_point('featuretools_dfs')
@@ -260,6 +261,10 @@ def dfs(dataframes=None,
 
     features = dfs_object.build_features(
         verbose=verbose, return_types=return_types)
+
+    assert (isinstance(features, list) and features != [] and
+            all([isinstance(feature, FeatureBase) for feature in features])), \
+        "No features can be generated from the specified primitives. Please make sure the primitives you are using are compatible with the variable types in your data."
 
     trans, agg, groupby, where = _categorize_features(features)
 
--- a/featuretools/synthesis/dfs.py
+++ b/featuretools/synthesis/dfs.py
@@ -276,7 +276,7 @@ def dfs(dataframes=None,
     if features_only:
         return features
 
-    assert (isinstance(features, list) and features != []), \
+    assert (features != []), \
         "No features can be generated from the specified primitives. Please make sure the primitives you are using are compatible with the variable types in your data."
 
     feature_matrix = calculate_feature_matrix(features,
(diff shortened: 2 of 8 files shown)


featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -147,7 +147,7 @@ def calculate_feature_matrix(features, entityset=None, cutoff_time=None, instanc
     """
     assert (isinstance(features, list) and features != [] and
             all([isinstance(feature, FeatureBase) for feature in features])), \
-        "features must be a non-empty list of features"
+        "No features can be generated from the specified primitives. Please make sure the primitives you are using are compatible with the variable types in your data."
```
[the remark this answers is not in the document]
mingdavidqi: moved it from calculate_feature_matrix to dfs

featuretools/synthesis/dfs.py
```diff
@@ -261,6 +262,10 @@ def dfs(dataframes=None,
     features = dfs_object.build_features(
         verbose=verbose, return_types=return_types)
 
+    assert (isinstance(features, list) and features != [] and
+            all([isinstance(feature, FeatureBase) for feature in features])), \
+        "No features can be generated from the specified primitives. Please make sure the primitives you are using are compatible with the variable types in your data."
+
```
rwedge, line 268: Let's move this check below where a `features_only` run of `dfs` would return so this check does not cause a `features_only` run of `dfs` to error if no features were created

featuretools/synthesis/dfs.py
```diff
@@ -261,6 +262,10 @@ def dfs(dataframes=None,
     features = dfs_object.build_features(
         verbose=verbose, return_types=return_types)
 
+    assert (isinstance(features, list) and features != [] and
+            all([isinstance(feature, FeatureBase) for feature in features])), \
```
rwedge, line 266: let's limit this to checking for an empty list

mingdavidqi: Will to move the checking after feature_only process resolve this?

rwedge: No, I'd also like to simplify the assert -- I'd like to tie this specific error message to the feature list being empty.  The other two situations will get caught by the check in `calculate_feature_matrix` , and would suggest some pretty strange things going on that might not fit this new error message

mingdavidqi: updated to assert only with features != []

featuretools/tests/synthesis/test_dfs_method.py
```diff
@@ -53,6 +56,44 @@ def datetime_es():
...
+        })
+        games = pd.DataFrame({
+            'id': range(5),
+            'home_team_id': [2, 2, 1, 0, 1],
+            'away_team_id': [1, 0, 2, 1, 0],
+            'home_team_score': [3, 0, 1, 0, 4],
+            'away_team_score': [2, 1, 2, 0, 0]
+        })
+        dataframes = {'teams': (teams, 'id', None, {'name': 'natural_language'}), 'games': (games, 'id')}
+        relationships = [('teams', 'id', 'games', 'home_team_id')]
```
rwedge, line 75: Let's move the lines defining the dataset outside of the `with` block since they aren't going to error and don't need patching

featuretools/tests/synthesis/test_dfs_method.py
```diff
@@ -53,6 +56,44 @@ def datetime_es():
...
+        dfs(dataframes, relationships, target_dataframe_name="teams", features_only=False)
+
+
+def test_dfs_empty_features_features_only():
```
rwedge, line 79: instead of a second test, what if the first test had a structure like this

```
with patch(DFSObject):
    test features_only=True returns empty list
    with pytest.raises(AssertionError):
        test features_only=False raises expected error message
```

that way we'd test if we can return an empty list of features in the features_only case

featuretools/tests/synthesis/test_dfs_method.py
mingdavidqi: moved out of with statement

featuretools/tests/synthesis/test_dfs_method.py
mingdavidqi: combined test cases for either feature_only or not

featuretools/tests/synthesis/test_dfs_method.py
```diff
@@ -53,6 +56,28 @@ def datetime_es():
...
+        'id': range(5),
+        'home_team_id': [2, 2, 1, 0, 1],
+        'away_team_id': [1, 0, 2, 1, 0],
+        'home_team_score': [3, 0, 1, 0, 4],
+        'away_team_score': [2, 1, 2, 0, 0]
+    })
+    dataframes = {'teams': (teams, 'id', None, {'name': 'natural_language'}), 'games': (games, 'id')}
+    relationships = [('teams', 'id', 'games', 'home_team_id')]
+    features = dfs(dataframes, relationships, target_dataframe_name="teams", features_only=True)
+    assert (isinstance(features, list) and features != []), True
```
rwedge, line 75: I'd like the patch of `DeepFeatureSynthesis` returning `[]` to also apply to this run dfs / assert code so we can test that an empty feature list is fine if `features_only=True`

we'd need to separate the `with` block into two blocks: the patch context which would apply to both `dfs` calls, and the `pytest.raises` with block only applying to the `features_only=False` call

mingdavidqi: patch DeepFeatureSynthesis to return []

rwedge (approved):
Looks good!  Congrats on your first contribution to Featuretools!

Issue #1820, opened 2021-12-28
Remove extra NaN checking in LatLong primitives


gsheni:
- After we have 1 NaN representation for LatLong columns in an EntitySet, we can remove the NaN checking we added.
- Some places where we added this
  - https://github.com/alteryx/featuretools/pull/1814/files#diff-a11c00320a03fc0c77c9e6fe1ceb63a873b641e491bbfd976426af9f33446ff3R945
  - https://github.com/alteryx/featuretools/pull/1814/files#diff-a11c00320a03fc0c77c9e6fe1ceb63a873b641e491bbfd976426af9f33446ff3R970
  - https://github.com/alteryx/featuretools/pull/1814/files#diff-a11c00320a03fc0c77c9e6fe1ceb63a873b641e491bbfd976426af9f33446ff3R1035

PR #1924, opened 2022-02-28
Normalize LatLong NaN values during EntitySet creation

### LINKED ISSUE #1819: Change LatLong values from NaN to tuple of NaNs during EntitySet creation
- To give LatLong a consistent NaN representation and make for fast primitive calculations, we should use the tuple of NaN
- This means that during the EntitySet creation process, any single np.nan values should be changed to (np.nan, np.nan)
- Woodwork will allow for both NaN representations (np.nan & tuple(np.nan, np.nan))
  - https://github.com/alteryx/woodwork/issues/1226

dvreed77:
Normalize LatLong NaN values during EntitySet creation

Fixes #1819 and #1820


featuretools/entityset/entityset.py
```diff
@@ -1776,6 +1784,29 @@ def _add_references_to_metadata(self, dataframe):
             metadata.update(entityset_id=self.id)
         _ES_REF[self.id] = self
 
+    def _normalize_dataframe(self, dataframe):
```
rwedge, line 1787: `_normalize_dataframe` might not be the best term to use since `EntitySet.normalize_dataframe` already exists and is used to split one dataframe into two dataframes

featuretools/entityset/entityset.py
```diff
@@ -1776,6 +1784,29 @@ def _add_references_to_metadata(self, dataframe):
...
+        def replace(x, is_koalas=False):
+            if not isinstance(x, (list, tuple, np.ndarray)) and pd.isna(x):
+                if is_koalas:
```
rwedge, line 1790: maybe change to `is_spark`

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -2461,3 +2461,47 @@ def test_setitem(add_dataframe):
...
+
+    dataframes = {"latLong": (latlong_df,)}
+
+    relationships = []
+
+    es = EntitySet("latlong-test", dataframes, relationships)
+
+    normalized_df = to_pandas(es["latLong"])
+
+    expected_df = to_pandas(
```
rwedge, line 2480: to_pandas given a pandas dataframe immediately returns the dataframe so we can skip it

featuretools/entityset/entityset.py
```diff
@@ -1776,6 +1784,29 @@ def _add_references_to_metadata(self, dataframe):
...
+                    return [np.nan, np.nan]
+                else:
+                    return (np.nan, np.nan)
+            else:
+                return x
+
+        for column in dataframe.columns:
+            if isinstance(dataframe.ww._schema.columns[column].logical_type, LatLong):
```
rwedge, line 1798: we could use
```
for column, logical_type in dataframe.ww.logical_types:
```

featuretools/tests/entityset_tests/test_es.py
dvreed77: @rwedge I am parameterizing pandas, dask, and scala. Is there a way I can check which type it is and only run to_pandas if its not pandas?

dvreed77: arg! yes of course, thanks

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -2461,3 +2461,43 @@ def test_setitem(add_dataframe):
...
+
+    es = EntitySet("latlong-test", dataframes, relationships)
+
+    normalized_df = to_pandas(es["latLong"])
+
+    expected_df = pd.DataFrame(
+        {"idx": [0, 1, 2], "latLong": [(np.nan, np.nan), (1, 2), (np.nan, np.nan)]}
+    )
+
+    pd.testing.assert_frame_equal(normalized_df, expected_df)
```
thehomebrewnerd, line 2484: It's interesting to me that for the spark test were are comparing these two dataframes, but `pd.testing_assert_frame_equal` considers them equal. I guess that's OK?

```
(Pdb) normalized_df
   idx       latLong
0    0  [None, None]
1    1    [1.0, 2.0]
2    2  [None, None]
(Pdb) expected_df
   idx     latLong
0    0  (nan, nan)
1    1      (1, 2)
2    2  (nan, nan)
(Pdb) pd.testing.assert_frame_equal(normalized_df, expected_df)
(Pdb) 
```

dvreed77: I think this comes down to the inconsistent way python compares NaNs especially when they are in tuples and lists. I can add more checks, but I think it might be overkill

thehomebrewnerd: I'm not overly concerned about it because for our purposes those two are equal. More concerned that we would get something that the assertion things is equal when they are really unequal. As long as we don't get false positives here, I think we are ok.

featuretools/primitives/standard/latlong_transform_primitives.py
```diff
@@ -53,8 +53,10 @@ def __init__(self, unit="miles"):
 
     def get_function(self):
         def cityblock(latlong_1, latlong_2):
-            lat_1s, lon_1s = _deconstrct_latlongs(latlong_1)
-            lat_2s, lon_2s = _deconstrct_latlongs(latlong_2)
+            lat_1s = np.array([x[0] for x in latlong_1])
+            lon_1s = np.array([x[1] for x in latlong_1])
+            lat_2s = np.array([x[0] for x in latlong_2])
+            lon_2s = np.array([x[1] for x in latlong_2])
```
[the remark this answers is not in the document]
gsheni: Might be worth look into this for 1 hour and seeing if we can come up with something.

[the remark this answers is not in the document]
dvreed77: I figured it out

thehomebrewnerd (approved):
Looks good to me once all the tests pass!

PR #1927, opened 2022-03-01
Update test cases to cover main.py file

### LINKED ISSUE #1885: No code coverage on __main__.py
https://github.com/alteryx/featuretools/pull/1882 Is passing for all CI checks except for code coverage, where suddenly there's no coverage of `__main__.py`. That PR's scipy update could be to blame, or it could be some untracked change (setuptools 60.8 vs setuptools 60.7).

We should determine why coverage was lost--though the "coverage" was just an import--and improve our tests so that `__main__.py` is truly covered.

mingdavidqi:
ASAIK, pytest normally avoids __main__.py files because the name __main__.py doesn't match the test_*.py or *_test.py patterns wich pytest is looking for.

No idea why it's covered (42.86%) before and why it's gone now.

Tried to write some test cases to cover __main__.py file and updated codecov-action @ v2 beacause of https://github.com/marketplace/actions/codecov
 
Fixes #1885

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -8,6 +8,7 @@ Future Release
     * Enhancements
     * Fixes
         * Updated the conda install commands to specify the channel (:pr:`1917`)
+        * Update test cases to cover __main__.py file (:pr:`1927`)
     * Changes
     * Documentation Changes
         * Add time series guide (:pr:`1896`)
@@ -15,7 +16,7 @@ Future Release
     * Testing Changes
 
     Thanks to the following people for contributing to this release:
-    :user:`tamargrey`, :user:`kushal-gopal`, :user:`rwedge`
+    :user:`tamargrey`, :user:`kushal-gopal`, :user:`rwedge`, :user:`mingdavidqi`
 
 v1.6.0 Feb 17, 2022
 ===================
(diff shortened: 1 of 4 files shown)


thehomebrewnerd:
@mingdavidqi I just realized something. I think you might just need to add an `__init__.py` file to the `cli_tests` folder. I think we ran into a similar issue a while back with Woodwork. I don't recall the details now, but I think there was some issue with tests not being used for coverage if they were in a folder without an `__init__.py` file.

Not completely sure this is it, but maybe something to look into.

test-requirements.txt
```diff
@@ -1,6 +1,6 @@
 pip>=21.3.1
 wheel>=0.33.1
-pytest>=5.2.0
+pytest==7.0.1
```
thehomebrewnerd, line 3: Just curious - are we doing something that requires us to use this specific version of pytest? If we need to upgrade to version 7, can we make this `pytest>=7.0.1` instead?

docs/source/release_notes.rst
```diff
@@ -8,6 +8,7 @@ Future Release
     * Enhancements
     * Fixes
         * Updated the conda install commands to specify the channel (:pr:`1917`)
+        * Update test cases to cover __main__.py file (:pr:`1927`)
```
thehomebrewnerd, line 11: I think this entry belongs in the `Testing` section. 

Also, I notice below that the entry for updating the DFS error message ended up in the `Documentation Changes` section, but seems like that should be in the `Changes` section instead. Can you also fix that in this PR?

mingdavidqi: made changes

thehomebrewnerd (approved):
LGTM!

PR #1929, opened 2022-03-07
Remove list_variable_types utility

### LINKED ISSUE #1613: Remove list_variable_types util function
### Remove list_variable_types util function

As part of the Woodwork integration update to Featuretools, the `list_variable_types` utility function was updated to return the output of `woodwork.list_logical_types` along with raising a `FutureWarning` that the function has been deprecated. At some point in the future we should remove this function completely. This will allow us to completely remove the `variable_types` folder as well as the `tests/variables` folder, both of which only contain code related to the `list_variable_types` function.

tamargrey:
Removes the deprecated utility `list_variable_types`.

Closes #1613


rwedge (approved):
LGTM

featuretools/tests/requirement_files/minimum_test_requirements.txt
```diff
@@ -11,3 +11,5 @@ boto3==1.17.46
 composeml==0.8.0
 urllib3==1.26.5
 pyarrow==3.0.0
+responses<0.19.0
```
rwedge, line 14: @gsheni will adding a requirement to this file lead to any issues if the minimum dependency checker decides to update the file?  Is it possible it would get overwritten if the dependency checker isn't tracking it?

gsheni: This is not the way to do this. You need to put `responses>=0.X.X,<0.19.0` in the main `requirements.txt`/`test-requirements.txt`. Then in `minimum_test_requirements.txt` you need to put the lowest `responses` requirement.

tamargrey: So the min responses version on the latest moto is `0.9.0` and it has been since moto v1.3.5 (our min version of moto). I'll make the changes to test-requirements and minimum_test_requirements.

rwedge (approved):

PR #1940, opened 2022-03-09
Add support for Python 3.10

### LINKED ISSUE #1803: Add Python 3.10 support

tamargrey:
Closes #1803

### DIFF
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python_version: ["3.7", "3.8", "3.9", "3.10"]
+        python_version: ["3.7", "3.8", "3.9"]
     steps:
       - name: Set up python ${{ matrix.python_version }}
         uses: actions/setup-python@v2
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python_version: ["3.7", "3.8", "3.9"]
+        python_version: ["3.7", "3.8", "3.9", 3.10]
     steps:
       - name: Set up python ${{ matrix.python_version }}
         uses: actions/setup-python@v2
--- a/.github/workflows/build_docs.yaml
+++ b/.github/workflows/build_docs.yaml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python_version: ["3.7", "3.8", "3.9", 3.10]
+        python_version: ["3.7", "3.8", "3.9", "3.10"]
     steps:
       - name: Set up python ${{ matrix.python_version }}
         uses: actions/setup-python@v2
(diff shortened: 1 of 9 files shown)


.github/workflows/build_docs.yaml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python_version: ["3.7", "3.8", "3.9"]
+        python_version: ["3.7", "3.8", "3.9", "3.10"]
```
tamargrey, line 16: Not sure we can actually do this since 3.10 isn't supported by koalas and we use koalas in our docs. Woodwork only runs the docs build for 3.8 in its CI. Is it better to match Woodwork and only run on 3.8 or have all but 3.10 be run?

gsheni: Have all but 3.10 run (since its not slowing us down, and increase our testing scenarios).

gsheni: Do we need to put a note somewhere in the docs that 3.10 isn't support for Koalas (yet)?

rwedge: That also means the 3.10 unit tests koalas run is pretty much useless

Can we look into excluding them from the test runs?

Not exactly what we want but gives an idea:
https://github.community/t/how-to-conditionally-include-exclude-items-in-matrix-eg-based-on-branch/16853/6

tamargrey: Seems like we would want to add 

```
      exclude:
        - python_version: "3.10"
        - libraries: "koalas"
```

tamargrey: 90a255c2b63349fb7e7c39902a3eefb6c47209d7

oops

```
[Invalid workflow file: .github/workflows/unit_tests_with_latest_deps.yml#L18](https://github.com/alteryx/featuretools/actions/runs/1959930281/workflow)
The workflow is not valid. .github/workflows/unit_tests_with_latest_deps.yml (Line: 18, Col: 7): Unexpected value 'exclude'
```

rwedge: I think it would be

```
exclude:
  - python_version: "3.10"
    libraries: "koalas"
```

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-configurations-from-a-matrix

rwedge: using a dash mark for each would exclude all 3.10, and all koalas

tamargrey: that is indeed what happened 🤦‍♀️  thank you!

docs/source/install.md
```diff
@@ -1,6 +1,6 @@
 # Install
 
-Featuretools is available for Python 3.7, 3.8, and 3.9. It can be installed from [pypi](https://pypi.org/project/featuretools/), [conda-forge](https://anaconda.org/conda-forge/featuretools), or from source.
+Featuretools is available for Python 3.7, 3.8, 3.9, and 3.10. It can be installed from [pypi](https://pypi.org/project/featuretools/), [conda-forge](https://anaconda.org/conda-forge/featuretools), or from source.
```
gsheni, line 3:
```suggestion
Featuretools is available for Python 3.7, 3.8, 3.9, and 3.10. It can be installed from [pypi](https://pypi.org/project/featuretools/), [conda-forge](https://anaconda.org/conda-forge/featuretools), or from [source](https://github.com/alteryx/featuretools).
```

gsheni (dismissed):
looks good assuming tests pass and my 1 suggestion is applied.

docs/source/install.md
tamargrey: done [f1aa760](https://github.com/alteryx/featuretools/pull/1940/commits/f1aa760d5730604cf65728488d03295cac627e09)

gsheni (approved):

PR #1948, opened 2022-03-11
Add sqrt, log, sin, cos and tan transform primitives

### LINKED ISSUE #1943: Add Sine, Cosine, Tangent, SquareRoot, Log primitives
- We can add primitives that calculate the since, cosine, and tangent of a numeric column
- Let's make sure we do this in a fast way
- An example how another library that uses Featuretools does it:
  - https://github.com/tvdboom/ATOM/blob/7a5ac34b43de1c72cdb22eb6cae60462da75a77c/atom/feature_engineering.py#L379

tvdboom:
- Fixes #1943


tvdboom:
I couldn't find a test where all primitives are systematically tested. Do I need to add a separate test for each or how should it be done?

docs/source/release_notes.rst
```diff
@@ -6,6 +6,7 @@ Release Notes
 Future Release
 ==============
     * Enhancements
+        * Added the SquareRoot, NaturalLogarithm, Since, Cosine and Tangent primitives (:pr:`1948`)
```
thehomebrewnerd, line 9: Spelling error - `Since` -> `Sine`

thehomebrewnerd:
> I couldn't find a test where all primitives are systematically tested. Do I need to add a separate test for each or how should it be done?

@tvdboom First, thanks for the contribution!

Second, the transform primitives should all get tested in `test_transform_features.py` in the `test_init_and_name` test: https://github.com/alteryx/featuretools/blob/393759398131a99361aa77f2cf3bc2b1aebf6a53/featuretools/tests/primitive_tests/test_transform_features.py#L71

That test should make sure that a feature is created using the primitive and that it calculates without error. That test does not actually test that the calculation is being performed correctly, but since you are using standard numpy functions, I don't think we need to worry about testing the results for correctness in this case, outside of the docstring tests.

tvdboom:
@thehomebrewnerd So if I understood correctly, since `test_init_and_name` fetches all transform primitives, the new primitives are already being tested, so I don't need to add anything else for that. What about the documentation? I added the primitives to the API reference. Does it need to be added somewhere else?

thehomebrewnerd:
@tvdboom Did you by chance enable github actions to run on your fork of the repo? I am getting dozens of notifications that some of our automated workflows are running and failing on your fork. Unfortunately your forked repo doesn't have the secrets setup for these workflows to succeed and they will continue to fail and send the notifications of the failure.

If actions are enabled, could you please disable them in the settings page?

image

On another note, I'll take a final look through your PR today, but I think we should be close to approving and merging.

thehomebrewnerd (approved):
Looks good. Thanks again for the contribution!

thehomebrewnerd:
@tvdboom Can you please review my message above about the github actions? I am continuing to get dozens of failed action run notices in my email daily about actions failing on your fork of the repo? So far I have not found a way to disable these notifications since they are coming from a repo you own.

Can you please disable actions on your fork of this repo by going to Settings -> Actions -> General in GitHub?

PR #1950, opened 2022-03-11
Remove wheel from setup_requires

thehomebrewnerd:
Need to remove `wheel` from `setup_requires` in `setup.cfg` as this will cause issues with building distribution with `python -m build`.

### DIFF
--- a/.github/workflows/install_test.yml
+++ b/.github/workflows/install_test.yml
@@ -24,11 +24,11 @@ jobs:
         uses: actions/checkout@v2
       - name: Build featuretools package
         run: make package_featuretools
-      - name: Install complete version of featuretools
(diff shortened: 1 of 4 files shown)


.github/workflows/install_test.yml
```diff
@@ -24,11 +24,11 @@ jobs:
         uses: actions/checkout@v2
       - name: Build featuretools package
         run: make package_featuretools
-      - name: Install complete version of featuretools
+      - name: Install complete version of featuretools from sdist
         run: |
           pip config --site set global.progress_bar off
           python -m pip install --upgrade pip
```
gsheni, line 30:
```suggestion
```

gsheni (approved):

PR #1961, opened 2022-03-15
Fix non-deterministic test in test_es.py

thehomebrewnerd:
Fix non-deterministic test in `test_es.py`. Failure showed up after switching from Koalas to the pyspark pandas API and failed because rows were returned in a different order with spark.


rwedge (approved):
LGTM

PR #1970, opened 2022-03-18
Remove make_trans_primitive and make_agg_primitive functions

### LINKED ISSUE #1953: Remove make_trans_primitive and make_agg_primitive function
- We recommend defining primitives via class and not using the function
- We should verify that our documentation/guides do not use make_trans_primitive/make_agg_primitive
- We want to do this because some users run into problems with these functions (https://github.com/alteryx/featuretools/issues/1942)

thehomebrewnerd:
### Remove `make_trans_primitive` and `make_agg_primitive` functions

Closes #1953


thehomebrewnerd:
Changing from `make_trans_primitive` and `make_agg_primitive` function to defining custom classes resulted in several new uncovered lines of code. This seems to be because many of the primitives that were defined using these functions were only tested in DFS and were never used in calculate feature matrix so many of the primitive functions are not being called.

We have a few options for this - I'm not sure what's best:
1. Leave everything as is with the uncovered new lines
2. Remove the unused primitive functions
3. Mark these primitive functions with `# pragma: no cover` and leave in place

rwedge:
https://github.com/alteryx/featuretools/blob/ff67dd33597997b3ec4409b7e3e20ebfb54ecc03/featuretools/primitives/base/primitive_base.py#L85-L92

the `continue` line is no longer getting covered

tamargrey (approved):
lgtm!

PR #1973, opened 2022-03-21
Conditionally run scheduled workflows

### LINKED ISSUE #1955: Prevent workflows from running on forked repos
By default our scheduled workflows such as Auto Approve Dependency PRs and Latest Dependency Checker should not run on forked repos, but I believe those runs can be enabled by the forked repo owner.

image

If the user does enable this, the scheduled workflows will run, but fail because the secrets used by those workflows are not available in the forked repo.

We should investigate and implement a way to prevent these workflows from running in a forked repo, perhaps by using some conditional logic checking the that the repo owner (`if: github.owner == alteryx`), but there might be other methods that could be used as well to prevent workflow execution.

thehomebrewnerd:
### Conditionally run scheduled workflows

Closes #1955 

Update scheduled workflows to run only if repository_owner is alteryx. This should prevent these scheduled workflows from running on repo forks and failing repeatedly.

### DIFF
--- a/.github/workflows/latest_dependency_checker.yml
+++ b/.github/workflows/latest_dependency_checker.yml
@@ -8,6 +8,7 @@ on:
   workflow_dispatch:
 jobs:
   build:
+    if: ${{ github.repository_owner == 'alteryx' }}
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/auto_approve_dependency_PRs.yml
+++ b/.github/workflows/auto_approve_dependency_PRs.yml
@@ -12,6 +12,7 @@ on:
       - completed
 jobs:
   build:
+    if: ${{ github.repository_owner == 'alteryx' }}
     runs-on: ubuntu-latest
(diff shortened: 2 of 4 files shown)


.github/workflows/latest_dependency_checker.yml
```diff
@@ -8,6 +8,7 @@ on:
   workflow_dispatch:
 jobs:
   build:
+    if: ${{ github.repository_owner == 'alteryx' }}
```
gsheni, line 11: Let's do this for the 3rd scheduled workflow:
https://github.com/alteryx/featuretools/blob/main/.github/workflows/test_with_woodwork_main_branch.yml

thehomebrewnerd: Updated [3e6e807](https://github.com/alteryx/featuretools/pull/1973/commits/3e6e80772cd124cb4d15043a4316a1881203b033)

gsheni:
@thehomebrewnerd let's add this to Woodwork and EvalML.

PR #2003, opened 2022-03-31
Update UnusedPrimitiveWarning text

### LINKED ISSUE #1141: Catch multiple versions of a primitive, one unused
Right now in the check for unused primitives, we determine if a feature has been explored using `feature.get_name()`, and we create the sets of used primitives with `feature.primitive.name`. 

Ideally both situations would be able to differentiate between two separate instances of the same primitive that were passed into `dfs` with different parameters, such as `Entropy(dropna = False)` and `Entropy(dropna = True)`, and in order to do that, we need to add information about the primitive's arguments.

thehomebrewnerd:
### Update UnusedPrimitiveWarning text

Closes #1141 

Add clarifying text to UnusedPrimitiveWarning to explain behavior if multiple instances of a primitive are unused.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -6,6 +6,7 @@ Release Notes
 Future Release
 ==============
     * Enhancements
+        * Improve ``UnusedPrimitiveWarning`` with additional information (:pr:`2003`)
     * Fixes
     * Changes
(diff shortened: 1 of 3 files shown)


rwedge (approved):
LGTM

PR #2019, opened 2022-04-19
Update primitive matching to consider all inputs

### LINKED ISSUE #1828: Update DFS primitive matching to consider all input types, not just first
### Update DFS primitive matching to consider all input types, not just first

Currently if a primitive has multiple valid input types, only the first type is used during primitive matching. For example, consider the `Not` transform primitive:
```
class Not(TransformPrimitive):
    """Negates a boolean value.
    Examples:
        >>> not_func = Not()
        >>> not_func([True, True, False]).tolist()
        [False, False, True]
    """
    name = "not"
    input_types = [[ColumnSchema(logical_type=Boolean)], [ColumnSchema(logical_type=BooleanNullable)]]
    return_type = ColumnSchema(logical_type=BooleanNullable)
    compatibility = [Library.PANDAS, Library.DASK, Library.KOALAS]
    description_template = "the negation of {}"

    def generate_name(self, base_feature_names):
        return u"NOT({})".format(base_feature_names[0])

    def get_function(self):
        return np.logical_not
```

The input types for this primitive indicate that it will work on both `Boolean` and `BooleanNullable` column types. However, when running DFS to generate features, only the first input type (`Boolean`) is considered and no features are generated from `BooleanNullable` column types. This can be seen clearly in the following example:

```python
import featuretools as ft
import pandas as pd

df = pd.DataFrame({
    "index": [0, 1, 2],
    "boolean": [True, False, True],
    "boolean_nullable": [True, False, True],
})

es = ft.EntitySet()
es.add_dataframe(
    dataframe=df,
    dataframe_name="bools",
    index="index",
    logical_types={"boolean": "Boolean", "boolean_nullable": "BooleanNullable"}
)

features = ft.dfs(entityset=es,
                  target_dataframe_name="bools",
                  trans_primitives=["not"],
                  features_only=True)
features
```
```
[<Feature: boolean>, <Feature: boolean_nullable>, <Feature: NOT(boolean)>]
```
The `Not` primitive only generated a feature from the column with the `Boolean` logical type, even though `BooleanNullable` is defined as a valid input type.

Ideally DFS would consider all valid input types when matching primitive inputs to valid columns, and not just the first input. We should review the current primitive matching code and update, if possible, so that all valid input types are considered when generating features.

thehomebrewnerd:
### Update primitive matching to consider all inputs

Closes #1828


thehomebrewnerd:
When testing this change, two new issues were discovered that can cause feature value calculation to fail in some cases. These two issues were not caused by this change and are documented in #2020 and #2021. 

Previous to this PR, users would have had to manually define the features in order for these bugs to show up, but with this PR, DFS could potentially generate features that could not be calculated without error.

rwedge:
Do you think we should have an alert of some kind in the release notes about the number of features created by DFS potentially changing?

thehomebrewnerd:
> Do you think we should have an alert of some kind in the release notes about the number of features created by DFS potentially changing?

Sounds like a good idea. I'll add that.

thehomebrewnerd:
> > Do you think we should have an alert of some kind in the release notes about the number of features created by DFS potentially changing?
> 
> Sounds like a good idea. I'll add that.

@rwedge Added a note in the release notes about number of features increasing in some cases with this change.

docs/source/release_notes.rst
```diff
@@ -21,6 +22,11 @@ Future Release
     Thanks to the following people for contributing to this release:
     :user:`dvreed77`, :user:`gsheni`, :user:`rwedge`, :user:`thehomebrewnerd`
 
+Note
+++++
+* The update to the DFS algorithim in this release may cause the numer of features returned
```
dvreed77, line 27: number misspelled

thehomebrewnerd: Fixed - and also fixed `algorithim`

dvreed77 (dismissed):
lgtm

thehomebrewnerd:
I ran the `ecommerce_many_features_test` in the performance tests against this branch. Initially that test highlighted some problems because of new features that were generated that failed on computation. After merging in the PRs to fix those issues, the `ecommerce_many_features_test` performance test completes successfully. Results comparing this branch to current main are below:

image

rwedge (approved):
LGTM

PR #2022, opened 2022-04-20
Remove boolean inputs from MultiplyNumeric

### LINKED ISSUE #2020: MultiplyNumeric primitive fails when multiplying IntegerNullable column with Boolean column
Features calculated with the MultiplyNumeric primitive fail when multiplying an IntegerNullable column with a Boolean column. 

Pandas does not support multiplying an `Int64` with a `bool` or `boolean` column and raises a `TypeError: can only perform ops with numeric values` when this is attempted. Since the `MultiplyNumeric` primitive currently allows for multiplication between boolean and integer columns, this can result in an error when attempting to calculate feature values in some cases.

We should consider removing the boolean input types from the `MultiplyNumeric` primitive as the name of the primitive implies that all the inputs should be numeric. This would resolve the issue and make the behavior consistent with `AddNumeric`, `SubtractNumeric` and `DivideNumeric`, none of which accept boolean inputs.

```python
import featuretools as ft

es = make_ecommerce_entityset(with_integer_time_index=True)
es['customers'].ww.set_types(logical_types={'cancel_date': 'IntegerNullable'})

multiply_feature = ft.Feature(es['customers'].ww['cancel_date']) * ft.Feature(es["customers"].ww["loves_ice_cream"])

fm = ft.calculate_feature_matrix(entityset=es, features=[multiply_feature])
```
```
TypeError: can only perform ops with numeric values
```

thehomebrewnerd:
### Remove boolean inputs from MultiplyNumeric

Closes #2020

This PR updates the `MultiplyNumeric` primitive to have behavior consistent with the other `*Numeric` primitives. The boolean input types have been removed from the valid `input_types`. This makes the behavior consistent with the name and the other primitives and will allow multiplication of two numeric columns only.


rwedge (approved):
Looks good

PR #2025, opened 2022-04-21
More ordinal comparison fixes

thehomebrewnerd:
### More ordinal comparison fixes

Implements additional changes to these primitives to enable comparison between ordinal columns that have different order values:
- `GreaterThan`
- `GreaterThanEqualTo`
- `LessThan`
- `LessThanEqualTo`


featuretools/primitives/standard/binary_transform.py
```diff
@@ -36,7 +36,21 @@ class GreaterThan(TransformPrimitive):
     description_template = "whether {} is greater than {}"
 
     def get_function(self):
-        return np.greater
+        def greater_than(val1, val2):
+            if pdtypes.is_categorical_dtype(val1) and pdtypes.is_categorical_dtype(
```
rwedge, line 40: should we have the top level logic be 

```
if val_1 is categorical or val_2 is categorical:
    ....
return val_1 > val_2
```
so that for the non-categorical case we exit after examining one if statement instead of 2?

thehomebrewnerd: I think we could, but if I'm thinking about this right if we did that the logic might a little harder to follow.

We have a few cases to handle:
1. One input is categorical, the other is not -> `nan`
2. Both inputs are categorical and categories are not equal -> `nan`
3. Both inputs are categorical and categories are equal -> `val1 > val2`
4. Inputs are both numeric or both datetime -> `val1 > `val2`

I think if we moved the `or` statement to the top we would have to do something like this and handle the case where they are both equal inside that first conditional:
```python
if val1 is categorical or val2 is categorical:
    if val1 is categorical and val2 is categorical:
        if val1.categories == val2.categories:
            return val1 > val2
    return np.nan
return val1 > val2
```

I guess it's debatable whether that is more clear or less clear, but I find it a little harder to follow. I think the "special case" that should work but doesn't is a little more hidden in the updated flow.

If you prefer that approach I can update - don't have a strong preference. Or is there something better yet that I'm not seeing?

rwedge: What if we keep the if / elif structure but store the `is_categorical` calls in variables so we aren't potentially making them multiple times

thehomebrewnerd: Sure, we can do that. I was thinking those calls would be pretty fast so didn't worry about it, but maybe that's not the case. Storing in variables certainly won't hurt. I'll update.

thehomebrewnerd: I think that change will also make the code a little easier to read.

thehomebrewnerd: Updated: [3b26676](https://github.com/alteryx/featuretools/pull/2025/commits/3b2667664f4ee44141a3a550b7ad4bc926dc7b66)

rwedge (approved):

PR #2030, opened 2022-04-25
Woodwork version 0.16 fixes

thehomebrewnerd:
Updates for compatibility for WW version 0.16

- bump min WW requirement to 0.16.1
- Update serialization to specify filename when deserializing from parquet
- Fix test in `test_transform_features.py`


featuretools/entityset/serialize.py
```diff
@@ -13,10 +13,10 @@
 ps = import_or_none("pyspark.pandas")
 
 FORMATS = ["csv", "pickle", "parquet"]
-SCHEMA_VERSION = "7.0.0"
+SCHEMA_VERSION = "7.1.0"
```
rwedge, line 16: i think if we're unable to deserialize an entityset parquet serialized with schema version 7.0.0 we should consider this a breaking change and bump schema version to 8.0.0

gsheni (approved):

rwedge (approved):
LGTM

PR #2033, opened 2022-04-25
Update unit tests with Woodwork main branch workflow name

gsheni:


.github/workflows/unit_tests_with_woodwork_main_branch.yml
```diff
@@ -1,7 +1,5 @@
 name: Test Featuretools with Woodwork main Branch
 on:
-  schedule:
```
gsheni, line 3: This will now happen on each commit for Woodwork
- https://github.com/alteryx/woodwork/commit/463f26e42891063afa18433dde6b578e8d0be05e#diff-3e48e8ad887017d16610e21247065df151dcfba00ecff81f6360fe52ec24be56R16

PR #2035, opened 2022-04-26
Add MultiplyNumericBoolean primitive that allows for multiplication between a numeric an a boolean column

### LINKED ISSUE #2026: Need a way to create a seed feature that is the multiplication of a numeric column with a boolean column.
PR #756 implemented a change to close issue #752 in which boolean inputs were added as valid input types to the `MultiplyNumeric` primitive. This was done to allow seed features to be created easily that were the product of a numeric column and a boolean column, as described in #752. 

With the updates being implemented in #2019 this becomes problematic because pandas does not allow for the multiplication between boolean columns and columns with a new nullable dtype such as `Int64` or `Float64`. The changes in #2019 can create situations where DFS will generate a feature automatically that cannot be calculated. Previously because DFS only used the first inputs to a primitive, these problematic features were note created.

PR #2022 removes the boolean inputs from the `MultiplyNumeric` primitive. This was done to resolve the issue created in #2019 and also makes the primitive behavior more consistent with its name and `AddNumeric`, `SubtractNumeric` and `DivideNumeric` primitives which do not accept boolean inputs.

With these changes a new solution is needed to address the original concern in issue #752. Possible solutions would be to implement a `MultiplyBooleanNumeric` primitive, but this would need to properly handle multiplication between an `IntegerNullable` and `Boolean` column, which was problematic in the prior implementation of `MultiplyNumeric`. Another solution might be to add a `Mask` primitive that takes two columns and returns a series that contains the values in the first column where the corresponding mask column is True and some default/placeholder value where the mask values are False. There might be other viable solutions as well.

thehomebrewnerd:
### Add MultiplyNumericBoolean primitive that allows for multiplication between a numeric an a boolean column

Replaces functionality that is removed in #2022 with a more appropriately named primitive.

Closes #2026

### DIFF
--- a/featuretools/tests/primitive_tests/test_transform_features.py
+++ b/featuretools/tests/primitive_tests/test_transform_features.py
@@ -700,6 +700,8 @@ def test_boolean_multiply(boolean_mult_es):
     es = boolean_mult_es
     to_test = [
         ("numeric", "numeric"),
+        ("numeric", "bool"),
+        ("bool", "numeric"),
         ("bool", "bool"),
     ]
     features = []
--- a/featuretools/tests/primitive_tests/test_transform_features.py
+++ b/featuretools/tests/primitive_tests/test_transform_features.py
@@ -1568,3 +1568,22 @@ def test_comparisons_with_ordinal_valid_inputs_that_dont_work_but_should(pd_es):
     fm = to_pandas(fm)
     for col in feature_cols:
         assert fm[col].isnull().all()
+
+def test_multiply_numeric_boolean():
+    df = pd.DataFrame({
+    'id': [0, 1, 2],
+    'val_no_nan': [100, 200, 300],
+    'val_with_nan': [100, 200, pd.NA],
+    'mask_no_nan': [True, False, True],
+    'mask_with_nan': pd.Series([pd.NA, False, pd.NA], dtype='boolean'),
+    }) 
+
+    es = ft.EntitySet()
+    es.add_dataframe(dataframe_name="df", dataframe=df)
+    es["df"].ww.set_types(logical_types={'val_no_nan': 'Integer',
+                                     'val_with_nan': 'IntegerNullable',
+                                     'mask_no_nan': 'Boolean',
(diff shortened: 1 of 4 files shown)


featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -699,7 +701,6 @@ def test_boolean_multiply(boolean_mult_es):
     to_test = [
         ("numeric", "numeric"),
         ("numeric", "bool"),
-        ("bool", "numeric"),
```
rwedge, line 702: why was this case removed?

thehomebrewnerd: We can't swap inputs with the new `MultiplyBooleanNumeric` primitive like we could with the previous `MultiplyNumeric` primitive. The first input needs to be the numeric column and the second the boolean column.

Or if you prefer, we could add logic in the primitive to determine which is which based on dtype and act accordingly?

rwedge: what if we add logic in the `__mul__` override to pass in [numeric, bool] in that order instead of always doing [self, other]?

thehomebrewnerd: Updated the primitive logic and inputs to handle this case since updating `__mul__` could result in two features with the same name.

rwedge (dismissed):
Looks good, assuming tests pass

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -698,6 +700,8 @@ def test_boolean_multiply(boolean_mult_es):
     es = boolean_mult_es
     to_test = [
         ("numeric", "numeric"),
+        ("numeric", "bool"),
```
thehomebrewnerd, line 703: @rwedge These were removed in the PR with changes to `MultiplyNumeric`, so had to add them back in now.

rwedge (approved):

dvreed77 (approved):
lgtm

PR #2039, opened 2022-04-26
Use codecov v3 and drop token

rwedge:


gsheni:
@rwedge are still going to do this?

gsheni (approved):

PR #2046, opened 2022-04-29
Delete setup.py, MANIFEST.in and move configuration to pyproject.toml

gsheni:
- Delete setup.py and move some configuration to pyproject.toml

> With the release of pip v19.1.1 , editable installs are again allowed in the presence of pyproject.toml. 

- Additionally, deleted MANIFEST.in because of the following:
  - https://packaging.python.org/en/latest/guides/using-manifest-in/#how-files-are-included-in-an-sdist
- following files are included in a source distribution:
  > all files specified by the package_data and data_files setup() arguments
  > the file specified by the license_file option in setup.cfg (setuptools 40.8.0+)
- deleted codecov because we don't actually use it


rwedge:
The oldest pip version I could get to make an editable install was 21.1, if that makes any difference

gsheni:
@rwedge I think that is fine.

I also deleted Manifest.in, see my comment on PR desc.

gsheni:
@rwedge @thehomebrewnerd ready for review

setup.cfg
```diff
@@ -104,12 +107,15 @@ complete =
     %(nlp_primitives)s
     %(spark)s
     %(autonormalize)s
-
+
```
rwedge, line 110: empty text

pyproject.toml
```diff
@@ -3,4 +3,22 @@ requires = [
     "setuptools >= 47",
     "wheel"
 ]
-build-backend = "setuptools.build_meta"
\ No newline at end of file
+build-backend = "setuptools.build_meta"
+
+[tool.pytest.ini_options]
+minversion = "6.0"
```
rwedge, line 9: should this be either left out or changed to 7.1.2?

gsheni: pytest only supports config in pyproject.toml at a min of 6.0
- https://docs.pytest.org/en/6.2.x/customize.html#pyproject-toml
I took this as a chance to upgrade our test dependencies (putting it to the latest one), since we hardly do it.

gsheni: Removed it.

rwedge (dismissed):
Looks good, assuming tests pass

thehomebrewnerd (approved):
Looks good to me. I tested both building and installing locally and all seemed to work as expected.

PR #2055, opened 2022-05-05
Fix Woodwork main install in test with WW main branch

thehomebrewnerd:
Updates how Woodwork is installed from main to prevent reinstalling other packages (like pandas) along with the Woodwork install.


rwedge (approved):
Looks good

PR #2056, opened 2022-05-11
Update DateToHoliday and DistanceToHoliday primitives to work with timezone-aware inputs

thehomebrewnerd:
This PR updates the `DateToHoliday` and `DistanceToHoliday` primitives to work with timezone-aware inputs and adds new tests to verify the correct answer is returned.

### DIFF
--- a/featuretools/primitives/standard/datetime_transform_primitives.py
+++ b/featuretools/primitives/standard/datetime_transform_primitives.py
@@ -98,7 +98,7 @@ class DateToHoliday(TransformPrimitive):
         def date_to_holiday(x):
             holiday_df = self.holidayUtil.to_df()
             df = pd.DataFrame({"date": x})
-            df.date = df.date.dt.normalize().astype("datetime64")
+            df.date = df.date.dt.normalize().dt.date.astype("datetime64[ns]")
 
             df = df.merge(
                 holiday_df, how="left", left_on="date", right_on="holiday_date"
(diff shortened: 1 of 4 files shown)


featuretools/primitives/standard/datetime_transform_primitives.py
```diff
@@ -201,7 +201,7 @@ def distance_to_holiday(x):
             df["x_index"] = df.index  # store original index as a column
             df = df.dropna()
             df = df.sort_values("date")
-            df.date = df.date.dt.normalize()
+            df.date = df.date.dt.normalize().dt.date.astype("datetime64[ns]")
```
rwedge, line 204: do we need to normalize if we are also extracting the date?

featuretools/tests/primitive_tests/test_datetoholiday_primitive.py
```diff
@@ -109,3 +109,26 @@ def test_multiple_countries():
...
+
+def test_with_timezone_aware_datetimes():
+    df = pd.DataFrame(
+        {
+            "non_timezone_aware_with_time": pd.date_range(
+                "2018-07-03 09:00", periods=3
+            ),
+            "non_timezone_aware_no_time": pd.date_range("2018-07-03", periods=3),
+            "timezone_aware_with_time": pd.date_range(
+                "2018-07-03 09:00", periods=3
```
rwedge, line 122: should we test with times where if converted to UTC, some times would change dates?

featuretools/primitives/standard/datetime_transform_primitives.py
thehomebrewnerd: No, the normalize calls aren't necessary anymore. Removed. Also updated to (I think) improve readability of these lines a little bit.

rwedge (approved):
Looks good

PR #2108, opened 2022-06-15
Adding DayOfYear primitive

sbadithe:
* Adds DayOfYear primitive that computes the day number from 1-365, given the datetime


gsheni:
You will need to add this primitive to docs/source/api_reference.rst

PR #2124, opened 2022-06-17
Adding IsYearEnd and IsYearStart primitives

### LINKED ISSUE #2060: Add IsYearStart primitive
- This primitive returns the `is_year_start` on a datetime column

### LINKED ISSUE #2061: Add IsYearEnd primitive
- This primitive determines the `is_year_end` of a datetime column

sbadithe:
Adding the following primitives: IsYearEnd, IsYearStart

Fixes #2061
Fixes #2060

### DIFF
--- a/featuretools/primitives/standard/datetime_transform_primitives.py
+++ b/featuretools/primitives/standard/datetime_transform_primitives.py
@@ -409,6 +409,58 @@ class IsWeekend(TransformPrimitive):
         return is_weekend
 
 
+class IsYearEnd(TransformPrimitive):
+    """Determines if a date falls on the end of a year.
+
+    Examples:
+        >>> from datetime import datetime
+        >>> dates = [datetime(2019, 12, 31),
+        ...          datetime(2019, 1, 1),
+        ...          datetime(2019, 11, 30)]
+        >>> is_year_end = IsYearEnd()
+        >>> is_year_end(dates).tolist()
+        [True, False, False]
+    """
+
+    name = "is_year_end"
+    input_types = [ColumnSchema(logical_type=Datetime)]
+    return_type = ColumnSchema(logical_type=BooleanNullable)
+    compatibility = [Library.PANDAS, Library.DASK, Library.SPARK]
+    description_template = "whether {} occurred on the end of a year"
+
+    def get_function(self):
+        def is_year_end(vals):
+            return vals.dt.is_year_end
+
+        return is_year_end
+
+
+class IsYearStart(TransformPrimitive):
+    """Determines if a date falls on the start of a year.
+
+    Examples:
+        >>> from datetime import datetime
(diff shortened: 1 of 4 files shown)


gsheni (approved):

gsheni (dismissed):

featuretools/tests/primitive_tests/test_transform_primitive.py
```diff
@@ -174,6 +176,22 @@ def test_is_quarter_start():
     np.testing.assert_array_equal(iqs_bools, correct_bools)
 
 
+def test_is_year_end():
+    is_year_end = IsYearEnd()
+    dates = pd.Series([datetime(2020, 12, 31), datetime(2020, 1, 1)])
```
gsheni, line 181: What happens if there is a NaN?

featuretools/primitives/standard/datetime_transform_primitives.py
```diff
@@ -461,6 +461,58 @@ def is_weekend(vals):
...
 
 
+class IsYearEnd(TransformPrimitive):
+    """Determines if a date falls on the end of a year.
+
+    Examples:
+        >>> from datetime import datetime
+        >>> dates = [datetime(2019, 12, 31),
+        ...          datetime(2019, 1, 1),
+        ...          datetime(2019, 11, 30)]
```
gsheni, line 471: Can we include an example with NaN?

gsheni: We should make sure we did this with the previous Datetime primitives we merged in

sbadithe: General question regarding the behavior of primitives with NaN: 

If I use the Month primitive on the following data: 

`pd.Series([datetime(2020, 3, 3), np.nan]) 
`
I get `[3.0, nan]`

If I pass `pd.Series([np.nan])` in, I get an error. 

I get analogous behavior for Year() and the like. I just wanted to confirm that this is indeed the correct expected behavior when writing tests.

thehomebrewnerd: You should always expect the series dtype for `Datetime` columns to be `datetime64[ns]` based on the behavior of Woodwork, which enforces that dtype for the series.

If you don't otherwise specify the dtype a series of `pd.Series([np.nan])` will have a dtype of `float64` and you can't perform datetime operations on a `float64` column.

I haven't tested this out, but I think if you convert `pd.Series([np.nan])` to a datetime column first, the `Month` primitive will no longer error.

dvreed77: Good call.  I confirmed this doesn't error
```python
s = pd.Series([np.nan]).astype('datetime64[ns]')

ft.primitives.Month()(s)
```

dvreed77 (approved):
lgtm

PR #2127, opened 2022-06-21
Improve feature deserialization to use common primitive instances

### LINKED ISSUE #2115: Feature serialization and deserialization improvements
The serialization and deserialization of features should be improved.

For serialization we should update the approach so we are not duplicating primitive serialized primitive information. Currently if there are features that share the same primitive, each serialized feature contains the information needed to deserialize the primitive. This primitive information only needs to be stored once and not duplicated. We should separate primitive information from feature information and just link from the feature to the corresponding primitive information.

For deserialization we currently create separate instances of primitives for each feature. For features that share the same primitive (with the same arguments) this means we have separate, but identical instances of the primitive. We should update feature deserialization so that features that share a primitive reference the same primitive instance rather than creating unique instances.

thehomebrewnerd:
Updates feature deserialization to use a common primitive instance for features that share a primitive instead of creating a unique primitive instance for each feature.

Partially closes #2115


featuretools/tests/primitive_tests/test_feature_serialization.py
```diff
@@ -304,3 +308,154 @@ def test_serialize_url(es):
...
+    new_sum_features = [f for f in deserialized_features if f.primitive.name == "sum"]
+
+    # After deserialization all features that share a primitive should use the same primitive instance
+    new_is_null_primitive = new_is_null_features[0].primitive
+    new_sum_primitive = new_sum_features[0].primitive
+    assert all([f.primitive is new_is_null_primitive for f in new_is_null_features])
+    assert all([f.primitive is new_sum_primitive for f in new_sum_features])
+
+
+def test_deserializer_uses_common_primitive_instances_with_args(es, tmpdir):
```
rwedge, line 350: Can start using tmp_path instead of tmpdir:

https://docs.pytest.org/en/latest/how-to/tmp_path.html#the-tmpdir-and-tmpdir-factory-fixtures

rwedge (approved):
Looks good, left one comment about adopting a new style change for pytest

featuretools/tests/primitive_tests/test_feature_serialization.py
thehomebrewnerd: Updated the two new tests to use `tmp_path` 

[81568d0](https://github.com/alteryx/featuretools/pull/2127/commits/81568d038ddd8ec9d5fad2d457963dbb621c9a4d)

PR #2130, opened 2022-06-22
Implement IsWorkingHours and IsLunchTime

### LINKED ISSUE #2076: Add IsLunchTime primitive
- This primitive determine if the datetime is = 12 and on a weekday

### LINKED ISSUE #2078: Add IsWorkingHours primitive
- this primitive determines if it is between 8AM and 6PM & a weekday that is not a holiday
- this primitive should be configurable with start and end time

sbadithe:
Adds the following datetime transform primitives: 
  -- `IsWorkingHour` : Checks if hour is between configurable start and end times
  -- `IsLunchTime` : Checks if hour is equal to configurable lunch time hour

Fixes #2076 
Fixes #2078


sbadithe:
Should I include holiday checking for IsLunchTime?

gsheni:
@sbadithe I would make holidays an optional argument, and default to False. You have lunch on the weekends :-)

sbadithe:
> @sbadithe I would make `include_holidays` an optional argument, and default to False. You have lunch on the weekends :-)

featuretools/primitives/standard/datetime_transform_primitives.py
```diff
@@ -459,6 +512,52 @@ def is_weekend(vals):
...
+        start_time (int): Start hour of workday. Default is 8 (8am)
+        end_time (int): End hour of workday. Default is 18 (6pm)
+        country (str): Country to use for determining Holidays.
+            Default is 'US'. Should be one of the available countries here:
+            https://github.com/dr-prodigy/python-holidays#available-countries
+
+    Examples:
+        >>> from datetime import datetime
+        >>> dates = [datetime(2022, 6, 21, 16, 3, 3),
+        ...          datetime(2019,1,3,4,4,4),
```
dvreed77, line 528: fix formatting here to include white space

featuretools/tests/primitive_tests/test_transform_primitive.py
```diff
@@ -174,6 +176,93 @@ def test_is_quarter_start():
...
+    actual = ilt(dates)
+    expected = [False, True, False, False]
+    np.testing.assert_array_equal(actual, expected)
+
+
+def test_is_lunch_time_include_holidays():
+    ilt = IsLunchTime(include_weekends=False, include_holidays=True)
+    dates = pd.Series(
+        [
+            datetime(2022, 6, 26, 12, 12, 12),
```
dvreed77, line 213: can you include a comment that is something like `Wed, Jun 26 2022 - (Holiday Name)`

gsheni:
@sbadithe Can you post the runtimes and snipped of code you used here?

sbadithe:
I populated the test array like this.
```
dates = []
for i in range(2000000): 
    dates.append(datetime(randint(2000, 2021),randint(1, 12), 3, randint(1, 23), 3, 4))
```
I timed the new primitives against a standard one such as  `month`. 

What I found was Month took on average 2.27s per loop, whereas IsLunchTime took 2.31s, and IsWorkingHours took 2.49s per loop.

featuretools/primitives/standard/datetime_transform_primitives.py
```diff
@@ -331,6 +334,69 @@ def is_leap_year(vals):
...
+        if not self.include_weekends:
+            l_time = l_time and (v.dayofweek < 5)
+        if not self.include_holidays:
+            l_time = l_time and not (v.normalize().isin(self.holidays_df.dates))
+        return l_time
+
+    def get_function(self):
+        def is_lunch_time(vals):
+            if isinstance(vals, pd.Series) or isinstance(vals, dd.dataframe.Series):
+                mask = vals.dt.hour == 12
```
dvreed77, line 387: I like this, but use better names.  `is_12_mask` and `is_weekday_mask`: then you have `mask = is_12_mask & is_weekday_mask`

sbadithe: Got it. I separated out the mask creation now, with more descriptive names

featuretools/primitives/standard/datetime_transform_primitives.py
```diff
@@ -331,6 +334,72 @@ def is_leap_year(vals):
...
+        [True, False, False, False]
+    """
+
+    name = "is_lunch_time"
+    input_types = [ColumnSchema(logical_type=Datetime)]
+    return_type = ColumnSchema(logical_type=BooleanNullable)
+    compatibility = [Library.PANDAS, Library.DASK, Library.SPARK]
+    description_template = "whether {} falls during lunch time"
+
+    def __init__(self, country="US", include_weekends=True, include_holidays=False):
```
thehomebrewnerd, line 366: A couple thoughts here...

- I think we should also include the hour as a configurable parameter as I'm not sure the noon hour is always considered lunch time in all contexts. We can default to `hour=12`.
- Just wondering about the logic for having `include_weekends` default to `True`, but `include_holidays` default to False? If we are trying to return only True in situations where it is lunch hour on a "standard" workday, then I think both should default to False.

featuretools/primitives/standard/datetime_transform_primitives.py
```diff
@@ -331,6 +334,72 @@ def is_leap_year(vals):
...
+        self.country = country
+        years_list = [1950 + x for x in range(150)]
+        self.federal_holidays = getattr(holidays, self.country)(years=years_list)
+        self.include_weekends = include_weekends
+        self.include_holidays = include_holidays
+        self.holidays_df = pd.DataFrame(
+            sorted(self.federal_holidays.items()), columns=["dates", "names"]
+        )
+        self.timestamp_dates = [pd.Timestamp(i) for i in self.holidays_df.dates]
```
thehomebrewnerd, line 375: If we have `include_holidays` default to `False` we only need to do the holiday related setup in this block when a user creates an instance with `include_holidays=True`. Otherwise we are doing unnecessary work here that never gets used later.

thehomebrewnerd: Also `self.timestamp_dates` only seems to be used in the `spark_mask` function, so there we don't need to calculate it here and store it on the primitive object - we can calculate that inside `spark_mask` instead.

featuretools/primitives/standard/datetime_transform_primitives.py
```diff
@@ -331,6 +334,72 @@ def is_leap_year(vals):
...
+        if not self.include_weekends:
+            lunch_mask = lunch_mask and (v.dayofweek < 5)
+        if not self.include_holidays:
+            holiday_mask = v.normalize() in self.timestamp_dates
+            lunch_mask = lunch_mask and not holiday_mask
+        return lunch_mask
+
+    def get_function(self):
+        def is_lunch_time(vals):
+            if isinstance(vals, pd.Series) or isinstance(vals, dd.dataframe.Series):
```
thehomebrewnerd, line 388: I think we should flip the logic around here. The code in the first block should be our default, and we only do special handling for spark.  Also, maybe use our custom `is_instance` function for checking the type of the series.

```python
if is_instance(vals, ps, "Series"):
   # code for Spark series here
else:
    # default code for pandas and Dask here
```

thehomebrewnerd (commented):
I've left a few initial comments on the current code, but in general I feel like we might be trying to do too much with these primitives by including logic about weekends and holidays. Those behaviors are already incorporated in other primitives and we are duplicating some of that work here. 

Maybe we should discuss this more, but my suggestion would be to keep these simple and focus only on the lunch time and working hours components. If users wanted to create a feature that incorporates both elements a custom feature can be created combining the output of `IsLunchTime` and `IsWeekend`,  for example, to determine if a particular date falls during lunch hours on a weekend.

If we decide to leave these primitives as they are with the weekend and holiday params, can try to import our existing primitive code, particularly that in `IsWeekend` and `DateToHoliday`, rather than duplicating it here with a different implementation?

featuretools/primitives/standard/datetime_transform_primitives.py
ozzieD: I agree with both of those both points

sbadithe:
Changes: 

`IsLunchTime` : checks if `dt.hour` is equal to a configurable lunch time
`IsWorkingHours` : checks if `dt.hour` is between configurable start and configurable end hour

featuretools/primitives/standard/datetime_transform_primitives.py
```diff
@@ -331,6 +331,39 @@ def is_leap_year(vals):
         return is_leap_year
 
 
+class IsLunchTime(TransformPrimitive):
+    """Determines if a datetime falls during configurable lunch hour
+
+    Args:
+        lunch_hour (int): Time when lunch is taken. Defaults to 12
```
thehomebrewnerd, line 338: Maybe we should specify what are valid values here. For example, to set 1pm as the lunch hour, I assume users should pass in `13`?

featuretools/primitives/standard/datetime_transform_primitives.py
```diff
@@ -331,6 +331,39 @@ def is_leap_year(vals):
...
+
+    Examples:
```
thehomebrewnerd, line 340: Also would be good to include a doctest example of how to use the primitive with a custom hour set.

featuretools/primitives/standard/datetime_transform_primitives.py
```diff
@@ -459,6 +492,40 @@ def is_weekend(vals):
...
+        end_time (int): End hour of workday. Default is 18 (6pm)
+
+    Examples:
+        >>> from datetime import datetime
+        >>> dates = [datetime(2022, 6, 21, 16, 3, 3),
+        ...          datetime(2019, 1, 3, 4, 4, 4),
+        ...          datetime(2022, 1, 1, 12, 1, 2)]
+        >>> iwh = IsWorkingHours()
+        >>> iwh(dates).tolist()
+        [True, False, True]
```
thehomebrewnerd, line 509: Similar to above, but mention the inputs that are valid when customizing the hours and include a doctest example for customizing the values.

thehomebrewnerd (commented):
Just a couple more minor things, but looking pretty good to me.

docs/source/release_notes.rst
```diff
@@ -5,6 +5,7 @@ Release Notes
 Future Release
 ==============
     * Enhancements
+        * Add ``IsWorkingHours`` and ``IsLunchTime`` transform primitives (:pr:`2130`)
```
thehomebrewnerd, line 8: This got missed earlier, but lines 16 and 17 thanking the contributors below should be indented with a single indent so they align with the first `*` in front of `* Testing Changes`.

featuretools/tests/primitive_tests/test_transform_primitive.py
```diff
@@ -177,6 +179,64 @@ def test_is_quarter_start():
...
+
+
+def test_is_working_hours_standard_hours():
+    is_working_hours = IsWorkingHours()
+    dates = pd.Series(
+        [
+            datetime(2022, 6, 21, 16, 3, 3),
+            datetime(2019, 1, 3, 4, 4, 4),
+            datetime(2022, 1, 1, 12, 1, 2),
+        ]
```
thehomebrewnerd, line 219: One last item - we should add a `nan` value here like we do in `IsLunchHour` just to confirm this primitive doesn't have issues with missing values.

thehomebrewnerd (approved):

PR #2136, opened 2022-06-23
Feature and primitive serialization and deserialization improvements

thehomebrewnerd:
### Serialization and deserialization improvements

Implements several changes to improve serialization and deserialization of Features:
- Uses a common primitive object when deserializing features that use the same primitive rather than creating separate primitive instances for each primitive (#2127)
- Serialize each primitive only once rather than saving duplicate primitive information for features that use the same primitive
- Allow users to easily rename Feature output column names and serialize/deserialize these custom names properly.

### DIFF
--- a/featuretools/feature_base/feature_base.py
+++ b/featuretools/feature_base/feature_base.py
@@ -784,8 +784,8 @@ class AggregationFeature(FeatureBase):
             "where": self.where and self.where.unique_name(),
             "use_previous": self.use_previous and self.use_previous.get_arguments(),
         }
-        if self._names and self.number_output_features > 1:
-            arg_dict["feature_names"] = self._names
+        if self.number_output_features > 1:
+            arg_dict["feature_names"] = self.get_feature_names()
         return arg_dict
 
     def relationship_path_name(self):
@@ -839,8 +839,8 @@ class TransformFeature(FeatureBase):
             "base_features": [feat.unique_name() for feat in self.base_features],
             "primitive": self.primitive,
         }
-        if self._names and self.number_output_features > 1:
-            arg_dict["feature_names"] = self._names
+        if self.number_output_features > 1:
+            arg_dict["feature_names"] = self.get_feature_names()
         return arg_dict
 
 
@@ -906,8 +906,8 @@ class GroupByTransformFeature(TransformFeature):
             "primitive": self.primitive,
             "groupby": self.groupby.unique_name(),
         }
-        if self._names and self.number_output_features > 1:
-            arg_dict["feature_names"] = self._names
+        if self.number_output_features > 1:
+            arg_dict["feature_names"] = self.get_feature_names()
         return arg_dict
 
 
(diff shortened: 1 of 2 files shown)


featuretools/feature_base/feature_base.py
```diff
@@ -103,6 +101,33 @@ def get_feature_names(self):
...
+            raise ValueError(
+                "Number of names provided must match the number of output features:"
+                f" {num_new_names} name(s) provided, {self.number_output_features} expected."
+            )
+
+        if len(set(names)) != num_new_names:
+            raise ValueError("Provided output feature names must be unique.")
+
+        self._names = names
+
```
ozzieD, line 130: An observation I have with this code base is that we are doing getters and setters like this:
```
class MyClass(object):
    def get_my_attr(self):
        return self._my_attr

    def set_my_attr(self, value):
        self._my_attr = value
```

When we really should be doing it the more pythonic way: 
```
class MyClass(object):
    @property
    def my_attr(self):
        return self._my_attr

    @my_attr.setter
    def my_attr(self, value):
        self._my_attr = value
```
anyway we could start using this pattern here? and maybe modify`get_feature_names` to just be

`@property`
`def feature_names` etc...

featuretools/primitives/utils.py
```diff
@@ -357,7 +357,13 @@ class PrimitivesDeserializer(object):
     """
```
ozzieD, line 357: misspelling on line 355

featuretools/primitives/utils.py
```diff
@@ -357,7 +357,13 @@ class PrimitivesDeserializer(object):
...
 
     def __init__(self):
-        self.class_cache = {}  # (class_name, module_name) -> class
+        # Cache to avoid repeatedly searching for primitive class
+        # (class_name, module_name) -> class
+        self.class_cache = {}
+
+        # Cache to use existing primitive instances
+        # primitive key -> primitive instance
+        self.instance_cache = {}
```
ozzieD, line 366: Are we using instance_cache anywhere?

featuretools/feature_base/feature_base.py
thehomebrewnerd: Yes, I think `column_names` is better than `feature_names`.

gsheni: @ozzieD this is a great suggestion and definitely something we should do. Let's do this in a follow-up issue/PR so that we can get the release out today.

@thehomebrewnerd Can you create the issue?

ozzieD: Yeah I have no issue with keeping it the way it is now and addressing it in a separate PR.

thehomebrewnerd: https://github.com/alteryx/featuretools/issues/2161

rwedge (commented):
Let's mention `set_feature_names` in the release notes as an improvement

thehomebrewnerd:
> Let's mention `set_feature_names` in the release notes as an improvement

Separated that PR out into a separate entry in the Enhancements section instead of lumping it in with the other changes in the Changes section.

ozzieD (approved):

PR #2137, opened 2022-06-23
Fixes for numpy 1.23.0

thehomebrewnerd:
### Fixes for numpy 1.23.0

Implement fixes for compatibility with numpy 1.23.0

With numpy 1.23.0 the `np.divide` function used in the `DivideNumeric` primitive now generates an error with pyspark inputs.
```
np.divide(df["ser1"], df["ser2"])

NotImplementedError: pandas-on-Spark objects currently do not support <ufunc 'divide'>.
```

To resolve, the primitive was updated to divide the two series directly rather than using the np.divide function.

Also had to update the doctest for the `NaturalLogarithm` primitive to round primitive output to account for slightly different floating point output with the new version of numpy.


gsheni (approved):

featuretools/primitives/standard/binary_transform.py
```diff
@@ -881,7 +881,10 @@ def __init__(self, commutative=False):
         self.commutative = commutative
 
     def get_function(self):
-        return np.divide
+        def divide_numeric(val1, val2):
```
gsheni, line 884: When primitives directly return a function, it tends to be faster. Is there no way to keep using `np.divide`?

thehomebrewnerd: image

Not sure it makes a difference in this particular case - perhaps because we are using optimized approaches in both cases (numpy vs pandas)?

gsheni (approved):

PR #2142, opened 2022-06-23
Allow users to directly set feature output column names and save during serialization

thehomebrewnerd:
### Allow users to directly set feature output column names and save during serialization

Adds a `set_feature_names` method to `FeatureBase` to directly allow for setting of feature column names to override the default name. Also updates serialization to save and restore these names.


featuretools/tests/computational_backend/test_calculate_feature_matrix.py
```diff
@@ -2266,3 +2275,25 @@ def multi_cum_sum(x):
...
+    direct_feat = ft.Feature(pd_es["sessions"].ww["device_name"], "log")
+    trans_feat = ft.Feature(pd_es["log"].ww["value"], primitive=Negate)
+    multi_output_agg_feat = ft.Feature(
+        pd_es["log"].ww["product_id"],
+        parent_dataframe_name="customers",
+        primitive=NMostCommon(n=2),
+    )
+
+    direct_name = ["device_direct"]
+    direct_feat.set_feature_names(direct_name)
```
rwedge, line 2290: for single output features, do we want `set_feature_names` and `rename` to both work?

rwedge (commented):
Can we test that a feature stacked on a renamed column from a feature with multiple outputs uses the renamed name?

featuretools/tests/computational_backend/test_calculate_feature_matrix.py
thehomebrewnerd: Not sure - are you suggesting that we just allow `set_feature_names` to work for multi-output features and renaming feature columns for single output should be done through `rename` as you can be done currently?

thehomebrewnerd: As I think about it I think I like the idea of having `set_feature_names` operate only on multi-output features - seems more consistent with the method name, and would reduce the scope of the change. We could add a check in the method to raise an error if `number_output_features == 1`.

@rwedge What do you think about that approach?

thehomebrewnerd: @rwedge Sounds good. I'll make this change and add the additional test case you recommended.

thehomebrewnerd:
> Can we test that a feature stacked on a renamed column from a feature with multiple outputs uses the renamed name?

@rwedge I added a test that checks the name of a feature created by stacking on a `FeatureOutputSlice` of a multi-output feature. Also updated tests to include a multi-output GroupbyTransformFeature.

featuretools/feature_base/feature_base.py
```diff
@@ -107,19 +107,26 @@ def get_feature_names(self):
     def set_feature_names(self, names):
         """Set new values for the feature column names, overriding the default values.
         Number of names provided much match the number of output columns defined for
-        the feature.
+        the feature. Only works for features that have more than one output column. Use
+        ``Feature.rename`` for changing the column name for single output features.
```
rwedge, line 111:
```suggestion
        ``Feature.rename`` to change the column name for single output features.
```

rwedge (approved):

PR #2147, opened 2022-06-25
Allow DFS to take in strings or datetime values into cutoff_time argument

### LINKED ISSUE #2014: dfs `cutoff_time` should take a datetime string
- As a user, I wish featuretools `dfs` would take a string as cutoff_time aswell as a datetime object

#### Code Example

```python
                      target_dataframe_name='customers',
                      cutoff_time="2014-1-1 05:00",
                      instance_ids=[1],
                      cutoff_time_in_index=True)

```

as well as 

```python
                      target_dataframe_name='customers',
                      cutoff_time=dateutil.parser.parse("2014-1-1 05:00"),
                      instance_ids=[1],
                      cutoff_time_in_index=True)
```

sbadithe:
Fixes #2014 

Makes it possible for users to pass in strings, or datetime values into dfs's `cutoff_time` argument

### DIFF
--- a/featuretools/synthesis/dfs.py
+++ b/featuretools/synthesis/dfs.py
@@ -245,12 +245,16 @@ def dfs(
         entityset = EntitySet("dfs", dataframes, relationships)
 
     if isinstance(cutoff_time, str):
-        try: 
+        try:
             cutoff_time = dateutil.parser.parse(cutoff_time)
         except dateutil.parser.ParserError:
-            raise ValueError(f"The provided cutoff_time of {cutoff_time} is either an invalid date or in an unknown format")
-        except OverflowError: 
-            raise OverflowError("The parsed date would exceed the largest valid integer") 
+            raise ValueError(
+                f"The provided cutoff_time of {cutoff_time} is either an invalid date or in an unknown format"
+            )
+        except OverflowError:
+            raise OverflowError(
+                "The parsed date would exceed the largest valid integer"
+            )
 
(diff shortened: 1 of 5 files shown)


featuretools/synthesis/dfs.py
```diff
@@ -242,6 +244,9 @@ def dfs(
     if not isinstance(entityset, EntitySet):
         entityset = EntitySet("dfs", dataframes, relationships)
 
+    if isinstance(cutoff_time, str):
+        cutoff_time = dateutil.parser.parse(cutoff_time)
```
gsheni, line 248: What if this raises a ParserError or OverflowError?

sbadithe: Thanks. I added handling for parseutil's ParserError and a plain OverflowError (looks like parseutil doesn't have a OverflowError defined, as far as I can tell). In the event of a ParserError is it best to just raise a ValueError?

gsheni: Not sure. @thehomebrewnerd any clue on what we usually do here?

sbadithe: I used `dateutil.parser.parse` only because it was in the issue description; I can change it to `pandas.to_datetime` right now. If `pandas.to_datetime` fails, it returns `pd.NaT`. In that case, what would be the most sensible behavior for the function?

dvreed77 (approved):
lgtm

PR #2155, opened 2022-06-28
Add periods parameter to Diff primitive and add DiffDatetime primitive

### LINKED ISSUE #1805: Add `gap` parameter to Diff primitive
The `Diff` primitive is potentially useful for time series problems, since it can indicate whether the relationship between two rows is increasing or decreasing. However, `Diff`'s current implementation subtracts the current row from the row directly above it. This means that it cannot be used for a time series problem or it'd expose the target variable. 

If we add a `gap` parameter that shifts the data by a certain amount, this could allow `Diff` to be used for time series problems. The default for the `gap` parameter would be `0` to keep the behavior the same.

ozzieD:
- Fixes #1805
- Added `periods` parameter in `Diff` so we can shift the difference on time series data
- After discussing with team, we added a separate `DiffDatetime` primitive


featuretools/primitives/standard/transform_primitive.py
```diff
@@ -261,9 +280,12 @@ class Diff(TransformPrimitive):
     uses_full_dataframe = True
     description_template = "the difference from the previous value of {}"
 
+    def __init__(self, periods=0):
+        self.periods = periods
+
     def get_function(self):
         def pd_diff(values):
-            return values.diff()
+            return values.shift(self.periods).diff()
```
tamargrey, line 288: Was using `diff(periods=self.periods)` not doable in the end?

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -253,6 +257,21 @@ class Diff(TransformPrimitive):
...
         [nan, 9.0, -7.0, 1.0, 11.0]
+
+        You can specify the number of periods to shift the values
+
+        >>> values = [1, 2, 4, 7, 11, 16]
+        >>> diff_periods = Diff(periods = 1)
+        >>> diff_periods(values).tolist()
+        [nan, nan, 1.0, 2.0, 3.0, 4.0]
+
+        Using datetimes
```
tamargrey, line 268: I think we'd need to add an input_type to allow datetime columns to actually be used by this primitive. So instead of `input_types = [ColumnSchema(semantic_tags={"numeric"})]` it could be `input_types = [[ColumnSchema(semantic_tags={"numeric"})], [ColumnSchema(logical_type=Datetime)]]`.

We might want to consider requiring it be a time index even, since something like diff only is relevant when there's an ordering to the data...

featuretools/primitives/standard/transform_primitive.py
ozzieD: Doesn't really do what we want. using periods in diff simply just skips which row you subtract from current row. So instead of row - row-1. Period =2 would do row -row-2. Using shift does more of what we were describing and its what evalml using in `compute_delays`. https://github.com/alteryx/evalml/blob/01da9097683c61aad0c2466e93590222113971a3/evalml/pipelines/components/transformers/preprocessing/time_series_featurizer.py#L208

tamargrey: Ahh that sounds familiar now that you mentions. I feel like I ran into that while creating the issue.

featuretools/primitives/standard/transform_primitive.py
ozzieD: Yeah I was confused by this since technically I can use a datetime series no problem. I will add that in.

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -253,6 +257,21 @@ class Diff(TransformPrimitive):
         >>> values = [1, 10, 3, 4, 15]
         >>> diff(values).tolist()
         [nan, 9.0, -7.0, 1.0, 11.0]
+
+        You can specify the number of periods to shift the values
+
+        >>> values = [1, 2, 4, 7, 11, 16]
+        >>> diff_periods = Diff(periods = 1)
+        >>> diff_periods(values).tolist()
+        [nan, nan, 1.0, 2.0, 3.0, 4.0]
```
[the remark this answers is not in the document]
ozzieD: First row that's output even with periods=0 (current default behavior) will always be nan since the previous row doesn't exist.

featuretools/primitives/standard/transform_primitive.py
tamargrey: Yep the input types won't stop you from calling the primitive's function on whatever series you choose, but it will impact which columns a primitive is applied to automatically in dfs.

gsheni:
Are we good to merge this? Or would you like another review?

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -242,32 +243,81 @@ class Diff(TransformPrimitive):
...
     """
 
     name = "diff"
-    input_types = [ColumnSchema(semantic_tags={"numeric"})]
+    input_types = [
+        [ColumnSchema(semantic_tags={"numeric"})],
+        [ColumnSchema(semantic_tags={"time_index"})],
```
tamargrey, line 272: You'll still need to remove this column schema, or dfs will apply `Diff` to a datetime time index as well. If you're trying to specify that diff should only apply to numeric time indices, you could do `ColumnSchema(semantic_tags={"numeric", "time_index"})`

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -347,12 +348,18 @@ def test_diff(pd_es):
     )
     diff2 = ft.Feature(value, groupby=customer_id_feat, primitive=Diff)
 
-    feature_set = FeatureSet([diff1, diff2])
+    diff_periods = ft.Feature(
+        value, groupby=customer_id_feat, primitive=Diff(periods=1)
```
tamargrey, line 352: I like how you split the diff datetime tests into two below -- could you pull out the periods checks into a second `test_diff_shift` test as well?

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -421,6 +421,65 @@ def test_diff_single_value_is_nan(pd_es):
...
+def test_diff_datetime(pd_es):
+    diff = ft.Feature(
+        pd_es["log"].ww["datetime"],
+        primitive=DiffDatetime,
+    )
+    feature_set = FeatureSet([diff])
+    calculator = FeatureSetCalculator(pd_es, feature_set=feature_set)
+    df = calculator.run(np.array(range(15)))
+    vals = df[diff.get_name()].tolist()
+    print(vals)
```
tamargrey, line 433: extra print statement

featuretools/primitives/standard/transform_primitive.py
ozzieD: forgot to remove this. Thanks!

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -261,13 +273,53 @@ class Diff(TransformPrimitive):
...
+
     def get_function(self):
         def pd_diff(values):
-            return values.diff()
+            return values.shift(self.periods).diff()
 
         return pd_diff
 
 
+class DiffDatetime(Diff):
```
tamargrey, line 286: Can you add this to the api_reference under Cumulative Transform Primitives?

tamargrey (dismissed):
Small comment about the api reference, but other than that, looks good!

featuretools/primitives/standard/transform_primitive.py
ozzieD: Done!

tamargrey (approved):
lgtm!

PR #2166, opened 2022-07-01
clean primitive_tests

### LINKED ISSUE #1511: standardize in-library imports within the test cases
The idea is that unless a test case is testing that `ft.some_function` is importable, it shouldn't use `ft.` within the test case and should instead import `some_function` at the top of the file

sbadithe:
Partially fixes #1511 in the primitive_tests folder.

### DIFF
--- a/featuretools/tests/primitive_tests/test_agg_feats.py
+++ b/featuretools/tests/primitive_tests/test_agg_feats.py
@@ -8,7 +8,15 @@ import pytest
 from woodwork.column_schema import ColumnSchema
 from woodwork.logical_types import Datetime
 
-import featuretools as ft
+from featuretools import (
+    AggregationFeature,
+    Feature,
+    IdentityFeature,
+    Timedelta,
+    calculate_feature_matrix,
+    dfs,
+    primitives,
+)
 from featuretools.entityset.relationship import RelationshipPath
 from featuretools.primitives import (
     Count,
@@ -47,22 +55,20 @@ def test_primitive():
 
 
 def test_get_depth(es):
-    log_id_feat = ft.IdentityFeature(es["log"].ww["id"])
-    customer_id_feat = ft.IdentityFeature(es["customers"].ww["id"])
-    count_logs = ft.Feature(
-        log_id_feat, parent_dataframe_name="sessions", primitive=Count
-    )
-    sum_count_logs = ft.Feature(
+    log_id_feat = IdentityFeature(es["log"].ww["id"])
+    customer_id_feat = IdentityFeature(es["customers"].ww["id"])
+    count_logs = Feature(log_id_feat, parent_dataframe_name="sessions", primitive=Count)
+    sum_count_logs = Feature(
         count_logs, parent_dataframe_name="customers", primitive=Sum
     )
     num_logs_greater_than_5 = sum_count_logs > 5
-    count_customers = ft.Feature(
+    count_customers = Feature(
         customer_id_feat,
         parent_dataframe_name="régions",
         where=num_logs_greater_than_5,
         primitive=Count,
     )
-    num_customers_region = ft.Feature(count_customers, dataframe_name="customers")
+    num_customers_region = Feature(count_customers, dataframe_name="customers")
 
     depth = num_customers_region.get_depth()
     assert depth == 5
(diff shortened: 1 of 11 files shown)


tamargrey:
Looks like theres a couple of final uses of the `ft` import in https://github.com/alteryx/featuretools/blob/Clean-imports/featuretools/tests/utils_tests/test_gen_utils.py

ozzieD (dismissed):

sbadithe:
> Looks like theres a couple of final uses of the `ft` import in https://github.com/alteryx/featuretools/blob/Clean-imports/featuretools/tests/utils_tests/test_gen_utils.py

I might be misunderstanding this, but since we import `list_logical_types, list_semantic_tags` from woodwork as well, do we want to keep the `ft.` import to disambiguate between the woodwork version and featuretools version?

thehomebrewnerd:
> > Looks like theres a couple of final uses of the `ft` import in https://github.com/alteryx/featuretools/blob/Clean-imports/featuretools/tests/utils_tests/test_gen_utils.py
> 
> I might be misunderstanding this, but since we import `list_logical_types, list_semantic_tags` from woodwork as well, do we want to keep the `ft.` import to disambiguate between the woodwork version and featuretools version?

Can we use the `import ... as ...` syntax for the Woodwork import to do something like `from woodwork import list_logical_types as ww_list_logical_types` and keep the FT versions as their default names?

docs/source/release_notes.rst
```diff
@@ -27,6 +27,7 @@ v1.11.0 Jun 30, 2022
     * Testing Changes
         * Fix pandas warning and reduce dask .apply warnings (:pr:`2145`)
         * Pin graphviz version used in windows tests (:pr:`2159`)
+        * Standardize imports in `primitive_tests` folder (:pr:`2166`)
```
thehomebrewnerd, line 30: This change looks bigger than just the `primitive_tests` folder, so maybe update the entry accordingly. Also, this needs to move to the Future Release section (which needs to be uncommented) and not in the `v1.11.0` release section.

sbadithe:
> > > Looks like theres a couple of final uses of the `ft` import in https://github.com/alteryx/featuretools/blob/Clean-imports/featuretools/tests/utils_tests/test_gen_utils.py
> > 
> > 
> > I might be misunderstanding this, but since we import `list_logical_types, list_semantic_tags` from woodwork as well, do we want to keep the `ft.` import to disambiguate between the woodwork version and featuretools version?
> 
> Can we use the `import ... as ...` syntax for the Woodwork import to do something like `from woodwork import list_logical_types as ww_list_logical_types` and keep the FT versions as their default names?

It should be fixed now. Thanks for the suggestion!

featuretools/tests/entityset_tests/test_es.py
```diff
@@ -25,7 +25,7 @@
     SubRegionCode,
 )
 
-import featuretools as ft
+from featuretools import Relationship, demo
```
thehomebrewnerd, line 28: I think we should change `demo` here to do a `from featuretools.demo import load_retail` to stick with this theme - import the function directly, not the entire demo module.

featuretools/tests/primitive_tests/test_features_deserializer.py
```diff
@@ -3,7 +3,14 @@
...
 import pytest
 
-import featuretools as ft
+from featuretools import (
+    AggregationFeature,
+    Feature,
+    IdentityFeature,
+    TransformFeature,
+    __version__,
+    primitives,
```
thehomebrewnerd, line 12: I suggest importing the primitive we need directly instead of importing the entire primitives module.

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -7,7 +7,16 @@
...
-import featuretools as ft
+from featuretools import (
+    AggregationFeature,
+    EntitySet,
+    Feature,
+    IdentityFeature,
+    TransformFeature,
+    calculate_feature_matrix,
+    dfs,
+    primitives,
```
thehomebrewnerd, line 18: Import specific primitives instead of module.

featuretools/tests/primitive_tests/test_agg_feats.py
```diff
@@ -8,7 +8,15 @@
...
 
-import featuretools as ft
+from featuretools import (
+    AggregationFeature,
+    Feature,
+    IdentityFeature,
+    Timedelta,
+    calculate_feature_matrix,
+    dfs,
+    primitives,
```
thehomebrewnerd, line 18: Import specific primitives instead of primitives module

featuretools/tests/computational_backend/test_feature_set_calculator.py
```diff
@@ -7,8 +7,14 @@
...
 
-import featuretools as ft
-from featuretools import Timedelta
+from featuretools import (
+    AggregationFeature,
+    EntitySet,
+    Feature,
+    Timedelta,
+    calculate_feature_matrix,
+    primitives,
```
thehomebrewnerd, line 16: Import specific primitives instead of module.

thehomebrewnerd (commented):
Also need to remove the `docs/.notes.swp` file from this PR.

featuretools/tests/primitive_tests/test_transform_features.py
sbadithe: We access `dir(primitives)` in this file, is it ok to leave 'primitives' in the import statement?

thehomebrewnerd: I didn't catch that. In that case, yes, it is fine to leave the import, but would confirm we aren't accessing any primitives with something like `primitives.Year`.

PR #2167, opened 2022-07-05
Remove 24th hour from and add 0th hour to PartOfDay

tamargrey:
Fixes bug where any timestamps from the 0th hour of a day wouldn't get transformed correctly to `midnight`


thehomebrewnerd (approved):
LGTM!

Issue #1806, opened 2021-12-09
Add a `RollingTrend` transform primitive


tamargrey:
The concept of `Trend` is important for time series problems, and Featuretools' current `Trend` primitive is an AggregationPrimitive, which means that it can't be used in the single table case of just having a datetime column and a target variable column.

In order to be used for time series feature engineering, we could create a `RollingTrend` primitive that lets users set a `gap` and `window_length` to calculate a trend over their specified rolling window across the dataset. 

The primitive could look something like 

```python

class RollingTrend(TransformPrimitive):
    name = "rolling_trend"
    input_types = [ColumnSchema(logical_type=Datetime, semantic_tags={'time_index'}), ColumnSchema(semantic_tags={'numeric'})]
    return_type = ColumnSchema(logical_type=Double, semantic_tags={'numeric'})

    def __init__(self, window_length=3, gap=0, min_periods=0):
        self.window_length = window_length
        self.gap = gap
        self.min_periods = min_periods

    def get_function(self):
        def rolling_trend(datetime, numeric):
            x = pd.Series(numeric.values, index=datetime.values)
            rolled_series = _roll_series_with_gap(x,
                                                  self.window_length,
                                                  gap=self.gap,
                                                  min_periods=self.min_periods)
            return rolled_series.apply(Trend.get_function).values
        return rolling_trend

```

PR #2170, opened 2022-07-05
Add RollingTrend primitives and refactor some util methods

ozzieD:
Resolves: https://github.com/alteryx/featuretools/issues/1806

also fix code coverage issues


featuretools/tests/primitive_tests/test_rolling_primitive.py
```diff
@@ -2,15 +2,17 @@
 
 import pandas as pd
 import pytest
+from numpy import isnan, nan
```
sbadithe, line 5: I think if we are importing `pandas as pd` we should import `numpy as np` to be consistent.

sbadithe, line 5: I think if we are importing `pandas as pd` we should import `numpy as np` to be consistent.

featuretools/primitives/api.py
```diff
@@ -1,4 +1,5 @@
 # flake8: noqa
+from .rolling_primitive_utils import apply_roll_with_offset_gap, roll_series_with_gap
```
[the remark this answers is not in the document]
ozzieD: So I actually want to have a discussion on this. I'm curious as to why we have an `api.py` file instead of just doing an `__all__` in `__init__.py` since this is essentially the same thing and we could get rid of the flake8 ignores. My intention with this was to just make the imports simpler but if this is only for exposing user facing functions then I'd be okay with removing.

featuretools/tests/primitive_tests/test_primitive_utils.py
```diff
@@ -37,14 +37,16 @@
     get_transform_primitives,
 )
 from featuretools.primitives.base import PrimitiveBase
+from featuretools.primitives.rolling_primitive_utils import (
```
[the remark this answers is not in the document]
ozzieD: Yeah I can pull those tests out into their own file.

featuretools/primitives/standard/rolling_transform_primitive.py
```diff
@@ -421,13 +425,80 @@ def _pandas_std(series):
...
+
+    Description:
+        Given a list of numbers and a corresponding list of
+        datetimes, return a rolling slope of the linear trend
+        of values, starting at the row `gap` rows away from the current row and looking backward
+        over the specified time window (by `window_length` and `gap`).
+
+        Input datetimes should be monotonic.
+
+    Examples:
```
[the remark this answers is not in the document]
ozzieD: whoops just missed that.

featuretools/primitives/standard/rolling_transform_primitive.py
```diff
@@ -421,13 +425,80 @@ def _pandas_std(series):
...
+        self.min_periods = min_periods
+
+    def get_function(self):
+        def rolling_trend(datetime, numeric):
+            x = pd.Series(numeric.values, index=datetime.values)
+            rolled_series = roll_series_with_gap(
+                x, self.window_length, gap=self.gap, min_periods=self.min_periods
+            )
+            if isinstance(self.gap, str):
+                additional_args = (self.gap, np.mean, self.min_periods)
```
tamargrey, line 492: `(self.gap, np.mean, self.min_periods)` is still calculating the average when an offset string gap is passed in. Should be `(self.gap, calculate_trend, self.min_periods)`

ozzieD: oof

docs/source/release_notes.rst
```diff
@@ -10,6 +10,7 @@ Future Release
     * Enhancements
         * Add ``IsWorkingHours`` and ``IsLunchTime`` transform primitives (:pr:`2130`)
         * Add periods parameter to ``Diff`` and add ``DiffDatetime`` primitive (:pr:`2155`)
+        * Add ``RollingTrend`` primitve (:pr:`2170`)
```
dvreed77, line 13: spelling

tamargrey (approved):
LGTM!

featuretools/primitives/api.py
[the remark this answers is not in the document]
dvreed77: I'm not sure if this is what @thehomebrewnerd is alluding to, but if these functions aren't public, should we keep the `_` at the begining?

featuretools/primitives/standard/aggregation_primitives.py
```diff
@@ -14,7 +14,7 @@
 )
 
 from featuretools.primitives.base.aggregation_primitive_base import AggregationPrimitive
-from featuretools.utils import convert_time_units
+from featuretools.utils import calculate_trend, convert_time_units
```
dvreed77, line 17: How does this work? Aren't these utils in time_utils.py?

featuretools/primitives/utils.py
```diff
@@ -398,179 +397,6 @@ def _find_class_in_descendants(self, search_key):
                 return cls
 
 
-def _roll_series_with_gap(series, window_size, gap=0, min_periods=1):
```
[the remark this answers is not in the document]
ozzieD: Yeah I would like that as well. It would just be inconsistent with how the repo is currently structured since all agg and trans primitives are just listed in a single file. I'd be down with this if we did this for the rest of the primitives. I was going to take it a step further and just put these util methods in a utils class where they are accessible as staticmethods but I might be going too far on the OOP side of the spectrum lol.

docs/source/release_notes.rst
ozzieD: oops

featuretools/primitives/utils.py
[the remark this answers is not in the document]
dvreed77: Naw, you get basically the same namespacing concept you want by leveraging python modules.

featuretools/primitives/standard/aggregation_primitives.py
ozzieD: I took out the guts of `Trend` in agg prims, slightly modified it so it could be used for the Rolling version, and placed it into `calculate_trend`. I took `calculate_trend` its helper functions away from `aggregation_primitive.py` into `time_utils.py` since I felt like that's where they kind of belonged based on the name of the file.

sbadithe (dismissed):

sbadithe:
Can probably get the Woodwork integration fix so you don't have to cap Woodwork before merging.

sbadithe:
Woodwork tests should fix hopefully with #2182!

gsheni (approved):

gsheni (approved):

PR #2177, opened 2022-07-11
Add pre-commit hooks for linting

gsheni:
- This makes it so that linting is automatically run right before you commit code:
- We use this https://pre-commit.com/
- Example of this:
Screen Shot 2022-07-11 at 5 30 10 PM


gsheni:
@sbadithe Can checkout this branch and make sure the steps in `contributing.md` work for you? They are as follows:
```
pip install pre-commit
pre-commit install
pre-commit run
```

sbadithe:
> @sbadithe Can checkout this branch and make sure the steps in `contributing.md` work for you? They are as follows:
> 
> ```
> pip install pre-commit
> pre-commit install
> pre-commit run
> ```

Commands run for me, but it skips all the checks as it says there are no files to check

gsheni:
@sbadithe Yes, it will only run on the changed files

contributing.md
```diff
@@ -100,6 +100,12 @@ Before starting major work, you should touch base with the maintainers of Featur
   # view docs locally
   open build/html/index.html
   ```
+* Before you commit, a few lint fixing hooks will run. You can also manually run these.
+  ```bash
+  pip install pre-commit
+  pre-commit install
```
thehomebrewnerd, line 106: Do we need to list the `pip install pre-commit` and `pre-commit install` commands or will that be covered when users run `make installdeps`?

I created a fresh environment and just ran `make installdeps` and it seems like everything worked.

PR #2187, opened 2022-07-19
2179 Restructure Directories

### LINKED ISSUE #2179: Restructure primitives folder and various util functions
The current structure of the primitive's folder makes it confusing to work with and harder to scale. @dvreed77  and I think a better structure for the primitives would be:

```
primitives
│  __init__.py ( remove all api.py since its not needed)
└───aggregation
|   |    __init__.py 
│  └───standard
│     │    __init__.py 
│     │   *each agg primitive listed in its own file*
└───transform
│   │    __init__.py 
│   └───standard
|       |    __init__.py 
│       │   *each trans primitive listed in its own file from `transform_primitive.py`*
│   └───cumulative
|       |    __init__.py 
│       │   *each cum trans primitive listed in its own file from `cum_transform_feature.py`*
│   └───datetime
|       |    __init__.py 
|       |    utils.py 
│       │   *each datetime transform primitive listed in its own file from `datetime_transform_primitive.py`*
│   └───latlong
|       |    __init__.py 
|       |    utils.py 
│       │   *each latlong transform primitive listed in its own file from `latlong_transform_primitive.py`*
│   └───rolling
|       |    __init__.py 
|       |    utils.py 
│       │   *each rolling transform primitive listed in its own file from `rolling_transform_feature.py`*

### PR DESCRIPTION
**Pull request by dvreed77**
Restructure Directories and Files to be more modular.

fixes #2179

### CODE REVIEW

File: featuretools/primitives/aggregation/api.py [hunk 1]
```diff
@@ -0,0 +1,7 @@
+from featuretools.primitives.aggregation import *
+from warnings import warn
+
+warn(
```
gsheni, line 4: What is this? Users have to change their import statements just because we re-organized directories? 
I don't think that is ideal

dvreed77: Only if they happen to be importing that far down from `featuretools.aggregation.api`

rwedge:
Why is `primitives.base` being renamed to `primitives.core`?

dvreed77:
> Why is `primitives.base` being renamed to `primitives.core`?

WIll change that back, was getting a little trigger happy :)

rwedge:
The idea behind `primitives.standard` is that if a primitive was in that module, it was one of the primitives that was part of the basic featuretools installation.  Since `featuretools.primitives` loads in primitives from other libraries via an entry point, `featuretools.primitives` isn't a good indicator of that.

gsheni:
@dvreed77 do we still want to do this or should we close this PR?

thehomebrewnerd:
Closing in favor of #2331

PR #2194, opened 2022-07-19
Run codecov on py310

rwedge:


rwedge:
Looks like the codecov misses I saw the other day were due to bugfixes in later python versions fixing false misses

PR #2200, opened 2022-07-22
Add unit tests to enforce primitive docstring verb usage

### LINKED ISSUE #2199: Add unit test to enforce primitive descriptions
- We want our primitive descriptions to be consistent and easy to understand.
- In order to do this, we can add the following test (for all primitives)
```python
docstring = primitive.__doc__
short_description = docstring.splitlines()[0]
first_word = short_description.split(' ', 1)[0]
valid_verbs = ['Calculates', 'Determines', 'Transforms', 'Computes',
                         'Shifts', 'Extracts', 'Applies']
assert any(s in first_word for s in valid_verbs)
```
- We can add more valid verbs if they are good verbs

sbadithe:
Fixes #2199 

-- Adds unit test that checks if primitive's docstring begins with verb in set of `valid_verbs`.
-- Rewords docstrings for primitives that would fail test above.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -13,6 +13,7 @@ Future Release
         * Add create feedstock PR workflow (:pr:`2181`)
         * Limit performance tests to python 3.8 (:pr:`2198`)
         * Add test to ensure primitive docstrings use standardized verbs (:pr:`2200`)
+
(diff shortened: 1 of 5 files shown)


gsheni (approved):
lgtm

PR #2210, opened 2022-07-28
[DRAFT] Add `series_library` to `PrimitiveBase`

### LINKED ISSUE #2165: Add series_library argument to transform primitives
- https://github.com/alteryx/featuretools/pull/2111

sbadithe:
- Fixes #2165 

The main purpose of this PR is to add a `series_library` attribute to `PrimitiveBase` and `FeatureBase`. In order to do this, it:  
- Adds `series_library` as an attribute on `primitive_dict` in the serialization and deserialization process 
- Adds branching logic to `feature_base` to determine the `series_library` based on the type of dataframe
- Refactors tests in `test_feature_serialization.py` and `test_features_serializer.py` to serialize the specific primitive instance on the feature, instead of just a new instance of the primitive
- Changes `Library` to be a string Enum to enable serialization

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,13 +7,14 @@ Future Release
 ==============
     * Enhancements
     * Fixes
+        * Fix compatibility issues with holidays 0.15 (:pr:`2254`)
     * Changes
         * Update release notes to make clear conda release portion (:pr:`2249`)
     * Documentation Changes
     * Testing Changes
 
     Thanks to the following people for contributing to this release:
-    :user:`gsheni`
+    :user:`cp2boston`, :user:`gsheni`
 
 v1.13.0 Aug 18, 2022
 ====================
--- a/featuretools/primitives/standard/datetime_transform_primitives.py
+++ b/featuretools/primitives/standard/datetime_transform_primitives.py
@@ -1,4 +1,3 @@
-import holidays
 import numpy as np
 import pandas as pd
 from woodwork.column_schema import ColumnSchema
@@ -991,24 +990,12 @@ class IsFederalHoliday(TransformPrimitive):
 
     def __init__(self, country="US"):
         self.country = country
-        try:
-            self.holidays = holidays.country_holidays(country=self.country)
-        except NotImplementedError:
-            available_countries = (
-                "https://github.com/dr-prodigy/python-holidays#available-countries"
-            )
-            error = "must be one of the available countries:\n%s" % available_countries
-            raise ValueError(error)
-        years_list = [1950 + x for x in range(150)]
-        self.federal_holidays = getattr(holidays, country)(years=years_list)
+        self.holidayUtil = HolidayUtil(country)
 
     def get_function(self):
         def is_federal_holiday(x):
-            holidays_df = pd.DataFrame(
-                sorted(self.federal_holidays.items()),
-                columns=["dates", "names"],
-            )
-            is_holiday = x.dt.normalize().isin(holidays_df.dates)
+            holidays_df = self.holidayUtil.to_df()
+            is_holiday = x.dt.normalize().isin(holidays_df.holiday_date)
             if x.isnull().values.any():
                 is_holiday = is_holiday.astype("object")
                 is_holiday[x.isnull()] = np.nan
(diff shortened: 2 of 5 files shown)


dvreed77:
I'm trying to think of a good way to test this.  

We need to make sure series is set before we get the function in the `feature_set_calculator`

featuretools/feature_base/feature_base.py
```diff
@@ -47,6 +51,13 @@ def __init__(
         # initialize if not already initialized
         if not isinstance(primitive, PrimitiveBase):
             primitive = primitive()
+
```
thehomebrewnerd, line 54: What if the user has already instantiated the primitive before creating the Feature? I don't think the series type will get set in that case.

sbadithe: If I am understanding correctly: this should be outside the `if-block`, right?

thehomebrewnerd: Right - we always want to set the series type on the primitive, not just when it is instantiated during Feature creation.

featuretools/feature_base/feature_base.py
```diff
@@ -47,6 +51,13 @@ def __init__(
...
+            # default library is PANDAS
+            if isinstance(dataframe, dd.DataFrame):
+                primitive.set_series_library(Library.DASK)
+            elif is_instance(dataframe, ps, "DataFrame"):
```
dvreed77, line 58: I think we want to set the instance even if primitive is initialized

rwedge (commented):
What if the primitive was instantiated before the feature was created?

Does the `series_library` information persist after using `ft.save_features` and `ft.load_features`?

featuretools/primitives/base/primitive_base.py
```diff
@@ -156,3 +156,6 @@ def flatten_nested_input_types(input_types):
                 sub_input for input_obj in input_types for sub_input in input_obj
             ]
         return input_types
+
+    def set_series_library(self, series_library: str):
```
thehomebrewnerd, line 160: Is there a reason to have this as a method rather than just setting the attribute directly?

rwedge:
> We need to make sure series is set before we get the function in the `feature_set_calculator`

What if near the start of `calculate_feature_matrix` we looped through the features and ensured `series_library` was set to the right library type?

sbadithe:
> What if the primitive was instantiated before the feature was created?
> 
> Does the `series_library` information persist after using `ft.save_features` and `ft.load_features`?

Could I test this by adding a test to `test_feature_serialization.py`?

ozzieD:
It seems like you may have to add `series_library` to the serializer now.

setup.cfg
```diff
@@ -44,7 +44,7 @@ include_package_data = True
 install_requires =
     click >= 7.0.0
     cloudpickle >= 1.5.0
-    dask[dataframe] >= 2021.10.0
+    dask[dataframe] == 2021.10.0
```
gsheni:
```suggestion
    dask[dataframe] >= 2021.10.0, <2022.08.0
```

sbadithe: thanks!

setup.cfg
```diff
@@ -45,7 +45,7 @@ install_requires =
     click >= 7.0.0
     cloudpickle >= 1.5.0
     dask[dataframe] >= 2021.10.0, <2022.8.0
-    distributed >= 2021.10.0
+    distributed >= 2021.10.0, <2022.8.0
```
gsheni, line 48: I added this.

featuretools/feature_base/features_deserializer.py
```diff
@@ -140,6 +140,8 @@ def _deserialize_feature(self, feature_name):
 
         args = feature_dict["arguments"]
         feature = cls.from_dictionary(args, self.entityset, dependencies, primitive)
+        if "series_library" in feature_dict:
+            feature.primitive.series_library = feature_dict["series_library"]
```
dvreed77, line 144: This works, but wondering if this is the best place to set series library on the primitivie

sbadithe: As in, within `_deserialize_feature`?

sbadithe:
Not sure what happened here...but I didn't intend to do that

Issue #2213, opened 2022-08-01
refactor re-used Rolling primitive logic into separate util function


ozzieD:
After closing and discussing https://github.com/alteryx/featuretools/issues/2205
We determined the best idea was to strip out the common logic for rolling primitives into a separate util function. This would clean up much of the duplicated code and make it a little easier for us and users to add rolling primitives, especially if we apply type hints to the function.

```
def apply_rolling_agg_to_series(
    series: Series, agg_func: Callable[Series], window_size: Union[int, str], gap: Union[int, str]=0, min_periods: int=1, ignore_window_nans: bool =False
):
    rolled_series = roll_series_with_gap(series, window_size, gap, min_periods)
    if isinstance(gap, str):
        additional_args = (gap, agg_func, min_periods)
        return rolled_series.apply(
            apply_roll_with_offset_gap,
            args=additional_args,
        ).values
    applied_rolled_series = rolled_series.apply(agg_func)

    # The shift made to account for gap adds NaNs to the rolled series
    # Those values get counted towards min_periods when they shouldn't.
    # So we need to replace any of those partial values with NaNs
    if ignore_window_nans:
        if not min_periods:
            # when min periods is 0 or None it's treated the same as if it's 1
            num_nans = gap
        else:
            num_nans = min_periods - 1 + gap
        rolled_series.iloc[range(num_nans)] = np.nan
    return applied_rolled_series.values
```

PR #2218, opened 2022-08-03
Extract rolling primitives logic into a general function

ozzieD:
resolves: https://github.com/alteryx/featuretools/issues/2213


featuretools/tests/primitive_tests/test_rolling_primitive_utils.py
```diff
@@ -475,4 +585,78 @@ def test_roll_series_with_non_offset_string_inputs(rolling_series_pd):
...
+            min_periods=1,
+        ).max()
+
+
+@pytest.mark.parametrize(
+    "primitive",
+    [RollingCount, RollingMax, RollingMin, RollingMean, RollingSTD, RollingTrend],
+)
+@patch("featuretools.primitives.rolling_primitive_utils.apply_roll_with_offset_gap")
+def test_no_call_to_apply_roll_with_offset_gap_with_numeric(
```
ozzieD, line 601: this test was moved to the utils test since it is testing the utils function more than the actual primitives

featuretools/tests/primitive_tests/test_rolling_primitive_utils.py
```diff
@@ -442,31 +542,41 @@ def count_wrapper(sub_s):
...
 
     error = "Cannot roll series. The specified window length, test, is not a valid offset alias."
     with pytest.raises(ValueError, match=error):
-        roll_series_with_gap(rolling_series_pd, window_size="test", gap="7D")
+        roll_series_with_gap(
+            rolling_series_pd,
+            window_size="test",
+            gap="7D",
+            min_periods=1,
+        )
```
ozzieD, line 579: warning lots of black, pre-commit linting changes

featuretools/primitives/standard/rolling_transform_primitive.py
```diff
@@ -618,34 +589,17 @@ def __init__(self, window_length=3, gap=0, min_periods=0):
 
     def get_function(self):
         def rolling_count(datetime):
+            def _pandas_count(series):
+                return series.count()
```
ozzieD, line 593: I followed what was already in RollingStd but I feeling like we could simplify this by just passing in the lamda function in `apply_rolling_agg_to_series` instead.

featuretools/primitives/rolling_primitive_utils.py
```diff
@@ -153,24 +140,104 @@ def _check_window_size(window_size):
...
+    agg_func: Callable[[Series], float],
+    window_size: Union[int, str],
+    gap: Union[int, str] = 0,
+    min_periods: int = 1,
+    ignore_window_nans: bool = False,
+) -> np.ndarray:
+    """Applies a given aggregation function to a rolled series.
+
+    Args:
+        series (Series): The series over which rolling windows will be created. Must be numeric in nature
```
sbadithe, line 177: Nit but is it possible to specify what "numeric in nature" means (like with explicit types)?

featuretools/primitives/rolling_primitive_utils.py
```diff
@@ -153,24 +140,104 @@ def _check_window_size(window_size):
...
+        numpy.ndarray: The array of rolling calculated values.
+
+    Note:
+        Certain operations, like `pandas.core.window.rolling.Rolling.count` that can be performed
+        on the Rolling object returned here may treat NaNs as periods to include in window calculations.
+        So a window [NaN, 1, 3]  when `min_periods=3` will proceed with count, saying there are three periods
+        but only two values and would return count=2. The calculation `max` on the other hand,
+        would say that there are not three periods in that window and would return max=NaN.
+        Most rolling calculations act this way. The implication of that here is that in order to
+        achieve the gap, we insert NaNs at the beinning of the series, which would cause `count` to calculate
```
sbadithe, line 211:
```suggestion
        achieve the gap, we insert NaNs at the beginning of the series, which would cause `count` to calculate
```

featuretools/primitives/rolling_primitive_utils.py
```diff
@@ -153,24 +140,104 @@ def _check_window_size(window_size):
...
+        on windows that technically should not have the correct number of periods. Any primitive that uses this function
+        should determine whether `ignore_window_nans` should be set to `true`.
+
+    Note:
+        Only offset aliases with fixed frequencies can be used when defining gap and window_length.
+        This means that aliases such as `M` or `W` cannot be used, as they can indicate different
+        numbers of days. ('M', because different months are different numbers of days;
```
sbadithe, line 218:
```suggestion
        numbers of days. ('M', because different months have different numbers of days;
```

featuretools/primitives/rolling_primitive_utils.py
```diff
@@ -153,24 +140,104 @@ def _check_window_size(window_size):
...
+        series (Series): The series over which rolling windows will be created. Must be numeric in nature
+            and have a DatetimeIndex.
+        agg_func (callable[Series -> float]): The aggregation function to apply to a rolled series.
+        window_size (int, string): Specifies the amount of data included in each window.
+            If an integer is provided, will correspond to a number of rows. For data with a uniform sampling frequency,
+            for example of one day, the window_length will correspond to a period of time, in this case,
+            7 days for a window_length of 7.
+            If a string is provided, it must be one of pandas' offset alias strings ('1D', '1H', etc),
+            and it will indicate a length of time that each window should span.
+            The list of available offset aliases, can be found at
```
sbadithe, line 186:
```suggestion
            The list of available offset aliases can be found at
```

featuretools/primitives/rolling_primitive_utils.py
```diff
@@ -153,24 +140,104 @@ def _check_window_size(window_size):
...
+            https://pandas.pydata.org/pandas-docs/stable/user_guide/timeseries.html#offset-aliases
+        gap (int, string, optional): Specifies a gap backwards from each instance before the
+            window of usable data begins. If an integer is provided, will correspond to a number of rows.
```
sbadithe, line 189:
```suggestion
            window of usable data begins. If an integer is provided, will correspond to the number of rows.
```

featuretools/primitives/rolling_primitive_utils.py
```diff
@@ -153,24 +140,104 @@ def _check_window_size(window_size):
...
+            If a string is provided, it must be one of pandas' offset alias strings ('1D', '1H', etc),
+            and it will indicate a length of time between a target instance and the beginning of its window.
```
sbadithe, line 191:
```suggestion
            and it will indicate the length of time between a target instance and the beginning of its window.
```

featuretools/primitives/rolling_primitive_utils.py
```diff
@@ -153,24 +140,104 @@ def _check_window_size(window_size):
...
+
+    Returns:
+        numpy.ndarray: The array of rolling calculated values.
+
+    Note:
+        Certain operations, like `pandas.core.window.rolling.Rolling.count` that can be performed
+        on the Rolling object returned here may treat NaNs as periods to include in window calculations.
+        So a window [NaN, 1, 3]  when `min_periods=3` will proceed with count, saying there are three periods
+        but only two values and would return count=2. The calculation `max` on the other hand,
+        would say that there are not three periods in that window and would return max=NaN.
```
sbadithe, line 209:
```suggestion
        would not recognize NaN as a valid period, and would therefore return `max=NaN` as the window has less valid periods (two, in this case) than `min_periods` (three, in this case).
```

sbadithe (commented):
just minor grammar suggestions

featuretools/primitives/rolling_primitive_utils.py
ozzieD: Yeah this is carried over from the other doc strings. I'm not a fan of the wording either. I could say: "The series must have a DatetimeIndex and a numeric column."

featuretools/primitives/rolling_primitive_utils.py
ozzieD: First one actually makes more sense in the context of how gap works.

featuretools/primitives/rolling_primitive_utils.py
ozzieD: I like this much better

featuretools/primitives/rolling_primitive_utils.py
```diff
@@ -136,14 +123,14 @@ def apply_roll_with_offset_gap(window, gap_offset, reducer_fn, min_periods):
     return reducer_fn(window)
 
 
-def _check_window_size(window_size):
+def _check_window_size(window_size: Union[int, str]) -> None:
```
tamargrey, line 126: I'm wondering if were's a custom type we could create for this `Union[int, str])` (which shows up 7 times in this PR) to encompass the idea of it being either an offset string or number of rows.

featuretools/tests/primitive_tests/test_rolling_primitive_utils.py
tamargrey: Do you know if there were any non linting changes to the actual test checks? I haven't seen any, but just want to make sure I'm not missing anything.

featuretools/primitives/rolling_primitive_utils.py
```diff
@@ -153,24 +140,106 @@ def _check_window_size(window_size):
...
+    agg_func: Callable[[Series], float],
+    window_size: Union[int, str],
+    gap: Union[int, str] = 0,
+    min_periods: int = 1,
+    ignore_window_nans: bool = False,
+) -> np.ndarray:
+    """Applies a given aggregation function to a rolled series.
+
+    Args:
+        series (Series): The series over which rolling windows will be created. The series must have a numeric column
```
tamargrey, line 177: Idk that a Series can be described as containing a numeric column since it kind of is just a column? Maybe something like  "The series must have numeric values and a DatetimeIndex." Feel free to ignore this if it's overly pedantic.

featuretools/primitives/rolling_primitive_utils.py
```diff
@@ -153,24 +140,106 @@ def _check_window_size(window_size):
...
+            and have a DatetimeIndex.
+        agg_func (callable[Series -> float]): The aggregation function to apply to a rolled series.
+        window_size (int, string): Specifies the amount of data included in each window.
```
tamargrey, line 180: This is completely my fault, because I had to have been the one who originally did this, but I just noticed that `window_size` and `window_length` are kind of used interchangeably. Would we be able to use just one (I think I'm partial to window_length)?

tamargrey (commented):
Thank you for doing this! Makes the logic around rolling primitives feel a lot more manageable

featuretools/primitives/rolling_primitive_utils.py
ozzieD: Personally, I like to avoid custom typing unless the type gets pretty crazy. Because python is dynamically typed having a simple union type like this is fine. Also once we eventually introduce mypy or pyright into the repo, I think having a simpler type will be much easier to deal with. I'm open to other opinions though.

tamargrey: That makes sense to me!

featuretools/primitives/rolling_primitive_utils.py
ozzieD: Makes sense to me I'll change it.

featuretools/primitives/rolling_primitive_utils.py
ozzieD: wow I didn't even notice reading it hahaha. I'm fine with either. I'll change it to window_length.

featuretools/tests/primitive_tests/test_rolling_primitive_utils.py
ozzieD: Only change is specifying the min_periods parameter for `roll_series_with_gap`. I rolled the default parameters up to `apply_rolling_agg_to_series` so existing functions like `roll_series_with_gap` no longer need a default value.

gsheni (approved):

PR #2230, opened 2022-08-07
Improve version comparison in check_schema_version util function 

### LINKED ISSUE #2154: Improve version comparison in check_schema_version util function
In the `check_schema_version` utility function, there is custom code to determine whether saved schema versions are older or newer than the current schema version. This comparison could likely be simplified significantly by using the packaging library performing the version comparison instead of the custom code.

Current code:
```
        current = SCHEMA_VERSION.split(".")
        saved = version_string.split(".")

        warning_text_upgrade = (
            "The schema version of the saved %s"
            "(%s) is greater than the latest supported (%s). "
            "You may need to upgrade featuretools. Attempting to load %s ..."
            % (cls_type, version_string, SCHEMA_VERSION, cls_type)
        )
        for c_num, s_num in zip_longest(current, saved, fillvalue=0):
            if c_num > s_num:
                break
            elif c_num < s_num:
                warnings.warn(warning_text_upgrade)
                break

        warning_text_outdated = (
            "The schema version of the saved %s"
            "(%s) is no longer supported by this version "
            "of featuretools. Attempting to load %s ..."
            % (cls_type, version_string, cls_type)
        )
        # Check if saved has older major version.
        if current[0] > saved[0]:
            logger.warning(warning_text_outdated)
```

If we use packaging as described here (https://stackoverflow.com/a/11887885) we could simplify the comparisons to something similar to `version.parse(saved) < version.parse(current)` and avoid splitting the version into components and looping over the version lists.

sbadithe:
- Fixes #2154


featuretools/utils/gen_utils.py
```diff
@@ -4,8 +4,8 @@
 import sys
 import warnings
 from enum import Enum
-from itertools import zip_longest
 
+from packaging.version import parse
```
gsheni, line 8: `packaging` needs to be added to setup.cfg in `install_requires`

gsheni (changes requested):
once `setup.cfg` is updated, I'll approve

setup.cfg
```diff
@@ -49,6 +49,7 @@ install_requires =
     holidays >= 0.13
     numpy >= 1.21.0
     pandas >= 1.3.0
+    packaging >= 20.0
```
sbadithe, line 52: It looks like `distributed` has `packaging >= 20.0` so I put this as the minimal version. It's also the release of `packaging` that supports Python3.8

gsheni (approved):

Issue #1781, opened 2021-11-12
Add Lag primitive


gsheni:
- Add Lag primitive, which can be helpful for time-series problems
```python
from featuretools.primitives.base import TransformPrimitive
from featuretools.variable_types import Variable


class LagByPeriods(TransformPrimitive):
    """Shifts an array of values by a specified number of periods.

    Args:
        periods (int): The number of periods by which to shift the input.
            Default is 1.

        fill_value (int, float, string, bool): The value to use to fill in
            the gaps left after shifting the input. Default is None.

    Examples:
        >>> lag = Lag()
        >>> lag([1, 2, 3, 4, 5]).tolist()
        [nan, 1.0, 2.0, 3.0, 4.0]

        You can specify the number of periods to shift the values

        >>> lag_periods = Lag(periods=3)
        >>> lag_periods([1, 2, 3, 4, 5]).tolist()
        [nan, nan, nan, 1.0, 2.0]

        You can specify the fill value to use

        >>> lag_fill_value = Lag(fill_value=100)
        >>> lag_fill_value([1, 2, 3, 4]).tolist()
        [100, 1, 2, 3]
    """
    name = "lag"
    input_types = [Variable]

    def __init__(self, periods=1, fill_value=None):
        self.periods = periods
        self.fill_value = fill_value

    def get_function(self):
        def lag(x):
            return x.shift(periods=self.periods, fill_value=self.fill_value)
        return lag
```

gsheni:
Unit Tests
```python
import numpy as np
import pandas as pd

from .utils import BaseTestTransform, find_applicable_primitives, valid_dfs

from premium_primitives.lag import Lag


class TestLag(BaseTestTransform):
    primitive = Lag

    # test datetimes

    def test_regular(self):
        primitive_instance = self.primitive()
        primitive_func = primitive_instance.get_function()
        array = pd.Series([1, 2, 3, 4])
        answer = primitive_func(array)
        correct_answer = pd.Series([np.nan, 1, 2, 3])
        pd.testing.assert_series_equal(answer, correct_answer)

    def test_period(self):
        primitive_instance = self.primitive(periods=3)
        primitive_func = primitive_instance.get_function()
        array = pd.Series([1, 2, 3, 4])
        answer = primitive_func(array)
        correct_answer = pd.Series([np.nan, np.nan, np.nan, 1])
        pd.testing.assert_series_equal(answer, correct_answer)

    def test_negative_period(self):
[...]

PR #2252, opened 2022-08-22
Add generic Lag primitive

ozzieD:
### Pull Request Description
Resolves https://github.com/alteryx/featuretools/issues/1781


featuretools/primitives/standard/transform_primitive.py
```diff
@@ -584,49 +587,75 @@ def email_address_to_domain(emails):
         return email_address_to_domain
 
 
-class NumericLag(TransformPrimitive):
+class Lag(TransformPrimitive):
```
gsheni, line 590: What if a user is already be using the `NumericLag` primitive?

ozzieD: Good question, I actually wanted to bring this up. Do we have users using featuretools for time series currently? If not I say just go ahead and replace it but if not we can just revert this and add a DeprecationWarning and remove it for V2. Thoughts?

gsheni: We could add a DeprecationWarning  if the user uses `NumericLag`, but still get `Lag` primitive into Featuretools. 
OR, we could update `dfs` to support the old primitive name. 

@rwedge any suggestion here?

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -584,49 +587,75 @@ def email_address_to_domain(emails):
...
+        [
+            ColumnSchema(logical_type=Boolean),
+            ColumnSchema(semantic_tags={"time_index"}),
+        ],
+        [
+            ColumnSchema(logical_type=BooleanNullable),
+            ColumnSchema(semantic_tags={"time_index"}),
+        ],
+        [
+            ColumnSchema(logical_type=AgeNullable),
```
tamargrey, line 637: Do we want `AgeFractional` as well? (I imagine `Age` wasn't doable bc it used the non nullable `int64` dtype?)

featuretools/primitives/standard/transform_primitive.py
rwedge: I'm in favor of deprecating `NumericLag`.  I think updating `dfs` to map "numeric_lag" to the `Lag` primitive would be problematic.  `Lag` has more accepted inputs and could create more features than the user was expecting, if we magically swapped in `Lag` for them.  It also would hinder custom primitives from using "numeric_lag" for as long as we have this `dfs` shortcut, even after we remove the original primitive

ozzieD: Added a `FutureWarning` to the primitive. Similar to how pandas does their warnings.

rwedge: Can we add a test checking the warning works as expected?

ozzieD: Done

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -623,10 +631,108 @@ class NumericLag(TransformPrimitive):
...
+        ],
+        [
+            ColumnSchema(logical_type=Unknown),
+            ColumnSchema(semantic_tags={"time_index"}),
+        ],
+    ]
+    return_type = None
+    uses_full_dataframe = True
+
+    def __init__(self, periods=1, fill_value=None):
```
thehomebrewnerd, line 729: Not sure having a fill_value makes sense now. You couldn't specify a numeric fill value and have it work on boolean columns for example. There probably isn't much harm in leaving it with the default at `None`, but if users specified a value, I think they will have trouble unless the restrict the application of their primitive instance to only certain columns that are compatible with their specified fill value.

featuretools/tests/primitive_tests/test_primitive_utils.py
```diff
@@ -227,18 +227,19 @@ def test_check_input_types():
     sem_tag_type_checks = set()
     unique_input_types = set()
     expected_log_in_check = {
-        "boolean_nullable",
```
thehomebrewnerd, line 230: Is there a reason for changing this test?

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -623,10 +631,108 @@ class NumericLag(TransformPrimitive):
...
+        self.periods = periods
+        self.fill_value = fill_value
+
+    def get_function(self):
+        def lag(input_col, time_index):
```
thehomebrewnerd, line 734: I don't think this matters much, but we have tended to pass the time index as the first argument for primitives that use it along with another data column

featuretools/tests/primitive_tests/test_transform_primitive.py
```diff
@@ -747,79 +748,124 @@ def test_trans_primitives_can_init_without_params():
...
+
+    correct_answer = pd.Series([np.nan, "hello", "world", "foo"])
+    pd.testing.assert_series_equal(answer, correct_answer)
+
+
+def test_lag_with_bools():
+    primitive_instance = Lag()
+    primitive_func = primitive_instance.get_function()
+
+    array = pd.Series([True, False, True, False])
```
thehomebrewnerd, line 852: I think we should explicitly set the series dtype and test with both `bool` and `boolean` here.

thehomebrewnerd, line 852: I think we should explicitly set the series dtype and test with both `bool` and `boolean` here.

featuretools/tests/primitive_tests/test_transform_primitive.py
```diff
@@ -747,79 +748,124 @@ def test_trans_primitives_can_init_without_params():
...
+
+    correct_answer = pd.Series([np.nan, True, False, True])
+    pd.testing.assert_series_equal(answer, correct_answer)
+
+
+def test_lag_with_floats():
+    primitive_instance = Lag()
+    primitive_func = primitive_instance.get_function()
+
+    array = pd.Series([1.23, 2.45, 3.56, 4.98])
```
thehomebrewnerd, line 865: Let's set the dtype to be explicit about what we are testing.

featuretools/tests/primitive_tests/test_transform_primitive.py
```diff
@@ -747,79 +748,124 @@ def test_trans_primitives_can_init_without_params():
...
 
     correct_answer = pd.Series([np.nan, 1, 2, 3])
     pd.testing.assert_series_equal(answer, correct_answer)
+
+
+def test_lag_with_strings():
+    primitive_instance = Lag()
+    primitive_func = primitive_instance.get_function()
+
+    array = pd.Series(["hello", "world", "foo", "bar"])
```
thehomebrewnerd, line 839: I'm thinking this test might go away based on previous comments, but let's explicitly set the dtype to `string` which is what we would expect from a Woodwork series.

featuretools/tests/primitive_tests/test_transform_primitive.py
```diff
@@ -747,79 +748,124 @@ def test_trans_primitives_can_init_without_params():
...
+    primitive_instance = Lag()
+    primitive_func = primitive_instance.get_function()
+
+    array = pd.Series([1.23, 2.45, 3.56, 4.98])
+    time_array = pd.Series(pd.date_range(start="2020-01-01", periods=4, freq="D"))
+
+    answer = pd.Series(primitive_func(array, time_array))
+
+    correct_answer = pd.Series([np.nan, 1.23, 2.45, 3.56])
+    pd.testing.assert_series_equal(answer, correct_answer)
```
thehomebrewnerd, line 871: After this can we add a test with a `categorical` dtype input to make sure that work as we expect it to?

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -1570,22 +1578,77 @@ def test_cfm_with_lag_and_non_nullable_column(pd_es):
...
         fm["NUMERIC_LAG(datetime, products.rating, periods=5)"]
         .head(periods)
         .isnull()
         .all()
     )
 
-    assert "NUMERIC_LAG(datetime, products.rating, periods=5)" in fm.columns
+
+def test_cfm_with_lag(pd_es):
+    # fill nans so we can use non nullable numeric logical type in the EntitySet
```
thehomebrewnerd, line 1596: I think we might want to consider using a different dataframe here that contains all the major logical types we are concerned with: `Double`, `Integer`, `IntegerNullable`, `Boolean`, `BooleanNullable` and `Categorical` just to make sure we have everything covered and we get the expected column types in the feature matrix.

thehomebrewnerd (commented):
Happy to discuss any of my comments in more detail if that would be helpful.

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -623,10 +631,108 @@ class NumericLag(TransformPrimitive):
...
+
+        You can specify the fill value to use
+
+        >>> lag_fill_value = Lag(fill_value=100)
+        >>> lag_fill_value([1, 2, 3, 4], pd.Series(pd.date_range(start="2020-01-01", periods=4, freq='D'))).tolist()
+        [100, 1, 2, 3]
+    """
+
+    name = "lag"
+    input_types = [
```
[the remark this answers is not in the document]
tamargrey: @ozzieD I forgot about return types when suggesting you expand the specified logical types. I think nate's probably right that we can keep this simple.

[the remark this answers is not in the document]
ozzieD: @thehomebrewnerd So I confirmed Integer and Boolean inputs do work because `init_ww_and_concat_fm` will replace the output type with the nullable counterparts. @rwedge mentioned this would be a problem with stacked primitives.

featuretools/primitives/standard/transform_primitive.py
ozzieD: Yeah I'm just never a fan of restricting functionality. I could put guard rails in by throwing TypeErrors if fill_value type doesn't match the input type. Thoughts?

featuretools/primitives/standard/transform_primitive.py
ozzieD: This was done because of the way `return_type=None` works. It grabs the first input type so it was assuming `time_index` for the output if I did it that way.

thehomebrewnerd: Got it. Didn't realize that is how the return type was getting set, but that makes sense.

featuretools/primitives/standard/transform_primitive.py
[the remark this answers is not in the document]
thehomebrewnerd: @rwedge Can you provide a little more detail on the stacking problem and is it something we need to be concerned with in this PR?

featuretools/primitives/standard/transform_primitive.py
thehomebrewnerd: I don't really like the idea of complicating the primitive calculation with a bunch of type checking. If we are going to leave this in, maybe we just update the docstring to inform the users that errors will result if they attempt to use a fill value that is incompatible with the data types of the columns to which the primitive is applied?

Not completely sure, but I think the will end up getting a Woodwork initialization error in the event of incompatible data types. That error might not be super-helpful in troubleshooting the problem though.

featuretools/primitives/standard/transform_primitive.py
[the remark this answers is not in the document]
rwedge: @thehomebrewnerd with `return_type = None` features created by this primitive will use the first input's column schema as their return type.  When returning the final feature matrix, `init_ww_and_concat_fm` will convert some columns with non-nullable return types to nullable types if there are nulls present.  However, this does not happen inside `FeatureSetCalculator` after values are calculated, so if a primitive unable to handle nulls stacked on this primitive it could receive null values as input.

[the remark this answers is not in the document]
thehomebrewnerd: @rwedge I see. Is that enough of an edge case for now that we can wait to address, or do you think that is something we should resolve now?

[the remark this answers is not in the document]
rwedge: @thehomebrewnerd a sanity check would be to see if there are any agg primitives in our primitive libraries that would encounter this scenario

featuretools/primitives/standard/transform_primitive.py
ozzieD: After discussing we are removing `fill_value`. Ideas to reintroduce are captured here: https://github.com/alteryx/featuretools/issues/2266

featuretools/tests/primitive_tests/test_primitive_utils.py
ozzieD: Just listed in types in alphabetical order. I was working on this file when replacing Lag with NumericLag but I can just revert these changes to make this MR less confusing.

featuretools/tests/primitive_tests/test_transform_primitive.py
ozzieD: Don't think there is a difference between the to both lead to `Series.dtype("bool")`

featuretools/tests/primitive_tests/test_transform_primitive.py
ozzieD: done

featuretools/tests/primitive_tests/test_transform_primitive.py
ozzieD: done

featuretools/tests/primitive_tests/test_transform_primitive.py
ozzieD: Series.astype("categorical") errors

featuretools/primitives/standard/transform_primitive.py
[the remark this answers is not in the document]
ozzieD: Changed inputs to 
- ColumnSchema(semantic_tags={"category"})
- ColumnSchema(semantic_tags={"numeric"})
- ColumnSchema(logical_type=BooleanNullable)
- ColumnSchema(logical_type=Boolean)

Tried using a bunch of aggs with Lag and haven't seen any issues

featuretools/tests/primitive_tests/test_transform_primitive.py
thehomebrewnerd: Sorry, I meant to say `category` here. I got the WW logical type and the pandas dtype mixed up.

featuretools/tests/primitive_tests/test_transform_primitive.py
thehomebrewnerd: I don't think there is a big difference, if any, but I was mostly thinking of making sure our tests cover all the possible series dtypes that the primitive function could see.

featuretools/tests/primitive_tests/test_transform_primitive.py
```diff
@@ -837,6 +824,7 @@ def test_lag_with_strings():
     primitive_func = primitive_instance.get_function()
 
     array = pd.Series(["hello", "world", "foo", "bar"])
+    array.astype("string")
```
thehomebrewnerd, line 827: This isn't actually changing the type of `array` as you are essentially throwing away the return value from the method call(this is not an in place operation).

You can update the line above to be 
```
array = pd.Series(["hello", "world", "foo", "bar"], dtype="string")
```

featuretools/tests/primitive_tests/test_transform_primitive.py
```diff
@@ -845,11 +833,26 @@ def test_lag_with_strings():
     pd.testing.assert_series_equal(answer, correct_answer)
 
 
+def test_lag_with_categories():
+    primitive_instance = Lag()
+    primitive_func = primitive_instance.get_function()
+
+    array = pd.Series(["cow", "cow", "pig", "pig"])
+    array.astype("category")
```
thehomebrewnerd, line 841: Same as other comment

featuretools/tests/primitive_tests/test_transform_features.py
ozzieD: rewrote this test to add the above types to the dataframe and check for their existence

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -1570,26 +1578,110 @@ def test_cfm_with_lag_and_non_nullable_column(pd_es):
...
+        target_dataframe_name="new_log",
+        entityset=pd_es,
+        agg_primitives=[],
+        trans_primitives=[lag_primitive],
+        cutoff_time=cutoff_times,
+    )
+    # Integer
+    assert fm["LAG(value, datetime, periods=5)"].head(periods).isnull().all()
+    assert fm["LAG(value, datetime, periods=5)"].isnull().sum() == periods
     assert (
```
thehomebrewnerd, line 1634: This isn't a big deal or a blocker in my opinion, but instead of checking the type string, you could also check that the logical type is an instance of the expected logical type class like is done on L1619.

ozzieD: yup just was lazy about importing :P. 

Switched!

thehomebrewnerd (approved):
Looks good, assuming CI tests pass.

PR #2254, opened 2022-08-22
2253 holidays library update causes lookup failures

cp2boston:
### Pull Request Description

- Updates to requirements files for holidays==0.15
- Update HolidayUtil to convert old countries to new country and subdivision
- Update to `def test_holiday_out_of_range()` to use correct number of days

-----

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,13 +7,14 @@ Future Release
 ==============
     * Enhancements
     * Fixes
+        * Fix changes resulting from python-holidays updates (:pr:`2251`)
     * Changes
         * Update release notes to make clear conda release portion (:pr:`2249`)
     * Documentation Changes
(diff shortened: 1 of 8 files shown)


rwedge (commented):
are we planning to drop support for older versions of the holidays package?  If so we should update the requirements in setup.cfg

docs/source/release_notes.rst
```diff
@@ -7,13 +7,14 @@ Future Release
 ==============
     * Enhancements
     * Fixes
+        * Fix changes resulting from python-holidays updates (:pr:`2254`)
```
rwedge, line 10:
```suggestion
        * Fix compatibility issues with holidays 0.15 (:pr:`2254`)
```

rwedge (approved):
left a suggestion on the release note wording but otherwise LGTM

PR #2291, opened 2022-09-19
Updates for pandas 1.5.0 compatibility

thehomebrewnerd:
Various changes needed for compatibility with pandas 1.5.0.


featuretools/primitives/standard/transform_primitive.py
```diff
@@ -653,8 +653,8 @@ class Lag(TransformPrimitive):
         You can specify the number of periods to shift the values
 
         >>> lag_periods = Lag(periods=3)
-        >>> lag_periods(["hello", "world", "test", "foo", "bar"], pd.Series(pd.date_range(start="2020-01-01", periods=5, freq='D'))).tolist()
-        [nan, nan, nan, 'hello', 'world']
+        >>> lag_periods([True, False, False, True, True], pd.Series(pd.date_range(start="2020-01-01", periods=5, freq='D'))).tolist()
+        [nan, nan, nan, True, False]
     """
```
ozzieD, line 658: was something wrong with the strings?

ozzieD (approved):

featuretools/primitives/standard/transform_primitive.py
thehomebrewnerd: Not wrong exactly, but complicated - using strings this way with the new version of pandas introduced `None` instead of `nan`, which would mean this docstring test would pass with old versions of pandas but not version `1.5.0`.

Also based on the specified input types we wouldn't really be calling this function with a `object` type input like we were doing with the string input. So, in order to allow this test to pass on old and new versions of pandas, I switched to an input that is in our supported input types and gives consistent output between pandas versions.

thehomebrewnerd:
From what I can tell, the change in coverage is due a change in an error raised by pandas. What was previously an `OverflowError` is now a `ValueError`. We have logic in place for handling both, but because coverage is running on the new version of pandas, the logic branch for handling the `OverflowError` doesn't get hit now, but it would with older versions of pandas.

featuretools/primitives/standard/transform_primitive.py
```diff
@@ -653,8 +653,8 @@ class Lag(TransformPrimitive):
         You can specify the number of periods to shift the values
 
         >>> lag_periods = Lag(periods=3)
-        >>> lag_periods(["hello", "world", "test", "foo", "bar"], pd.Series(pd.date_range(start="2020-01-01", periods=5, freq='D'))).tolist()
-        [nan, nan, nan, 'hello', 'world']
+        >>> lag_periods([True, False, False, True, True], pd.Series(pd.date_range(start="2020-01-01", periods=5, freq='D'))).tolist()
```
gsheni, line 656: Can we have a string lag and a boolean lag in the example?

thehomebrewnerd: We could, but I'm not sure it's so simple: https://github.com/alteryx/featuretools/pull/2291#discussion_r975598983

If we added a string lag, the doctest wouldn't pass for all of the pandas version we currently support because of the `None` vs `nan` output difference. Unless we added some extra conversion to the output to make sure we always get the same missing value representation.

thehomebrewnerd: Just want to point out, that once Woodwork gets initialized in the feature matrix we will always end up with the same missing value representation, but since that doesn't happen at the primitive function level, this difference in output will be present when calling it like we do in the doctest examples.

thehomebrewnerd: Added comment: https://github.com/alteryx/featuretools/pull/2291/commits/b8e5cfe04986468df1b984613588a25d2a1019a6

gsheni (approved):

PR #2301, opened 2022-09-22
Update allowed types for instance_vals in query_by_values

### LINKED ISSUE #1272: Update allowed types for instance_vals in query_by_values
Currently the `query_by_values` method lists four allowable input types for the `instance_vals` parameter: `pd.Dataframe, pd.Series, list[str] or str`

The current Featuretools code will only call this method with a `pd.Series` input during normal operation, although there are several unit tests that call the method with a list used to specify `instance_vals`. There is code in the `_vals_to_series` method that handles the conversion of these different types. As there is no current need in Featuretools for this code to handle `pd.DataFrame` or `str` inputs, these types could be removed. 

Optionally, support for list input could be dropped if the corresponding unit tests are updated, which would leave only `pd.Series` as a valid input type for `instance_vals`.

Making these changed would simplify the code base and test suite. Any users that might be calling `query_by_values` directly with one of the dropped types, would simply have to convert their values to a pd.Series before the call, but otherwise the functionality and output would remain unchanged.

Kryvonis:
- Leave only `pd.Series` as a valid input type for `instance_vals` in `Entityset`.
- Fixes #1272

### DIFF
--- a/featuretools/entityset/entityset.py
+++ b/featuretools/entityset/entityset.py
@@ -1600,7 +1600,7 @@ class EntitySet(object):
 
         Args:
             dataframe_name (str): The id of the dataframe to query
-            instance_vals (pd.Dataframe, pd.Series, list[str] or str) :
+            instance_vals (pd.Series) :
                 Instance(s) to match.
             column_name (str) : Column to query on. If None, query on index.
             columns (list[str]) : Columns to return. Return all columns if None.
@@ -1618,8 +1618,6 @@ class EntitySet(object):
         dataframe = self[dataframe_name]
         if not column_name:
             column_name = dataframe.ww.index
-
-        instance_vals = _vals_to_series(instance_vals, column_name)
 
         training_window = _check_timedelta(training_window)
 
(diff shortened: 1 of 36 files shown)


gsheni:
@Kryvonis thank you for putting up this pull request. I'm curious, why would it be beneficial to support less types?

thehomebrewnerd:
> @Kryvonis thank you for putting up this pull request. I'm curious, why would it be beneficial to support less types?

@gsheni See this issue for context: https://github.com/alteryx/featuretools/issues/1272

Kryvonis:
Sorry, my fault. I didn't know about that rule. I tried to amend email but Git add new commits. Do I need to remove previous commits?

gsheni:
@Kryvonis You can also create a new branch off Featuretools `main`, copy the changes from here, and put up a new pull request in a new branch. Be sure to use the same author email on the commits as your GitHub account.

Kryvonis:
I think I already fixed author commits but I don't know is there ok that some tests are failed

sbadithe:
Hi Kryvonis, thanks for doing this! 

If I am understanding this correctly, I believe one of the reasons the tests are failing is that the `query_by_values` method is being called with Dask or PySpark series arguments for the `instance_vals` parameter. Also, I believe the lines 1874 and 1877 in `_vals_to_series` are doing necessary logic. One way to handle this would be to keep the `_vals_to_series` method. We could remove the single value and dataframe cases and keep the other ones.

@thehomebrewnerd @rwedge

featuretools/entityset/entityset.py
```diff
@@ -1619,8 +1619,6 @@ def query_by_values(
         if not column_name:
             column_name = dataframe.ww.index
 
-        instance_vals = _vals_to_series(instance_vals, column_name)
```
sbadithe, line 1622: I suspect we need this function call still. I think within the call, we can remove the check for `pd.dataframe`.

featuretools/entityset/entityset.py
```diff
@@ -1850,35 +1848,6 @@ def replace(x, is_spark=False):
...
-        return None
-
-    # If this is a single value, make it a list
-    if not hasattr(instance_vals, "__iter__"):
-        instance_vals = [instance_vals]
-
-    # convert iterable to pd.Series
-    if isinstance(instance_vals, pd.DataFrame):
-        out_vals = instance_vals[column_id]
-    elif is_instance(instance_vals, (pd, dd, ps), "Series"):
```
sbadithe, line 1868: I think we need this function, in particular this check, along with L1858

PR #2331, opened 2022-10-21
Restructure primitives directory to use individual primitives files

### LINKED ISSUE #2179: Restructure primitives folder and various util functions

thehomebrewnerd:
### Restructure primitives directory to use individual primitives files

Closes #2179


sbadithe (commented):
Is it possible that we could run into issues with file handling and/or imports if the filename shares its name with a Python built-in? i.e. `max.py` and `max`.

thehomebrewnerd:
> Is it possible that we could run into issues with file handling and/or imports if the filename shares its name with a Python built-in? i.e. `max.py` and `max`.

That was a problem with `and` and `or`, and that's why I appended `_primitive` to the filenames, but I don't think there is an issue with `max`...if there is, I didn't run into it.

featuretools/primitives/standard/aggregation_primitives/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+from featuretools.primitives.standard.aggregation_primitives.skew import Skew
+from featuretools.primitives.standard.aggregation_primitives.std import Std
+from featuretools.primitives.standard.aggregation_primitives.sum import Sum
+from featuretools.primitives.standard.aggregation_primitives.time_since_first import (
+    TimeSinceFirst,
+)
+from featuretools.primitives.standard.aggregation_primitives.time_since_last import (
+    TimeSinceLast,
+)
+from featuretools.primitives.standard.aggregation_primitives.trend import Trend
```
ozzieD, line 57: These can all be relative imports

thehomebrewnerd: We have a pre-commit hook that converts them all to absolute.

featuretools/primitives/standard/aggregation_primitives/__init__.py
```diff
@@ -0,0 +1,57 @@
+# flake8: noqa
```
ozzieD, line 1: I'm pretty sure this line can be deleted in every __init__.py file. I added an ignore to these files in `.flake8`

featuretools/primitives/base/__init__.py
```diff
@@ -1,2 +1,4 @@
 # flake8: noqa
-from featuretools.primitives.base.api import *
+from featuretools.primitives.base.aggregation_primitive_base import AggregationPrimitive
+from featuretools.primitives.base.primitive_base import PrimitiveBase
+from featuretools.primitives.base.transform_primitive_base import TransformPrimitive
```
ozzieD, line 4: same here with relative imports

featuretools/primitives/standard/aggregation_primitives/__init__.py
ozzieD: shucks, no worries.

sbadithe:
> > Is it possible that we could run into issues with file handling and/or imports if the filename shares its name with a Python built-in? i.e. `max.py` and `max`.
> 
> That was a problem with `and` and `or`, and that's why I appended `_primitive` to the filenames, but I don't think there is an issue with `max`...if there is, I didn't run into it.

I think it might be good to treat `all`, `any`, `max`, `min`, and `sum` the same way to prevent any potential naming issues in the future.

ozzieD:
Since we have better namespacing we can rename rolling_primitives_utils.py to utils.py

featuretools/primitives/standard/aggregation_primitives/__init__.py
thehomebrewnerd: We can't delete it in the files that use the `from ... import *` approach or we get this linting error:

```
featuretools/primitives/standard/__init__.py:4:1: F403 'from featuretools.primitives.standard.transform_primitives import *' used; unable to detect undefined names
```

I was able to remove it from a couple other places though.

thehomebrewnerd:
> Since we have better namespacing we can rename rolling_primitives_utils.py to utils.py

Done!

thehomebrewnerd:
> I think it might be good to treat `all`, `any`, `max`, `min`, and `sum` the same way to prevent any potential naming issues in the future.

@sbadithe I renamed all the associated files to include the `_primitive` suffix.

thehomebrewnerd:
> Thoughts on shortening the primitive submodules a bit?
> 
> Going from
> 
> `featuretools.primitives.standard.transform_primitives.binary_transform_primitives`
> 
> to
> 
> `featuretools.primitives.standard.transform.binary`
> 
> Not as descriptive but transform and primitive are still in the submodule path

@rwedge Not opposed to it. Would help make some of the imports more manageable and not split over so many lines I think.

featuretools/primitives/standard/transform/natural_language/num_characters.py
```diff
@@ -0,0 +1,29 @@
+from woodwork.column_schema import ColumnSchema
```
sbadithe, line 1: I like that `num_characters` and `num_words` are moved here

featuretools/primitives/__init__.py
```diff
@@ -1,10 +1,19 @@
+# flake8: noqa
```
ozzieD, line 1: should be able to delete this

featuretools/primitives/standard/__init__.py
```diff
@@ -1,2 +1,5 @@
 # flake8: noqa
```
ozzieD, line 1: same here

featuretools/primitives/standard/transform/__init__.py
```diff
@@ -0,0 +1,31 @@
+# flake8: noqa
```
ozzieD, line 1: same here

featuretools/primitives/__init__.py
thehomebrewnerd: Can't remove it here because this line `from featuretools.primitives.standard import *` causes a lint failure

featuretools/primitives/standard/__init__.py
thehomebrewnerd: Same issue - the `import *` causes linting problems

featuretools/primitives/standard/transform/__init__.py
thehomebrewnerd: Same issue - the `import *` causes linting problems

ozzieD: got it

ozzieD:
I'm wondering if we can put all the stray transform primitives in their own folders as well. I'm okay with this as is. But maybe we could but the email and url ones in thier respective folders. Maybe the Cosine, Tangent, Percentile etc.. in one called math.

ozzieD (approved):

PR #2343, opened 2022-10-28
Add Expanding Primitives 

### LINKED ISSUE #2242: Add Expanding* primitives for time series feature engineering
Add `Expanding*` versions of the `Rolling*` primitives for time series feature engineering. 

The current rolling primitives, such as rolling max, use data contained within a window to determine the value along with a gap, which allows them to be used to create features from a target column without including recent values. We could add corresponding `Expanding*` primitives that include a gap parameter that uses all of the previous history excluding the gap rows to calculate the values. This could be implemented as new primitives or by including an optional gap parameter in the existing `Cumulative*` primitives or by updating the `Rolling*` primitives to use all history prior to the gap rows. The fact that the current `Cumulative*` primitives do not have a gap option prevents them from being used to create time series features from a target column as they would leak the target value at each time step. Similarly, since the current `Rolling*` primitives don't have an option to use all history they cannot be used for this purpose.

See technique 5 in this article for more info: https://www.analyticsvidhya.com/blog/2019/12/6-powerful-feature-engineering-techniques-time-series/

sbadithe:
- Fixes #2242


featuretools/tests/conftest.py
```diff
@@ -731,7 +731,7 @@ def divide_by_zero_es_spark(divide_by_zero_es_pd):
 
 
 @pytest.fixture
-def rolling_series_pd():
+def window_series_pd():
```
sbadithe, line 734: I think we can use this fixture for all window operations

featuretools/primitives/standard/transform/time_series/expanding/expanding_count.py
```diff
@@ -0,0 +1,91 @@
...
+    def __init__(self, gap=1, min_periods=0):
+        self.gap = gap
+        self.min_periods = min_periods
+
+    def get_function(self):
+        def expanding_count(datetime):
+            x = pd.Series(1, index=datetime)
+            if isinstance(self.gap, int):
+                x = x.shift(self.gap)
+            else:
```
sbadithe, line 84: I think the non-integer gap is responsible for much of the complexity in the rolling primitives, but I'm not sure how much benefit there is to the user.

featuretools/primitives/standard/transform/time_series/expanding/expanding_count.py
```diff
@@ -0,0 +1,91 @@
...
+        Input datetimes should be monotonic.
+
+    Args:
+        gap (int, string, optional): Specifies a gap backwards from each instance before the
+            usable data begins. If an integer is provided, it will correspond to a number of rows.
+            If a string is provided, it must be one of pandas' offset alias strings ('1D', '1H', etc),
+            and it will indicate a length of time between a target instance and the beginning of its window.
+            Defaults to 1.
+        min_periods (int, optional): Minimum number of observations required for performing calculations
+            over the window. min_periods must always be less than or equal to the length of the series. (TODO: Check? Gap?)
```
sbadithe, line 26: fixing

docs/source/api_reference.rst
```diff
@@ -219,6 +219,19 @@ Email and URL Transform Primitives
     URLToTLD
 
 
+Expanding Transform Primitives
```
thehomebrewnerd, line 222: I'm not exactly clear on the best way to organize all this, but these are primarily being added for time series problems, so it might make sense to add them to the Time Series Transform Primitive section instead of creating a new one here.

featuretools/primitives/standard/transform/time_series/__init__.py
```diff
@@ -23,3 +23,11 @@
 from featuretools.primitives.standard.transform.time_series.rolling_trend import (
     RollingTrend,
 )
+from featuretools.primitives.standard.transform.time_series.expanding import (
```
thehomebrewnerd, line 26: If we are going to create a new folder for the expanding primitives, I think we should be consistent and do the same thing with our rolling* primitives so we have a consistent organizational structure.

That could be a separate small MR.

featuretools/primitives/standard/transform/time_series/utils.py
```diff
@@ -236,3 +237,16 @@ def apply_rolling_agg_to_series(
...
+def _apply_gap_for_expanding_primitives(
+    x: Series,
+    gap: Union[int, str],
+) -> Optional[Series]:
+    if isinstance(gap, int):
+        return x.shift(gap)
+    else:
+        raise NotImplementedError(
+            "We currently do not support string offsets for the gap parameter in "
+            "Expanding primitives",
```
thehomebrewnerd, line 251: Minor nitpick, but I would change this wording just a bit:

`String offsets are not supported for the gap parameter in Expanding primitives.`

featuretools/primitives/standard/transform/time_series/utils.py
```diff
@@ -236,3 +237,16 @@ def apply_rolling_agg_to_series(
...
+
+
+def _apply_gap_for_expanding_primitives(
+    x: Series,
+    gap: Union[int, str],
+) -> Optional[Series]:
+    if isinstance(gap, int):
+        return x.shift(gap)
+    else:
+        raise NotImplementedError(
```
thehomebrewnerd, line 249: I think this should be a `TypeError` instead of a `NotImplementedError`. My understanding is that `NotImplementedErrors` should typically be used with abstract base classes: https://docs.python.org/3/library/exceptions.html#NotImplementedError

featuretools/tests/primitive_tests/test_rolling_primitive_utils.py
```diff
@@ -20,16 +20,16 @@
 from featuretools.tests.primitive_tests.utils import get_number_from_offset
 
 
-def test_get_rolled_series_without_gap(rolling_series_pd):
+def test_get_rolled_series_without_gap(window_series_pd):
```
thehomebrewnerd, line 23: Do you think it would be more appropriate for this entire file to be named `test_time_series_primitive_utils.py` now?

featuretools/primitives/standard/transform/time_series/__init__.py
sbadithe: I agree

featuretools/tests/primitive_tests/test_rolling_primitive_utils.py
sbadithe: I think the name can stay the same. We still only test `rolling` functionality in that file, to my knowledge. I renamed the fixture in `conftest.py` to reuse it for the Expanding suite (and maybe Exponential and Time Series down the line).

thehomebrewnerd: On that, I think we probably should add a targeted unit test to test the new `_apply_gap_for_expanding_primitives` function you added. This is currently tested in more complicated tests, but I think simple targeted tests for that specific functionality would be better - both that it raises an error with invalid input and that we get back the expected series with the proper input.

featuretools/primitives/standard/transform/time_series/expanding/expanding_count.py
```diff
@@ -0,0 +1,68 @@
...
+        >>> expanding_count(times).tolist()
+        [nan, nan, 2.0, 3.0, 4.0]
+    """
+
+    name = "expanding_count"
+    input_types = [ColumnSchema(logical_type=Datetime, semantic_tags={"time_index"})]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={"numeric"})
+    uses_full_dataframe = True
+
+    def __init__(self, gap=1, min_periods=0):
```
thehomebrewnerd, line 58: Shouldn't min periods be 1?

featuretools/primitives/standard/transform/time_series/expanding/expanding_count.py
```diff
@@ -0,0 +1,68 @@
...
+        min_periods (int, optional): Minimum number of observations required for performing calculations
+            over the window. Defaults to 1.
+
+
+    Examples:
+        >>> import pandas as pd
+        >>> expanding_count = ExpandingCount()
+        >>> times = pd.date_range(start='2019-01-01', freq='1min', periods=5)
+        >>> expanding_count(times).tolist()
+        [0.0, 1.0, 2.0, 3.0, 4.0]
```
thehomebrewnerd, line 34: With a default gap of 1, why is the first value here `0.0` instead of `nan`? There shouldn't be any data to count before the first point.

featuretools/primitives/standard/transform/time_series/expanding/expanding_count.py
```diff
@@ -0,0 +1,68 @@
...
+        >>> import pandas as pd
+        >>> expanding_count = ExpandingCount(min_periods=3)
+        >>> times = pd.date_range(start='2019-01-01', freq='1min', periods=5)
+        >>> expanding_count(times).tolist()
+        [nan, nan, 2.0, 3.0, 4.0]
+    """
+
+    name = "expanding_count"
+    input_types = [ColumnSchema(logical_type=Datetime, semantic_tags={"time_index"})]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={"numeric"})
```
thehomebrewnerd, line 55: Shouldn't the return type be IntegerNullable - we should only get whole numbers and `nan` values, right?

featuretools/primitives/standard/transform/time_series/expanding/expanding_max.py
```diff
@@ -0,0 +1,71 @@
...
+from woodwork.logical_types import Datetime, Double
+
+from featuretools.primitives.base.transform_primitive_base import TransformPrimitive
+from featuretools.primitives.standard.transform.time_series.utils import (
+    _apply_gap_for_expanding_primitives,
+)
+
+
+class ExpandingMax(TransformPrimitive):
+    """Computes the expanding minimum of events over a given window.
```
thehomebrewnerd, line 12: "minimum" should be "maximum"

featuretools/primitives/standard/transform/time_series/expanding/expanding_max.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Description:
+        Given a list of datetimes, return an expanding minimum starting
```
thehomebrewnerd, line 15: minimum -> maximum

featuretools/primitives/standard/transform/time_series/expanding/expanding_max.py
```diff
@@ -0,0 +1,71 @@
...
+    Args:
+        gap (int, optional): Specifies a gap backwards from each instance before the
+            usable data begins. Corresponds to number of rows. Defaults to 1.
+        min_periods (int, optional): Minimum number of observations required for performing calculations
+            over the window. Defaults to 1.
+
+
+    Examples:
+        >>> import pandas as pd
+        >>> expanding_min = ExpandingMax()
```
thehomebrewnerd, line 31: min -> max

featuretools/primitives/standard/transform/time_series/expanding/expanding_max.py
```diff
@@ -0,0 +1,71 @@
...
+        >>> times = pd.date_range(start='2019-01-01', freq='1min', periods=5)
+        >>> expanding_min(times, [2, 4, 6, 7, 2]).tolist()
+        [nan, 2.0, 4.0, 6.0, 7.0]
+
+        We can also control the gap before the expanding calculation.
+
+        >>> import pandas as pd
+        >>> expanding_min = ExpandingMax(gap=0)
```
thehomebrewnerd, line 39: min -> max

featuretools/primitives/standard/transform/time_series/expanding/expanding_max.py
```diff
@@ -0,0 +1,71 @@
...
+        >>> times = pd.date_range(start='2019-01-01', freq='1min', periods=5)
+        >>> expanding_min(times, [2, 4, 6, 7, 2]).tolist()
+        [2.0, 4.0, 6.0, 7.0, 7.0]
+
+        We can also control the minimum number of periods required for the rolling calculation.
+
+        >>> import pandas as pd
+        >>> expanding_min = ExpandingMax(min_periods=3)
```
thehomebrewnerd, line 47: min -> max

featuretools/primitives/standard/transform/time_series/expanding/expanding_max.py
```diff
@@ -0,0 +1,71 @@
...
+        >>> expanding_min(times, [2, 4, 6, 7, 2]).tolist()
+        [nan, nan, nan, 6.0, 7.0]
+    """
+
+    name = "expanding_max"
+    input_types = [
+        ColumnSchema(logical_type=Datetime, semantic_tags={"time_index"}),
+        ColumnSchema(semantic_tags={"numeric"}),
+    ]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={"numeric"})
```
thehomebrewnerd, line 58: I think the return type should be just `ColumnSchema(semantic_tags={"numeric"})` here since the values could be integers or floats.

featuretools/primitives/standard/transform/time_series/expanding/expanding_mean.py
```diff
@@ -0,0 +1,71 @@
...
+from woodwork.logical_types import Datetime, Double
+
+from featuretools.primitives.base.transform_primitive_base import TransformPrimitive
+from featuretools.primitives.standard.transform.time_series.utils import (
+    _apply_gap_for_expanding_primitives,
+)
+
+
+class ExpandingMean(TransformPrimitive):
+    """Computes the expanding minimum of events over a given window.
```
thehomebrewnerd, line 12: "minimum" -> "mean"

featuretools/primitives/standard/transform/time_series/expanding/expanding_mean.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Description:
+        Given a list of datetimes, return an expanding minimum starting
```
thehomebrewnerd, line 15: minimum -> mean

featuretools/primitives/standard/transform/time_series/expanding/expanding_min.py
```diff
@@ -0,0 +1,70 @@
...
+        >>> expanding_min(times, [5, 4, 3, 2, 1]).tolist()
+        [nan, nan, nan, 3.0, 2.0]
+    """
+
+    name = "expanding_min"
+    input_types = [
+        ColumnSchema(logical_type=Datetime, semantic_tags={"time_index"}),
+        ColumnSchema(semantic_tags={"numeric"}),
+    ]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={"numeric"})
```
thehomebrewnerd, line 57: Probably can drop the logical type from the schema here.

thehomebrewnerd:
> To summarize current changes regarding return LogicalType:
> 
> Trend, Standard Dev: Double Count, Max, Mean, Min: No LogicalType specified

I think it should be this:
Mean, Trend, STD: Double (highly unlikely that any of this will return all whole numbers)
Count: IntegerNullable (count will always be a whole number)
Max, Min: No logical type (min/max could be float or int depending on type of input data)

featuretools/primitives/standard/transform/time_series/expanding/expanding_count.py
sbadithe: It turns out that calling shift on a `pd.date_range` object does not insert `nan` values at the beginning

sbadithe:
> > To summarize current changes regarding return LogicalType:
> > Trend, Standard Dev: Double Count, Max, Mean, Min: No LogicalType specified
> 
> I think it should be this: Mean, Trend, STD: Double (highly unlikely that any of these will return all whole numbers) Count: IntegerNullable (count will always be a whole number) Max, Min: No logical type (min/max could be float or int depending on type of input data)

Sorry about that. This makes sense to me

featuretools/tests/primitive_tests/test_expanding_primitives.py
```diff
@@ -133,3 +169,43 @@ def test_expanding_primitives_throw_warning(window_series_pd, primitive):
             numeric=window_series_pd,
             datetime=window_series_pd.index,
         )
+
+
+def test_apply_gap_for_expanding_primitives_throws_warning(window_series_pd):
```
thehomebrewnerd, line 174: `_warning` here should be `_error`

featuretools/tests/primitive_tests/test_expanding_primitives.py
```diff
@@ -133,3 +169,43 @@ def test_expanding_primitives_throw_warning(window_series_pd, primitive):
...
+    with pytest.raises(TypeError):
```
thehomebrewnerd, line 175: Since we are testing that we raise a custom TypeError here, we should also check that the error message matches what we expect. Without that, we don't know for sure that the error was the one we raised, or if it was coming from something else.

thehomebrewnerd (commented):
Just a couple small comments, but otherwise this is looking pretty good to me.

thehomebrewnerd (approved):

PR #2359, opened 2022-11-03
Add RateOfChange primitive

### LINKED ISSUE #2176: Add RateOfChange primitive
Add a primitive that computes the rate of change of a numeric value with respect to a time index column. Values would be computed by dividing the change in value by the change in time for consecutive data points.

Could also add an Acceleration primitive that simply computes the rate of change of `RateOfChange`.

thehomebrewnerd:
### Add RateOfChange primitive

Closes #2176 

Adds a new `RateOfChange` primitive that calculates the rate of change of a value in units per second, using the time index and a list of numeric values.


featuretools/primitives/standard/transform/numeric/rate_of_change.py
```diff
@@ -0,0 +1,34 @@
...
+    ]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={"numeric"})
+    uses_full_dataframe = True
+    description_template = "the rate of change of of {}"
+
+    def get_function(self):
+        def rate_of_change(values, time):
+            time_delta = time.diff().dt.total_seconds()
+            value_delta = values.diff()
+            return value_delta / time_delta
```
gsheni, line 31: Curious, what is the difference between your approach and this pandas function:
- https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.pct_change.html

thehomebrewnerd: Based on a quick read of the documentation, that function appears to just be calculating a percent change between consecutive values. This primitive calculates the rate of change between points (think velocity).

Take this dataframe as an example: 
```
values = pd.Series([0, 30, 180, -90, 0, 240])
time = pd.Series(pd.date_range(start='2019-01-01', freq='1min', periods=6))
df = pd.DataFrame({"values": values, "time": time})
```

If you use the pandas `pct_change` method you get this:
```
0    NaN
1    inf
2    5.0
3   -1.5
4   -1.0
5    inf
```

If you compute the rate of change using the elapsed time between points from the time index you get this:
```
0    NaN
1    0.5
2    2.5
3   -4.5
4    1.5
5    4.0
```

thehomebrewnerd: If the values were a distance measurement recorded in `meters`, the pandas function just tells you how much a value increased or decreased as a percentage of the previous one (`%`). This primitive will give you the rate of change of distance, which is a velocity in this case measured in units of `meters per second`.

Said another way, this primitive is basically giving you the first derivative of a set of values that are ordered by time.

featuretools/primitives/standard/transform/numeric/rate_of_change.py
```diff
@@ -0,0 +1,35 @@
+from woodwork.column_schema import ColumnSchema
+from woodwork.logical_types import Double
+
+from featuretools.primitives.base import TransformPrimitive
+
+
+class RateOfChange(TransformPrimitive):
+    """Computes the rate of change of a value per second.
```
sbadithe, line 8: I think it may be interesting in the future to have the unit of time be configurable

sbadithe, line 8: I think it may be interesting in the future to have the unit of time be configurable

featuretools/primitives/standard/transform/numeric/rate_of_change.py
```diff
@@ -0,0 +1,35 @@
...
+        >>> results
+        [nan, 0.5, 2.5, -4.5, 1.5]
+    """
+
+    name = "rate_of_change"
+    input_types = [
+        ColumnSchema(semantic_tags={"numeric"}),
+        ColumnSchema(semantic_tags={"time_index"}),
+    ]
+    return_type = ColumnSchema(logical_type=Double, semantic_tags={"numeric"})
```
sbadithe, line 25: Is there a rate semantic tag? Might be something to look at in the future.

sbadithe (approved):
lgtm! Just left a few comments about potential functionality in the future.

featuretools/primitives/standard/transform/numeric/rate_of_change.py
thehomebrewnerd: I also thought about that, but I don't think having a configurable unit would add predictive value since changing the unit effectively just scales the output by a fixed value. Might be useful in other circumstances though. My preference is to keep this simple until we identify a specific need for the additional complexity.

sbadithe: I agree. I think it would be more of a convenience for the user if they wanted to examine their feature matrix. For example, if the user had data sampled daily.

ozzieD (approved):

PR #2365, opened 2022-11-08
Add OneDigitPostalCode and TwoDigitPostalCode primitives

### LINKED ISSUE #2283: Add primitive for 1 digit Postal Code Prefix (US-only)
- As a user of Featuretools, I would like to do feature engineering for Postal Codes in USA. 
- I would like to extract the 1 digit prefix:
Screen Shot 2022-09-12 at 10 36 12 AM

### LINKED ISSUE #2284: Add primitive for 2 digit Postal Code Prefix (US-only)
- As a user of Featuretools, I would like to do feature engineering for Postal Codes in USA.
- I would like to extract the 2 digit prefix:
ZIP_Code_zones svg

sbadithe:
- Fixes #2283 
- Fixes #2284


featuretools/tests/conftest.py
```diff
@@ -746,6 +748,35 @@ def rolling_outlier_series_pd():
     )
 
 
+@pytest.fixture
```
dvreed77, line 751: Do we need these as all separate fixtures, if we are only using the dataframe fixture?

docs/source/api_reference.rst
```diff
@@ -280,6 +280,13 @@ NaturalLanguage Transform Primitives
    UpperCaseCount
    WhitespaceCount
 
+Postal Primitives
```
thehomebrewnerd, line 283: Maybe call this "Postal Code Primitives"?

docs/source/release_notes.rst
```diff
@@ -17,6 +17,7 @@ v1.18.0 Nov 15, 2022
     * Enhancements
         * Add ``RollingOutlierCount`` primitive (:pr:`2129`)
         * Add ``RateOfChange`` primitive (:pr:`2359`)
+        * Add ``OneDigitPostalCode`` and ``TwoDigitPostalCode`` primitives (:pr:`2365`)
```
thehomebrewnerd, line 20: This needs to move to the Future Release section now.

featuretools/tests/conftest.py
thehomebrewnerd: Along those lines - we also might want to test with Dask and Spark inputs, if since we are saying these primitives are also compatible with those libraries.

Also, does Woodwork force the type of the data once initialized? Is the data always stored as a string after Woodwork initialization? If so, there is no real need to test different input types (int vs string) here as we are just essentially testing that Woodwork is converting as we expect upon initialization, and that should be handled in the WW tests.

sbadithe: I think you are right that in the test cases here, the dtype is always a string categorical. 

However, I found this edge case where the dtype can be a float categorical: 

``` 
ser = pd.Series([12345, 67890]).astype('category')
ser = ww.init_series(ser, logical_type='PostalCode') 
```

In this case it appears in my jupyter notebook the type each element is a float.
 
I can update the tests here to reflect that case. I suspect this is related to the fact that there are multiple types of data that can hold the categorical dtype in pandas. If I remove the call to `astype`, the dtype is coerced to a string categorical. Woodwork seems to only check that the dtype is categorical, but not which type of categorical.

thehomebrewnerd: I think this might be a Woodwork issue. I think we should always expect the data to be of the same type once Woodwork is initialized. Let's file an issue in WW for this edge case. If we are coercing other types to string, I think we should do the same with a categorical input as well.

sbadithe: Done! I've also updated the tests to test this particular edge case for now.

featuretools/primitives/standard/transform/postal/one_digit_postal_code.py
```diff
@@ -0,0 +1,34 @@
...
+
+    Examples:
+        >>> one_digit_postal_code = OneDigitPostalCode()
+        >>> one_digit_postal_code(['92432', '34514']).tolist()
+        ['9', '3']
+    """
+
+    name = "one_digit_postal_code"
+    input_types = [ColumnSchema(logical_type=PostalCode)]
+    compatibility = [Library.PANDAS, Library.DASK, Library.SPARK]
```
thehomebrewnerd, line 23: If we are including Dask and Spark here, I think we should include tests with Dask and Spark inputs to explicitly test those cases. Also true for `TwoDigitPostalCode`.

featuretools/primitives/standard/transform/postal/two_digit_postal_code.py
```diff
@@ -0,0 +1,34 @@
...
+    name = "two_digit_postal_code"
+    input_types = [ColumnSchema(logical_type=PostalCode)]
+    compatibility = [Library.PANDAS, Library.DASK, Library.SPARK]
+    return_type = ColumnSchema(logical_type=Categorical, semantic_tags={"category"})
+    description_template = "The two digit postal code prefix of {}"
+
+    def get_function(self):
+        def two_digit_postal_code(postal_codes):
+            def transform_postal_code(pc):
+                return str(pc)[:2] if not pd.isna(pc) else pc
```
thehomebrewnerd, line 30: Minor nitpick, but I think it would be a little more clear to user `pd.notna(pc)` instead of `not pd.isna(pc)`.

featuretools/primitives/standard/transform/postal/one_digit_postal_code.py
```diff
@@ -0,0 +1,34 @@
...
+    return_type = ColumnSchema(logical_type=Categorical, semantic_tags={"category"})
+    description_template = "The one digit postal code prefix of {}"
+
+    def get_function(self):
+        def one_digit_postal_code(postal_codes):
+            def transform_postal_code(pc):
+                return str(pc)[0] if not pd.isna(pc) else pc
```
thehomebrewnerd, line 30: Minor nitpick, but I think it would be a little more clear to user `pd.notna(pc)` instead of `not pd.isna(pc)`.

featuretools/primitives/standard/transform/postal/two_digit_postal_code.py
```diff
@@ -0,0 +1,34 @@
...
+
+            return postal_codes.apply(transform_postal_code)
```
thehomebrewnerd, line 32: An alternate approach here, which might simplify the code is to use `.str.slice` to get the values. So instead of using `.apply(transform_postal_code)` you could just return `postal_codes.str.slice(start=0, stop=2)` and eliminate the need to define the custom function for `.apply`.

Not sure if there is a performance advantage of one over the other, but there should be less code with `.str.slice` assuming it works in all cases (might not work for the integer categorical edge case, but I'm not 100% convinced we will should cover that yet).

sbadithe:
Spent a bit of time last night trying to see if there is a better way to structure these tests. There's some repetition here, but I think this is the best I found so far.

featuretools/primitives/standard/transform/postal/two_digit_postal_code.py
sbadithe: I will run some perf tests and get back to you on this.

featuretools/tests/conftest.py
```diff
@@ -746,6 +748,62 @@ def rolling_outlier_series_pd():
...
+
+@pytest.fixture(
+    params=[
+        "postal_code_series_string_pd",
+        "postal_code_series_int_pd",
+        "postal_code_series_null_pd",
+    ],
+)
+def postal_code_series_dask(request):
+    dd = pytest.importorskip("dask.dataframe", reason="Dask not installed, skipping")
```
thehomebrewnerd, line 802: Dask is a core requirement and should always be installed.

thehomebrewnerd:
> Spent a bit of time last night trying to see if there is a better way to structure these tests. There's some repetition here, but I think this is the best I found so far.

I think one way to clean up the number fixtures would be to create a dataframe instead of individual series objects, then iterate over the dataframe columns in the tests.

featuretools/primitives/standard/transform/postal/two_digit_postal_code.py
sbadithe: It looks like performance is nearly identical

featuretools/tests/conftest.py
```diff
@@ -746,6 +747,52 @@ def rolling_outlier_series_pd():
...
+    ans = ps.from_pandas(postal_code_dataframe)
+    return ans
+
+
+@pytest.fixture
+def postal_code_dataframe_dask(postal_code_dataframe):
+    ans = dd.from_pandas(
+        postal_code_dataframe,
+        npartitions=1,
+    ).compute()
```
thehomebrewnerd, line 781: This compute call converts this into a pandas dataframe

thehomebrewnerd: Another note - in both this fixture and the spark fixture, you are losing the WW initialization. You will either need to reinitialize if that is important or handle initialization for all the fixtures inside the tests.

featuretools/tests/conftest.py
```diff
@@ -746,6 +747,52 @@ def rolling_outlier_series_pd():
     )
 
 
+@pytest.fixture
+def postal_code_dataframe():
```
thehomebrewnerd, line 751: Our convention has been to use the suffix `_pd` to denote the pandas version of the fixture, and then we drop all suffixes in the fixture that gets used for the tests. So I would rename this to `postal_code_dataframe_pd` and rename the `postal_code_dataframes` fixture to `postal_code_dataframe`.

featuretools/tests/primitive_tests/test_postal_primitives.py
```diff
@@ -0,0 +1,54 @@
...
+import pandas as pd
+from woodwork.accessor_utils import _is_spark_series
+
+from featuretools.primitives.standard.transform.postal import (
+    OneDigitPostalCode,
+    TwoDigitPostalCode,
+)
+
+
+def one_digit_postal_code_test(postal_series):
```
thehomebrewnerd, line 11: What are we doing in this test and `two_digit_postal_code_test`? There are no assertions?

sbadithe: Those are helper functions I wrote to simplify the logic in the primitive tests. They return the actual and expected values to be compared in the tests functions. I would be happy to move that logic back to the test functions if it would be cleaner.

thehomebrewnerd: Oh, I see now. I missed that before, obviously.

I think it is fine to use helper functions, but how about this:
- Rename them to something that doesn't include the `test` terminology. Reading the function name made me think we were testing something inside the function.
- Try to combine these into a single function, maybe pass in the primitive as a parameter instead of hard-coding the primitive name inside the helper function.
- I'm a little concerned that we are manipulating the output of the primitive function before we do our assertions. Is there a way we can update so that the expected values are *exactly* what we expect from the primitive function? That makes it much more clear what we are actually asserting, in my opinion.

sbadithe: I refactored the primitive to return `pd.NA` in the case of encountering a null value. I then just convert the Dask or Pyspark objects to a Pandas series, and used the Pandas testing utility. It is cleaner now because we don't have to deal separately with the case of `pd.NA` (if we use `np.testing` methods, we need to convert `pd.NA` to `np.nan`)

featuretools/tests/primitive_tests/test_postal_primitives.py
```diff
@@ -0,0 +1,44 @@
...
+    if _is_spark_series(postal_series):
+        expected = postal_series.apply(op).to_pandas()
+        actual = actual.to_pandas()
+    else:
+        expected = postal_series.apply(op)
+
+    if _is_dask_series(postal_series):
+        actual = actual.compute()
+        expected = expected.compute()
+    return actual, expected
```
thehomebrewnerd, line 26: Sorry if it feels like I'm being picky here, but I think I would use this function to only get the expected values. Call the primitive function directly in the test to get the actual values.

Also, I think we should remove the `.compute()` and `to_pandas()` calls in this function and use our testing utility function `to_pandas` in `es_utils.py` to handle all of the conversions to pandas.

If we remove those calls and use `to_pandas()` instead, do we even need to retain this function? Maybe not since most of the logic here seems to be handling conversion to pandas.

featuretools/tests/primitive_tests/test_postal_primitives.py
```diff
@@ -0,0 +1,23 @@
...
+from featuretools.primitives.standard.transform.postal import (
+    OneDigitPostalCode,
+    TwoDigitPostalCode,
+)
+from featuretools.tests.testing_utils.es_utils import to_pandas
+
+
+def test_one_digit_postal_code(postal_code_dataframe):
+    primitive = OneDigitPostalCode().get_function()
+    for _, row in postal_code_dataframe.iterrows():
```
thehomebrewnerd, line 12: Shouldn't we be iterating over the columns in the dataframe and not the rows?

PR #2380, opened 2022-11-23
Fix `base_of_exclude` handling

### LINKED ISSUE #2323: Fix `base_of_exclude` handling
In `_build_transform_features` in `deep_feature_synthesis.py`, we need to handle the case where `base_of_exclude` is defined in a custom primitive. The use case for this is if the user wants to block certain primitives from stacking on other ones during feature generation.

sbadithe:
- closes #2323 
- currently, the `check_stacking` function is only called during the generation of aggregation features. This means that features such as `base_of_exclude` and `stack_on_exclude` do not work when creating transform features. This PR addresses that bug. It also does some refactoring by moving the `stack_on_exclude` attribute and related member data off of AggregationPrimitiveBase and onto PrimitiveBase, seeing as the functionality provided does not need to be unique to AggregationPrimitives.


featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -1099,14 +1113,16 @@ def check_stacking(primitive, inputs):
...
+        # However, we want to check stacking rules with the primitive the DirectFeature is based on.
+        f_primitive = _find_root_primitive(feature)
+
         if not primitive_stack_on_self and isinstance(f_primitive, primitive_class):
             return False
 
         if isinstance(f_primitive, tup_primitive_stack_on_exclude):
             return False
 
-        # R TODO: handle this
```
sbadithe, line 1109: After discussing with @rwedge, determined that this todo was already completed

docs/source/release_notes.rst
```diff
@@ -9,6 +9,7 @@ Future Release
         * Add ``OneDigitPostalCode`` and ``TwoDigitPostalCode`` primitives (:pr:`2365`)
         * Add ``ExpandingCount``, ``ExpandingMin``, ``ExpandingMean``, ``ExpandingMax``, ``ExpandingSTD``, and ``ExpandingTrend`` primitives (:pr:`2343`)
     * Fixes
+        * Fix ``base_of_exclude`` handling in DeepFeatureSynthesis for TransformPrimitives (:pr:`2380`)
```
rwedge, line 12:
```suggestion
        * Fix DeepFeatureSynthesis to consider the ``base_of_exclude`` family of attributes when creating transform features(:pr:`2380`)
```

rwedge (approved):
LGTM

PR #2389, opened 2022-12-08
Only allow nonnegative version numbers

sbadithe:
- The test in question was testing invalid version numbers in the case where the minor version number was 0. The latest release of the `packaging` library started throwing an error for this case. This PR resolves that by taking the max of (0, minor-1) to prevent negative version numbers.


rwedge:
Does it make sense for these tests to be tied to the actual schema version number?  Should we mock the schema version instead to get consistent test coverage?

thehomebrewnerd:
@sbadithe When you pull in the latest changes from `main` you should also update `pyproject.toml` to remove the upper bound restriction on `packaging`.

featuretools/utils/gen_utils.py
```diff
@@ -43,23 +44,22 @@ def find_descendents(cls):
 
 def check_schema_version(cls, cls_type):
```
rwedge, line 45: remove in favor of function in `schema_utils.py`

gsheni (approved):

PR #2393, opened 2022-12-08
Update MultiplyNumericBoolean to work with more inputs

### LINKED ISSUE #2390: MultiplyNumericBoolean primitive doesn't work properly with `BooleanNullable` inputs on recent pandas versions
An error is raised by the MultiplyNumericBoolean primitive when using a `BooleanNullable` input. This seems to be a problem with more recent pandas versions as the error occurs on pandas 1.5.1 and 1.5.2, but was not present on 1.4.1.

```python
import pandas as pd
from featuretools.primitives import MultiplyNumericBoolean

n = pd.Series([1, 2, 3], dtype="float")
b = pd.Series([True, False, True], dtype="boolean")

prim = MultiplyNumericBoolean()
prim(n, b)
```
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
...
TypeError: Invalid value '0' for dtype boolean
```

Possible Solution: Instead of using the current np.where and masking approach, maybe we can just convert the boolean input series into an `Int64` series and multiply that converted series with the numeric series.

thehomebrewnerd:
### Update MultiplyNumericBoolean to work with more inputs

Certain combination of input dtypes would result in errors with newer pandas versions. This PR refactors the calculation in `MultiplyNumericBoolean` to resolve this issue.

Fixes #2390

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -10,6 +10,7 @@ Future Release
         * Add ``ExpandingCount``, ``ExpandingMin``, ``ExpandingMean``, ``ExpandingMax``, ``ExpandingSTD``, and ``ExpandingTrend`` primitives (:pr:`2343`)
     * Fixes
         * Fix DeepFeatureSynthesis to consider the ``base_of_exclude`` family of attributes when creating transform features(:pr:`2380`)
+        * Fix bug in ``MultiplyNumericBoolean`` primitive that cause error with certain input dtype combinations (:pr:`2393`)
     * Changes
     * Documentation Changes
     * Testing Changes
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -10,7 +10,7 @@ Future Release
         * Add ``ExpandingCount``, ``ExpandingMin``, ``ExpandingMean``, ``ExpandingMax``, ``ExpandingSTD``, and ``ExpandingTrend`` primitives (:pr:`2343`)
     * Fixes
         * Fix DeepFeatureSynthesis to consider the ``base_of_exclude`` family of attributes when creating transform features(:pr:`2380`)
-        * Fix bug in ``MultiplyNumericBoolean`` primitive that cause error with certain input dtype combinations (:pr:`2393`)
+        * Fix bug in ``MultiplyNumericBoolean`` primitive that can cause and error with certain input dtype combinations (:pr:`2393`)
     * Changes
     * Documentation Changes
     * Testing Changes
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -10,7 +10,7 @@ Future Release
         * Add ``ExpandingCount``, ``ExpandingMin``, ``ExpandingMean``, ``ExpandingMax``, ``ExpandingSTD``, and ``ExpandingTrend`` primitives (:pr:`2343`)
     * Fixes
         * Fix DeepFeatureSynthesis to consider the ``base_of_exclude`` family of attributes when creating transform features(:pr:`2380`)
-        * Fix bug in ``MultiplyNumericBoolean`` primitive that can cause and error with certain input dtype combinations (:pr:`2393`)
+        * Fix bug in ``MultiplyNumericBoolean`` primitive that can cause an error with certain input dtype combinations (:pr:`2393`)
     * Changes
     * Documentation Changes
     * Testing Changes
--- a/featuretools/primitives/standard/transform/binary/multiply_numeric_boolean.py
+++ b/featuretools/primitives/standard/transform/binary/multiply_numeric_boolean.py
@@ -1,5 +1,3 @@
-import numpy as np
-import pandas as pd
 import pandas.api.types as pdtypes
 from woodwork.column_schema import ColumnSchema
 from woodwork.logical_types import Boolean, BooleanNullable
@@ -60,8 +58,6 @@ class MultiplyNumericBoolean(TransformPrimitive):
                 bools = ser2
                 vals = ser1
             result = vals * bools.astype("Int64")
-            # Replace all pd.NA with np.nan to avoid WW init error
-            result = result.replace({pd.NA: np.nan})
             return result
 
         return multiply_numeric_boolean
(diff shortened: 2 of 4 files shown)


featuretools/primitives/standard/transform/binary/multiply_numeric_boolean.py
```diff
@@ -17,13 +17,14 @@ class MultiplyNumericBoolean(TransformPrimitive):
...
     Examples:
+        >>> import pandas as pd
         >>> multiply_numeric_boolean = MultiplyNumericBoolean()
         >>> multiply_numeric_boolean([2, 1, 2], [True, True, False]).tolist()
         [2, 1, 0]
         >>> multiply_numeric_boolean([2, None, None], [True, True, False]).tolist()
         [2.0, nan, nan]
-        >>> multiply_numeric_boolean([2, 1, 2], [True, True, None]).tolist()
-        [2.0, 1.0, nan]
+        >>> multiply_numeric_boolean([2, 1, 2], pd.Series([True, True, pd.NA], dtype="boolean")).tolist()
```
thehomebrewnerd, line 26: With the new code the old boolean series had a dtype of object and pandas wasn't converting this to `Int64`. Updated here to use `boolean` since we should always get WW initialized series as inputs during normal CFM operations and that will have a dtype of either `bool` or `boolean`.

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -1808,7 +1808,7 @@ def test_multiply_numeric_boolean():
 
     multiply_numeric_boolean = MultiplyNumericBoolean()
     for input in test_cases:
-        vals = pd.Series(input["val"])
+        vals = pd.Series(input["val"]).astype("Int64")
```
thehomebrewnerd, line 1811: If we don't covert here the input with a single `pd.NA` value ends up with a dtype of `object` which would never happen in real usage. An `object` input to the primitive also causes an error.

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -1818,6 +1818,38 @@ def test_multiply_numeric_boolean():
...
+            expected = pd.Series([1, 0, 3])
+            pd.testing.assert_series_equal(actual, expected, check_dtype=False)
+
+
+def test_multiply_numeric_boolean_multiple_dtypes_with_nulls():
+    # Test with null values
+    vals = pd.Series([np.nan, 2, 3])
+    bools = pd.Series([True, False, pd.NA], dtype="boolean")
+    multiply_numeric_boolean = MultiplyNumericBoolean()
+    multiply_numeric_boolean = MultiplyNumericBoolean()
```
sbadithe, line 1844: double initialization here and on L1843

docs/source/release_notes.rst
```diff
@@ -10,6 +10,7 @@ Future Release
         * Add ``ExpandingCount``, ``ExpandingMin``, ``ExpandingMean``, ``ExpandingMax``, ``ExpandingSTD``, and ``ExpandingTrend`` primitives (:pr:`2343`)
     * Fixes
         * Fix DeepFeatureSynthesis to consider the ``base_of_exclude`` family of attributes when creating transform features(:pr:`2380`)
+        * Fix bug in ``MultiplyNumericBoolean`` primitive that can cause and error with certain input dtype combinations (:pr:`2393`)
```
sbadithe, line 13:
```suggestion
        * Fix bug in ``MultiplyNumericBoolean`` primitive that can cause an error with certain input dtype combinations (:pr:`2393`)
```

featuretools/tests/primitive_tests/test_transform_features.py
thehomebrewnerd: Removed

docs/source/release_notes.rst
thehomebrewnerd: Fixed

featuretools/primitives/standard/transform/binary/multiply_numeric_boolean.py
```diff
@@ -53,15 +54,12 @@ class MultiplyNumericBoolean(TransformPrimitive):
...
-                mask = ser2
+                bools = ser2
                 vals = ser1
-            vals_not_null = vals.notnull()
-            # Only apply mask where the input is not null
-            mask = mask.where(vals_not_null)
-            result = vals.where(mask, mask.replace({False: 0}))
+            result = vals * bools.astype("Int64")
             # Replace all pd.NA with np.nan to avoid WW init error
             result = result.replace({pd.NA: np.nan})
```
sbadithe, line 64: Do we need this line? In the doctest, the result has a `pd.NA`.

thehomebrewnerd: That's a good question. On the surface it seems like it is no longer necessary, but let me look a bit closer. 

One thing I'm concerned about - the doc test doesn't actually initialize Woodwork, that happens later in the CFM process and this comment said it was done to avoid a WW init error. The doctest wouldn't cover that, and also the doctest example doesn't contain a `np.nan` because the return type is `Int64` which cannot hold `np.nan`, so the replace call does nothing in that case.

thehomebrewnerd: I ran DFS/CFM using a dataframe that had all the column dtype combinations that we should run into, and it worked successfully. That leads me to believe this `replace` call is no longer required. I'll try to remove and see if our CI passes for the min package versions as well.

Here are the features that were generated and computed successfully:
```
[<Feature: int>,
 <Feature: Int_no_null>,
 <Feature: Int_null>,
 <Feature: float>,
 <Feature: float_null>,
 <Feature: bool>,
 <Feature: boolean_no_null>,
 <Feature: boolean_null>,
 <Feature: bool * Int_no_null>,
 <Feature: bool * Int_null>,
 <Feature: bool * float>,
 <Feature: bool * float_null>,
 <Feature: bool * int>,
 <Feature: boolean_no_null * Int_no_null>,
 <Feature: boolean_no_null * Int_null>,
 <Feature: boolean_no_null * float>,
 <Feature: boolean_no_null * float_null>,
 <Feature: boolean_no_null * int>,
 <Feature: boolean_null * Int_no_null>,
 <Feature: boolean_null * Int_null>,
 <Feature: boolean_null * float>,
 <Feature: boolean_null * float_null>,
 <Feature: boolean_null * int>]
```

PR #2397, opened 2022-12-10
Remove extra calls to `all`, `.values()`, and `.keys()` when constructing `ignore_columns` in `DeepFeatureSynthesis`

sbadithe:
- In the `init` method for `DeepFeatureSynthesis`, we perform a type check on `ignore_columns` to make sure it maps strings to list of strings 
- Currently, we iterate multiple times over the input to perform this type checking, then iterate again to create the `ignore_columns` data structure. 
- By refactoring, we can reduce this to iterating once over the input. This reduces code complexity and computation required

### DIFF
--- a/featuretools/synthesis/deep_feature_synthesis.py
+++ b/featuretools/synthesis/deep_feature_synthesis.py
@@ -1263,23 +1263,19 @@ def _features_have_same_path(input_features):
 
 
 def _build_ignore_columns(input_dict: Dict[str, List[str]]) -> DefaultDict[str, set]:
-    """Iterates over the input dictionary and builds the ignore_columns data structure"""
+    """Iterates over the input dictionary to build the ignore_columns dictionary.
+    Expects the input_dict's keys to be strings, and values to be lists of strings
+    Throws a TypeError if they are not.
+    """
     ignore_columns = defaultdict(set)
     if input_dict is not None:
         for df_name, cols in input_dict.items():
-            _validate_ignore_columns_entry(df_name, cols)
+            if not isinstance(df_name, str) or not isinstance(cols, list):
+                raise TypeError("ignore_columns should be dict[str -> list]")
+            elif not all(isinstance(c, str) for c in cols):
+                raise TypeError("list in ignore_columns must only have string values")
             ignore_columns[df_name] = set(cols)
     return ignore_columns
-
-
-def _validate_ignore_columns_entry(df_name: str, col: List[str]) -> None:
-    """Check that ignore_columns dictionary maps strings to list of strings,
-    raise a TypeError if it does not.
-    """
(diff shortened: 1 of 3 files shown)


featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -194,18 +194,8 @@ def __init__(
...
-                isinstance(i, list) for i in ignore_columns.values()
-            ):
-                raise TypeError("ignore_columns should be dict[str -> list]")
-            # check if list values are all of type str
-            elif not all(
-                all(isinstance(v, str) for v in value)
-                for value in ignore_columns.values()
-            ):
-                raise TypeError("list values should be of type str")
             for df_name, cols in ignore_columns.items():
```
sbadithe, line 197: What about refactoring this line and the next two into its own function `_make_ignore_columns` (or something similar). It would help us separate implementation details from the algorithm itself. 

More generally, I am curious to hear thoughts about refactoring chunks like 

``` 
self.agg_primitives = sorted(
            [
                check_primitive(
                    p,
                    "aggregation",
                    aggregation_primitive_dict,
                    transform_primitive_dict,
                )
                for p in agg_primitives
            ],
        )
```

into their own functions? We could then have a more testable API that might be easier to reason about (i.e. we could reason about what we want `self.agg_primitives` to be rather than how do it).

ozzieD: Yeah, any consolidation of duplicated code is a win in my book, what you have here already is a good step in making our code more testable, readable, and efficient. Also, I prefer to keep mrs small so if we go down the path of refactoring something else let us do it in a separate mr.

sbadithe: I agree. I will create an issue to track that refactoring.

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -1276,6 +1262,26 @@ def _features_have_same_path(input_features):
     return True
 
 
+def _build_ignore_columns(input_dict: Dict[str, List[str]]) -> DefaultDict[str, set]:
+    """Iterates over the input dictionary and builds the ignore_columns data structure"""
+    ignore_columns = defaultdict(set)
+    if input_dict is not None:
+        for df_name, cols in input_dict.items():
+            _validate_ignore_columns_entry(df_name, cols)
```
rwedge, line 1270: if `_build_ignore_columns` is now separated out, should we move the code for `_validate_ignore_columns_entry` back in?

sbadithe: I think that makes sense

rwedge (approved):
LGTM

Issue #2070, opened 2022-05-11
Add BoxCox primitive


gsheni:
- https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.boxcox.html
![scipy-stats-boxcox-1](https://user-images.githubusercontent.com/8726321/167942768-de82060a-383f-4495-911e-5b0d49516d97.png)

ozzieD:
Also add to recommended primitives

PR #2398, opened 2022-12-12
Add a get_recommend_primitives function for a single table EntitySet

ozzieD:
- Add a `get_recommend_primitives` function which given a single table entity set, provides a list of recommended primitives.


featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,261 @@
...
+    "rolling_min",
+    "rolling_std",
+    "rolling_trend",
+    "lag",
+]
+
+
+# TODO: Support multi-table
+def get_recommended_primitives(
+    es: EntitySet,
```
rwedge, line 113: I'd suggest `entityset` instead of `es` to be consistent with other functions in the library

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,261 @@
...
+            ),
+        )
+
+    if valid_time_series_primitives:
+        recommended_primitives.update(valid_time_series_primitives)
+    return list(recommended_primitives)
+
+
+def _recommend_non_numeric_primitives(
+    es: EntitySet,
```
rwedge, line 185: same as above

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,261 @@
...
+
+    Args:
+        es (EntitySet): EntitySet that only contains one dataframe.
+        target_dataframe_name (str): Name of target dataframe to access in `es`.
+        valid_primitives (List[str]): List of primitives to calculate and check output features.
+    """
+
+    recommended_non_numeric_primitives = set()
+    # Only want to run feature generation on non numeric primitives
+    es_without_numerics = EntitySet("ES without numerics")
```
rwedge, line 203: Instead of making a second entityset, we could use the `ignore_columns` parameter in `dfs` to filter out all numeric columns for that `dfs` run

There's an issue with how the second ES is created that we can avoid by not making a second one

ozzieD: ahh nice yeah that would work better

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,261 @@
...
+    "greater_than_scalar",
+    "less_than_scalar",
+    "modulo_by_feature",
+    "time_since",
+    "savgol_filter",
+    "isin",
+    "numeric_bin",
+]
+
+OTHER_PRIMITIVES_TO_EXCLUDE = [
```
sbadithe, line 69: this is a little mysterious to me, can we change the name or add a comment explaining why these primitives are excluded?

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,261 @@
...
+]
+
+DEFAULT_EXCLUDED_PRIMITIVES = (
+    REQUIRED_INPUT_PRIMITIVES
+    + DEPRECATED_PRIMITIVES
+    + ORDERED_PRIMITIVES
+    + OTHER_PRIMITIVES_TO_EXCLUDE
+)
+
+TIME_SERIES_PRIMITIVES = [
```
sbadithe, line 100: Just curious, is the plan to include Expanding primitives here?

ozzieD: I think so I wanted to get the group's perspective but I'm going to assume the answer is yes and make this change.

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,264 @@
...
+        recommended_primitives.update(valid_time_series_primitives)
+    return list(recommended_primitives)
+
+
+def _recommend_non_numeric_primitives(
+    entityset: EntitySet,
+    target_dataframe_name: str,
+    valid_primitives: List[str],
+) -> set:
+    """Get a set of recommended non-numeric primitives given an entity set.
```
sbadithe, line 193: Should we add the other arguments in the docstring, like 

`Given an entity set, target_dataframe_name, and valid_primitives, returns a set of recommended non-numeric primitives`

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,264 @@
...
+
+
+DEPRECATED_PRIMITIVES = [
+    # // as of premium - primitives 0.14.0 and feature - tools 0.26.1
+    "multiply_boolean",  # functionality duplicated by 'and' primitive
+    "numeric_lag",  # deperecated and replaced with `lag`
+]
+
+REQUIRED_INPUT_PRIMITIVES = [
+    # // as of premium - primitives 0.14.0 and feature - tools 0.26.1
```
thehomebrewnerd, line 44: Remove comment - doesn't apply here.

featuretools/utils/recommend_primitives.py
thehomebrewnerd: I think we need to include all the time series only primitives, because if we don't they won't get added to the list of excluded primitives for non-time series problems, correct? Also we have a RollingOutlierCount primitive that needs to be included.

We really need to have a better way to identify these primitives longer term. I'm afraid manually maintaining this list will lead to problems long term (but that is out of scope for this PR).

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,264 @@
...
+    )
+    skew_numeric_primitives = set(["square_root", "natural_logarithm"])
+    valid_skew_primtives = skew_numeric_primitives.intersection(valid_primitive_names)
+
+    time_series_primitives = set(TIME_SERIES_PRIMITIVES)
+    valid_time_series_primitives = time_series_primitives.intersection(
+        valid_primitive_names,
+    )
+
+    numeric_tags_only_df = entityset[target_dataframe_name].ww.select("numeric")
```
thehomebrewnerd, line 173: Would it be better to move this selection process into the `_recommend_skew_numeric_primitives` function?

featuretools/utils/recommend_primitives.py
ozzieD: Added a comment

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,264 @@
...
+    "rolling_trend",
+    "lag",
+]
+
+
+# TODO: Support multi-table
+def get_recommended_primitives(
+    entityset: EntitySet,
+    target_dataframe_name: str,
+    is_time_series: bool,
```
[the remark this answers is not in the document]
ozzieD: yeah I was also iffy on this name. `include_time_series_primitives` is better

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,273 @@
...
+
+    Note:
+        This function currently only works for single table and will only recommend transform primitives.
+    """
+    recommended_primitives = set()
+
+    if not is_time_series:
+        excluded_primitives += TIME_SERIES_PRIMITIVES
+
+    if is_time_series:
```
sbadithe, line 147:
```suggestion
   else:
```

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,273 @@
...
+    for col in numerics_only_df:
+        # Shouldn't recommend log, sqrt if nans, zeros and negative numbers are present
+        contains_nan = numerics_only_df[col].isnull().values.any()
+        all_above_zero = (numerics_only_df[col] > 0).all()
+        if all_above_zero and not contains_nan:
+            skew = numerics_only_df[col].skew()
+            # We currently don't have anything in featuretools to automatically handle left skewed data as well as skewed data with negative values
+            if skew > 0.5 and skew < 1 and "square_root" in valid_skew_primtives:
+                recommended_skew_primitives.add("square_root")
+                # TODO: Add Box Cox here when available
```
sbadithe, line 269: Should we make issues to track these?

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,273 @@
...
+DEFAULT_EXCLUDED_PRIMITIVES = (
+    REQUIRED_INPUT_PRIMITIVES
+    + DEPRECATED_PRIMITIVES
+    + ORDERED_PRIMITIVES
+    + OTHER_PRIMITIVES_TO_EXCLUDE
+)
+
+# TODO: Make this list more dynamic
+TIME_SERIES_PRIMITIVES = [
+    "expanding_count",
```
sbadithe, line 102: @ozzieD @thehomebrewnerd 
Should exponential primitives be included here?

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,273 @@
+from typing import List
+
+import pandas as pd
+
+from featuretools.computational_backends import calculate_feature_matrix
+from featuretools.entityset import EntitySet
+from featuretools.primitives.utils import get_transform_primitives
+from featuretools.synthesis import dfs, get_valid_primitives
+
+ORDERED_PRIMITIVES = [  # primitives that require ordering
```
sbadithe, line 10: I think this name could be confused for being an ordered list of primitives

featuretools/utils/recommend_primitives.py
thehomebrewnerd: No - those primitives are not created properly to be used for time series problems where a measurement delay might be present.

featuretools/utils/recommend_primitives.py
ozzieD: we have it already : https://github.com/alteryx/featuretools/issues/2070

featuretools/utils/recommend_primitives.py
ozzieD: I didn't want to add the extra parameters to `_recommend_skew_numeric_primitives` if we didn't need them, also we may end up using the numeric for something else in the future.

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,264 @@
...
+) -> List[str]:
+    """Get a list of recommended primitives given an entity set.
+
+    Description:
+        Given a single table entity set with a `target_dataframe_name`, `is_time_series` specified
+        and a list of any primitives in `excluded_primitives` to not be included in the final recommendation list.
+
+    Args:
+        entityset (EntitySet): EntitySet that only contains one dataframe.
+        target_dataframe_name (str): Name of target dataframe to access in `entityset`.
```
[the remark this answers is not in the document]
ozzieD: I'm fine with that,  I figured I'd keep this in case we do support multi-table in the future. But if it's not looking like this will be done in the near future I can remove this parameter.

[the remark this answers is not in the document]
thehomebrewnerd: I'd vote for removing it until we need it since we don't have near-term plans for multi-table.

featuretools/utils/recommend_primitives.py
sbadithe: Oh, I meant to to make sure we add it back here!

featuretools/utils/recommend_primitives.py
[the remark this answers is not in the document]
ozzieD: So yesterday, I did the former but just added some logic and exceptions if the number of DataFrames in the entityset is 0 or greater than 1. lmk what you think.

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,264 @@
...
+def get_recommended_primitives(
+    entityset: EntitySet,
+    target_dataframe_name: str,
+    is_time_series: bool,
+    excluded_primitives: List[str] = DEFAULT_EXCLUDED_PRIMITIVES,
+) -> List[str]:
+    """Get a list of recommended primitives given an entity set.
+
+    Description:
+        Given a single table entity set with a `target_dataframe_name`, `is_time_series` specified
```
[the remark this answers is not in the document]
ozzieD: Cleaned up the wording.

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,264 @@
...
+    "rolling_mean",
+    "rolling_min",
+    "rolling_std",
+    "rolling_trend",
+    "lag",
+]
+
+
+# TODO: Support multi-table
+def get_recommended_primitives(
```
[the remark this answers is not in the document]
ozzieD: added to `api_reference.rst`

docs/source/release_notes.rst
```diff
@@ -30,6 +31,7 @@ v1.20.0 Jan 5, 2023
         * Fix to eliminate fragmentation ``PerformanceWarning`` in ``feature_set_calculator.py`` (:pr:`2424`)
         * Fix serialization of ``NumberOfCommonWords`` feature with custom word_set (:pr:`2432`)
         * Improve edge case handling in NaturalLanguage primitives by standardizing delimiter regex (:pr:`2423`)
+        * Fix typo in ``_handle_binary_comparison`` function name and update ``set_feature_names`` docstring (:pr:`2388`)
```
rwedge, line 35: duplicate of line 28

ozzieD: yup was just fixing this, idk what happened with my merge lol

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,244 @@
...
+        )
+
+    target_dataframe_name = es_dataframe_list[0]
+
+    recommended_primitives = set()
+
+    if not include_time_series_primitives:
+        excluded_primitives += TIME_SERIES_PRIMITIVES
+    else:
+        recommended_primitives.add("lag")
```
rwedge, line 118: I think this could be removed and Lag would still get added, if valid, along with the other valid time series primitives

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,244 @@
...
+        numerics_only_df (pd.DataFrame): Woodwork initialized dataframe of origin features with only `numeric` semantic tags
+        valid_skew_primitives (set): set of valid skew primitives (square_root or natural_logarithm) to potentially recommend.
+
+    Note:
+        We currently only have primitives to address right skewness.
+    """
+    recommended_skew_primitives = set()
+    for col in numerics_only_df:
+        # Shouldn't recommend log, sqrt if nans, zeros and negative numbers are present
+        contains_nan = numerics_only_df[col].isnull().values.any()
```
rwedge, line 233:
```suggestion
        contains_nan = numerics_only_df[col].isnull().any()
```

rwedge (commented):
General thought, should we import primitive classes and use Primitive.name instead?  Primitive string names have changed in the past

featuretools/utils/recommend_primitives.py
ozzieD: I think this was left over from previous logic, good catch

featuretools/tests/utils_tests/test_recommend_primitives.py
```diff
@@ -0,0 +1,195 @@
...
+            "num_characters",
+        ],
+    )
+    assert expected_recomendations == actual_recomendations
+
+
+def test_get_recommended_primitives(make_es):
+    ecom_es_customers = EntitySet()
+    ecom_es_customers.add_dataframe(make_es["customers"])
+    prims_to_exclude = DEFAULT_EXCLUDED_PRIMITIVES + ["latlong_to_city"]
```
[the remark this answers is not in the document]
ozzieD: There is an issue with the reverse geocoder in`latlong_to_city` but it doesn't apply to these tests. I'll remove.

ozzieD:
> General thought, should we import primitive classes and use Primitive.name instead? Primitive string names have changed in the past

I guess my only concern would be trying to import primitives from the outside libraries

featuretools/tests/utils_tests/test_recommend_primitives.py
```diff
@@ -0,0 +1,195 @@
...
+        "num_unique_separators",
+        "age",
+        "year",
+        "is_leap_year",
+        "days_in_month",
+        "is_free_email_domain",
+        "number_of_unique_words",
+    ]
+    for prim in expected_recomendations:
+        assert prim in actual_recomendations
```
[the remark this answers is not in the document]
ozzieD: That would fail min requirements without premium primitives installed

rwedge:
> > General thought, should we import primitive classes and use Primitive.name instead? Primitive string names have changed in the past
> 
> I guess my only concern would be trying to import primitives from the outside libraries

I guess related to that, it could useful for constants like `TIME_SERIES_PRIMITIVES` to be adjustable by the user to include their own primitives

featuretools/tests/utils_tests/test_recommend_primitives.py
```diff
@@ -0,0 +1,192 @@
...
+def test_recommend_non_numeric_primitives(make_es):
+    ecom_es_customers = EntitySet("ES without numerics")
+    ecom_es_customers.add_dataframe(make_es["customers"])
+    valid_primitives = [
+        "day",
+        "num_characters",
+        "natural_logarithm",
+        "sine",
+    ]
+    actual_recomendations = _recommend_non_numeric_primitives(
```
sbadithe, line 95:
```suggestion
    actual_recommendations = _recommend_non_numeric_primitives(
```

featuretools/tests/utils_tests/test_recommend_primitives.py
```diff
@@ -0,0 +1,192 @@
...
+        "customers",
+        valid_primitives,
+    )
+    expected_recomendations = set(
+        [
+            "day",
+            "num_characters",
+        ],
+    )
+    assert expected_recomendations == actual_recomendations
```
sbadithe, line 106:
```suggestion
    assert expected_recommendations == actual_recommendations
```

featuretools/tests/utils_tests/test_recommend_primitives.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def test_get_recommended_primitives_time_series(make_es):
+    ecom_es_log = EntitySet()
+    ecom_es_log.add_dataframe(make_es["log"])
+    ecom_es_log["log"].ww.set_time_index("datetime")
+    actual_recomendations_ts = get_recommended_primitives(
```
sbadithe:
```suggestion
    actual_recommendations_ts = get_recommended_primitives(
```

Tiny typo with with some more occurrences in this test

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,242 @@
...
+        if (
+            f.primitive.name is not None
+            and f.primitive.name not in recommended_non_numeric_primitives
+        ):
+            try:
+                matrix = calculate_feature_matrix([f], entityset)
+                for f_name in f.get_feature_names():
+                    if len(matrix[f_name].unique()) > 1:
+                        recommended_non_numeric_primitives.add(f.primitive.name)
+            except Exception:  # If error in calculating feature matrix pass on the recommendation
```
sbadithe, line 205: minor styling question, but in general, do we prefer this approach or `contextlib.suppress`?

featuretools/tests/utils_tests/test_recommend_primitives.py
```diff
@@ -0,0 +1,192 @@
...
+        "number_of_unique_words",
+    ]
+    for prim in expected_recommendations:
+        assert prim in actual_recommendations
+
+    for ts_prim in TIME_SERIES_PRIMITIVES:
+        assert ts_prim not in actual_recommendations
+
+
+def test_get_recommended_primtives_exclude(make_es):
```
sbadithe, line 161:
```suggestion
def test_get_recommended_primitives_exclude(make_es):
```

featuretools/utils/recommend_primitives.py
ozzieD: I've always just used `try,except`.

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,242 @@
...
+
+    Args:
+        entityset (EntitySet): EntitySet that only contains one dataframe.
+        include_time_series_primitives (bool): Whether or not time-series primitives should be considered.
+        excluded_primitives (List[str]): List of transform primitives to exclude from recommendations.
+
+    Note:
+        The main objective of this function is to recommend primitives that could potentially provide important features to the modeling process.
+        Non-numeric primitives do a great job in mainly serving as a way to extract information from origin features that may essentially be meaningless by themselves (e.g., NaturalLaguange, Datetime, LatLong).
+        That is why they are the main focus of this function. Numeric transform primitives are very case-by-case dependent and therefore it is hard to mathmatically quantify which should be recommended.
```
sbadithe, line 96:
```suggestion
        That is why they are the main focus of this function. Numeric transform primitives are very case-by-case dependent and therefore it is hard to mathematically quantify which should be recommended.
```

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,242 @@
...
+        "is_first_occurrence",
+        "is_last_occurrence",
+        "time_since_previous",
+    ]
+)
+
+
+DEPRECATED_PRIMITIVES = [
+    "multiply_boolean",  # functionality duplicated by 'and' primitive
+    "numeric_lag",  # deperecated and replaced with `lag`
```
sbadithe, line 26:
```suggestion
    "numeric_lag",  # deprecated and replaced with `lag`
```

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,242 @@
...
+        Given a single table entity set with a `target_dataframe_name` and an applicable list of `valid_primitives`,
+        get a set of primitives which produce non-unique features.
+
+    Args:
+        entityset (EntitySet): EntitySet that only contains one dataframe.
+        target_dataframe_name (str): Name of target dataframe to access in `entityset`.
+        valid_primitives (List[str]): List of primitives to calculate and check output features.
+    """
+
+    recommended_non_numeric_primitives = set()
```
sbadithe, line 179: I really like that the functions are typed. Could we add a type annotation here to make it clear this set will hold strings? I think it could aid comprehension.

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,242 @@
...
+        get a set of primitives which could be applied to address right skewness.
+
+    Args:
+        numerics_only_df (pd.DataFrame): Woodwork initialized dataframe of origin features with only `numeric` semantic tags
+        valid_skew_primitives (set): set of valid skew primitives (square_root or natural_logarithm) to potentially recommend.
+
+    Note:
+        We currently only have primitives to address right skewness.
+    """
+    recommended_skew_primitives = set()
```
sbadithe, line 228: Same comment as above

docs/source/api_reference.rst
```diff
@@ -513,3 +513,4 @@ Primitive Util Methods
 
     list_primitives
     summarize_primitives
+    get_recommended_primitives
```
thehomebrewnerd, line 516: Big nitpick, but could we put these in alphabetical order:

```suggestion
    get_recommended_primitives
    list_primitives
    summarize_primitives
```

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,242 @@
...
+    excluded_primitives: List[str] = DEFAULT_EXCLUDED_PRIMITIVES,
+) -> List[str]:
+    """Get a list of recommended primitives given an entity set.
+
+    Description:
+        This function works by first getting a list of valid primitives withholding any primitives specified in `excluded_primitives` that could be applied to a single-table EntitySet.
+        Secondly, engineered features are created for non-numeric fields and are checked for non-uniqueness. If the feature is non-unique, it is added to the recommendation list.
+        Then, numeric fields are checked for skewness. Depending on how skew a column is `square_root` or `natural_logarithm` will be recommended.
+        Lastly if `include_time_series_primitives` is specified as `True`, `Lag` will always be recommended,
+        as well as all Rolling and Expanding primitives if numeric columns are present.
```
thehomebrewnerd, line 86: I'm not sure how much we have studied this, but including all Rolling and Expanding primitives might be overkill. I'm ok with leaving this as is for now, but might be something we want to look into more in the future.

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,242 @@
...
+        include_time_series_primitives (bool): Whether or not time-series primitives should be considered.
+        excluded_primitives (List[str]): List of transform primitives to exclude from recommendations.
+
+    Note:
+        The main objective of this function is to recommend primitives that could potentially provide important features to the modeling process.
+        Non-numeric primitives do a great job in mainly serving as a way to extract information from origin features that may essentially be meaningless by themselves (e.g., NaturalLaguange, Datetime, LatLong).
+        That is why they are the main focus of this function. Numeric transform primitives are very case-by-case dependent and therefore it is hard to mathematically quantify which should be recommended.
+        Therefore, we decided to only look at transform primitives that address skewed numeric columns as this is a standard and quantifiable transformation step. The only exception to this rule being
+        for time series problems. Because there are so few primitives that are only applicable for time series, we recommend users try all of them.
```
thehomebrewnerd, line 98: Wording nitpick - maybe rewrite to remove the "we decided..." and "we recommend..." text.

Maybe something like
```
Therefore, only transform primitives that address skewed numeric columns are included, as this is a standard and quantifiable transformation step.
```

and 

```
Because there are so few primitives that are only applicable for time series, all of them are included in the recommended primitives list.
```

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,242 @@
...
+
+    Note:
+        This function currently only works for single table and will only recommend transform primitives.
+    """
+    es_dataframe_list = list(entityset.dataframe_dict)
+    if len(es_dataframe_list) == 0:
```
thehomebrewnerd, line 104: Instead of this dict to list conversion you could just us `es.dataframes` here. `es.dataframes` is a list of all the dataframes in the EntitySet. Then below when you get the target dataframe name you could grab it with `es.dataframes[0].ww.name`.

Not sure if that's better or not...

featuretools/utils/recommend_primitives.py
thehomebrewnerd: Do we know of a specific case where this Exception gets thrown or are we just including this in case this happens? Just thinking that if this step fails, that probably means we have another bug somewhere that we need to address.

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,242 @@
...
+    ]
+
+    recommended_primitives.update(
+        _recommend_non_numeric_primitives(
+            entityset,
+            target_dataframe_name,
+            valid_primitive_names,
+        ),
+    )
+    skew_numeric_primitives = set(["square_root", "natural_logarithm"])
```
thehomebrewnerd, line 140: In terms of separation of concerns, would it make more sense for the `_recommend_skew_numeric_primitives` function to contain this list? We have this list here and the rest of the logic in the other function, which means you need to look at two functions to understand what gets recommended. Same thing for the time series logic.

Just thinking maybe it would make more sense for us to always call those functions and have the logic for what to do based on the inputs all contained within those functions.

featuretools/tests/utils_tests/test_recommend_primitives.py
```diff
@@ -0,0 +1,192 @@
...
+        _recommend_skew_numeric_primitives(
+            right_skew_moderate_and_heavy_df,
+            valid_skew_primtives,
+        )
+        == valid_skew_primtives
+    )
+
+
+def test_recommend_non_numeric_primitives(make_es):
+    ecom_es_customers = EntitySet("ES without numerics")
```
thehomebrewnerd, line 87: nit: You don't need to supply an EntitySet name here

thehomebrewnerd (commented):
Just a few relatively minor things, but generally looking pretty good to me.

featuretools/utils/recommend_primitives.py
ozzieD: yeah I know after our discussions we wanted to add all of them. I'm sure after feature selection these may be dropped anyway.

featuretools/utils/recommend_primitives.py
ozzieD: ahh i see, I'll change that

featuretools/utils/recommend_primitives.py
ozzieD: In terms of the first comment, yeah this happened in `latlong_to_city` because of an issue with its reverse geocoder. I put this in place just in case there are primitives with issues now or in the future.

ozzieD: As for the second comment, done. It doe make it look cleaner. The time series stuff is a one liner so I didn't put it in it's own function. lmk what you think

thehomebrewnerd: I like it!

thehomebrewnerd (approved):

featuretools/utils/recommend_primitives.py
```diff
@@ -0,0 +1,242 @@
...
+    Description:
+        This function works by first getting a list of valid primitives withholding any primitives specified in `excluded_primitives` that could be applied to a single-table EntitySet.
+        Secondly, engineered features are created for non-numeric fields and are checked for non-uniqueness. If the feature is non-unique, it is added to the recommendation list.
+        Then, numeric fields are checked for skewness. Depending on how skew a column is `square_root` or `natural_logarithm` will be recommended.
+        Lastly if `include_time_series_primitives` is specified as `True`, `Lag` will always be recommended,
+        as well as all Rolling and Expanding primitives if numeric columns are present.
+
+    Args:
+        entityset (EntitySet): EntitySet that only contains one dataframe.
+        include_time_series_primitives (bool): Whether or not time-series primitives should be considered.
```
gsheni, line 88: Default value should be mentioned in docstring

ozzieD: I didn't have a default originally but I guess I'll set it to `False`

featuretools/utils/recommend_primitives.py
```diff
@@ -195,8 +196,9 @@ def _recommend_non_numeric_primitives(
                 for f_name in f.get_feature_names():
                     if len(matrix[f_name].unique()) > 1:
                         recommended_non_numeric_primitives.add(f.primitive.name)
-            except Exception:  # If error in calculating feature matrix pass on the recommendation
-                pass
+            except Exception as e:  # If error in calculating feature matrix pass on the recommendation
+                logger = logging.getLogger("featuretools")
+                logger.error(f"Exception in {f.primitive.name}: {str(e)}")
```
rwedge, line 201: What about a message like this?

"Exception with feature {f.get_name()} with primitive {f.primitive.name}: {str(e)}"

ozzieD: Changed

PR #2400, opened 2022-12-13
Refactor building groupby features if `require_direct_input`  

### LINKED ISSUE #2402: Reduce traversals over features in `_build_transform_features`
- Currently, when building `groupby` features, we iterate three times over the `input_features`.
https://github.com/alteryx/featuretools/blob/9ce731ee3a38a646732d1d9cd8a50d208d4ccd3b/featuretools/synthesis/deep_feature_synthesis.py#L764 

 I think this can be reduced to one. Here is an example of what we can reduce it to: 
```
if require_direct_input:
                            path = input_features[0].relationship_path
                            found_direct = False
                            same_paths = True
                            for feature in input_features:
                                if isinstance(feature, DirectFeature):
                                    found_direct = True
                                    if feature.relationship_path != path:
                                        same_paths = False
                                        break
                            if not found_direct or not same_paths:
                                continue
```
Some questions to ask: 
- Do we get any performance benefit? (I'm thinking it's unlikely) 
- Which one is cleaner and clearer to read? 
- Are there other areas of our code that make use of a similar pattern?

sbadithe:
- We were looping over the `input_features` twice because of the call to `_features_have_same_path`, but I think one traversal should do the trick
- Since the checks for being a DirectFeature and having the same path are so tightly coupled (this function is still doing exactly what it was before), I think it should be okay to move it into one function, but I'm open to ideas here
- Helps reduce complexity, and we can delete the `_features_have_same_path` function
- In `_build_transform_features`, removes function call and iterates once over the input_features. This reduces from three traversals to one (closes #2402)

### DIFF
--- a/featuretools/synthesis/deep_feature_synthesis.py
+++ b/featuretools/synthesis/deep_feature_synthesis.py
@@ -1272,6 +1272,7 @@ def _all_direct_and_same_path(input_features) -> bool:
             return False
     return True
 
+
 def _direct_of_dataframe(feature, parent_dataframe):
     return (
         isinstance(feature, DirectFeature)
--- a/featuretools/synthesis/deep_feature_synthesis.py
+++ b/featuretools/synthesis/deep_feature_synthesis.py
@@ -744,12 +744,12 @@ class DeepFeatureSynthesis(object):
                 any_direct_in_matching_input = None
                 all_direct_and_same_path_in_matching_input = None
                 if require_direct_input:
-                    any_direct_in_matching_input = any(
+                    if any_direct_in_matching_input := any(
                         isinstance(bf, DirectFeature) for bf in matching_input
-                    )
-                    all_direct_and_same_path_in_matching_input = (
-                        _all_direct_and_same_path(matching_input)
-                    )
+                    ):
+                        all_direct_and_same_path_in_matching_input = (
+                            _all_direct_and_same_path(matching_input)
+                        )
                 if any(True for bf in matching_input if bf.number_output_features != 1):
                     continue
                 for groupby in groupby_matches:
--- a/featuretools/synthesis/deep_feature_synthesis.py
+++ b/featuretools/synthesis/deep_feature_synthesis.py
@@ -8,7 +8,7 @@ from typing import Any, DefaultDict, Dict, List
 from woodwork.column_schema import ColumnSchema
 from woodwork.logical_types import Boolean, BooleanNullable
 
-from featuretools import primitives, FeatureBase
+from featuretools import primitives
 from featuretools.entityset.entityset import LTI_COLUMN_NAME
 from featuretools.entityset.relationship import RelationshipPath
 from featuretools.feature_base import (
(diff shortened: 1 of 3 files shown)


featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -1260,19 +1260,16 @@ def check_primitive(
...
+    """Given a list of features, returns True if they are all
+    DirectFeatures with the same relationship_path, and False if not
+    """
     path = input_features[0].relationship_path
-
-    for f in input_features[1:]:
+    for f in input_features:
+        if not isinstance(f, DirectFeature):
+            return False
         if f.relationship_path != path:
```
ozzieD, line 1271:
```suggestion
    for f in input_features:
        if not isinstance(f, DirectFeature) or f.relationship_path != path:
            return False
```

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -1260,19 +1260,16 @@ def check_primitive(
...
             return False
```
ozzieD, line 1272:
```suggestion
```suggestion
    for f in input_features:
        if not isinstance(f, DirectFeature) or f.relationship_path != path:
            return False
```
```

File: featuretools/synthesis/deep_feature_synthesis.py [hunk 3]
```diff
@@ -1260,19 +1260,16 @@ def check_primitive(
...
 
 
-def _all_direct_and_same_path(input_features):
-    return all(
-        isinstance(f, DirectFeature) for f in input_features
-    ) and _features_have_same_path(input_features)
-
-
-def _features_have_same_path(input_features):
+def _all_direct_and_same_path(input_features) -> bool:
```
ozzieD, line 1263: If were slowly adding typing to this repo can you also type out `input_features` as well?

featuretools/synthesis/deep_feature_synthesis.py
sbadithe: just realized we have another traversal in line 753, I think we can consolidate all of them

featuretools/synthesis/deep_feature_synthesis.py
sbadithe: Done! I'm not super confident in the signature though...

sbadithe: EDIT: I just typed it as list. If we type it as a list of FeatureBase, there are issues with circular imports.

ozzieD: Top of the file:
```
from typing import List, TYPE_CHECKING

if TYPE_CHECKING:
     from featuretools import FeatureBase
```

`def _all_direct_and_same_path(input_features: List["FeatureBase"]) -> bool:`

sbadithe: thanks!

sbadithe: looks like this gets flagged by codecov unfortunately

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -737,32 +738,49 @@ def _build_transform_features(
...
-                                for feature in (input_features)
-                                if isinstance(feature, DirectFeature)
-                            )
-                        ):
-                            continue
-                        new_f = GroupByTransformFeature(
-                            list(matching_input),
-                            groupby=groupby[0],
-                            primitive=groupby_prim,
+                if require_direct_input:
```
rwedge, line 744: this block should come after the  `if any(True for bf ...)` block that is just below it

rwedge (approved):
LGTM

PR #2404, opened 2022-12-14
Resolve Pyarrow Warning on Import

sbadithe:
- Resolves this warning: `WARNING:root:'PYARROW_IGNORE_TIMEZONE' environment variable was not set. It is required to set this environment variable to '1' in both driver and executor sides if you use pyarrow>=2.0.0. pandas-on-Spark will set it for you but it does not work if there is a Spark context already launched.
`
- Makes sure the user has not set this env variable, if they have, it doesn't touch it


rwedge:
Should we be silencing another library's warnings?

gsheni:
Yah agree with @rwedge . It's fine to do this in our CI or mention it in our contributing.md

PR #2408, opened 2022-12-16
Update RateOfChange input types to only accept Datetime time index

### LINKED ISSUE #2395: RateOfChange primitive does not work with a numeric time index
The `RateOfChange` primitive uses the pandas `.dt` accessor which only works with datetime columns. The primitive will currently accept a numeric time index column as an input though. When this happens, an error results.

We should either update the primitive to only work with a Datetime time index or to handle the numeric time index case.

```python
import pandas as pd
import featuretools as ft

df = pd.DataFrame({
    "id": [0, 1, 2, 3],
    "time_index": [1, 2, 3, 4],
    "values": [100, 200, 300, 400]
})

es = ft.EntitySet()
es.add_dataframe(
    dataframe=df,
    dataframe_name="df",
    index="id",
    time_index="time_index",
)

fm, features = ft.dfs(entityset=es, target_dataframe_name="df", trans_primitives=["rate_of_change"])
```

```
AttributeError("Can only use .dt accessor with datetimelike values")
```

thehomebrewnerd:
### Update RateOfChange input types to only accept Datetime time index

Fixes #2395

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,7 +7,8 @@ Future Release
 ==============
     * Enhancements
     * Fixes
-        * Fix typo in ``_handle_binary_comparison`` function name and update ``set_feature_names`` docstring (:pr:`2388`) 
+        * Fix typo in ``_handle_binary_comparison`` function name and update ``set_feature_names`` docstring (:pr:`2388`)
+        * Only allow Datetime time index as input to ``RateOfChange`` primitive (:pr:`2408`)
     * Changes
         * Iterate only once over ``ignore_columns`` in ``DeepFeatureSynthesis`` init (:pr:`2397`)
     * Documentation Changes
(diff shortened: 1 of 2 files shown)


sbadithe (approved):
Do we need to add a test that numeric time indexes don't work?

ozzieD (approved):

thehomebrewnerd:
> Do we need to add a test that numeric time indexes don't work?

I think that should already be covered by our features tests - we know DFS doesn't build features with input types that are not specified, so adding a specific test here seems unnecessary. The inclusion of all time index types in the primitive definition before was just an oversight.

sbadithe:
> > Do we need to add a test that numeric time indexes don't work?
> 
> I think that should already be covered by our features tests - we know DFS doesn't build features with input types that are not specified, so adding a specific test here seems unnecessary. The inclusion of all time index types in the primitive definition before was just an oversight.

Makes sense to me!

PR #2409, opened 2022-12-16
Update init calls in add_last_time_indexes to use init_with_parital_schema

### LINKED ISSUE #1463: Clean up `EntitySet.add_last_time_indexes` by initializing Woodwork with partial schema
Once Woodwork implements [this issue]( https://github.com/alteryx/woodwork/issues/965), we can clean up the Woodwork initialization in `add_last_time_indexes` to pass in the previous dataframe's table schema to keep that typing information but also perform inference on the new last time index column.

thehomebrewnerd:
### Update init calls in add_last_time_indexes to use init_with_parital_schema

Fixes #1463 

Note: The current Woodwork `init` call simply calls `init_with_partial_schema` and passes along all the specified kwargs, so this change does not alter any of the existing behavior, but just makes it more clear that we are reusing the existing schema information by calling `init_with_partial_schema`

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -7,15 +7,16 @@ Future Release
 ==============
     * Enhancements
     * Fixes
-        * Fix typo in ``_handle_binary_comparison`` function name and update ``set_feature_names`` docstring (:pr:`2388`) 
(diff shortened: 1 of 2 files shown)


tamargrey (approved):
Looks good! Only thing I can think to add might be a test that mocks out the woodwork init and confirms it's not called, but given how minor of a change this is, that's probably overkill.

ozzieD (approved):

Issue #2411, opened 2022-12-17
NumberOfWordsInQuotes freezes on twitter dataset


ozzieD:
NumberOfWordsInQuotes completely freezes up on this dataset 
[twitter_training.csv](https://github.com/alteryx/featuretools/files/10252617/twitter_training.csv)

```python
import featuretools as ft
from featuretools.primitives import NumberOfWordsInQuotes
df = pd.read_csv("./twitter_training.csv")
es = ft.EntitySet()
es.add_dataframe(
    dataframe=df,
    dataframe_name="df",
    index="id",
    make_index=True,
)

prim_instance = NumberOfWordsInQuotes()
get_fun = prim_instance.get_function()
answer = get_fun(es["df"]["text"])
```

PR #2413, opened 2022-12-19
Fix backtracking in NumberOfWordsInQuotes primitive

sbadithe:
- fixes catastrophic backtracking issue related to #2411
- adds the line that triggered catastrophic backtracking as test case

### DIFF
--- a/featuretools/tests/primitive_tests/natural_language_primitives_tests/test_number_of_words_in_quotes.py
+++ b/featuretools/tests/primitive_tests/natural_language_primitives_tests/test_number_of_words_in_quotes.py
@@ -26,6 +26,7 @@ class TestNumberOfWordsInQuotes(PrimitiveT):
                 '"user@email.com"',
                 '"https://alteryx.com"',
                 '"100,000"',
+                '""Word Word Word Word"" is the perfect conclusion to the ""Borderlands 3"" line, which focuses on the fans ""favorite character and gives the players the opportunity to close for a long time some very important questions about\'s character and the memorable scenery with which the players interact."',
             ],
         )
         expected = pd.Series([0, 5, 1, 0, 3, 6, 3, 1, 1, 1], dtype="Int64")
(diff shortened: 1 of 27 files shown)


featuretools/tests/primitive_tests/natural_language_primitives_tests/test_number_of_words_in_quotes.py
```diff
@@ -26,9 +26,10 @@ def test_regular_double_quotes_input(self):
                 '"user@email.com"',
                 '"https://alteryx.com"',
                 '"100,000"',
+                '"This Borderlands game here"" is the perfect conclusion to the ""Borderlands 3"" line, which focuses on the fans ""favorite character and gives the players the opportunity to close for a long time some very important questions about\'s character and the memorable scenery with which the players interact.',
             ],
         )
-        expected = pd.Series([0, 5, 1, 0, 3, 6, 3, 1, 1, 1], dtype="Int64")
+        expected = pd.Series([0, 5, 1, 0, 3, 6, 3, 1, 1, 1, 6], dtype="Int64")
```
gsheni, line 32: Can we add this string to other NaturalLauguage primitive tests?

sbadithe: We could, but I am concerned that what made this string an edge case for this primitive, may not make it an edge case for others. I think one possible longterm solution would be to have a single corpus (something like the twitter dataset), that we run all our applicable NLP primitives on. This would be a test not for correctness (which the individual test files could check) but that the primitives avoid failure. Basically programmatically doing what @ozzieD did when he ran the tests. We can exclude the tests from our regular test suite if efficiency is a concern.

gsheni (approved):

featuretools/tests/primitive_tests/natural_language_primitives_tests/test_number_of_words_in_quotes.py
sbadithe: how about making a test that imports all NatLang primitives and runs them on a fixture of strings we think the primitives could possibly hang or fail on (like all whitespace, or this string). Anytime we encounter an edge-case string, we could add it to the fixture. This would give us the confidence that the primitives won't hang. 


@gsheni @thehomebrewnerd

sbadithe: Sorry if this is kind of confusing, I can put up a POC in a different PR

gsheni: That sounds like a good idea and what I had in mind.

PR #2414, opened 2022-12-19
Fix for numpy 1.24.0

thehomebrewnerd:
### Fix for numpy 1.24.0

Use `"object"` string instead of `np.object` which has been removed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -75,7 +75,7 @@ test = [
 spark = [
     "woodwork[spark] >= 0.18.0",
     "pyspark >= 3.2.2",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -75,6 +75,7 @@ test = [
 spark = [
     "woodwork[spark] >= 0.18.0",
     "pyspark >= 3.2.2",
+    "numpy < 1.24.0",
```
thehomebrewnerd, line 78: I will create a new issue to remove this once pyspark is compatible

gsheni (approved):

PR #2418, opened 2022-12-19
Add TimeSinceLastFalse, TimeSinceLastMax, TimeSinceLastMin, TimeSinceLastTrue aggregation primitives

gsheni:


featuretools/primitives/standard/aggregation/time_since_last_true.py
```diff
@@ -0,0 +1,57 @@
...
+        ...          datetime(2010, 1, 1, 11, 57, 30)]
+        >>> booleans = [True, True, False]
+        >>> time_since_last_true(times, booleans, time=cutoff_time)
+        285.0
+    """
+
+    name = "time_since_last_true"
+    input_types = [
+        ColumnSchema(logical_type=Datetime, semantic_tags={"time_index"}),
+        ColumnSchema(logical_type=Boolean),
```
thehomebrewnerd, line 32: Not sure if this should be done in a separate PR, but maybe want to update this and TimeSinceLastFalse to also work with BooleanNullable inputs?

featuretools/primitives/standard/aggregation/time_since_last_true.py
```diff
@@ -0,0 +1,57 @@
...
+    def get_function(self):
+        def time_since_last_true(datetime_col, bool_col, time=None):
+            df = pd.DataFrame(
+                {
+                    "datetime": datetime_col,
+                    "bool": bool_col,
+                },
+            ).dropna()
+            if df.empty:
+                return np.nan
```
thehomebrewnerd, line 48: Not sure if it is critical, but should we add a test case to cover this line and the corresponding uncovered line in TimeSinceLastFalse?

gsheni: Added e6d22c1b46206ed53a3f

featuretools/primitives/standard/aggregation/time_since_last_false.py
```diff
@@ -0,0 +1,63 @@
...
+        def time_since_last_false(datetime_col, bool_col, time=None):
+            df = pd.DataFrame(
+                {
+                    "datetime": datetime_col,
+                    "bool": bool_col,
+                },
+            ).dropna()
+            if df.empty:
+                return np.nan
+            df["bool"] = df["bool"].astype("bool")
```
thehomebrewnerd, line 55: I think this line will cause problems with `BooleanNullable` inputs. However, I"m not real sure this is even necessary because the boolean column should always come in as either `bool` or `boolean` since we expect WW to be initialized on the series.

gsheni: Removed https://github.com/alteryx/featuretools/pull/2418/commits/71390a33ab03f681a5cd4e8e34397e2a12ae909e

thehomebrewnerd (approved):

PR #2423, opened 2022-12-20
Fix delimiter inconsistency in Natural Language primitives

### LINKED ISSUE #2419: Standardize Regular Expressions for Natural Language Primitives
Multiple primitives make use of regular expressions. Some of them define what punctuation to delimit on. We should standardize these. This would give users the confidence that Primitive A does not consider a string to be one word, while Primitive B considers it to be two. We could define the regexes in a common file and import them in the primitives.

This was originally an issue on `nlp-primitives` before the primitives were moved here.

sbadithe:
closes #2419 
closes alteryx/featuretools#2425 
closes alteryx/featuretools#2426

Changes: 

After this change, the default delimiters for all primitives are only whitespace, unless manually overrided. This means primitives will leave words like "alteryx.com" and "1,000" as one word, unless explicitly told not to. 

This results in adjusting test cases for some primitives. Furthermore, it results in a change to NumberOfCommonWords where words are split on whitespace and then stripped of punctuation before being checked for membership in the common word set.


featuretools/primitives/standard/transform/natural_language/number_of_common_words.py
```diff
@@ -46,9 +49,11 @@ class NumberOfCommonWords(TransformPrimitive):
 
     def __init__(
         self,
-        word_set=set(common_words_1000),
```
sbadithe, line 49: Having a mutable default parameter is unsafe

featuretools/primitives/standard/transform/natural_language/median_word_length.py
```diff
@@ -16,13 +19,13 @@ class MedianWordLength(TransformPrimitive):
 
     Args:
         delimiters_regex (str): Delimiters as a regex string for splitting text into words.
-            The default delimiters include "- [].,!?;\\n".
+            The default delimiters include [-.!?]\n\t
 
     Examples:
         >>> x = ['This is a test file', 'This is second line', 'third line $1,000', None]
```
sbadithe, line 25: this test was breaking up "$1,000" into "$1" and "000". After this change, it stays as one word

featuretools/primitives/standard/transform/natural_language/number_of_common_words.py
sbadithe: Leaving this alone for now because it triggers serialization test failures, but I think we should fix in future

featuretools/primitives/standard/transform/natural_language/constants.py
```diff
@@ -1,4 +1,11 @@
-common_words_1000 = [
+from string import punctuation
+
+DELIMITERS = " \n\t"
+DELIMITERS = f"[{DELIMITERS}]"
```
rwedge, line 4: two assignments seems unnecessary

featuretools/primitives/standard/transform/natural_language/total_word_length.py
```diff
@@ -30,12 +33,11 @@ class TotalWordLength(TransformPrimitive):
 
     default_value = 0
 
-    def __init__(self, delimiters_regex=r"[- \[\].,!\?;\n]"):
-        self.delimiters_regex = delimiters_regex
+    def __init__(self, do_not_count=PUNCTUATION_AND_WHITESPACE):
```
rwedge, line 36: I think changing the parameter name means we'll need to bump the feature schema since old copies of this primitive that have a custom regex will no longer work.

rwedge (commented):
For `NumberOfUniqueWords` should we test for `\t` somewhere?

featuretools/primitives/standard/transform/natural_language/total_word_length.py
sbadithe: Oops, I meant to revert this change. I changed the behavior of this primitive to strip punctuation, and to split on the delimiters, so I think we can keep the parameter name.

docs/source/release_notes.rst
```diff
@@ -3,15 +3,23 @@
 Release Notes
 -------------
 
+Breaking Changes
+++++++++++++++++
+* The `delimiter_regex` parameter for ``TotalWordLength`` has been renamed to `do_not_count`.
+  Old features that had a non-default value for the parameter will no longer load.
```
rwedge, line 9:
```suggestion
  Old saved features that had a non-default value for the parameter will no longer load.
```

docs/source/release_notes.rst
```diff
@@ -3,15 +3,23 @@
 Release Notes
 -------------
 
+Breaking Changes
```
rwedge, line 6: the Breaking Changes section should go between the Future Release section and the most recent release

docs/source/release_notes.rst
```diff
@@ -3,15 +3,23 @@
...
+++++++++++++++++
+* The `delimiter_regex` parameter for ``TotalWordLength`` has been renamed to `do_not_count`.
```
rwedge, line 8: Let's put this line in the `Changes` section as well to track PR number

docs/source/release_notes.rst
```diff
@@ -3,15 +3,23 @@
...
+  Old features that had a non-default value for the parameter will no longer load.
+
+
 Future Release
 ==============
     * Enhancements
         * Add ``TimeSinceLastFalse``, ``TimeSinceLastMax``, ``TimeSinceLastMin``, and ``TimeSinceLastTrue`` primitives (:pr:`2418`)
         * Add ``MaxConsecutiveFalse``, ``MaxConsecutiveNegatives``, ``MaxConsecutivePositives``, ``MaxConsecutiveTrue``, ``MaxConsecutiveZeros``, ``NumConsecutiveGreaterMean``, ``NumConsecutiveLessMean`` (:pr:`2420`)
+        * Improve edge case handling in NaturalLanguage primitives by refactoring delimiter (:pr:`2423`)
```
rwedge, line 17: Let's remove this in favor of the fixes entry with the same wording

PR #2424, opened 2022-12-22
Fix fragmentation PerformanceWarning in feature_set_calculator.py

### LINKED ISSUE #2405: [refactor] PerformanceWarning: DataFrame is highly fragmented during DFS
During the Deep Feature Synthesis python outputs the following warning:
> python3.9/site-packages/featuretools/computational_backends/feature_set_calculator.py:819: **PerformanceWarning:** DataFrame is highly fragmented.  This is usually the result of calling `frame.insert` many times, which has poor performance.  Consider joining all columns at once using pd.concat(axis=1) instead. To get a de-fragmented frame, use `newframe = frame.copy()`
  return data.assign(**new_cols)
```

```python
es = ft.EntitySet("test_data")

es.entity_from_dataframe(dataframe=test_df, entity_id = "test_", index = "id")

# Run deep feature synthesis with transformation primitives
feature_matrix, feature_defs = ft.dfs(entityset = es,
                                      trans_primitives = ['add_numeric', 'multiply_numeric']
                                     ,target_entity="test_"
                                     )

```

<details>

Featuretools version: 0.27.1

SYSTEM INFO
-----------
python: 3.9.15.final.0
python-bits: 64
OS: Darwin
OS-release: 21.6.0
machine: x86_64
processor: i386
byteorder: little
LC_ALL: None
LANG: None
LOCALE: None.UTF-8

INSTALLED VERSIONS
------------------
numpy: 1.23.4
pandas: 1.4.2
tqdm: 4.64.1
PyYAML: 6.0
cloudpickle: 2.0.0
dask: 2022.12.0
distributed: 2022.12.0
psutil: 5.9.0
pip: 22.2.2
setuptools: 65.5.0

</details>

thehomebrewnerd:
Fixes #2405 

- Fix fragmentation PerformanceWarning in feature_set_calculator.py
- Fix warning related to deprecation of `pandas.util`
- Fix warning related to setting as type with `"datetime"` without units


featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -936,7 +939,16 @@ def update_feature_columns(feature_data, data):
...
+    if isinstance(data, pd.DataFrame):
+        return pd.concat([data, pd.DataFrame(new_cols, index=data.index)], axis=1)
+    else:
+        for name, col in new_cols.items():
+            col.name = name
+            if isinstance(data, dd.DataFrame):
+                data = dd.concat([data, col], axis=1)
+            else:
+                data = ps.concat([data, col], axis=1)
+        return data
```
thehomebrewnerd, line 951: I don't really like this approach for Dask/Spark, but we can't create a new dataframe from the columns here like we can for pandas, so we have to concat the columns one at a time. Maybe there is a better way, but I didn't come up with one off the top of my head.

featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -936,7 +939,16 @@ def update_feature_columns(feature_data, data):
         data.update(new_cols)
         return data
 
-    return data.assign(**new_cols)
+    if isinstance(data, pd.DataFrame):
+        return pd.concat([data, pd.DataFrame(new_cols, index=data.index)], axis=1)
+    else:
```
sbadithe, line 944: If we return in the `if-block`, do we need the `else-block`? It would save us an indent if we remove

featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -917,11 +925,6 @@ def wrap(df):
     return wrap
 
 
-def set_default_column(frame, f):
```
sbadithe, line 920: Just curious, why did we delete this function?

featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -936,7 +939,16 @@ def update_feature_columns(feature_data, data):
...
 
-    return data.assign(**new_cols)
+    if isinstance(data, pd.DataFrame):
+        return pd.concat([data, pd.DataFrame(new_cols, index=data.index)], axis=1)
+    else:
+        for name, col in new_cols.items():
+            col.name = name
+            if isinstance(data, dd.DataFrame):
+                data = dd.concat([data, col], axis=1)
+            else:
```
sbadithe, line 949: Should we have an explicit check to see if `ps` is defined and if `isinstance(data, ps.DataFrame)`?

The case I am thinking about is, if data is a `pd.Series` for example, it would fall into this else block.

featuretools/computational_backends/feature_set_calculator.py
sbadithe, line 944: Styling question: Should we have an `else` block here?

featuretools/primitives/standard/transform/binary/greater_than.py
```diff
@@ -40,11 +41,11 @@ def greater_than(val1, val2):
             val2_is_categorical = pdtypes.is_categorical_dtype(val2)
             if val1_is_categorical and val2_is_categorical:
                 if not all(val1.cat.categories == val2.cat.categories):
-                    return np.nan
+                    return val1.where(pd.isnull, np.nan)
```
sbadithe, line 44: Curious, what is this change for?

featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -936,7 +939,16 @@ def update_feature_columns(feature_data, data):
         data.update(new_cols)
         return data
 
-    return data.assign(**new_cols)
```
dvreed77, line 939: I assume this assign is doing the inserts under the hood that the warning originates from?

thehomebrewnerd: Yeah, correct

featuretools/computational_backends/feature_set_calculator.py
thehomebrewnerd: Doesn't seem necessary. I will remove.

featuretools/computational_backends/feature_set_calculator.py
thehomebrewnerd: @rwedge Since this was a change you made before I picked this up, do you recall your thinking here?

rwedge: @thehomebrewnerd @sbadithe this code snippet is no longer used multiple times and is only two lines so I put it back in the function that still uses it

featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -507,10 +507,10 @@ def _calculate_transform_features(
         for f in features:
             # handle when no data
             if frame_empty:
-                set_default_column(frame, f)
-
+                feature_values.append(
+                    (f, [[] for _ in range(f.number_output_features)]),
```
rwedge, line 511:
```suggestion
                    (f, [f.default_value for _ in range(f.number_output_features)]),
```

I think this would be better default handling

featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -932,11 +935,22 @@ def update_feature_columns(feature_data, data):
             new_cols[name] = value
 
     # Handle the case where a dict is being updated
-    if isinstance(data, dict):
-        data.update(new_cols)
-        return data
+    if isinstance(df, dict):
```
ozzieD, line 938: is it a common occurrence where `df` is a `dict`.  Asking cause it feels weird to name this `df` when there are cases where it's not actually a DataFrame.

thehomebrewnerd: That's fair. I didn't think through this case when I renamed previously. I will switch this back to `data`.

ozzieD (approved):

gsheni (approved):

PR #2429, opened 2022-12-23
Add global test for NaturalLanguage primitives

sbadithe:
- As per discussion in #2413, this adds a test that checks strings that have caused errors in the past, ~~as well as randomized test inputs~~ to make sure none of our NaturalLanguage primitives hang or fail on generated input.
- Adds `get_natural_language_primitives` and refactors `get_transform_primitives` and `get_aggregation_primitives` utility functions


docs/source/release_notes.rst
```diff
@@ -3,15 +3,17 @@
...
-.. Future Release
-  ==============
+Future Release
+==============
     * Enhancements
     * Fixes
     * Changes
     * Documentation Changes
     * Testing Changes
+        * Add global NatLang test with problematic strings (:pr:`2429`)
```
gsheni, line 13:
```suggestion
        * Add global Natural Language test with problematic strings (:pr:`2429`)
```

ozzieD:
I wonder why code cov is looking through test files?

featuretools/tests/primitive_tests/natural_language_primitives_tests/test_natural_language_perf.py
```diff
@@ -0,0 +1,62 @@
...
+    PunctuationCount,
+    TitleWordCount,
+    TotalWordLength,
+    UpperCaseCount,
+    WhitespaceCount,
+)
+
+TIMEOUT_THRESHOLD = 20
+
+if system == "Linux" or system == "Darwin":
```
ozzieD, line 25: Do Windows users skip these tests?

rwedge: I think the regular CI runs are skipping these tests as well.  The "unit tests core" job is reporting the same number of tests on main and this PR

rwedge:
> I wonder why code cov is looking through test files?

Codecov is configured to check our test files as well

featuretools/tests/primitive_tests/natural_language_primitives_tests/test_natural_language_primitives_terminate.py
```diff
@@ -0,0 +1,64 @@
...
+    PunctuationCount,
+    TitleWordCount,
+    TotalWordLength,
+    UpperCaseCount,
+    WhitespaceCount,
+)
+
+TIMEOUT_THRESHOLD = 20
+
+if system == "Linux" or system == "Darwin":
```
rwedge, line 25: Check out pytest's skipif: https://docs.pytest.org/en/6.2.x/reference.html?highlight=pytest%20skip#pytest-mark-skipif-ref

Also I think given these codecov warnings, `os.sytems` isn't "Linux" for our standard CI runs

featuretools/tests/primitive_tests/natural_language_primitives_tests/test_natural_language_perf.py
sbadithe: @ozzieD Yes. This is because `SIGALRM` is only available on UNIX systems.

featuretools/primitives/utils.py
```diff
@@ -17,28 +18,47 @@
...
-                if attribute.name:
-                    transform_primitives.add(attribute)
-    return {prim.name.lower(): prim for prim in transform_primitives}
+    """Returns all transform primitives, regardless
+    of compatibility
+    """
+    return _get_primitives(featuretools.primitives.TransformPrimitive)
+
+
+def get_natural_language_primitives():
```
thehomebrewnerd, line 48: I'm wondering about the need to create this new public method? If the only reason we are doing this is for running a single test, maybe this should just be part of the test, or perhaps a specific test fixture instead?

gsheni: I can see it being useful if someone wanted to feed all NL primitives to DFS.

thehomebrewnerd: Sure, but that seems beyond the scope of this MR.

My current thoughts:
1. Let's not complicate our API with new methods unless we have a specific need for the functionality
2. If we do add this method for the purpose above, we should update our API docs to show we added this.
3. If we add a method like this, should it be specific to NaturalLanguage or should it be something more generic? Should it match on only logical types or should it match on `ColumnSchema` which really are what fully define the type? I think we should answer those questions before we implement a new public method.

featuretools/tests/primitive_tests/natural_language_primitives_tests/test_natural_language_primitives_terminate.py
```diff
@@ -0,0 +1,39 @@
...
+import pandas as pd
+import pytest
+
+from featuretools.primitives.utils import get_natural_language_primitives
+
+TIMEOUT_THRESHOLD = 20
+
+
+@pytest.mark.skipif(
+    system() == "Linux" or system() == "Darwin",
```
rwedge, line 13: this will skip on Linux or Darwin

featuretools/tests/primitive_tests/natural_language_primitives_tests/test_natural_language_primitives_terminate.py
```diff
@@ -0,0 +1,39 @@
...
+            f"NaturalLanguage primitive took longer than {TIMEOUT_THRESHOLD}"
+        )
+
+    signal.signal(signal.SIGALRM, handle_SIGALRM)
+    for text in strings_that_have_triggered_errors_before:
+        signal.alarm(TIMEOUT_THRESHOLD)
+        try:
+            primitive().get_function()(pd.Series(text))
+        except TimeoutError:
+            raise TimeoutError(
```
rwedge, line 34: if we're going to raise the error anyway, does this need to be a try/except block?

sbadithe:
> @sbadithe Did you look at `pytest-timeout` as an option here? Could we use this instead of trying to implement this manually? The downside is that we would need to add another testing dependency.
> 
> https://github.com/pytest-dev/pytest-timeout

Is it ok if it hasn't been updated since January 2022 on pypi?

thehomebrewnerd:
> Is it ok if it hasn't been updated since January 2022 on pypi?

@sbadithe In my opinion, yes. This is a simple library, so I wouldn't expect many updates. Also, this is a test only dependency, so we can always replace in the future if needed and it won't break FT. Finally, it is maintained by the pytest folks. All those factors combined make it a low risk add IMHO.

gsheni (approved):

gsheni (approved):

PR #2459, opened 2023-01-21
Fix bug with `NumWords`; add test suite

### LINKED ISSUE #2457: `NumWords` returns wrong answer when text with multiple spaces is passed in
```
 NumWords().get_function()(pd.Series(["hello   world"]))
```
Returns 4. Adding another space would return 5. 

The issue is with how the number of words is counted. Consecutive spaces should be collapsed into one.

sbadithe:
- Fix #2457 
- Add missing test suite for `NumWords`

### DIFF
--- a/featuretools/primitives/standard/transform/natural_language/num_words.py
+++ b/featuretools/primitives/standard/transform/natural_language/num_words.py
@@ -1,3 +1,4 @@
+import re
 from string import punctuation
 from typing import Iterable
 
@@ -13,7 +14,8 @@ from featuretools.utils.gen_utils import Library
 
 
 class NumWords(TransformPrimitive):
-    """Determines the number of words in a string.
+    """Determines the number of words in a string. Words are sequences of characters
+    delimited by whitespace.
 
     Examples:
         >>> num_words = NumWords()
@@ -38,7 +40,7 @@ class NumWords(TransformPrimitive):
                 if not isinstance(elem, Iterable):
                     return pd.NA
                 return sum(
-                    1 for word in elem.split(DELIMITERS) if word.strip(punctuation)
+                    1 for word in re.split(DELIMITERS, elem) if word.strip(punctuation)
                 )
 
             return array.apply(_get_number_of_words)
--- a/featuretools/primitives/standard/transform/natural_language/num_words.py
+++ b/featuretools/primitives/standard/transform/natural_language/num_words.py
@@ -1,6 +1,6 @@
 import re
 from string import punctuation
-from typing import Iterable
+from typing import Optional
 
 import pandas as pd
 from woodwork.column_schema import ColumnSchema
@@ -34,10 +34,10 @@ class NumWords(TransformPrimitive):
 
     def get_function(self):
         def word_counter(array):
-            def _get_number_of_words(elem: pd.Series):
+            def _get_number_of_words(elem: Optional[str]):
                 """Returns the number of words in given element,
-                or null given non-iterable input"""
-                if not isinstance(elem, Iterable):
+                or pd.NA given null input"""
+                if pd.isna(elem):
                     return pd.NA
                 return sum(
                     1 for word in re.split(DELIMITERS, elem) if word.strip(punctuation)
--- a/featuretools/primitives/standard/transform/natural_language/num_words.py
+++ b/featuretools/primitives/standard/transform/natural_language/num_words.py
@@ -6,8 +6,11 @@ from woodwork.column_schema import ColumnSchema
 from woodwork.logical_types import NaturalLanguage
 
 from featuretools.primitives.base import TransformPrimitive
+from featuretools.primitives.standard.transform.natural_language.constants import (
+    DELIMITERS,
+)
 from featuretools.utils.gen_utils import Library
-from featuretools.primitives.standard.transform.natural_language.constants import DELIMITERS
+
 
 class NumWords(TransformPrimitive):
     """Determines the number of words in a string by counting the spaces.
@@ -35,9 +38,6 @@ class NumWords(TransformPrimitive):
                     return pd.NA
                 return sum(1 for word in elem if len(word.strip(punctuation)))
 
-            return array\
-                .str\
-                .split(DELIMITERS)\
-                .apply(_get_number_of_words)
+            return array.str.split(DELIMITERS).apply(_get_number_of_words)
 
         return word_counter
--- a/featuretools/primitives/standard/transform/natural_language/num_words.py
+++ b/featuretools/primitives/standard/transform/natural_language/num_words.py
@@ -3,7 +3,7 @@ from typing import Iterable
 
 import pandas as pd
 from woodwork.column_schema import ColumnSchema
-from woodwork.logical_types import NaturalLanguage
+from woodwork.logical_types import IntegerNullable, NaturalLanguage
 
 from featuretools.primitives.base import TransformPrimitive
 from featuretools.primitives.standard.transform.natural_language.constants import (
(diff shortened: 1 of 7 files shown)


featuretools/primitives/standard/transform/natural_language/num_words.py
```diff
@@ -25,6 +32,12 @@ class NumWords(TransformPrimitive):
 
     def get_function(self):
         def word_counter(array):
-            return array.fillna("").str.count(" ") + 1
+            def _get_number_of_words(elem: pd.Series):
+                """Returns the number of words or null given non-iterable input"""
+                if not isinstance(elem, Iterable):
```
sbadithe, line 37: One question we need to decide is what to return give null input. Prior to this change, we returned 0 given null input. If we choose to return null, we should update `NumCharacters` as well to keep the primitives uniform.

featuretools/primitives/standard/transform/natural_language/num_words.py
```diff
@@ -19,12 +26,18 @@ class NumWords(TransformPrimitive):
...
+                    return pd.NA
+                return sum(1 for word in elem if len(word.strip(punctuation)))
+
+            return array.str.split(DELIMITERS).apply(_get_number_of_words)
```
thehomebrewnerd, line 41: I think we are sort of dividing this primitive logic up between the `word_counter` function and the `_get_number_of_words` function by splitting on `DELIMTERS` before calling `.apply`.

Can we refactor this to call `.apply` on the original series, and add the logic for splitting on `DELIMITERS` to the _get_number_of_words` function instead?

Maybe something like this:
```
    def get_function(self):
        def word_counter(array):
            def _get_number_of_words(elem):
                """Returns the number of words or null given non-iterable input"""
                if not isinstance(elem, str):
                    return pd.NA
                return sum(1 for word in re.split(DELIMITERS, elem) if word.strip(punctuation))

            return array.apply(_get_number_of_words)

        return word_counter
```

featuretools/primitives/standard/transform/natural_language/num_words.py
```diff
@@ -19,12 +26,18 @@ class NumWords(TransformPrimitive):
...
     description_template = "the number of words in {}"
 
     def get_function(self):
         def word_counter(array):
-            return array.fillna("").str.count(" ") + 1
+            def _get_number_of_words(elem: pd.Series):
+                """Returns the number of words or null given non-iterable input"""
+                if not isinstance(elem, Iterable):
+                    return pd.NA
+                return sum(1 for word in elem if len(word.strip(punctuation)))
```
thehomebrewnerd, line 39: Do you need the `len()` check here or is it enough to check on the truthy value of a string?

featuretools/tests/primitive_tests/natural_language_primitives_tests/test_num_words.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class TestNumberOfUniqueWords(PrimitiveT):
+    primitive = NumWords
+
+    def test_general(self):
+        x = pd.Series(
+            [
+                "test test test test",
+                "test TEST test TEST,test test test",
```
thehomebrewnerd, line 19: is `TEST,test` considered one word here? Is that really what we want?

featuretools/tests/primitive_tests/natural_language_primitives_tests/test_num_words.py
```diff
@@ -0,0 +1,76 @@
...
+                "and subsequent lines...",
+            ],
+        )
+        expected = pd.Series([4, 6, 3])
+        actual = self.primitive().get_function()(x)
+        pd.testing.assert_series_equal(actual, expected, check_names=False)
+
+    def test_special_characters_and_whitespace(self):
+        x = pd.Series(["50% 50 50% \t\t\t\n\n", "a test* test"])
```
thehomebrewnerd, line 28: What about numbers like `$5,031.40` is that one word? I think maybe we should test for that type of scenario as well.

featuretools/tests/primitive_tests/natural_language_primitives_tests/test_num_words.py
sbadithe: Hmm, I was under the impression that we decided not to split on punctuation in the discussion for an earlier primitive.

thehomebrewnerd: Perhaps. `test,test` seems like two words to me though since I don't know of any words that contain a comma. It's probably hard to cleanly handle 100% of the cases though. As long as our docstring clearly indicates that we don't consider punctuation as a separator between words, only spaces, I think we can leave this as it is.

featuretools/primitives/standard/transform/natural_language/num_words.py
```diff
@@ -19,12 +28,21 @@ class NumWords(TransformPrimitive):
...
     input_types = [ColumnSchema(logical_type=NaturalLanguage)]
-    return_type = ColumnSchema(semantic_tags={"numeric"})
+    return_type = ColumnSchema(logical_type=IntegerNullable, semantic_tags={"numeric"})
     compatibility = [Library.PANDAS, Library.DASK, Library.SPARK]
     description_template = "the number of words in {}"
 
     def get_function(self):
         def word_counter(array):
-            return array.fillna("").str.count(" ") + 1
+            def _get_number_of_words(elem: pd.Series):
```
thehomebrewnerd, line 37: `elem` should either be a string or `pd.NA`, but not a `pd.Series`.

featuretools/primitives/standard/transform/natural_language/num_words.py
```diff
@@ -19,12 +28,21 @@ class NumWords(TransformPrimitive):
...
+                """Returns the number of words in given element,
+                or null given non-iterable input"""
+                if not isinstance(elem, Iterable):
```
thehomebrewnerd, line 40: Can't we just check if `elem` is of type `str` here instead without importing Iterable? Since the input of this primitive is a Natural Language column with a `string` dtype, I think the only possible options here are `str` or `pd.NA`. If you change this, please also update the docstring accordingly.

featuretools/tests/primitive_tests/test_transform_features.py
```diff
@@ -1120,7 +1120,7 @@ def test_text_primitives(es):
         sort_index=True,
     )
 
-    word_counts = [514, 3, 3, 644, 1268, 1269, 177, 172, 79, 240, 1239, 3, 3, 3, 3]
+    word_counts = [532, 3, 3, 653, 1306, 1305, 174, 173, 79, 246, 1253, 3, 3, 3, 3]
```
thehomebrewnerd, line 1123: One final question. In the original issue it says that adding whitespace characters with the old approach would cause counts to increase, which was not accurate. In this test, we are seeing increases in counts. Do we understand why this is and are these correct? Do we expect word counts to increase with this change?

sbadithe: I believe this to be a consequence of how we were handling tabs and newlines. In the old way, we counted the whitespace " " to infer the number of words. However, this would lead to the string ```soft drinks:\n\nDiet Coke``` (taken from the entityset) being inferred as 3 words. Our current implementation splits the string, and would count the above as having 4 words.

thehomebrewnerd: Got it. So now because we also split on tabs and new line characters we end up with more words, even though we now properly handle multiple spaces in a row.

thehomebrewnerd (approved):

featuretools/tests/primitive_tests/test_transform_features.py
sbadithe: Yep, to summarize for future reference, these changes 
(1) decrease our count because we stop recognizing words like "hello   world" as having 4 words simply because it has 3 spaces
(2) adds to our count because we now split on tabs and newlines, which we weren't considering previously 
I think this explains why some of our tests have increased number of words, and some have decreased number of words, and this seems to hold up under manual inspection.

PR #2463, opened 2023-01-25
Add test suite for`NumCharacters`and modify docstring

### LINKED ISSUE #2462: `NumCharacters` does not have test suite
We should add a test suite for the `NumCharacters` primitive

sbadithe:
- Fixes #2462 
- Refactors ``NumCharacters`` to return null given null input

### DIFF
--- a/featuretools/selection/selection.py
+++ b/featuretools/selection/selection.py
@@ -170,16 +170,20 @@ def remove_highly_correlated_features(
         more_complex_name = columns_to_check[i]
         more_complex_col = fm_to_check[more_complex_name]
 
-        # Convert boolean column to be float64
-        if pd.api.types.is_bool_dtype(more_complex_col):
+        # Convert boolean or Int64 column to be float64
+        if pd.api.types.is_bool_dtype(more_complex_col) or isinstance(
+            more_complex_col.dtype, pd.Int64Dtype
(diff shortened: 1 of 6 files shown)


featuretools/selection/selection.py
```diff
@@ -170,16 +170,20 @@ def remove_highly_correlated_features(
         more_complex_name = columns_to_check[i]
         more_complex_col = fm_to_check[more_complex_name]
 
-        # Convert boolean column to be float64
-        if pd.api.types.is_bool_dtype(more_complex_col):
+        # Convert boolean or Int64 column to be float64
+        if pd.api.types.is_bool_dtype(more_complex_col) or isinstance(
```
sbadithe, line 174: Changing `NumCharacters` to return `IntegerNullable` triggered an error here. Int64 doesn't seem to support the `corr` operation. There is a `_is_integer_dtype` function, but that would cast a wider net and grab series with the `int64` dtype. This way is a little more selective and does the conversion less.

thehomebrewnerd (approved):
This looks good to me. My only minor nitpick is that is unrelated to this change, is that perhaps we should update the primitive docstring to make it more clear that this primitive counts all characters, including whitespace and punctuation, but that's not a big deal.

PR #2469, opened 2023-01-30
Remove autouse from PrimitiveT.es fixture

rwedge:
The test class's test methods are not making use of autouse so this PR turns off autouse


gsheni:
Added in https://github.com/alteryx/featuretools/pull/2507

PR #2474, opened 2023-02-01
Change `TotalWordLength` and `UpperCaseWordCount` to return IntegerNullable

sbadithe:
IntegerNullable seems more appropriate, as fractional values seem not to be possible.

### DIFF
--- a/featuretools/tests/primitive_tests/test_upper_case_word_count.py
+++ b/featuretools/tests/primitive_tests/test_upper_case_word_count.py
@@ -24,7 +24,6 @@ class TestUpperCaseWordCount:
             primitive_func(x),
             answers,
             check_names=False,
-            check_dtype=False,
         )
 
     def test_nan(self):
@@ -42,5 +41,4 @@ class TestUpperCaseWordCount:
             primitive_func(x),
             answers,
             check_names=False,
-            check_dtype=False,
         )
--- a/featuretools/tests/primitive_tests/test_upper_case_word_count.py
+++ b/featuretools/tests/primitive_tests/test_upper_case_word_count.py
@@ -19,7 +19,7 @@ class TestUpperCaseWordCount:
             dtype="string",
         )
         primitive_func = self.primitive().get_function()
-        answers = pd.Series([2, 1, 2, 3, 1])
+        answers = pd.Series([2, 1, 2, 3, 1], dtype='Int64')
         pd.testing.assert_series_equal(
             primitive_func(x),
             answers,
@@ -36,7 +36,7 @@ class TestUpperCaseWordCount:
             dtype="string",
         )
         primitive_func = self.primitive().get_function()
-        answers = pd.Series([pd.NA, 0, 2])
+        answers = pd.Series([pd.NA, 0, 2], dtype='Int64')
         pd.testing.assert_series_equal(
             primitive_func(x),
             answers,
--- a/featuretools/tests/primitive_tests/test_upper_case_word_count.py
+++ b/featuretools/tests/primitive_tests/test_upper_case_word_count.py
@@ -19,9 +19,9 @@ class TestUpperCaseWordCount:
             dtype="string",
         )
         primitive_func = self.primitive().get_function()
-        answers = pd.Series([2, 1, 2, 3, 1], dtype='Int64')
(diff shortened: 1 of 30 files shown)


featuretools/tests/primitive_tests/test_upper_case_word_count.py
```diff
@@ -19,9 +19,12 @@ def test_strings(self):
             dtype="string",
         )
         primitive_func = self.primitive().get_function()
-        answers = pd.Series([2.0, 1.0, 2.0, 3.0, 1.0])
+        answers = pd.Series([2, 1, 2, 3, 1])
```
gsheni, line 22:
```suggestion
        answers = pd.Series([2, 1, 2, 3, 1], dtype='Int64')
```

featuretools/tests/primitive_tests/test_upper_case_word_count.py
```diff
@@ -34,7 +37,10 @@ def test_nan(self):
             dtype="string",
         )
         primitive_func = self.primitive().get_function()
-        answers = pd.Series([pd.NA, 0.0, 2.0])
+        answers = pd.Series([pd.NA, 0, 2])
```
gsheni, line 40:
```suggestion
        answers = pd.Series([pd.NA, 0, 2], dtype='Int64')
```

featuretools/tests/primitive_tests/test_upper_case_word_count.py
```diff
@@ -19,9 +19,12 @@ def test_strings(self):
...
         )
         primitive_func = self.primitive().get_function()
-        answers = pd.Series([2.0, 1.0, 2.0, 3.0, 1.0])
+        answers = pd.Series([2, 1, 2, 3, 1], dtype='Int64')
         pd.testing.assert_series_equal(
-            primitive_func(x), answers, check_names=False, check_dtype=False
+            primitive_func(x),
+            answers,
+            check_names=False,
+            check_dtype=False,
```
gsheni, line 27:
```suggestion
            check_dtype=True,
```

featuretools/tests/primitive_tests/test_upper_case_word_count.py
```diff
@@ -34,7 +37,10 @@ def test_nan(self):
...
         )
         primitive_func = self.primitive().get_function()
-        answers = pd.Series([pd.NA, 0.0, 2.0])
+        answers = pd.Series([pd.NA, 0, 2], dtype='Int64')
         pd.testing.assert_series_equal(
-            primitive_func(x), answers, check_names=False, check_dtype=False
+            primitive_func(x),
+            answers,
+            check_names=False,
+            check_dtype=False,
```
gsheni, line 45:
```suggestion
            check_dtype=True,
```

sbadithe: removed `check_dtype=False` as it defaults to True

featuretools/tests/primitive_tests/test_upper_case_word_count.py
```diff
@@ -19,12 +19,11 @@ def test_strings(self):
             dtype="string",
         )
         primitive_func = self.primitive().get_function()
-        answers = pd.Series([2.0, 1.0, 2.0, 3.0, 1.0])
+        answers = pd.Series([2, 1, 2, 3, 1], dtype="Int64")
         pd.testing.assert_series_equal(
-            primitive_func(x),
+            primitive_func(x).astype("Int64"),
```
gsheni, line 24: So this means our primitive is not returning the expected data type?

sbadithe: Yes, the dtype in the first test is `int64` (I think as there are no null values), and the second test is `object` , which I am looking into

gsheni: I think when we run DFS/CFM, we do cast the columns to match the primitive return_type. So this might be out of scope, and we can keep the `astype`

sbadithe: I think I agree. I think the type inference in DFS/CFM is working as all the tests are passing

gsheni (approved):

PR #2476, opened 2023-02-01
Fix latest dependency checker install command

gsheni:


.github/workflows/latest_dependency_checker.yaml
```diff
@@ -22,7 +22,8 @@ jobs:
       - name: Update dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install -e .[spark, test]
```
thehomebrewnerd, line 25: Looks like the previous command was `python -m pip install .[spark,test]` without a space between. Would that work here? Might be nice to install them both together as the dependency resolver might arrive at different package versions when we do a separate install - I don't think pip takes into account already installed packages for subsequent install commands.

.github/workflows/latest_dependency_checker.yaml
```diff
@@ -22,7 +22,8 @@ jobs:
...
+          python -m pip install -e ".[spark]"
```
gsheni, line 25:
```suggestion
          python -m pip install -e ".[spark,test]"
```

.github/workflows/latest_dependency_checker.yaml
gsheni: Fixed

pyproject.toml
```diff
@@ -104,13 +104,13 @@ docs = [
...
 ]
 dev = [
     "ruff >= 0.0.239",
     "black[jupyter] >= 23.1.0",
     "pre-commit >= 2.20.0",
-    "featuretools[docs, spark, test]",
+    "featuretools[docs,spark,test]",
 ]
 complete = [
     "featuretools[autonormalize, nlp, sklearn, spark, sql, tsfresh, updater]",
```
gsheni, line 116:
```suggestion
    "featuretools[autonormalize,nlp,sklearn,spark,sql,tsfresh,updater]",
```

thehomebrewnerd (approved):

PR #2516, opened 2023-03-13
Implement New Single Table DFS Algorithm

### LINKED ISSUE #2487: Implement Single Table DFS Algorithm
# Summary
Algorithm for Faster Generate Complete Feature Definitions

# Big Changes
- No Data, Schema only
- features that require data are partially complete at this stage and filled out during calculate feature matrix
- Single Table only for now
- No Stacking
- Simple Dataclass Objects as much as possible
- More functional based approach for easier testing

# Inputs
- Data Schema - Simple JSON Array
  - column name - internal name preferable
  - column logical type
  - column semantic tags

# Outputs
- Features Array (simple objects)
  - Feature Name
  - Primitive
  - Base Columns (array of base columns)

# Flow

## Step 1 - Convert to Dataclasses:
Schema to Columns and Features. Note: this can probably be refined

## Step 2 - Group

Create Hashmap of:
- Logical Type -> List[Feature]
- Tag -> List[Feature]
- ANY -> List[Feature]

## Step 3 - Iterate through
given a list of approved primitives, iterate through each one.

## Step 4 - Iterate through input sets for each primitive
for example, a primitive may take (Boolean, Boolean) OR (Numeric, Numeric)

## Step 5 - For each input set, get list of features
- Using the shorthand of F1(B) to represent a boolean column with name of F1, and F2(N) to represent a numeric column with the name of F2

- our input features are `F1(N), F2(N), F3(B), F4(N), F5(B)`

- primitive signature is `N,B`

- list of feature combinations would be

```
F1, F3
F2, F3
F4, F3
F1, F5
F2, F5
F4, F5
```
## Step 5a - Index Input Set
Get types and positions of Input set

```
InputSet: [N,B,N] //2 numerics and a boolean
Output: {N: [0, 2], B: [1]}

InputSet: [C] //1 categorical
Output: {C: [0]}

InputSet: [(), ()] //2 of any type
Output: {'ANY': [0,1]}
```
## Step 5b - Get Combinations for each Type
Implementation

For each type from 5a, get combinations of features.

if > 1

if commutative

itertools.combinations

if not commutative

itertools.permutations

```
InputSet: [N,B,N], Commutative: False
---
N: [0,2]
Output: 
[
	(F1,F2),
	(F1,F4),
	(F2,F1),
	(F2,F4),
	(F4,F1),
	(F4,F2),
]
---
B: [1]
Output:
[
	(F3,)
	(F5,)
]
```
## Step 5c - Product of Combinations
```
Input:
combs = [
	[
		(F1,F2),
		(F1,F4),
		(F2,F1),
		(F2,F4),
		(F4,F1),
		(F4,F2),
	],
	[
		(F3,)
		(F5,)
	]
]

itertools.product(*combs) = 
[
	[(F1,F2), (F3,)],
	[(F1,F4), (F3,)],
	[(F2,F1), (F3,)],
	[(F2,F4), (F3,)],
	[(F4,F1), (F3,)],
	[(F4,F2), (F3,)],
	
	[(F1,F2), (F5,)],
	[(F1,F4), (F5,)],
	[(F2,F1), (F5,)],
	[(F2,F4), (F5,)],
	[(F4,F1), (F5,)],
	[(F4,F2), (F5,)],
]
```
## Step 5d - Flatten and Rearrange
```
[
	[F1, F3, F2],
	[F1, F3, F4],
	[F2, F3, F1],
	[F2, F3, F4],
	[F4, F3, F1],
	[F4, F3, F2],
	
	[F1, F5, F2],
	[F1, F5, F4],
	[F2, F5, F1],
	[F2, F5, F4],
	[F4, F5, F1],
	[F4, F5, F2],
]
```

# Acceptance Criteria
Featuretools includes a single separate public function that takes the following as input:
- woodwork schema object 
- a list of available primitives

and produces a list of features as output.

Each underlying private function will be unit tested.

Parity testing against current implementation will be done on 2 dataframes that are commonly used for testing within the library.

dvreed77:
### Pull Request Description
Implement New Single Table DFS Algorithm only using Schema as input

Fixes #2487


pyproject.toml
```diff
@@ -111,6 +111,7 @@ dev = [
     "black[jupyter] >= 23.1.0",
     "pre-commit >= 2.20.0",
     "featuretools[docs,spark,test]",
+    "typing_extensions==4.5.0"
```
gsheni, line 114:
```suggestion
    "typing_extensions == 4.5.0",
```

pyproject.toml
```diff
@@ -40,6 +40,7 @@ dependencies = [
     "scipy >= 1.4.0",
     "tqdm >= 4.32.0",
     "woodwork[dask] >= 0.18.0",
+    "typing_extensions == 4.5.0",
```
gsheni, line 43: Not a fan of adding dependencies to Featuretools, unless it's absolutely necessary.

featuretools/feature_discovery/type_defs.py
```diff
@@ -0,0 +1,179 @@
...
+from woodwork.logical_types import LogicalType
+
+import featuretools as ft
+from featuretools.feature_base.feature_base import FeatureBase
+from featuretools.primitives.base.primitive_base import PrimitiveBase
+
+ANY = "ANY"
+
+
+def get_all_primitives():
```
sbadithe, line 19: Can this change at runtime? Should we make this a constant?

featuretools/feature_discovery/type_defs.py
```diff
@@ -0,0 +1,179 @@
...
+    for attribute_string in dir(ft.primitives):
+        attribute = getattr(ft.primitives, attribute_string)
+        if isclass(attribute):
+            if issubclass(attribute, PrimitiveBase) and attribute.name:
+                primitives.add(attribute)
+    return {prim.__name__: prim for prim in primitives}
+
+
+def get_all_logical_types():
+    """Helper function that selects all primitives
```
sbadithe, line 33: I think this docstring has to be updated

featuretools/feature_discovery/type_defs.py
```diff
@@ -0,0 +1,179 @@
...
+        name=name,
+        logical_type=logical_type,
+        tags=tags,
+        primitive=primitive,
+        base_features=base_features,
+        # TODO: replace this with dataframe name?
+        df_id=None,
+    )
+
+    return f
```
sbadithe, line 179:
```suggestion
    return Feature(
        name=name,
        logical_type=logical_type,
        tags=tags,
        primitive=primitive,
        base_features=base_features,
        # TODO: replace this with dataframe name?
        df_id=None,
    )
```

featuretools/primitives/standard/transform/binary/subtract_numeric.py
```diff
@@ -32,6 +32,7 @@ class SubtractNumeric(TransformPrimitive):
     return_type = ColumnSchema(semantic_tags={"numeric"})
     compatibility = [Library.PANDAS, Library.DASK]
     description_template = "the result of {} minus {}"
+    commutative = True
```
sbadithe, line 35: This may be out of scope, but my initial thought when reading `commutative=True` is that the primitive is commutative

featuretools/feature_discovery/feature_discovery.py
```diff
@@ -0,0 +1,413 @@
...
+
+def get_primitive_return_type(primitive: Type[PrimitiveBase]) -> ColumnSchema:
+    if primitive.return_type is None:
+        return_type = primitive.input_types[0]
+        if isinstance(return_type, list):
+            return_type = return_type[0]
+    else:
+        return_type = primitive.return_type
+
+    return return_type
```
sbadithe, line 275:
```suggestion
    if primitive.return_type: 
       return primitive.return_type
    return_type = primitive.input_types[0]
    if isinstance(return_type, list):
       return_type = return_type[0]
    return return_type
```
Nit, but IMO this is easier to read. Feel free to disregard

featuretools/feature_discovery/feature_discovery.py
```diff
@@ -0,0 +1,413 @@
...
+    Args:
+        feature (Feature):
+
+    Returns:
+        List[str]
+            List of hashing keys
+    """
+    keys: List[str] = []
+    logical_type = feature.logical_type
+    lt_name = None
```
sbadithe, line 147:
```suggestion
    logical_type = feature.logical_type
    logical_type_name = None
```
Nit, If we are aliasing the first as `logical_type`, I think it is worth aliasing the latter correspondingly

featuretools/feature_discovery/feature_discovery.py
```diff
@@ -0,0 +1,413 @@
...
+
+    Args:
+        column_schema (ColumnSchema):
+
+    Returns:
+        List[str]
+            List of hashing keys
+    """
+    keys: List[str] = []
+    lt = column_schema.logical_type
```
sbadithe, line 28: Can we alias this uniformly across helper functions?

featuretools/feature_discovery/type_defs.py
```diff
@@ -0,0 +1,179 @@
...
+            base_features=base_features,
+            df_id=input_dict["df_id"],
+        )
+
+        assert hydrated_feature.id == input_dict["id"]
+
+        return hydrated_feature
+
+
+def convert_old_to_new(feature: FeatureBase):
```
sbadithe, line 152: Can we change this function name?

featuretools/feature_discovery/type_defs.py
```diff
@@ -0,0 +1,179 @@
...
+    base_features = [convert_old_to_new(x) for x in feature.base_features]
+
+    name = feature.get_name()
+    col_schema = feature.column_schema
+
+    logical_type = col_schema.logical_type
+    if logical_type is not None:
+        assert issubclass(type(logical_type), LogicalType)
+        logical_type = type(logical_type)
```
sbadithe, line 161:
```suggestion
    if (logical_type := col_schema.logical_type):
        assert issubclass(type(logical_type), LogicalType)
        logical_type = type(logical_type)
```
I _think_ you can do this(?)

featuretools/feature_discovery/feature_discovery.py
```diff
@@ -0,0 +1,413 @@
...
+from woodwork.logical_types import LogicalType
+from woodwork.table_schema import TableSchema
+
+from featuretools.feature_discovery.type_defs import ANY, Feature
+from featuretools.primitives.base.primitive_base import PrimitiveBase
+
+
+def column_to_keys(column_schema: ColumnSchema) -> List[str]:
+    """
+    Generate hashing keys from Columns Schema. For example:
```
sbadithe, line 15: I think this description would be a good name for the function

featuretools/feature_discovery/feature_discovery.py
```diff
@@ -0,0 +1,413 @@
...
+    out = {}
+    for column_schema in input_set:
+        for key in column_to_keys(column_schema):
+            out[key] = out.get(key, 0) + 1
+    return out
+
+
+def get_features(
+    feature_groups: Dict[str, List[Feature]],
+    input_set: List[ColumnSchema],
```
sbadithe, line 78: Might be missing something, but can we call this 'required_columns_set` or something? (Unless there is another meaning of `input_set` I am missing.)

featuretools/feature_discovery/feature_discovery.py
```diff
@@ -0,0 +1,413 @@
...
+        features2 = feature_groups[name]
+
+        if commutative:
+            prod_iter.append(combinations(features2, count))
+        else:
+            prod_iter.append(permutations(features2, count))
+
+    feature_combinations = product(*prod_iter)
+
+    return [[z for y in x for z in y] for x in feature_combinations]
```
sbadithe, line 128: This is a little confusing to me -- can we make the variable names more descriptive?

featuretools/tests/feature_discovery/test_feature_discovery.py
```diff
@@ -0,0 +1,384 @@
...
+        target_dataframe_name="nums",
+        trans_primitives=primitives,
+        features_only=True,
+    )
+
+    features_new = my_dfs(df.ww.schema, primitives)
+
+    feature_names_old = set([x.get_name() for x in features_old]) - input_feature_names
+
+    feature_names_new = set([x.name for x in features_new]) - input_feature_names
```
sbadithe, line 382:
```suggestion
    feature_names_old = set([x.get_name() for x in features_old]) - input_feature_names

    feature_names_new = set([x.name for x in features_new]) - input_feature_names
```
Do we need to subtract from both if we are comparing them for equality right after?

featuretools/tests/feature_discovery/test_type_defs.py
```diff
@@ -0,0 +1,172 @@
...
+        logical_type=Double,
+        base_features=[f1, f2],
+    )
+
+    input_dict = f_orig.to_dict()
+    f_from_dict = Feature.from_dict(input_dict)
+    assert f_orig == f_from_dict
+
+
+def test_convert_old_to_new():
```
sbadithe, line 134: Can we rename this test to be more specific?

featuretools/tests/feature_discovery/test_feature_discovery.py
```diff
@@ -0,0 +1,384 @@
...
+        (ColumnSchema(semantic_tags={"index"}), ["index"]),
+        (
+            ColumnSchema(logical_type=Double, semantic_tags={"index", "other"}),
+            ["Double,index", "Double,other"],
+        ),
+    ],
+)
+def test_column_to_keys(column_schema, expected):
+    actual = column_to_keys(column_schema)
+    assert set(actual) == set(expected)
```
sbadithe, line 54: Nit, but imo it might be cleaner to make the fixture a set so we are not mutating it in the assert

featuretools/feature_discovery/type_defs.py
```diff
@@ -0,0 +1,179 @@
...
+            "name": self.name,
+            "logical_type": self.logical_type.__name__ if self.logical_type else None,
+            "tags": list(self.tags),
+            "primitive": self.primitive.__name__ if self.primitive else None,
+            "base_features": [x.to_dict() for x in self.base_features],
+            "df_id": self.df_id,
+            "id": self.id,
+        }
+
+    @staticmethod
```
sbadithe, line 124: If this is a constructor, I think we can make it a `classmethod`(?)

sbadithe (commented):
Really cool! Just took an initial look and left some comments

featuretools/primitives/standard/transform/binary/subtract_numeric.py
thehomebrewnerd: Subtraction is not commutative though. Shouldn't this be `False`?

featuretools/feature_discovery/type_defs.py
dvreed77: I think staticmethod makes more sense here because we are not modifying any class properties

featuretools/primitives/standard/transform/binary/subtract_numeric.py
dvreed77: yeah, I agree, but it was set as true in the constructor, so just moved that out. Setting it to false would change a lot of tests potentially. 

I think its not a huge deal to leave it as true cause the difference would just be the sign, which shouldn't affect Machine learning

thehomebrewnerd: Agree on the impact on ML - probably no need to generate 2 perfectly correlated features. Maybe we should make a note of that logic here for future reference?

featuretools/tests/feature_discovery/test_feature_discovery.py
dvreed77: I don't think this is mutating, but could be missing something.

featuretools/feature_discovery/feature_discovery.py
dvreed77: done

featuretools/primitives/standard/transform/binary/subtract_numeric.py
dvreed77: MR opened here: https://github.com/alteryx/featuretools/pull/2527

featuretools/feature_discovery/type_defs.py
```diff
@@ -0,0 +1,176 @@
...
+        hydrated_feature = Feature(
+            name=input_dict["name"],
+            logical_type=logical_type,
+            tags=set(input_dict["tags"]),
+            primitive=primitive,
+            base_features=base_features,
+            df_id=input_dict["df_id"],
+        )
+
+        assert hydrated_feature.id == input_dict["id"]
```
ozzieD, line 128: This assert feels a bit redundant. Any reason this wouldn't be equal?

featuretools/feature_discovery/convertors.py
```diff
@@ -0,0 +1,192 @@
...
+    Returns:
+       List[FeatureBase]
+    """
+    feature_cache: FeatureCache = {}
+
+    converted_features: List[FeatureBase] = []
+    for feature in feature_list:
+        if feature.is_multioutput():
+            related_feature_ids = [f.id for f in feature.related_features]
+            if any([x in feature_cache for x in related_feature_ids]):
```
cp2boston, line 181: How large is the list?  Here, you're creating the entire list with the comprehension before applying the any().  If the list is large, you could use a generator instead, which would check each on the fly and stop with the first one found.

featuretools/feature_discovery/convertors.py
```diff
@@ -0,0 +1,192 @@
...
+        df_id=None,
+    )
+
+
+def get_base_features(
+    feature: Feature,
+    feature_cache: FeatureCache,
+) -> List[FeatureBase]:
+    """
+    Get base features off of cache. Handles the case where feature maybe
```
cp2boston, line 64: picky: maybe s/b may be

featuretools/feature_discovery/convertors.py
```diff
@@ -0,0 +1,192 @@
...
+    Returns:
+       FeatureBase
+    """
+
+    def rfunc(feature: Feature) -> FeatureBase:
+        # if feature has already been converted, return from cache
+        if feature.id in cache:
+            return cache[feature.id]
+
+        # if depth is 0, we are at an origin featire
```
cp2boston, line 145: picky: featire s/b feature

featuretools/feature_discovery/convertors.py
```diff
@@ -0,0 +1,192 @@
...
+
+import pandas as pd
+from woodwork.logical_types import LogicalType
+
+from featuretools.feature_base.feature_base import (
+    FeatureBase,
+    IdentityFeature,
+    TransformFeature,
+)
+from featuretools.feature_discovery.type_defs import Feature
```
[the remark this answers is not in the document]
dvreed77: yeah, any ideas for a good name?

featuretools/feature_discovery/convertors.py
dvreed77: good idea

featuretools/feature_discovery/feature_discovery.py
```diff
@@ -0,0 +1,571 @@
...
+    logical_type = column_schema.logical_type
+    tags = column_schema.semantic_tags
+    lt_key = None
+    if logical_type:
+        lt_key = type(logical_type).__name__
+
+    if len(tags) > 0:
+        for tag in tags:
+            if tag is not None:
+                keys.append(f"{lt_key},{tag}" if lt_key is not None else tag)
```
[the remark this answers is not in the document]
dvreed77: Interesting case. If ColumnA is defined as `ColumnSchema(logical_type=Double, semantic_tags={"index", "other"})` would that mean that it only fits input Features that match that signature exactly? 

So would we NOT want Feature(Double, ("index")) to match since it doesn't have the "other" tag?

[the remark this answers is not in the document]
rwedge: I think the question I'm trying to pose is how would we handle a primitive that required two semantic tags and a logical type for one of its inputs?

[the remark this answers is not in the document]
dvreed77: yeah, I think you're right. I was trying to find an existing primitive that does this, but I don't think it exists. In the OR case, the primitive would just have a different line item, but I think I do need to think about this AND case, and your suggestion is right.

I think that means I only need 1 key for `"{lt_key},{tag1},{tag2}"` and multiple keys like I'm doing now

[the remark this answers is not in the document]
dvreed77: refactored this keying to be part of the featurecollection creation

featuretools/feature_discovery/convertors.py
```diff
@@ -0,0 +1,178 @@
...
+        base_features (List[FeatureBase])
+
+    Returns:
+       FeatureBase
+    """
+    assert feature.primitive
+    assert isinstance(
+        feature.primitive,
+        TransformPrimitive,
+    ), "Only Transform Primitives"
```
sbadithe, line 78: Should we print what we got instead here?

featuretools/feature_discovery/convertors.py
```diff
@@ -0,0 +1,178 @@
...
+        logical_type=logical_type,
+        tags=tags,
+        primitive=primitive,
+        base_features=base_features,
+        # TODO: replace this with dataframe name?
+        df_id=None,
+    )
+
+
+def to_transform_feature(
```
sbadithe, line 59: Nit: just to be consistent, I think we should name this with the `convert x to y` format used in this PR

thehomebrewnerd (commented):
Just starting to go through this in detail, but wanted to leave a little initial feedback. I haven't gotten into the heart of the algorithm yet, but will as soon as I can.

Something else to think about here. I think we should decide what new functions we want to be part of the public API and which ones should be private. We should add the new public functions to the api reference document, and perhaps more clearly mark others as private. We should also think about adding some documentation about this new functionality (could create a separate ticket for that).

featuretools/feature_discovery/utils.py
```diff
@@ -0,0 +1,70 @@
...
+}
+inferred_tag_map[None] = set()
+
+
+def column_schema_to_keys(column_schema: ColumnSchema) -> List[str]:
+    """
+    Generate hashing keys from Columns Schema. For example:
+    - ColumnSchema(logical_type=Double) -> ["Double"]
+    - ColumnSchema(semantic_tags={"index"}) -> ["index"]
+    - ColumnSchema(logical_type=Double, semantic_tags={"index", "other"}) -> ["Double,index", "Double,other"]
```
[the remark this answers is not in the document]
dvreed77: This function does do that, but I forgot to update the doc string

featuretools/feature_discovery/LiteFeature.py
```diff
@@ -0,0 +1,260 @@
...
+        self.base_features = base_features if base_features else []
+        self.df_id = df_id
+        self.idx = idx if idx is not None else 0
+        self.related_features = related_features if related_features else set()
+
+        if self.primitive:
+            assert isinstance(self.primitive, PrimitiveBase)
+            assert (
+                len(self.base_features) > 0
+            ), "there must be base features if give a primitive"
```
sbadithe, line 58:
```suggestion
            ), "there must be base features if given a primitive"
```

featuretools/feature_discovery/LiteFeature.py
```diff
@@ -0,0 +1,260 @@
...
+        if df_id:
+            hash_msg.update(df_id.encode("utf-8"))
+
+        if primitive:
+            # TODO: hashing should be on primitive
+            hash_msg.update(hash_primitive(primitive)[0].encode("utf-8"))
+            commutative = primitive.commutative
+            assert (
+                len(base_features) > 0
+            ), "there must be base features if give a primitive"
```
sbadithe, line 98:
```suggestion
            ), "there must be base features if given a primitive"
```

featuretools/feature_discovery/feature_discovery.py
```diff
@@ -0,0 +1,456 @@
...
+
+            column_set = [ColumnSchema(semantic_tags={"numeric"}), ColumnSchema(semantic_tags={"numeric"})]
+            indexed_column_set = index_column_set(column_set)
+            [("numeric": 2)]
+    """
+    out = {}
+    for column_schema in column_set:
+        for key in column_schema_to_keys(column_schema):
+            out[key] = out.get(key, 0) + 1
+    return [(k, v) for k, v in out.items()]
```
sbadithe, line 42: Nit: I think you can do: 
```suggestion
    return list(out.items())
```

featuretools/feature_discovery/LiteFeature.py
```diff
@@ -0,0 +1,260 @@
...
+        if self.name:
+            return self.name
+        elif len(self.related_features) > 0:
+            return f"{self._gen_name}[{self.idx}]"
+        return self._gen_name
+
+    def get_depth(self) -> int:
+        return self.depth
+
+    def dependendent_primitives(self) -> Set[Type[PrimitiveBase]]:
```
sbadithe, line 176:
```suggestion
    def dependent_primitives(self) -> Set[Type[PrimitiveBase]]:
```

featuretools/feature_discovery/feature_discovery.py
```diff
@@ -0,0 +1,456 @@
...
+def get_features(
+    feature_collection: FeatureCollection,
+    column_keys: Tuple[Tuple[str, int]],
+    commutative: bool,
+) -> List[List[LiteFeature]]:
+    """
+    Calculates all LiteFeature combinations using the given hashmap of existing features, and the input set of required columns.
+
+    Args:
+        feature_groups (FeatureCollection):
```
sbadithe, line 54: docstring needs to be updated

featuretools/feature_discovery/feature_discovery.py
```diff
@@ -0,0 +1,456 @@
...
+
+    return column_sets
+
+
+def get_matching_features(
+    feature_collection: FeatureCollection,
+    primitive: PrimitiveBase,
+) -> List[List[LiteFeature]]:
+    """
+    For a given primitive, find all feature sets that can be used to create new feature
```
sbadithe, line 123: docstring needs to be updated

featuretools/feature_discovery/feature_discovery.py
```diff
@@ -0,0 +1,456 @@
...
+def features_from_primitive(
+    primitive: PrimitiveBase,
+    feature_collection: FeatureCollection,
+) -> List[LiteFeature]:
+    """
+    For a given primitive, creates all engineered features
+
+    Args:
+        primitive (Type[PrimitiveBase])
+        feature_groups (Dict[str, List[LiteFeature]]):
```
sbadithe, line 203: docstring needs to be updated here

featuretools/feature_discovery/convertors.py
```diff
@@ -0,0 +1,214 @@
...
+    else:
+        fb = fb.rename(feature.get_name())
+
+    return fb
+
+
+def convert_feature_to_featurebase(
+    feature: LiteFeature,
+    dataframe: pd.DataFrame,
+    cache: FeatureCache = {},
```
sbadithe, line 134: is default mutable argument here intentional?

featuretools/feature_discovery/LiteFeature.py
```diff
@@ -0,0 +1,260 @@
...
+    def __repr__(self) -> str:
+        base_features = ", ".join([f"{x.id[:5]}..." for x in self.base_features])
+        return f"LiteFeature(name='{self.get_name()}', logical_type={self.logical_type}, tags={self.tags}, primitive={self.get_primitive_name()}, base_features=[{base_features}], df_id=None, id='{self.id[:5]}...' n_output_features={self.n_output_features} idx={self.idx})"
+
+    @property
+    def column_schema(self) -> ColumnSchema:
+        return ColumnSchema(logical_type=self.logical_type, semantic_tags=self.tags)
+
+    def rename(self, name: str):
+        self.name = name
```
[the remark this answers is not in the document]
dvreed77: yes, I think you are right!

featuretools/feature_discovery/LiteFeature.py
```diff
@@ -40,37 +41,134 @@ def __init__(
...
+    def name(self):
+        if self._alias:
+            return self._alias
+        elif self.is_multioutput():
+            return f"{self._name}[{self.idx}]"
+        return self._name
+
+    @name.setter
+    def name(self, value):
+        self._alias = value
```
tamargrey, line 85: I think it's confusing to be able to set `self._name` with the `name` param at init but set `self._alias` at this setter and have the property be able to return the alias or the name. Why not keep them totally separate fields?

featuretools/feature_discovery/LiteFeature.py
```diff
@@ -0,0 +1,260 @@
...
+        }
+
+    def is_multioutput(self) -> bool:
+        return len(self.related_features) > 0
+
+    def copy(self) -> LiteFeature:
+        copied_feature = LiteFeature(
+            name=self.name,
+            logical_type=self.logical_type,
+            tags=self.tags,
```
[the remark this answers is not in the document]
dvreed77: Is this still an issue? Not exactly sure what you meant.

featuretools/feature_discovery/LiteFeature.py
```diff
@@ -40,37 +41,134 @@ def __init__(
...
+    def logical_type(self):
+        return self._logical_type
+
+    @logical_type.setter
+    def logical_type(self, _):
+        raise AttributeError("logical_type is immutable")
+
+    @property
+    def tags(self):
+        return self._tags
```
tamargrey, line 103: You'll need to return a copy of `self._tags` or a user can mutate it

```python
    f1 = LiteFeature(
        name="Column 1",
        logical_type=Double,
    )

    f1.tags.add("hi")
    assert "hi" in f1.tags
```

dvreed77: thank you! I think this also applies to what I just asked too ;)

featuretools/feature_discovery/LiteFeature.py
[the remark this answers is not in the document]
tamargrey: Ok so the vast majority of features will hit `self._tags = self._tags | self._logical_type.standard_tags` which makes a new set object, but in this very specific case, it doesn't hit that line and the tags object is the same 

```python
    f1 = LiteFeature("f1", Double)
    f2 = LiteFeature("f2", Double)

    f3 = LiteFeature(
        name="Column 1",
        base_features=[f2, f1],
        logical_type=Double,
        tags={"index"}
    )

    f_copy = f3.copy()

    f_copy.tags.add("hi")
    assert "hi" in f3._tags
```

I think that copying `tags` at the property would solve this case (assuming people don't go through `_tags`) but I think it's safest to copy any mutable objects when constructing the copied feature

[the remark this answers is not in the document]
tamargrey: Though if LiteFeatures are truly immutable, I think you're right that this isn't something to worry about since the fear would be that mutating the copied feature would mutate the original feature.

featuretools/feature_discovery/LiteFeature.py
```diff
@@ -0,0 +1,295 @@
...
+    def primitive(self, _):
+        raise AttributeError("primitive is immutable")
+
+    @property
+    def base_features(self):
+        return self._base_features
+
+    @base_features.setter
+    def base_features(self, _):
+        raise AttributeError("primitive is immutable")
```
tamargrey, line 123: should this message reference base_features?

featuretools/feature_discovery/LiteFeature.py
```diff
@@ -0,0 +1,295 @@
...
+    def depth(self):
+        return self._depth
+
+    @depth.setter
+    def depth(self, _):
+        raise AttributeError("depth is immutable")
+
+    @property
+    def related_features(self):
+        return self._related_features
```
tamargrey, line 159: Similar to `tags` this may also need a `self._related_features.copy()` to avoid mutations being passed to the LiteFeature

featuretools/feature_discovery/convertors.py
```diff
@@ -0,0 +1,214 @@
...
+) -> List[LiteFeature]:
+    """
+    Convert a FeatureBase object to a LiteFeature object
+
+    Args:
+        feature (FeatureBase):
+
+    Returns:
+       LiteFeature - converted LiteFeature object
+    """
```
thehomebrewnerd, line 31: Nit: docstring doesn't match function args and return types

featuretools/tests/feature_discovery/test_feature_discovery.py
```diff
@@ -0,0 +1,361 @@
...
+                ("f_3", "Boolean"),
+                ("f_4", "Boolean"),
+                ("f_5", "Double"),
+            ],
+            [Lag, Day, Absolute, AddNumeric],
+        ),
+    ],
+)
+@patch.object(LiteFeature, "_generate_hash", lambda x: x.name)
+def test_compare_dfs(col_defs, primitives):
```
thehomebrewnerd, line 317: Just wondering if we need a test where we exercise all transform primitives and compare the results so that if we ever make a change to the core DFS algo behavior that might impact some edge cases, we have confirmed we get the same behavior here? Not sure...might be overkill.

featuretools/feature_discovery/LiteFeature.py
```diff
@@ -0,0 +1,295 @@
...
+            df_id=self._df_id,
+            idx=self._idx,
+            related_features=self._related_features,
+        )
+
+        copied_feature.name = self._alias
+
+        return copied_feature
+
+    def __repr__(self) -> str:
```
tamargrey, line 293: Thoughts on simplifying the repr to be more similar to BaseFeatures? I think including all this info is making it hard to understand a short list of origin features (compare to the list of FeatureBase objects below)

image

featuretools/tests/feature_discovery/test_feature_discovery.py
tamargrey: I like this idea!

featuretools/tests/feature_discovery/test_convertors.py
```diff
@@ -0,0 +1,243 @@
...
+
+    fb = convert_feature_to_featurebase(f_2, df, {})
+
+    assert isinstance(fb, TransformFeature)
+    assert fb.get_name() == "ABSOLUTE(f_1)"
+    assert len(fb.base_features) == 1
+    assert fb.base_features[0].get_name() == "f_1"
+
+    f_2.name = "f_2"
+    fb = convert_feature_to_featurebase(f_2, df, {})
```
thehomebrewnerd, line 131: I'm not sure if this is the right place for this question, but what happens if we create the `f_2` feature which depends on `f_1`, and then after that `f_1` gets renamed (a new alias, I think), does everything still work as expected when it comes time to calculate feature values for `f_2`? What would happen to the default name of the `f_2` feature in this case?

featuretools/feature_discovery/LiteFeature.py
```diff
@@ -0,0 +1,295 @@
...
+        return self._id == other._id
+
+    def __lt__(self, other: LiteFeature):
+        return self._id < other._id
+
+    def __ne__(self, other):
+        return self._id != other._id
+
+    def __hash__(self):
+        return hash(self._id)
```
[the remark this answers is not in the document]
dvreed77: `id` is simply a string at this point, you still need to wrap it in the built in hash function. https://docs.python.org/3.5/reference/datamodel.html#object.__hash__

featuretools/feature_discovery/LiteFeature.py
```diff
@@ -0,0 +1,295 @@
...
+
+        hash_msg.update(str(idx).encode("utf-8"))
+
+        return hash_msg.hexdigest()
+
+    def __eq__(self, other: LiteFeature):
+        return self._id == other._id
+
+    def __lt__(self, other: LiteFeature):
+        return self._id < other._id
```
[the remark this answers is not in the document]
dvreed77: LiteFeatures need to be sortable to make consistent arrays of base_features that are used for downstream hashing

featuretools/tests/feature_discovery/test_feature_discovery.py
dvreed77: This is sort of tested in the looking glass tests

dvreed77: Ok, I'll put it here, if it becomes a problem we can always remove

featuretools/tests/feature_discovery/test_type_defs.py
```diff
@@ -0,0 +1,404 @@
...
+        logical_type=Double,
+        base_features=[f2, f1],
+    )
+
+    assert f6 != f7
+
+
+def test_feature_type_assertions():
+    with pytest.raises(
+        AssertionError,
```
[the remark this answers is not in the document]
dvreed77: yeah, I think you're right here. For the assertions that I have that aren't due to user error, but are used to enforce that I'm doing the right thing internally, I think I want to keep them assertions if you're ok with that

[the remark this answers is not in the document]
thehomebrewnerd: Yeah, I think it's fine to keep the internal checks in place as assertions since we would never expect those errors to get presented to the user.

featuretools/feature_discovery/LiteFeature.py
dvreed77: @thehomebrewnerd @tamargrey. I changed this allow users to set the alias with `set_alias`. This allows you to set the alias, but the getter on `name` still works the same. If an alias is present, it uses that, otherwise uses the name. Hope that addresses your concerns

featuretools/feature_discovery/LiteFeature.py
```diff
@@ -0,0 +1,295 @@
...
+            logical_type=self._logical_type,
+            tags=self._tags,
+            primitive=self._primitive,
+            base_features=[x.copy() for x in self._base_features],
+            df_id=self._df_id,
+            idx=self._idx,
+            related_features=self._related_features,
+        )
+
+        copied_feature.name = self._alias
```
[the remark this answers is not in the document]
dvreed77: switched to `set_alias`

featuretools/feature_discovery/LiteFeature.py
```diff
@@ -0,0 +1,295 @@
...
+
+        return copied_feature
+
+    def __repr__(self) -> str:
+        base_features = ", ".join([f"{x.id[:5]}..." for x in self._base_features])
+        return f"LiteFeature(name='{self.name}', logical_type={self.logical_type}, tags={self.tags}, primitive={self.get_primitive_name()}, base_features=[{base_features}], df_id=None, id='{self.id[:5]}...' n_output_features={self.n_output_features} idx={self.idx})"
```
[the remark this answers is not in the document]
dvreed77: done

featuretools/tests/feature_discovery/test_feature_discovery.py
dvreed77: done!

featuretools/feature_discovery/FeatureCollection.py
```diff
@@ -0,0 +1,249 @@
...
+    def __repr__(self):
+        return f"<FeatureCollection ({self.hash_key[:5]}) n_features={len(self.all_features)} indexed={self.indexed}>"
+
+    @property
+    def hash_key(self) -> str:
+        if self._hash_key is None:
+            if not self.sorted:
+                self.sort_features()
+            self._set_hash()
+        assert self._hash_key is not None
```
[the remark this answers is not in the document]
dvreed77: yeah, to enforce return type

featuretools/feature_discovery/feature_discovery.py
```diff
@@ -0,0 +1,377 @@
...
+            features = schema_to_features(df.ww.schema)
+
+    """
+    features = []
+    for col_name, column_schema in schema.columns.items():
+        assert isinstance(column_schema, ColumnSchema)
+
+        logical_type = column_schema.logical_type
+        assert logical_type
+        assert issubclass(type(logical_type), LogicalType)
```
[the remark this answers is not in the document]
dvreed77: yeah for type hinting

featuretools/feature_discovery/feature_discovery.py
```diff
@@ -0,0 +1,377 @@
...
+
+        tags = column_schema.semantic_tags
+        assert isinstance(tags, set)
```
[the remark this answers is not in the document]
dvreed77: whats wrong with this? I think I need it to assert the type, else its `tags: set[Unknown | Any] | set[str | Any | Unknown] | Unknown | set[Unknown] | set[str] | set[Unknown | str]`

featuretools/feature_discovery/feature_discovery.py
```diff
@@ -0,0 +1,377 @@
...
+def primitive_to_columnsets(primitive: PrimitiveBase) -> List[List[ColumnSchema]]:
+    column_sets = primitive.input_types
+    assert column_sets is not None
+    if not isinstance(column_sets[0], list):
+        column_sets = [primitive.input_types]
+
+    column_sets = cast(List[List[ColumnSchema]], column_sets)
+
+    # Some primitives are commutative, yet have explicit versions of commutative pairs (eg. MultiplyNumericBoolean),
+    # which would create multiple versions, so this resolved that.
```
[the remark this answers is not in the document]
dvreed77: Created a ticket here: https://github.com/alteryx/featuretools/issues/2559

featuretools/feature_discovery/convertors.py
```diff
@@ -0,0 +1,214 @@
...
+    return fb
+
+
+def convert_feature_to_featurebase(
+    feature: LiteFeature,
+    dataframe: pd.DataFrame,
+    cache: FeatureCache,
+) -> FeatureBase:
+    """
+    Recursively transforms a feature object into a Featurebase object
```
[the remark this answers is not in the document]
dvreed77: done

featuretools/feature_discovery/FeatureCollection.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def get_by_origin_feature_name(self, name: str) -> LiteFeature:
+        feature = self.by_name.get(name)
+        assert feature is not None
+        return feature
+
+    def get_dependencies_by_origin_name(self, name) -> Set[LiteFeature]:
+        origin_feature = self.by_name[name]
+
+        assert origin_feature, "no origin feature with that name exists"
```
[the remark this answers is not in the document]
dvreed77: fixed

featuretools/feature_discovery/FeatureCollection.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def get_by_tag(self, tag: str) -> Set[LiteFeature]:
+        return self.by_tag.get(tag, set())
+
+    def get_by_origin_feature(self, origin_feature: LiteFeature) -> Set[LiteFeature]:
+        return self.by_origin_feature.get(origin_feature, set())
+
+    def get_by_origin_feature_name(self, name: str) -> LiteFeature:
+        feature = self.by_name.get(name)
+        assert feature is not None
```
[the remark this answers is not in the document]
dvreed77: you're right. Fixed

featuretools/feature_discovery/FeatureCollection.py
```diff
@@ -0,0 +1,249 @@
...
+from featuretools.feature_discovery.utils import hash_primitive, logical_types_map
+from featuretools.primitives.base.primitive_base import PrimitiveBase
+from featuretools.primitives.utils import (
+    PrimitivesDeserializer,
+)
+
+
+class FeatureCollection:
+    def __init__(self, features: List[LiteFeature]):
+        self.all_features: List[LiteFeature] = features
```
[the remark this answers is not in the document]
dvreed77: done

featuretools/tests/feature_discovery/test_convertors.py
dvreed77: Changing the alias has the biggest impact when dealing with origin features and should be avoided as much as possible, but should be fine: 

Here is what would happen:
- You create an origin Feature based on column named "Column A"
- You create an engineered Feature using the Absolute primitive to create the name "ABSOLUTE(Column A)"
- You now have 2 features, and if you want to create a feature matrix off of this, you need to convert these back to FeatureBase features first using the conversion function. This will create 2 FeatureBase features with names above and everything should work as expected

Now if you want to change the alias AFTER feature discovery, here is what would happen.
- you set an alias on the origin LiteFeature, to give it a name of "f1", and alter the corresponding column on the dataframe
- you now need to convert the 2 LiteFeatures to FeatureBase features. The Origin Feature would create an IdentityFeature pointing to the "f1" column on the dataframe. Then it would create a TransformFeature, with the name of "ABSOLUTE(Column A)", which probably isn't what you want, but it will still be pointing to the correct base_feature.
- Essentially everything works as expected, its just the names won't be right.

thehomebrewnerd (approved):

PR #2522, opened 2023-03-15
Refactor `can_stack_primitive_on_inputs`

### LINKED ISSUE #2521: Refactor `can_stack_primitive_on_inputs` helper function
- Refactoring `can_stack_primitive_on_inputs` would help make `DeepFeatureSynthesis` more readable

sbadithe:
- closes #2521


featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -1116,43 +1159,34 @@ def can_stack_primitive_on_inputs(primitive, inputs):
...
+            tuple_primitive_stack_on_exclude,
         ):
             return False
 
-        if primitive_stack_on_self and isinstance(f_primitive, primitive_class):
-            continue
-
-        if tup_primitive_stack_on is None or isinstance(
+        # we permit stacking only if it is not prohibited and meets the criterion to be permitted
+        if _check_if_stacking_is_permitted(
```
dvreed77, line 1174: maybe I'm missing something, but prohibited and permitted should be mutually exclusive states, and so if its not prohibited, its permitted. Can you refactor even more to make this the case?

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -1116,43 +1159,34 @@ def can_stack_primitive_on_inputs(primitive, inputs):
...
         ):
             continue
 
-        if f_primitive.base_of is None:
-            continue
-        if primitive_class in f_primitive.base_of:
-            continue
-
+        # if we reach this case, we default to stacking not being permitted
         return False
```
ozzieD, line 1183: since we continue into a `return False` here we could probably 
just do: `return _check_if_stacking_is_permitted()`

rwedge:
We should test this branch with the many features ecommerce test to verify feature definitions are unchanged

featuretools/synthesis/deep_feature_synthesis.py
thehomebrewnerd: @dvreed77 I don't think prohibited and permitted in this case are mutually exclusive. I think we can have cases where both checks are `False` or both are `True` because the logic checks are different between the "permitted" and "prohibited" functions. We want prohibited to take precedence, so in the case that both checks are True or both checks are False, we want the result to be "prohibited".

I might be missing something still, but I think those conditions make it a little tricker to refactor this further.

thehomebrewnerd: I did add some additional comments to maybe clarify what is going on a little better. I'm open to other suggestions on how to improve further though.

featuretools/synthesis/deep_feature_synthesis.py
```diff
@@ -1094,17 +1094,60 @@ def _find_root_primitive(feature):
     return feature.primitive
 
 
+def _check_if_stacking_is_prohibited(
```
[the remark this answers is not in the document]
thehomebrewnerd: Added some type hints in [5178220](https://github.com/alteryx/featuretools/pull/2522/commits/51782202324e9d52db02f44974ab8d964fc76f65)

featuretools/synthesis/deep_feature_synthesis.py
thehomebrewnerd: We only want to return when `_check_if_stacking_is_permitted` returns `False`, otherwise we need to keep going to make sure it's `True` for all inputs.

featuretools/synthesis/deep_feature_synthesis.py
sbadithe: I agree with Nate here, I remember trying to find a way to clean it up before coming up with the original changes, but none were immediately apparent to me at the time. I think this PR was an initial attempt at refactoring the stacking logic at a surface level, mostly to play around with and discover the areas where we could refactor. A second, more fundamental refactor might be needed to simplify the deeper logic of how we handle stacking.

thehomebrewnerd:
> We should test this branch with the many features ecommerce test on looking glass to verify feature definitions are unchanged

Tested this against `v1.27.0` and the number of features generated for all tests remained the same.

Issue #2526, opened 2023-03-24
Subtract Numeric does not set class property commutative


dvreed77:
Many, if not all, of the pairwise primitives like, Equal, MultiplyNumeric, etc, set the commutative property on the class. Subtract numeric is only set during initialization. 

We should stick with True, since even though Subtraction is not commutative, the result should have little bearings on ML, etc.

#### Example

```python
class SubtractNumeric:
    ...
    commutative = True
    ...
```
</details>

PR #2527, opened 2023-03-24
Update SubtractNumeric Primitive with commutative class property

dvreed77:
Fixes: #2526


rwedge (commented):
Maybe we should add a simple test that verifies the `commutative` parameter for SubtractNumeric is working as intended

PR #2560, opened 2023-04-24
Make Dask an optional dependency

### LINKED ISSUE #1423: Make dask an optional dependency
- Currently, dask is required to be installed with Featuretools (its used for if n_jobs > 1 for DFS)
- However, dask can be made into an optional dependency, and only needed if n_jobs is greater than 1 when calling DFS.
- Therefore, we should create a `dask-requirements.txt`, and modify `setup.py` with an additional `extras_require`. 
- Additionally, DFS should be modified to look for Dask if n_jobs is greater than 1. 
- If Dask is not installed, DFS should raise a error, and direct the user to install Featuretools with Dask

thehomebrewnerd:
### Make Dask an optional dependency

Closes #1423

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,12 @@ python -m pip install "featuretools[nlp]"
 python -m pip install "featuretools[tsfresh]"
 ```
 
+**Dask Support** - Use Dask Dataframes to create EntitySets or run DFS with njobs > 1
+
+```
+python -m pip install "featuretools[dask]"
+```
+
 **SQL** - Automatic EntitySet generation from relational data stored in a SQL database: 
 
 ```
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -765,7 +765,7 @@ def parallel_calculate_chunks(
 ):
     import_or_raise(
         "dask",
-        "Dask must be installed to calculate feature matrix in parallel.",
+        "Dask must be installed to calculate feature matrix with n_jobs set to anything but 1",
     )
     from dask.base import tokenize
     from distributed import Future, as_completed
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -128,7 +128,7 @@ def calculate_feature_matrix(
             percentage of all rows. if None, and n_jobs > 1 it will be set to 1/n_jobs
 
         n_jobs (int, optional): number of parallel processes to use when
-            calculating feature matrix. Requires Dask.
+            calculating feature matrix. Requires Dask if not equal to 1.
 
         dask_kwargs (dict, optional): Dictionary of keyword arguments to be
             passed when creating the dask client and scheduler. Even if n_jobs
--- a/featuretools/computational_backends/calculate_feature_matrix.py
+++ b/featuretools/computational_backends/calculate_feature_matrix.py
@@ -764,7 +764,7 @@ def parallel_calculate_chunks(
     include_cutoff_time=True,
 ):
     import_or_raise(
-        "dask",
+        "distributed",
         "Dask must be installed to calculate feature matrix with n_jobs set to anything but 1",
     )
     from dask.base import tokenize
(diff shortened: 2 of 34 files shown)


featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -127,7 +128,7 @@ def calculate_feature_matrix(
             percentage of all rows. if None, and n_jobs > 1 it will be set to 1/n_jobs
 
         n_jobs (int, optional): number of parallel processes to use when
-            calculating feature matrix.
+            calculating feature matrix. Requires Dask.
```
gsheni, line 131:
```suggestion
            calculating feature matrix. Requires Dask if not equal to 1. 
```

featuretools/computational_backends/utils.py
```diff
@@ -213,20 +213,22 @@ def get_client_cluster():
     """
     Separated out the imports to make it easier to mock during testing
     """
-    from distributed import Client, LocalCluster
+    distributed = import_or_none("distributed")
+    Client = distributed.Client
```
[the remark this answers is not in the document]
thehomebrewnerd: This should never get called unless Dask is installed. The `import_or_raise` call added to `parallel_calculate_chunks` should guard against this getting called without Dask installed.

dvreed77 (approved):
lgtm, but would feel better with another approval

README.md
```diff
@@ -77,6 +77,12 @@ python -m pip install "featuretools[nlp]"
 python -m pip install "featuretools[tsfresh]"
 ```
 
+**Dask Support** - Use Dask Dataframes to create EntitySets or run DFS with njobs > 1
+
+```
+python -m pip install "featuretools[dask]"
```
gsheni, line 83: Also add to docs:
https://github.com/alteryx/featuretools/blob/main/docs/source/install.md

featuretools/tests/conftest.py
```diff
@@ -21,14 +19,24 @@
 
 @pytest.fixture()
 def dask_cluster():
-    with LocalCluster() as cluster:
-        yield cluster
+    distributed = pytest.importorskip(
+        "distributed",
+        reason="Dask not installed, skipping",
```
sbadithe, line 24: Would it be a good idea to have a Dask error message saved instead of copying it for all tests?

gsheni (approved):

featuretools/computational_backends/utils.py
```diff
@@ -213,20 +213,22 @@ def get_client_cluster():
...
+    LocalCluster = distributed.LocalCluster
 
     return Client, LocalCluster
 
 
 def _validate_cutoff_time(
-    cutoff_time: typing.Union[dd.DataFrame, pd.DataFrame, str, datetime],
+    cutoff_time: typing.Union[pd.DataFrame, str, datetime],
```
sbadithe, line 224: Given we handle the Dask case on L231, is `dd.DataFrame` still a valid type for `cutoff_time`?

featuretools/tests/conftest.py
thehomebrewnerd: We could do that, but the only people that ever see this message are developers, so I don't think it's critical that this message be the same everywhere.

featuretools/computational_backends/utils.py
thehomebrewnerd: Technically, yes, I think `dd.DataFrame` is still valid type as input to this function, but it will always get converted to a pandas dataframe before it gets used. We get an error here if Dask isn't installed though because `dd` becomes `None` and `None` doesn't have a `DataFrame` attribute.

If there is an easy way to fix that, I can update, but wasn't sure off the type of my head.

featuretools/tests/conftest.py
sbadithe: Makes sense. I was wondering more from a code cleanliness perspective.

featuretools/computational_backends/calculate_feature_matrix.py
```diff
@@ -762,6 +763,10 @@ def parallel_calculate_chunks(
     progress_callback=None,
     include_cutoff_time=True,
 ):
+    import_or_raise(
+        "dask",
```
rwedge, line 767: I think distributed would be a better check since it requires dask and is needed for this code section

thehomebrewnerd: Updated

PR #2585, opened 2023-07-17
Add pandas 2.0 support

### LINKED ISSUE #2532: Support pandas 2.0.0 in Featuretools
- pandas 2.0.0 was released and it broke some of our unit tests.
- For this issue, remove the pandas restriction for 2.0.0 in pyproject.toml, and run the unit tests `make test`.
- Then, look at the failing unit tests, and see what calls need to be updated.
- If needed, we can restrict the spark 's pandas to not install pandas 2.0.0 (in extra_requires).

### LINKED ISSUE #2556: Support dask 2023.4.0 in Featuretools

thehomebrewnerd:
### Add pandas 2.0 support
Closes #2532 
Closes #2556


featuretools/tests/primitive_tests/aggregation_primitive_tests/test_agg_primitives.py
```diff
@@ -239,11 +239,13 @@ def test_regular(self, dtype):
         given_answer = primitive_func(data)
         assert np.isclose(answer, given_answer, atol=0.01)
 
-        data = pd.Series([sqrt(x) for x in list(range(100))], dtype=dtype)
-        answer = -0.46
-        primitive_func = self.primitive().get_function()
-        given_answer = primitive_func(data)
-        assert np.isclose(answer, given_answer, atol=0.01)
+        if dtype == "float64":
```
sbadithe, line 242: This is not a blocker (more of an incidental comment), but seeing the `if` clause here makes me consider if we should eventually refactor this test. Given that the previous `data` objects are being cast to `float64` dtype as well, and the test does not require `if` clauses, it seems like there are three separate classes of behaviors here: 

- How the kurtosis method handles int64 dtype
- How the kurtosis primitive handles float64 dtype when the data "looks like" an int64 (evidently this doesn't seem to differ from the previous class) 
- How the kurtosis primitive handles float64 dtypes when the data is a "true" float 

Separating those might help to better document this primitive's behavior. In any case, this is probably something to revisit later in a future issue.

sbadithe (approved):

PR #2620, opened 2023-09-19
remove looking_glass_performance_tests

### LINKED ISSUE #2619: Remove old performance testing workflow (non airflow)
Due to current looking glass being archived, the workflow for triggering those tests will no longer work and has to be removed.

petejanuszewski1:
closes #2619 

removed looking_glass_performance_tests.yaml


thehomebrewnerd:
@petejanuszewski1 The latest release of `holidays` broke our unit tests. I think you are might need to pin `holidays<0.33` in `pyproject.toml` or alternatively update the failing tests before the CI checks will pass on this PR.

pyproject.toml
```diff
@@ -31,7 +31,7 @@ license = {text = "BSD 3-clause"}
 requires-python = ">=3.8,<4"
 dependencies = [
     "cloudpickle >= 1.5.0",
-    "holidays >= 0.13",
+    "holidays < 0.33",
```
thehomebrewnerd, line 34: I think we need to still include `>=0.13` here as older versions might be problematic.

`"holidays >= 0.13, <0.33"`

petejanuszewski1: ah, makes sense. I'll update.

thehomebrewnerd (approved):

PR #2657, opened 2024-02-09
Support pandas 2.2.0

### LINKED ISSUE #2649: Add support for pandas 2.2.0
There's a problem with calculating the feature matrix for agg primitives (at least sum, skew, std, and mean, and maybe more): https://github.com/alteryx/featuretools/actions/runs/7803132816/job/21282196680#step:10:1310

tamargrey:
closes #2649

Adds a workaround for https://github.com/pandas-dev/pandas/issues/57317

Also fixes the woodwork tests with main so they can be triggered.


featuretools/tests/computational_backend/test_feature_set_calculator.py
```diff
@@ -1293,3 +1294,26 @@ def error(s):
     # Calculating without precalculated features should error.
     with pytest.raises(RuntimeError, match=error_msg):
         FeatureSetCalculator(pd_es, feature_set=FeatureSet([direct])).run(instance_ids)
+
+
+def test_nunique_nested_with_agg_bug(pd_es):
+    """Pandas 2.2.0 has a bug where pd.Series.nunique produces columns with
+    the category dtype instead of int64 dtype, causing an error when we attempt
+    another aggregation"""
```
tamargrey, line 1302: The error message showing up was `TypeError: category type does not support sum operations`, but this is actually a bug in NumUnique, which was returning data with dtype `category` with numeric categories. The error would eventually show up because the mean, sum, std, etc aggregation functions cannot be made on non numeric data.

featuretools/computational_backends/feature_set_calculator.py
```diff
@@ -822,6 +822,21 @@ def last_n(df):
...
+                        if (
+                            isinstance(dtype, pd.CategoricalDtype)
+                            and str(dtype.categories.dtype) == "int64"
+                        ):
+                            cols_to_fix.append(col)
+
+                    if cols_to_fix:
+                        base_frame[cols_to_fix] = base_frame[cols_to_fix].astype(
+                            "int64",
+                        )
```
tamargrey, line 838: I ended up adding a fix for the bug where the groupby/aggregation is actually happening, because the `NumUnique` primitive itself just returns the pandas `pd.Series.nunique` function, and the bug is occurring at the actual aggregation step. This feels very removed from the actual bug, and I'm not fully in love with it as a workaround, so I'm open to other options. I'm a bit worried that it may be overly-broad; maybe there are numeric-category columns that we wouldn't want to turn into int64? But I didn't want to over engineer and add an explicit check for the nunique function.

tamargrey: Probably worth waiting to merge this MR till one of the pandas folks responds to the bug I opened up in case they have a better workaround than what I've found

tamargrey: Oh actually just got a response. It seems like just returning the string `"nunique"` in the primitive fixes this. Will investigate more on monday, but @thehomebrewnerd any idea why we weren't able to do that before? I see we do that for spark dataframes

thehomebrewnerd: I'm not sure off the top of my head. I think it's possible that we could have done this before, but the paradigm is for the `get_function` method of a primitive to return a function, so that's probably why we did it that way. There are probably a few cases where both a string and a function could work and this must be one of those cases.

tamargrey: The only resulting [failure](https://github.com/alteryx/featuretools/actions/runs/7850051684/job/21424702596?pr=2657) from making that change is in the docstring for the primitive, which must be the only place we attempt to calculate the value by instantiating and then using the primitive as a function. I'm leaning towards using 5dbc28d10e5ad9ca831e0d2cb14cf54a84f1ab24 as the fix here, as it is a much more minimal change than converting every numeric categorical column when we're calculating the feature matrix. 

@thehomebrewnerd any reason you can think of that we wouldn't be okay updating the docstring for now until the bug is fixed?

docs/source/release_notes.rst
```diff
@@ -8,6 +8,8 @@ Future Release
     * Enhancements
     * Fixes
         * Fix dependency issues (:pr:`2644`, :pr:`2656`)
+        * Add workaround for pandas 2.2.0 bug with nunique and unpin pandas (:pr:`2657`)
+        * Fix the tests that run with Woodwork main so they can be triggered (:pr:`2657`)
```
thehomebrewnerd, line 12: Nit: I think I'd put this in the Testing Changes section

tamargrey: done!

featuretools/computational_backends/feature_set_calculator.py
tamargrey: Decided to add a param to allow users to use `pd.Series.nunique` if need be: 6d83654f431d49500645f0385f6d6b1c97636b99. If the pandas devs have plans to fix this issue, we'll eventually remove it. If the new behavior will eventually become the behvaior even for `'nunique'`, we'll need to implement a fix in `feature_set_calculator.py` to either convert integer categorical data or call `s.groupby(…).nunique(…)` directly when need be (as that method can keep the `int64` because it is aware that `nunique` would want the eventual type to be int (whereas `.agg` has no knowledge of the operations it's calling)

thehomebrewnerd (approved):

PR #2702, opened 2024-04-09
Move flatten_list util function into feature_discovery

### LINKED ISSUE #2695: Investigate and remove Dask restriction in pyproject.toml
Need to determine what code changes are required to support newer versions of Dask, implement the required changes, and remove the version restrictions in pyproject.toml.

### LINKED ISSUE #2701: Fix import in feature_discovery module
In the `feature_discovery` module a utility function is used from the `tests` module. Because of the associated imports, this means that pytest must be installed in order to import and use any code from the module. We should move the utility function out of the `tests` module and into `feature_discovery` so it can be used without installing/importing pytest.

thehomebrewnerd:
Closes #2701 
Closes #2695


tamargrey (approved):

featuretools/feature_discovery/utils.py
```diff
@@ -75,3 +75,7 @@ def get_primitive_return_type(primitive: PrimitiveBase) -> ColumnSchema:
     if isinstance(return_type, list):
         return_type = return_type[0]
     return return_type
+
+
+def flatten_list(nested_list):
+    return [item for sublist in nested_list for item in sublist]
```
VascoSch92, line 81: Just for your information: you can improve the speed this method by using `itertools`

See https://note.nkmk.me/en/python-list-flatten/

PR #2722, opened 2024-05-09
Use filter arg in tarfile.extractall to prevent unsafe unarchival operations

### LINKED ISSUE #2723: Add filter arg to tarfile.extractall call to safely deserialize entitysets
Update `tarfile.extractall` call to safely deserialize entitysets.

thehomebrewnerd:
Closes #2723 

The filter argument was added in these Python versions: 3.9.17, 3.10.12, 3.11.4, 3.12.0. If the user is running an older Python version an error will be thrown and they will need to upgrade to one of these patch releases (or newer) to extract the archive.

### DIFF
--- a/docs/source/release_notes.rst
+++ b/docs/source/release_notes.rst
@@ -13,6 +13,7 @@ Future Release
         * Temporarily restrict Dask version (:pr:`2694`)
         * Remove support for creating ``EntitySets`` from Dask or Pyspark dataframes (:pr:`2705`)
         * Bump minimum versions of ``tqdm`` and ``pip`` in requirements files (:pr:`2716`)
+        * Use ``filter`` arg in call to ``tarfile.extractall`` to safely deserialize EntitySets (:pr:`2722`)
     * Documentation Changes
     * Testing Changes
         * Fix serialization test to work with pytest 8.1.1 (:pr:`2694`)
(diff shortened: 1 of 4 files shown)


machineFL:
No linked issues found. Please add the corresponding issues in the pull request description. <br/> [Use GitHub automation to close the issue when a PR is merged](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)

docs/source/release_notes.rst
```diff
@@ -13,6 +13,7 @@ Future Release
         * Temporarily restrict Dask version (:pr:`2694`)
         * Remove support for creating ``EntitySets`` from Dask or Pyspark dataframes (:pr:`2705`)
         * Bump minimum versions of ``tqdm`` and ``pip`` in requirements files (:pr:`2716`)
+        * Use ``filter`` arg in call to ``tarfile.extractall`` to safely deserialize EntitySets (:pr:`2722`)
```
jeremyliweishih, line 16: Would this be considered a breaking change?

jeremyliweishih (approved):
👍

docs/source/release_notes.rst
thehomebrewnerd: @jeremyliweishih I kinda debated about that - we aren't making any change to the API, so I decided against labeling it as such. It just might mean that if someone has an older environment and they upgrade FT, they might also need to upgrade Python, but only if they are using this EntitySet serialization functionality, which probably isn't all that common.

jeremyliweishih: makes sense to me!

PR #2773, opened 2026-06-30
TST: add tests validating primitive input_types shapes

### LINKED ISSUE #2086: Add tests that confirm primitive input_types are the expected shapes
There are a number of assumptions we make about the shape of Primitive `input_types` lists:
- Its either a list of ColumnSchema objects or a list of lists of ColumnSchema objects (and not a combination)
- All sub-lists are the same length
- No `input_types` list or sublist is empty 

As we may need to rely on these assumptions at some point, we should add tests that confirm these assumptions for all primitives, so that if we add a Primitive that breaks any of these assumptions in the future, we are notified.

Lakshmi786:
## Summary

Adds four parametrised pytest tests that validate the shape contracts
for `input_types` across all 203 registered primitives.

**Contracts tested:**
1. `input_types` is never empty
2. Items are uniformly `ColumnSchema` (flat) or `list` (nested) — never mixed
3. When nested, all sub-lists are the same length
4. When nested, no sub-list is empty

All 812 generated test cases pass on the current codebase.

Closes #2086

### DIFF
--- a/featuretools/tests/primitive_tests/test_input_types_shape.py
+++ b/featuretools/tests/primitive_tests/test_input_types_shape.py
@@ -1,0 +1,70 @@
+"""Tests that confirm primitive input_types lists have the expected shapes.
+
+Every Primitive's input_types must satisfy:
+  1. It is either a flat list of ColumnSchema, or a list of lists of ColumnSchema
+     (never a mix of the two).
+  2. When it is a list of lists, all sub-lists have the same length.
+  3. Neither the outer list nor any inner list is empty.
+"""
+import pytest
+from woodwork.column_schema import ColumnSchema
+
+from featuretools.primitives import get_aggregation_primitives, get_transform_primitives
+
+
+def _all_primitives():
+    prims = {}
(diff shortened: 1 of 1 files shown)


Lakshmi786:
Hi team! 👋

I am a ML/Data Science engineer working on improving open-source tooling. I noticed issue #2086 had been open for a while with no test coverage for `input_types` shape validation, so I went ahead and implemented it.

### What this PR adds
Four parametrised pytest tests that run against all **203** registered primitives and assert the contracts the codebase relies on for `input_types`:

1. The outer list is never empty
2. Items are uniformly `ColumnSchema` (flat) or `list` (nested) — never a mix
3. When nested, all sub-lists are the same length
4. When nested, no sub-list is empty

All **812** generated test cases pass on the current codebase.

Happy to adjust the test style, naming, or scope based on your feedback. Let me know if you'd like me to also add these checks to the `PrimitiveBase` class itself as a runtime guard.

Thanks for maintaining such a great library!

